PR #4082, opened 2023-09-23
Use paths for dev dependencies

sffc:
https://github.com/unicode-org/icu4x/issues/4065

Without this, I can't publish cleanly.

PR #5537, opened 2024-09-12
In zerovec, add the yoke version bound to the dev-dependency

musicinmybrain:
This allows the `dev-dependency` to appear in the normalized `Cargo.toml` in the published crate.

```toml
[dev-dependencies.yoke]
version = ">=0.6.0, <0.8.0"
features = ["derive"]
```

This is helpful when packaging for Fedora Linux, because the `dev-dependency` on `yoke` is the one that carries `features = ["derive"]`, needed for running various tests included in the published `zerovec` crate.

### DIFF
--- a/utils/zerovec/Cargo.toml
+++ b/utils/zerovec/Cargo.toml
@@ -51,7 +51,7 @@ rmp-serde = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
 potential_utf = { workspace = true, features = ["zerovec"] }
-yoke = { path = "../../utils/yoke", features = ["derive"] }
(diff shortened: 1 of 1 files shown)


Manishearth (approved):
Fine with this. dev-dependency stuff in published code is best-effort, but if it's just a small fix like this I'm all for it

sffc:
I think having versions in dev-dependencies prevents the crates from being publishable, but that constraint might only apply to "circular dev dependencies"?

See: https://github.com/unicode-org/icu4x/pull/4082

musicinmybrain:
Well, I did try running `cargo publish --dry-run` from within `utils/yoke/`, and that seemed to work as expected. I don’t know if there’s anything else I should try.

Note that here the `dev-dependency` on `yoke` is still a `path` dependency, it just *also* has a `version` specification.

sffc (approved):
OK

PR #5538, opened 2024-09-13
Partial datetime FFI migration

sffc:
#1317

It wasn't as hard as I thought it might be. Easier than panicking everywhere.


Manishearth (dismissed):

Manishearth (approved):

ffi/capi/src/errors.rs
```diff
@@ -241,6 +244,20 @@ impl From<icu_datetime::DateTimeError> for Error {
     }
 }
 
+#[cfg(feature = "datetime")]
+impl From<icu_datetime::LoadError> for Error {
```
robertbastian, line 248: issue: this is not using fine-grained error types

PR #5543, opened 2024-09-14
Remove 1.x DateTimeFormatter and TypedDateTimeFormatter

sffc:
#1317 

Also promotes NeoFormatter and TypedNeoFormatter from experimental. (doesn't rename them yet)


sffc:
I deleted the old formatter, and the code compiles + passes tests with `--all-features`. Need to still clean up features.

sffc:
There's more to delete, but I would rather check-in this big chunk than let it bitrot while I look through ever nook and cranny (which I intend to do before 2.0)

README.md
```diff
@@ -37,19 +37,16 @@ icu = "1.5.0"
...
-
-let dtf = DateTimeFormatter::try_new(&locale!("es").into(), options)
+let dtf = NeoFormatter::<NeoAutoDateTimeMarker>::try_new(&locale!("es").into(), NeoSkeletonLength::Long.into())
     .expect("locale should be present in compiled data");
 
 let date = DateTime::try_new_iso_datetime(2020, 9, 12, 12, 35, 0).expect("datetime should be valid");
 let date = date.to_any();
 
-let formatted_date = dtf.format_to_string(&date).expect("formatting should succeed");
+let formatted_date = dtf.convert_and_format(&date).to_string_lossy();
```
robertbastian, line 49: the `expect` that was here before fails?

sffc: It doesn't fail; the error is on the TryWriteable impl now, thus the `to_string_lossy`

sffc: This PR creates `to_string_lossy`, an associated function on FormattedNeoDateTime, because I had a lot of docs tests I had to migrate which wanted to use `assert_eq` and otherwise I had to do a convoluted TryWriteable call site. I would like to land this and debate the API later.

sffc:
It's mildly concerning that the issue in bd0affb434930c0c204b2b0576b98c24e0e65228 was only found by a tutorials test that happened to exercise the data markers incorrectly classified as "experimental"

Manishearth (approved):

PR #5546, opened 2024-09-16
Don't use input traits in neo

robertbastian:
We have `NeoGetField` for this. 

Also adds `format_timezone`.

### DIFF
--- a/components/datetime/src/format/datetime.rs
+++ b/components/datetime/src/format/datetime.rs
@@ -3,7 +3,7 @@
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
 use crate::fields::{self, Field, FieldLength, FieldSymbol, Second, Week, Year};
-use crate::input::{DateInput, ExtractedDateTimeInput, ExtractedTimeZoneInput, IsoTimeInput};
+use crate::input::ExtractedInput;
 use crate::pattern::runtime::{PatternBorrowed, PatternMetadata};
 use crate::pattern::{
     runtime::{Pattern, PatternPlurals},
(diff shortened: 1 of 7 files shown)


components/datetime/src/format/datetime.rs
```diff
@@ -67,7 +67,7 @@ use writeable::{Part, Writeable};
 /// ```
 #[derive(Debug, Copy, Clone)]
 pub struct FormattedDateTime<'l> {
```
sffc, line 69: Note: `FormattedDateTime` is being deleted in #5543. But, it looks like this PR is mostly independent of that.

components/datetime/src/format/neo.rs
```diff
@@ -2217,13 +2213,21 @@ where
     ///     "The date is: November 20, 1700 Anno Domini"
     /// );
     /// ```
-    pub fn format_date<T>(&self, datetime: &'a T) -> FormattedDateTimePattern<'a>
+    pub fn format_date<I>(&self, datetime: &'a I) -> FormattedDateTimePattern<'a>
```
sffc, line 2216: This function is obsolete and should be deleted. Well, it should be replaced with a format function that binds on all of the input markers of `R`, like in NeoFormatter.

sffc (dismissed):

components/datetime/src/format/neo.rs
robertbastian: How is this obsolete but `format_time` is not?

components/datetime/src/format/datetime.rs
robertbastian: I had this on top of #5543 but I don't want to block on it because it doesn't seem ready.

sffc: I wanted to finish it on Friday but ran out of time... now it will be bitrot. Oh well :)

sffc (approved):
Not the final solution, but an incrementally positive change

components/datetime/src/format/neo.rs
robertbastian: I see a use case for these, you have a type that implements time fields, but you only want to format the date.

sffc: The fields that get formatted are determined by the pattern you feed into this type, though.

PR #5549, opened 2024-09-17
`TimeZone` + `ResolvedTimeZone`

robertbastian:
#5533


components/timezone/src/time_zone.rs
```diff
@@ -21,123 +20,94 @@ use tinystr::TinyAsciiStr;
...
+    /// assert_eq!(tz1.offset().map(UtcOffset::offset_seconds), Some(7200));
+    /// assert_eq!(tz2.offset().map(UtcOffset::offset_seconds), Some(-9000));
+    /// assert_eq!(tz3.offset().map(UtcOffset::offset_seconds), Some(9000));
     /// ```
     ///
-    /// [`CustomZonedDateTime::try_iso_from_str`]: crate::CustomZonedDateTime::try_iso_from_str
+    /// [`ZonedDateTime::try_iso_from_str`]: crate::ZonedDateTime::try_iso_from_str
     #[cfg(feature = "compiled_data")]
     #[inline]
     pub fn try_from_str(s: &str) -> Result<Self, UnknownTimeZoneError> {
```
sffc, line 113: Thought: We should support parsing `-0500[America/Chicago]`

components/datetime/src/neo_marker.rs
```diff
@@ -1869,11 +1884,11 @@ macro_rules! impl_zone_marker {
...
-        /// let zone = CustomTimeZone::from_parts(
-        ///     -40, // offset eighths of hour
-        ///     tinystr!(8, "uschi"), // time zone ID
-        ///     tinystr!(4, "amce"), // metazone ID
-        ///     tinystr!(2, "dt"), // zone variant: daylight time
+        /// let zone = TimeZone::new("-05:00".parse().unwrap(), TimeZoneBcp47Id(tinystr!(8, "uschi")))
+        ///     .to_formattable_at(
+        ///         &Default::default(),
+        ///         &Default::default(),
+        ///         &icu_calendar::DateTime::try_new_iso_datetime(2024, 7, 1, 0, 0, 0).unwrap()
```
sffc, line 1891: Issue: this new code is less readable than the old code, which was already not very readable.

#5554

components/timezone/src/zoned_datetime.rs
```diff
@@ -4,35 +4,108 @@
...
+            time: self.time,
+            zone: self.zone,
+        }
+    }
+
+    /// Convert the [`ZonedDateTime`] to a [`FormattableZonedDateTime`].
+    pub fn to_formattable(
+        self,
+        metazone_calculator: &MetazoneCalculator,
+        zone_offset_calculator: &ZoneOffsetCalculator,
```
sffc, line 47: This feels obvious, but we should probably try to merge these down to one type. I can't think of any use case where someone would want to load a MetazoneCalculator and a ZoneOffsetCalculator from different data sources.

components/timezone/src/zoned_datetime.rs
```diff
@@ -4,35 +4,108 @@
...
+    pub fn to_calendar<A2: AsCalendar>(&self, calendar: A2) -> ZonedDateTime<A2> {
+        ZonedDateTime {
+            date: self.date.to_calendar(calendar),
+            time: self.time,
+            zone: self.zone,
+        }
+    }
+
+    /// Convert the [`ZonedDateTime`] to a [`FormattableZonedDateTime`].
+    pub fn to_formattable(
```
sffc, line 44: I like the data model, but I think we need to consider these conversion functions a bit more carefully.

To start, I think the functions could be

```rust
impl<A: AsCalendar> ZonedDateTime<A> {
    pub fn into_formattable(self) -> FormattableZonedDateTime<A> {
        MetazoneAndZoneOffsetCalculator::new().compute(self)
    }
}

// combined type: discuss below
impl MetazoneAndZoneOffsetCalculator {
    pub fn compute<A: AsCalendar>(&self, zdt: ZonedDateTime<A>) -> FormattableZonedDateTime<A> {
        // use this type's internal data to convert them
    }
}
```

So we have a nice, clean, ergonomic function for the common case, and a still-pretty-clean function that follows all ICU4X conventions for the dynamic data case. In general I think it has been a good call that we generally avoid functions that take dataful objects as arguments, and this lets us do that.

I'm not sure we need these functions on `TimeZone` itself, since you should convert to a `ZonedDateTime` first. However, we do need a `FormattableTimeZone`.

components/timezone/src/time_zone.rs
```diff
@@ -175,88 +141,100 @@ impl CustomTimeZone {
...
+
+    /// The BCP47 time-zone identifier
+    pub fn bcp47_id(&self) -> Option<TimeZoneBcp47Id> {
+        self.timezone.bcp47_id
+    }
+
+    /// The CLDR metazone identifier
+    pub fn metazone_id(&self) -> Option<MetazoneId> {
+        self.metazone_id
+    }
```
sffc, line 234: Issue: I do still feel it is useful/necessary to give a way to make one of these from IDs without data.

If we keep `CustomTimeZone`, it gives that low-level power for anyone who needs it.

sffc (changes requested):

components/timezone/src/zoned_datetime.rs
sffc: I think for now just keep `MetazoneCalculator` and add the zone offset data marker to its signature. It will load two data payloads. Eventually we should merge the data payloads into one.

components/timezone/src/time_zone.rs
robertbastian: We should, we already have the code through ixdtf.

components/timezone/src/time_zone.rs
robertbastian: There is already a way to make one of these IDs without data: `MetazoneId(tinystr!(_, "usce"))`. If we teach the datetime formatter to format this, our job is done. I *do not* want anyone to be able to construct the combination of `chzrh` and `usce`.

sffc: > If we teach the datetime formatter to format this

Ok.

robertbastian:
The build failures are FFI, which I can get to tomorrow.

ffi/capi/src/zoned_formatter.rs
```diff
@@ -57,20 +61,25 @@ pub mod ffi {
...
 
-        /// Formats a [`IsoDateTime`] and [`CustomTimeZone`] to a string.
+        #[cfg(feature = "compiled_data")]
+        /// Formats a [`IsoDateTime`] and [`TimeZone`] to a string.
         pub fn format_iso_datetime_with_custom_time_zone(
             &self,
             datetime: &IsoDateTime,
-            time_zone: &CustomTimeZone,
+            time_zone: &TimeZone,
+            timezone_calculator: &TimeZoneCalculator,
```
sffc, line 70: Nit: Do the same abstraction in FFI as you have in Rust, instead of passing the timezone calculator in here

components/timezone/src/calculator.rs
```diff
@@ -2,42 +2,47 @@
...
 use icu_provider::prelude::*;
 use zerovec::ule::AsULE;
 
-/// [`MetazoneCalculator`] uses data from the [data provider] to calculate metazone id.
+/// [`TimeZoneCalculator`] uses data from the [data provider] to compute [`ResolvedTimeZone`](crate::ResolvedTimeZone)s.
 ///
 /// [data provider]: icu_provider
 #[derive(Debug)]
-pub struct MetazoneCalculator {
+pub struct TimeZoneCalculator {
```
sffc, line 18: Discussion: Not sure about the name `TimeZoneCalculator`. I would expect such a name to be reserved for a TZDB calculator.

ffi/capi/src/zoned_formatter.rs
sffc: OK, I thought we were mostly finalized, but from https://github.com/unicode-org/icu4x/issues/5533 now I understand that we're not

PR #5552, opened 2024-09-18
Remove more obsolete datetime types

sffc:
#1317


sffc:
This covers all unused code outside of the `provider` module. I'll do that next but it's a bit more involved so I'll leave it for another PR.

Manishearth (approved):

PR #5553, opened 2024-09-18
Don't use 1.x baked data in icu_datetime

sffc:
#1317

Not a _super_ interesting milestone, but it's what I was able to finish today and it's landable


components/datetime/src/provider/mod.rs
```diff
@@ -46,44 +43,14 @@ const _: () = {
         pub use icu_datetime_data::icu_locale as locale;
     }
     make_provider!(Baked);
-    impl_buddhist_date_lengths_v1_marker!(Baked);
-    impl_buddhist_date_symbols_v1_marker!(Baked);
-    impl_chinese_date_lengths_v1_marker!(Baked);
-    impl_chinese_date_symbols_v1_marker!(Baked);
-    impl_coptic_date_lengths_v1_marker!(Baked);
-    impl_coptic_date_symbols_v1_marker!(Baked);
-    impl_dangi_date_lengths_v1_marker!(Baked);
```
robertbastian, line 55: If this data is not needed anymore, it should be removed from the `KEYS` array, so they don't get generated anymore. Later, the data markers should also be deleted.

Manishearth (approved):

components/datetime/src/provider/mod.rs
sffc: > Not a _super_ interesting milestone, but it's what I was able to finish today and it's landable

PR #5563, opened 2024-09-19
Bring time zone length behavior up to spec

sffc:
Time zone length works like all other fields, except that it always prefers short when used in a pattern with other fields.

It does create an edge case where the constructor can return `TypeTooNarrow`, which is shown via a docs test.


Manishearth (approved):
cumulative praise: I really appreciate how the neo design is much more uniform and easy to look at, the time zone code used to be much messier

PR #5565, opened 2024-09-20
Add content_locale member to LineBreakOptions

makotokato:
Fixes https://github.com/unicode-org/icu4x/issues/3284.


sffc (approved):

sffc:
The same Dart CI failure is on other PRs so I am inclined to ignore it.

PR #5567, opened 2024-09-20
Remove old datetime data from compiled_data and debug JSON

sffc:
#1317


sffc:
The keys-from-bin hard-coded .wasm file needs to be re-generated. Unfortunately re-generating it is not trivial and the gen.sh script isn't working for me.

Manishearth (approved):

sffc:
> The keys-from-bin hard-coded .wasm file needs to be re-generated. Unfortunately re-generating it is not trivial and the gen.sh script isn't working for me.

I'm fixing this in a follow-up PR

PR #5568, opened 2024-09-20
Update tutorial_buffer.wasm and expected data markers

sffc:
Follow-up from #5567


components/icu/src/datagen.rs
```diff
@@ -89,13 +89,22 @@ pub fn markers_for_bin(path: &Path) -> Result<HashSet<DataMarkerInfo>, DataError
...
+            crate::datetime::provider::neo::GregorianDateNeoSkeletonPatternsV1Marker::INFO,
+            crate::datetime::provider::neo::TimeNeoSkeletonPatternsV1Marker::INFO,
+            crate::datetime::provider::neo::DayPeriodNamesV1Marker::INFO,
+            crate::datetime::provider::neo::GregorianMonthNamesV1Marker::INFO,
+            crate::datetime::provider::neo::GregorianYearNamesV1Marker::INFO,
+            crate::datetime::provider::neo::WeekdayNamesV1Marker::INFO,
             crate::calendar::provider::WeekDataV1Marker::INFO,
+            crate::calendar::provider::WeekDataV2Marker::INFO,
             crate::decimal::provider::DecimalSymbolsV1Marker::INFO,
-            crate::plurals::provider::OrdinalV1Marker::INFO,
```
sffc, line 98: Seems like I have a little internal cleanup to do here: I shouldn't need both WeekDataV1 and WeekDataV2. OrdinalV1Marker was dropped and it hasn't been replaced yet, since neo datetime does not support week-of-year yet.

Manishearth (dismissed):

Manishearth (approved):

PR #5573, opened 2024-09-23
Collator preferences

robertbastian:
https://github.com/unicode-org/icu4x/issues/419

Based on #5779


components/decimal/src/lib.rs
```diff
@@ -170,3 +166,35 @@ impl FixedDecimalFormatter {
...
+impl ResolvedFixedDecimalFormatterPreferences {
+    fn as_data_id(&self) -> DataIdentifierBorrowed {
+        DataIdentifierBorrowed::for_marker_attributes_and_locale(
+            DataMarkerAttributes::from_str_or_panic(
+                self.numbering_system.0.as_str()
+            ),
+            &self.data_locale,
+        )
+    }
+}
```
sffc, line 200: Praise: I like the pattern of pre-computing the DataLocale and putting this code into an associated function.

components/decimal/src/lib.rs
```diff
@@ -170,3 +166,35 @@ impl FixedDecimalFormatter {
...
+    }
+);
+
+impl FixedDecimalFormatterPreferences<'_> {
+    /// TODO
+    pub fn resolve(self) -> ResolvedFixedDecimalFormatterPreferences {
+        #[allow(clippy::unwrap_used)] // temporary
+        ResolvedFixedDecimalFormatterPreferences {
+            data_locale: DataLocale::from(self.lid),
+            numbering_system: self.numbering_system.unwrap(),
```
sffc, line 186: Issue: it should just be stored as an Option, I think? We don't actually have a way to retrieve the resolved numbering system from data because that wasn't a design constraint until now. Maybe we should add it to DecimalSymbolsV2, #4437

utils/preferences/src/preferences.rs
```diff
@@ -24,44 +27,56 @@ pub trait PreferenceKey {
...
-                    )*
-                }
-            }
-        }
-
-        impl From<icu_locale_core::Locale> for $name {
-            fn from(loc: icu_locale_core::Locale) -> Self {
-
-                let lid = Some(loc.id);
+        impl<'a> From<&'a $crate::preferences::icu_locale_core::Locale> for $name<'a> {
```
sffc, line 77: Discussion: Do we want an impl from both `Locale` and `&Locale`?

components/collator/src/lib.rs
```diff
@@ -293,3 +293,64 @@ pub use options::MaxVariable;
...
+}
+
+use icu_preferences::extensions::unicode::keywords::CollationType;
+use icu_provider::DataIdentifierCow;
+use icu_provider::DataMarkerAttributes;
+
+impl<'a> From<&'a ResolvedCollatorPreferences> for DataIdentifierCow<'a> {
+    fn from(value: &'a ResolvedCollatorPreferences) -> Self {
+        Self::from_borrowed_and_owned(
+            match value.collation_type {
```
sffc, line 331: Question: should this match live here or in `icu_preferences` code?

utils/preferences/src/preferences.rs
robertbastian: Please post discussions in the issue

components/locale_core/src/preferences/extensions/unicode/keywords/collation.rs
```diff
@@ -40,3 +40,26 @@ enum_keyword!(
         /// Pinyin ordering for Latin, zhuyin order for Bopomofo and CJK characters (used in Chinese)
         ("zhuyin" => Zhuyin),
 }, "co");
+
+impl CollationType {
+    #[doc(hidden)]
+    pub fn as_str(self) -> &'static str {
```
robertbastian, line 46: @zbraniecki can this be generated? It's not trivial because of nesting, but I need this to create an allocation-free `DataMarkerAttributes`.

zbraniecki (dismissed):

components/locale_core/src/preferences/extensions/unicode/keywords/collation.rs
zbraniecki: Added in #5788

zbraniecki (dismissed):

PR #5589, opened 2024-09-24
Hide `icu::properties::runtime::*` again

robertbastian:


Manishearth (approved):

Manishearth (commented):
not a huge fan of always having the parse step but I think this is fine for now

Manishearth (changes requested):
Actually, no. I recall a major use case in actual code for this stuff was parsing a property and figuring out whether it is binary or enumerated, which dictates what code you use after that. Having a runtime way of talking about which property you have (which doesn't require parsing multiple times) is useful for that.

robertbastian:
This restores what is currently launched.

Manishearth (approved):

Manishearth (approved):

robertbastian:
#3573

PR #5590, opened 2024-09-24
Add `icu_provider::marker::ErasedMarker`

robertbastian:
Using an erased marker for storage is a common pattern across multiple components. Also removed some unnecessary `cast()`s in datetime.


sffc (commented):
LGTM

PR #5596, opened 2024-09-25
Deduplicating metazone data structs

robertbastian:
And cleaning up datagen.


provider/source/src/time_zones/convert.rs
```diff
@@ -40,500 +32,551 @@ fn type_fallback(zone_format: &ZoneFormat) -> Option<&String> {
...
-    let mut meta_zone_ids = BTreeMap::new();
-    for (meta_zone_id, meta_zone_id_data) in meta_zone_ids_resource.iter() {
-        meta_zone_ids.insert(meta_zone_id_data.long_id.to_string(), *meta_zone_id);
-    }
-    // TODO(#1781): Remove this special case once the short id is updated in CLDR
-    meta_zone_ids.insert("Yukon".to_owned(), MetazoneId(tinystr!(4, "yuko")));
-    meta_zone_ids
-}
-
-impl From<CldrTimeZonesData<'_>> for TimeZoneFormatsV1<'static> {
```
sffc, line 111: Nit: I prefer the coding style of `impl From<CldrStruct> for Icu4xStruct` because

1. The impl can be used even if not loading data; for example, the processing of some other key might depend on TimeZoneFormatsV1, and it shouldn't need to invoke `impl DataProvider` in order to get the struct. This happens a lot in datetime, for example. I already call `From` impls in some places that need data from elsewhere.
2. More focused code purpose. Keeping `impl DataProvider` small seems like a benefit.
3. Less indentation

provider/source/src/time_zones/mod.rs
```diff
@@ -5,106 +5,143 @@
...
-                        .dates("gregorian")
-                        .read_and_parse(req.id.locale, "timeZoneNames.json")?;
-
-                    let time_zone_names_resource = &resource.main.value.dates.time_zone_names;
+// TODO: cache these
+impl SourceDataProvider {
+    /// Returns a map from time zone long identifier to time zone BCP-47 ID.
+    ///
+    /// For example: "America/Chicago" to "uschi"
+    fn compute_bcp47_tzids_btreemap(&self) -> Result<BTreeMap<String, TimeZoneBcp47Id>, DataError> {
```
sffc, line 24: Issue: I really prefer this function working the way it had worked before.

You're now loading `timezone.json` multiple times, sometimes up to 3 times, at some call sites. It also makes this function do more than it is stated to do: it's now "load _and_ compute bcp47 tzids btreemap"

I acknowledge that there are some call sites that load `timezone.json` for only the purpose of invoking this function. I would be okay with a private function on SourceDataProvider that returns the timezone.json resource, which you can then pass down into this function. For example, you can change an existing call site like this:

```rust
// current
let resource: &cldr_serde::time_zones::bcp47_tzid::Resource =
    self.cldr()?.bcp47().read_and_parse("timezone.json")?;
let iana2bcp = &compute_bcp47_tzids_btreemap(&resource.keyword.u.time_zones.values);

// proposed
let resource = self.get_timezone_resource()?;
let iana2bcp = &compute_bcp47_tzids_btreemap(&resource.keyword.u.time_zones.values);
```

sffc (changes requested):
This PR isn't what I was expecting when I saw the title (I was hoping for the one that merges metazone and zone variant calculator) 😉 

Merging the data structs seems like a good idea. I wanted to approve this PR, but unfortunately I don't think the changes you made in `icu_provider_source` are an improvement. If you make those changes more minimal, I'll approve the PR.

I don't think we previously discussed this for 2.0, though. I'd like to focus on 2.0 tasks.

robertbastian:
They are definitely an improvement, because loading five JSON files and then just using one is stupid. This also lets us cache the btreemaps in a follow-up.

I consider this part of date time cleanup, which is 2.0.

sffc:
> They are definitely an improvement, because loading five JSON files and then just using one is stupid. This also lets us cache the btreemaps in a follow-up.

Huh? Where were we loading five JSON files and then just using one?

sffc:
Ok, I see, `CldrTimeZonesData` is not a good abstraction.

I think my other two comments still stand (keep conversion code and loading code separate) but I understand better now why you wanted to make these changes.

provider/source/src/time_zones/mod.rs
robertbastian: Abstracting this in a method lets us cache this in the future. I also don't want to repeat the JSON loading everywhere I need the BTreeMap, that's just code duplication.

provider/source/src/time_zones/convert.rs
robertbastian: > The impl can be used even if not loading data; for example, the processing of some other key might depend on TimeZoneFormatsV1, and it shouldn't need to invoke impl DataProvider in order to get the struct. This happens a lot in datetime, for example. I already call From impls in some places that need data from elsewhere.

If that happens we can factor out such code. At the moment there is already shared code that is factored out, but code specific to one data marker isn't.

> Keeping impl DataProvider small seems like a benefit.

Disagree. The data provider implementation should compute the data struct without tons of indirection. The previous code that used a macro and `From` implementations was hard to follow and debug, because cross-references don't work.

younies (commented):
Looks good, forwarding to @sffc for approval.

sffc (approved):
I don't agree with some of the coding decisions but I like consolidating the data structs.

provider/source/src/time_zones/mod.rs
sffc: I don't think we should make a separate method until we actually have caching in place, because in the interim it is less efficient because you need to re-load and re-parse the JSON multiple times.

provider/source/src/time_zones/convert.rs
sffc: DataProvider impls on SourceDataProvider should be short and focused: parse the locale, load the JSON it needs, call some conversion function, and return the payload. I find it hard to read SourceDataProvider code whose functions are multi-purpose, long, and deeply nested (10+ levels of indentation!). I prefer if procedural code (data provider impls) and logical code (map from JSON to ICU4X data struct) are separate functions. This should be the default, not only in cases where the code can be shared.

robertbastian: Map from JSON to data struct breaks down when the data struct mixes data from x JSON files plus TZDB. Defining intermediate structs like this had before is definitely worse than what I have implemented.

provider/source/src/time_zones/mod.rs
robertbastian: It was a separate method before, it's a separate method now. The JSON isn't parsed multiple times, it's cached.

PR #5598, opened 2024-09-25
CLDR 46 Beta2

robertbastian:


provider/source/tests/data/cldr/cldr-units-full/main/ar/units.json
```diff
@@ -1920,14 +1920,8 @@
...
           "speed-light-speed": {
-            "gender": "feminine",
-            "displayName": "سرعة الضوء",
-            "unitPattern-count-zero": "سرعة الضوء {0}",
-            "unitPattern-count-one": "سرعة الضوء {0}",
-            "unitPattern-count-two": "سرعة الضوء {0}",
-            "unitPattern-count-few": "سرعة الضوء {0}",
-            "unitPattern-count-many": "سرعة الضوء {0}",
-            "unitPattern-count-other": "سرعة الضوء {0}"
+            "displayName": "light",
```
robertbastian, line 1923: light speed broken in many languages @sffc @srl295

components/timezone/src/windows_tz.rs
robertbastian, line 1: @nekevss FYI

sffc (dismissed):
Weird that light speed is broken. Seems like something CLDR should fix before the release

provider/source/tests/data/cldr/cldr-units-full/main/ar/units.json
sffc: Confirmed that this is apparently WAI (this is a new item, and CLDR rolled back the localized data which was translated incorrectly across the board and they didn't have time to replace it with new data)

sffc (approved):

macchiati:
I'll add a note about this to the release page. Examining the data, many of
the translators misunderstood what was required, and so we ended up
dropping the data for some locales.

PR #5606, opened 2024-09-27
In zerovec, skip `check_sizes()` test on non-64-bit arches

### LINKED ISSUE #5605: zerovec: one test fails on 32-bit platforms
On 32-bit platforms, e.g. `i686`, one `zerovec` tests fails:

```
failures:
---- tests::check_sizes stdout ----
thread 'tests::check_sizes' panicked at src/lib.rs:541:9:
ZeroVec<u8> is of size 12
failures:
    tests::check_sizes
test result: FAILED. 64 passed; 1 failed; 1 ignored; 0 measured; 11 filtered out; finished in 0.00s
error: test failed, to rerun pass `--lib`
```

Looking at the `check_sizes()` tests, some but not all `check_size_of!()` invocations have multiple expected sizes,

https://github.com/unicode-org/icu4x/blob/7634c7e8a0c9cf1bf0f47d9f75588e5ddd73873d/utils/zerovec/src/lib.rs#L538-L555

presumably to cover different Rust versions as described in

https://github.com/unicode-org/icu4x/blob/7634c7e8a0c9cf1bf0f47d9f75588e5ddd73873d/utils/zerovec/src/lib.rs#L525-L526

but these do *not* appear to try to cover 32-bit architectures. Rather than adding the actual sizes for all of these on `i686`, I propose just disabling this test on 32-bit architectures by guarding it with `#[cfg(target_pointer_width = "64")]`. I will open a PR implementing this suggestion. (However, if you really prefer to maintain a list of expected sizes on 32-bit platforms, that would work too!)

musicinmybrain:
Since the expected sizes are based on 64-bit architectures, we need to either maintain additional expected sizes for 32-bit architectures, or (as implemented in this commit) just skip this test on architectures that are not 64-bit.

Fixes #5605.

### DIFF
--- a/utils/zerovec/src/lib.rs
+++ b/utils/zerovec/src/lib.rs
@@ -518,6 +518,8 @@ pub use zerovec_derive::make_ule;
 pub use zerovec_derive::make_varule;
 
 #[cfg(test)]
+// Expected sizes are based on a 64-bit architecture
+#[cfg(target_pointer_width = "64")]
 mod tests {
     use super::*;
     use core::mem::size_of;


Manishearth (approved):

sffc (commented):
Ok for now; we might want to consider changing these to check "size in number of words" instead of a byte count

PR #5619, opened 2024-09-30
More `foo32` consistency

robertbastian:
#2413

### DIFF
--- a/components/properties/src/code_point_set.rs
+++ b/components/properties/src/code_point_set.rs
@@ -142,17 +142,7 @@ impl<'a> CodePointSetDataBorrowed<'a> {
         self.set.contains(ch)
     }
 
-    /// Check if the set contains a character as a UTF32 code unit
-    ///
-    /// ```rust
-    /// use icu::properties::CodePointSetData;
-    /// use icu::properties::props::Alphabetic;
-    ///
-    /// let alphabetic = CodePointSetData::new::<Alphabetic>();
-    ///
-    /// assert!(!alphabetic.contains32(0x0A69));  // U+0A69 GURMUKHI DIGIT THREE
-    /// assert!(alphabetic.contains32(0x00C4));  // U+00C4 LATIN CAPITAL LETTER A WITH DIAERESIS
-    /// ```
+    /// See [`Self::contains`].
     #[inline]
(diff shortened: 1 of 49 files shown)


Manishearth (dismissed):

components/properties/src/code_point_set.rs
```diff
@@ -142,17 +142,7 @@ impl<'a> CodePointSetDataBorrowed<'a> {
...
-    /// ```rust
-    /// use icu::properties::CodePointSetData;
-    /// use icu::properties::props::Alphabetic;
-    ///
-    /// let alphabetic = CodePointSetData::new::<Alphabetic>();
-    ///
-    /// assert!(!alphabetic.contains32(0x0A69));  // U+0A69 GURMUKHI DIGIT THREE
-    /// assert!(alphabetic.contains32(0x00C4));  // U+00C4 LATIN CAPITAL LETTER A WITH DIAERESIS
-    /// ```
+    /// See [`Self::contains`].
```
sffc, line 145: Suggestion, here and elsewhere: Previously the docs said "as a UTF32 code unit" but now it doesn't say that any more.
```suggestion
    /// Returns whether the set contains a code point represented as a UTF-32 code unit.
    ///
    /// For an example, see [`Self::contains`].
```

robertbastian: A `u32` is not a UTF-32 code unit though, this was incorrect.

sffc: Huh?

It's important to say that the `u32` is a UTF-32 code unit. Otherwise, I don't know what it is: it could be a zero-padded ASCII, a Latin1 byte, a TinyStr, …

sffc (changes requested):
^

components/properties/src/code_point_set.rs
sffc: If you're bothered by "not all u32s are code points", you could stick our agreed-upon term "potential" in there somewhere

robertbastian: Please just approve this PR as this is blocking another PR. I can improve docs for all `32` methods in a follow-up.

sffc (approved):

PR #5620, opened 2024-09-30
Impl Serde for PackedPatternsV1 with packed-like human form

sffc:
Replaces #5592

This PR does _not_ export `impl Deserialize for PluralElements`, which was going to prevent #5592 from landing. However, if we ever add plural packed patterns, we're going to need _something_ like that.

(note on my coding velocity: this PR took me just over 3 hours to write, which I consider to be a lot of time for a serde impl. the previous PR took less than an hour.)

### DIFF
--- a/components/datetime/Cargo.toml
+++ b/components/datetime/Cargo.toml
@@ -87,6 +87,7 @@ datagen = [
     "dep:databake",
     "dep:litemap",
     "icu_calendar/datagen",
+    "icu_plurals/datagen",
     "icu_timezone/datagen",
     "serde",
     "std",
--- a/components/datetime/src/pattern/runtime/pattern.rs
+++ b/components/datetime/src/pattern/runtime/pattern.rs
@@ -108,6 +108,10 @@ impl<'data> Pattern<'data> {
             metadata: self.metadata,
         }
     }
+
+    pub fn as_ref(&self) -> Pattern<'_> {
+        self.as_borrowed().as_pattern()
+    }
 }
 
 impl PatternULE {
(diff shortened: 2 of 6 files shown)


sffc:
Review note: the last commit, 674c0b4, re-writes a lot of what I wrote in d679710. You may want to review both commits at the same time.

Manishearth (dismissed):

Manishearth (approved):

robertbastian (approved):
Is there any context for this? I don't understand why you are spending 3h on a pretty serde impl on an unused type that does not affect testdata.

sffc:
> Is there any context for this? I don't understand why you are spending 3h on a pretty serde impl on an unused type that does not affect testdata.

It will replace the thing currently called `PackedSkeletonDataV1` but I can't switch it out until all the impls are landed.

sffc:
I spent 3h on a pretty serde impl because my first attempt added API surface we didn't want to add, so I had to write a manual serde impl. It has been important for the datetime patterns to be human-readable in JSON, and they wouldn't have been human-readable if I had used the default human-readable serialization of the machine struct.

robertbastian:
> It will replace the thing currently called PackedSkeletonDataV1 but I can't switch it out until all the impls are landed.

Can you link an issue?

sffc:
> > It will replace the thing currently called PackedSkeletonDataV1 but I can't switch it out until all the impls are landed.
> 
> Can you link an issue?

It's done now. https://github.com/unicode-org/icu4x/pull/5648

PR #5627, opened 2024-10-01
Rename `UnicodeSetData` to `EmojiSetData`

robertbastian:
https://github.com/unicode-org/icu4x/issues/3573

#5588


Manishearth (approved):

tools/make/diplomat-coverage/src/allowlist.rs
```diff
@@ -327,6 +330,8 @@ lazy_static::lazy_static! {
         "icu::properties::CodePointMapDataBorrowed::static_to_owned",
         "icu::properties::CodePointSetData::as_borrowed",
         "icu::properties::CodePointSetDataBorrowed::static_to_owned",
```
sffc, line 332: replace instead of add?

sffc (approved):

Issue #4716, opened 2024-03-20
DatagenDriver::export should return a results bag


sffc:
Currently the function prints a bunch of stuff to the console and returns `()`. It should return an object with information about the job that completed (still logging the text for CLI users). For example:

1. Resolved options for DeduplicationStrategy
2. Information about the locales that were included or which failed to include
3. Exporter-specific stats like how many bytes were written (requires adding an associated type to DataExporter)

What needs to be done in 2.0:

1. Changing the signature of the `.export` function
2. Adding the associated type to DataExporter

What could be done sooner:

1. Make a new `.export_with_results` function
2. Resolved options and locale inclusion info

CC @robertbastian @zbraniecki

robertbastian:
I've done this without associated types for now, because we use `dyn DataExporter`.

PR #5629, opened 2024-10-01
Add `#[non_exhaustive]` return types to `DataExporter::close` and `ExportDriver::export`

robertbastian:
#4716, #3746


sffc (commented):
Thanks

Issue #5545, opened 2024-09-16
Rename TimeZoneFormatsV1 to TimeZoneEssentialsV1 and migrate to icu_pattern


sffc:
TimeZoneFormatsV1 has unused fields and it uses runtime string replacement instead of icu_pattern. This should be fixed. Shouldn't take long and it will be a pain to do it post-2.0.

sffc:
Also look into potentially-unused fields and whether we need them. https://github.com/unicode-org/icu4x/issues/4920

PR #5637, opened 2024-10-02
Use pattern in `TimeZoneEssentialsV1`

robertbastian:
#5545


provider/source/src/time_zones/convert.rs
```diff
@@ -49,32 +51,29 @@ impl DataProvider<TimeZoneFormatsV1Marker> for SourceDataProvider {
             .time_zone_names;
 
         fn parse_hour_format(hour_format: &str) -> (Cow<'static, str>, Cow<'static, str>) {
-            // e.g. "+HH:mm;-HH:mm" -> ("+HH:mm", "-HH:mm")
+            // e.g. "+HH:mm;-hh:mm" -> ("+H:m", "-h:m")
```
sffc, line 54: Thought: It makes sense what you did here, but lowercase "h" generally means 12-hour time, but that doesn't make sense for GMT offset formatting. This mapping you chose seems okay, though nonstandard.

sffc (approved):
Good work!

I opened a follow-up https://github.com/unicode-org/icu4x/issues/5638 to investigate whether we can support more use cases with the hours format.

provider/source/src/time_zones/convert.rs
robertbastian: `Hhm` are the only characters I'm sure are not used by patterns, so I went with this.

sffc: It would be slightly cleaner to use `MultiNamedPlaceholderPattern` for these, although it's slightly more data.

PR #5640, opened 2024-10-02
Delete more dead fixtures

robertbastian:


components/datetime/tests/fixtures/tests/components_with_zones.json
```diff
@@ -4,16 +4,6 @@
...
-                "components": {
-                    "weekday": "long",
-                    "month": "long",
-                    "day": "numeric-day-of-month",
-                    "year": "numeric",
-                    "hour": "two-digit",
-                    "minute": "two-digit",
-                    "second": "two-digit",
-                    "time_zone_name": "long-generic"
-                },
```
sffc, line 16: So, I had mixed feelings about deleting these, which is why I hadn't done it yet. I do think it is potentially useful, if we ever add classical skeletons back again, to have these fixtures, even though they are dead right now.

components/datetime/tests/fixtures/tests/components_with_zones.json
```diff
@@ -138,31 +88,5 @@
                 "en": "Tuesday, January 21, 2020, 08:25:07 GMT+05:00"
             }
         }
-    },
-    {
-        "description": "Date time example that includes a week-of for a locale with plural variants as well as one with non default WeekData",
```
sffc, line 143: Even though week-of-year is dead right now, it should be added back in a future release. #5643

PR #5641, opened 2024-10-02
Add short offset, specific location, and generic partial location formats

### LINKED ISSUE #5638: Support short localized GMT format and GMT format with seconds
https://unicode.org/reports/tr35/tr35-dates.html#Using_Time_Zone_Names says:

> **Localized GMT format:** A constant, specific offset from GMT (or UTC), which may be in a translated form. There are two styles for this. The first is used when there is an explicit non-zero offset from GMT; this style is specified by the <gmtFormat> element and <hourFormat> element. The long format always uses 2-digit hours field and minutes field, with optional 2-digit seconds field. The short format is intended for the shortest representation and uses hour fields without leading zero, with optional 2-digit minutes and seconds fields. The digits used for hours, minutes and seconds fields in this format are the locale's default decimal digits:
>
> - "GMT+03:30" (long)
> - "GMT+3:30" (short)
> - "UTC-03.00" (long)
> - "UTC-3" (short)
> - "Гринуич+03:30" (long)

However, currently we only support the "long" format, and we don't support offsets with hours only (the spec doesn't say how to do that: presumably use FixedDecimalFormatter directly??) or hours, minutes, and seconds (although there are no examples of this in the spec, and the data has only hours and minutes).

2.0 to check if we need to add data to TimeZoneEssentialsV1. Otherwise, 2.x priority because ECMAScript will need this.

CC @robertbastian

robertbastian:
Specific location and generic partial location don't have a symbol as far as I can tell, so they're not useable right now.

Fixes #5638


components/datetime/src/raw/neo.rs
```diff
@@ -500,7 +500,7 @@ impl ZonePatternSelectionData {
         components: NeoTimeZoneSkeleton,
         is_only_field: bool,
     ) -> Self {
-        let length = if is_only_field {
+        let length = if is_only_field || components.style == NeoTimeZoneStyle::Offset {
```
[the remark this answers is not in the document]
sffc: Yeah, this should be commented better. In a datetime format string, a time zone should generally use the short form if available, but as a standalone field, it should use the inherited length.

I think that should apply to Localized GMT, too? This corner of the spec is a little fuzzy and may change in the future, but I don't see why we should make Localized GMT have different behavior.

components/datetime/tests/fixtures/tests/components_with_zones.json
```diff
@@ -124,7 +124,7 @@
                     "hour": "two-digit",
                     "minute": "two-digit",
                     "second": "two-digit",
-                    "time_zone_name": "offset"
+                    "time_zone_name": "long-offset"
                 },
                 "semantic": {
```
sffc, line 129: In these cases, I suggest updating the test expectation instead of the inputs.

components/datetime/src/format/datetime.rs
```diff
@@ -712,174 +714,160 @@ where
...
     match time_zone {
         // `z..zzz`
-        ResolvedNeoTimeZoneSkeleton::SpecificShort => {
-            formatters.0 = Some(TimeZoneFormatterUnit::SpecificNonLocationShort(
-                SpecificNonLocationShortFormat {},
-            ));
-        }
+        ResolvedNeoTimeZoneSkeleton::SpecificShort => [
+            Some(TimeZoneFormatterUnit::SpecificNonLocationShort),
+            Some(TimeZoneFormatterUnit::LocalizedOffsetLong),
```
sffc, line 721: UTS 35 says that Specific Short should go to Short Localized GMT

https://unicode.org/reports/tr35/tr35-dates.html#goals

components/datetime/src/format/datetime.rs
```diff
@@ -712,174 +714,160 @@ where
...
-                GenericNonLocationShortFormat {},
-            ));
-            formatters.1 = Some(TimeZoneFormatterUnit::GenericLocation(
-                GenericLocationFormat {},
-            ));
-        }
+        ResolvedNeoTimeZoneSkeleton::GenericShort => [
+            Some(TimeZoneFormatterUnit::GenericNonLocationShort),
+            Some(TimeZoneFormatterUnit::GenericLocation),
+            Some(TimeZoneFormatterUnit::LocalizedOffsetLong),
```
sffc, line 734: UTS 35 says that Generic Short should go to Short Localized GMT (after location, which should usually not fail)

https://unicode.org/reports/tr35/tr35-dates.html#goals

components/datetime/src/options/components.rs
```diff
@@ -578,15 +578,15 @@ pub enum TimeZoneName {
...
-    //   https://github.com/unicode-org/cldr-json/blob/c23635f13946292e40077fd62aee6a8e122e7689/cldr-json/cldr-dates-full/main/es-MX/timeZoneNames.json#L13
-    /// Localized offset format, in the locale's preferred hour format. (e.g., GMT-0800),
-    Offset,
+    /// Long localized offset form, e.g. GMT-08:00
+    LongOffset,
+
+    // UTS-35 fields: O
+    // Per UTS-35: Short localized offset format
+    /// Short localized offset form, e.g. GMT-8
+    ShortOffset,
```
sffc, line 589: Note: this is components bag code, which is behind `#[cfg(feature = "datagen")]`

components/datetime/src/provider/date_time.rs
```diff
@@ -232,6 +232,6 @@ pub(crate) trait ZoneSymbols<'data> {
 
 impl<'data> ZoneSymbols<'data> for () {
     fn get_payloads(&self) -> TimeZoneDataPayloadsBorrowed<'data> {
-        TimeZoneDataPayloadsBorrowed::empty()
+        TimeZoneDataPayloadsBorrowed::default()
```
sffc, line 235: Nit: I prefer `::empty()` since "default" could mean either with-compiled-data or empty.

sffc (approved):

components/datetime/tests/fixtures/tests/components_with_zones.json
robertbastian: I had to split the offset field into `OffsetLong` and `OffsetShort`, so I had to change this

components/datetime/src/provider/date_time.rs
robertbastian: If there's a default call in a method that's not marked `compiled_data`, it cannot be compiled data. This is derivable so it simplifies the code.

components/datetime/src/options/components.rs
robertbastian: Ok? Should I not touch this?

components/datetime/src/format/datetime.rs
robertbastian: Done

components/datetime/src/format/datetime.rs
robertbastian: Done

components/datetime/src/raw/neo.rs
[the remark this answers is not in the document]
robertbastian: Ok

components/datetime/src/format/neo.rs
```diff
@@ -2002,24 +2002,41 @@ impl<R: DateTimeNamesMarker> RawDateTimeNames<R> {
             let field = match item {
                 FieldForDataLoading::Field(field) => field,
                 FieldForDataLoading::TimeZone(time_zone) => {
-                    self.load_time_zone_essentials(zone_essentials_provider, locale)?;
                     match time_zone {
                         // `z..zzz`
                         ResolvedNeoTimeZoneSkeleton::SpecificShort => {
+                            self.load_time_zone_essentials(zone_essentials_provider, locale)?;
```
sffc, line 2008: Suggestion: make a bool `needs_time_zone_essentials` and then have just one call site into `load_time_zone_essentials`.

components/datetime/src/time_zone.rs
```diff
@@ -745,59 +853,3 @@ impl FormatTimeZone for Bcp47IdFormat {
...
-/// Formats a time segment with optional zero-padding.
-fn format_time_segment<W: writeable::PartsWrite + ?Sized>(
-    sink: &mut W,
-    n: u8,
-    padding: ZeroPadding,
-) -> fmt::Result {
-    debug_assert!((0..60).contains(&n));
-    if padding == ZeroPadding::On && n < 10 {
-        sink.write_char('0')?;
-    }
```
sffc, line 723: Praise: Thanks for using the FixedDecimalFormatter here. I didn't realize we still had code that wasn't using it.

components/datetime/tests/patterns/tests/time_zones.json
```diff
@@ -882,5 +882,48 @@
...
+        "configs": [
+          {
+            "Iso8601": [
+              "Basic",
+              "Required",
+              "Optional"
+            ]
+          }
+        ],
+        "expected": ["+0545"]
```
sffc, line 905: Praise: Thanks for following and testing the spec.

Localized GMT:

> Note: The digits should be whatever are appropriate for the locale used to format the time zone, not necessarily from the western digits, 0..9. For example, they might be from ०..९.

ISO-8601 styles:

> Note: The digits in this case are always from the western digits, 0..9.

sffc (commented):
Good except we can't land this with the unwraps in the runtime formatting code

components/datetime/src/neo.rs
```diff
@@ -675,8 +675,8 @@ where
             &<R::Z as ZoneMarkers>::GenericShortV1Marker::bind(provider),
             &<R::Z as ZoneMarkers>::SpecificLongV1Marker::bind(provider),
             &<R::Z as ZoneMarkers>::SpecificShortV1Marker::bind(provider),
-            Some(loader), // fixed decimal formatter
-            Some(loader), // week calculator
+            loader, // fixed decimal formatter
+            loader, // week calculator
```
sffc, line 679: Thought: This regresses on https://github.com/unicode-org/icu4x/issues/4340 (the loaders are supposed to be optional and we just haven't plumbed it in yet), but it's labeled 2.0-stretch. I think it's fine to change this signature for now because the final design might not look like this.

components/datetime/src/options/components.rs
sffc: It's fine. Just a note.

sffc (approved):

sffc:
Also, what is SpecificLocation? I don't see it in UTS 35

robertbastian:
Well the data is in time zone essentials: https://cldr.unicode.org/translation/time-zones-and-city-names#:~:text=regionFormat%2Dstandard. It's like generic location, but includes standard/daylight information, so it is specific location even if UTS 35 doesn't know about it.

sffc (approved):

sffc:
Interesting, so there is Generic Partial Location which is in UTS 35 but is unreachable by any of the known pattern symbols, and then there is Specific Location which is not mentioned in UTS 35 at all (and is therefore not reachable from any pattern) but is in the translator guide book.

PR #5646, opened 2024-10-03
Backport CLDR 46

robertbastian:


sffc:
Just to note, I mentioned this in chat, but if we do this, I want to consider an approach where we don't ship new data in 1.x but instead patch datagen to work with CLDR 46.

robertbastian:
Why? I've done the work to ship CLDR 46, why wouldn't we?

sffc:
A few ways we could go:

1. Ship 1.6 of all crates
2. Ship 1.5.x of data crates and datagen
3. Ship 1.5.x-cldr46 of data crates and 1.5.x of datagen only

They all have different tradeoffs.

sffc:
- @robertbastian There are 2 fixes to datagen required. Likely subtags (@sffc: might be fixed in a future beta), and the time zone name conflict ("MET" conflicts with "Mexico"). On main, I changed the data struct version. In 1.5, I just ignore MET in datagen, but then also patch the lookup code.
- @sffc We weren't reading "MET" in CLDR 45, so it seems OK that we don't read it in the updated data?
- @robertbastian Well if we say it is CLDR 46, it's wrong that we don't accept it
- @sffc Patching icu_timezone in the way you described seems harmless
- @echeran It seems just saying 1.6 is easier/cleaner
- @robertbastian Yeah
- @Manishearth I feel like this is more along the line of "bug fix" more than "feature release".
- @sffc I'm not super happy about 1.6 because it's a lot of crates to ship. I'm not super happy about 1.5.x because it's instability. I think 1.5.x-cldr46 is a decent option: people who want it can get it via lockfile.
- @robertbastian I think it should be 1.6 because it's a fairly big behavior change. For example, root Arabic uses latin digits; I had to update test cases. I don't think shipping 1.6 is that much work.
- @Manishearth I wouldn't consider Craig a normal user. People who want stability already use lockfiles; it won't _change_ automatically. If you are a "I just want something to work" client, you just want a simple solution for data updates. In other words, the situations where lockfiles automatically update are situations where they probably want the update.
- @echeran What @robertbastian said strengthens how I feel here; different data generates different behavior. It makes sense that data updates change test behaviors. On the point of Craig not being a normal user, I don't think the comparison to the unicodetools project is apt. Craig (the Conformance project) is using these as public APIs as a normal user would do.
- @sffc ICU4C versions have a 1-to-1 mapping with CLDR versions. We also currently have that invariant in ICU4X. I don't think that is an invariant we should break. I think the operation of running `cargo update` is something clients should generally always be able to run without fear of breaking things. But CLDR updates are absolutely subject to breaking things. People should opt-in via Cargo.toml updates, not just relying on lockfiles.
- @Manishearth The fact that ICU is tied to CLDR releases is something that I do not like for Rust. On safe review and audits, the argument for updating data is easier than the arugment for updating code. Regarding the Craig's use case with Conformance project, caring about specific CLDR versions is not normal. If you care about this, then you're not running `cargo update`. I want the property of being update the data separately, but I'm not against just making the data update a 1.6 version (minor version, not patch version), specifically in baked data.

Conclusion for now:

- Land @robertbastian's PR
- Don't run `cargo publish` yet

robertbastian:
@sffc or @Manishearth to admin merge

PR #5648, opened 2024-10-03
Replace PackedSkeletonDataV1 with PackedPatternsV1

sffc:
The JSON file diffs look something like this:

image

So the metadata format changes, but the patterns themselves are the same. (I spot-checked but didn't do an exhaustive check)


sffc:
Note: the baked data has 2 layers of nesting. I could probably give it its own constructor, but the impact will be less than it was with the unit display names.

`icu::datetime::provider::PackedPatternsV1 { header: 524291u32, elements: unsafe { zerovec::VarZeroVec::from_bytes_unchecked(b"\x03\0\x1C\0/\0\0\x80 \x04\0\0 \x80@\x01\0\0,\0\0 \x80\x13\x01\0\0(\x80\x12\x01\0\0)\0\x80 \x03\0\0 \x80@\x01\0\0,\0\0 \x80\x13\x01\0\x80 \x01\0\0/\x80@\x01\0\0/\x80\x13\x01") } }`

Manishearth (approved):

PR #5651, opened 2024-10-04
Merge years into a single field

sffc:
#1317


sffc:
This is mostly a mechanical change.

sffc:
More to do, but this gets us most of the way to where we want to be for spec conformance with field sets. I think far enough for 2.0-beta purposes.

robertbastian (approved):

sffc:
The `datagen` commit is not due to the just-pushed change in ef110e0; it is due to earlier commits. Testdata had been regenerated in cf518a2, but baked data was stale.

(if you're curious why, I wasn't able to run `cargo make bakeddata` from an airplane because I didn't have the data files downloaded yet for CLDR 46)

robertbastian (approved):

PR #5652, opened 2024-10-04
Remove panics in tz datagen

robertbastian:


provider/source/src/time_zones/convert.rs
```diff
@@ -407,67 +409,67 @@ macro_rules! long_short_impls {
...
-                                                        }
-                                                    })
-                                                    .collect::<Vec<_>>(),
-                                            }
-                                        })
+                                .filter_map(|(key, metazone)| {
+                                    metazone.$field.as_ref().and_then(type_fallback).and_then(
+                                        |format| {
+                                            Some((
+                                                meta_zone_id_data.get(key)?.clone(),
```
sffc, line 422: Question: this one now drops metazones instead of panicking? And similar elsewhere?

sffc (approved):
It seems fine, but I have questions:

1. What problem is it fixing? We're generally strict in datagen, and this appears to make us less strict. (Please understand my perspective: you send me a 200-line PR with no description and no context, and I am supposed to make sense of it.)
2. Why is there a diff in the baked data but not the JSON data? Presumably there's some locale somewhere that is impacted by this?

sffc:
(I approve because I trust your judgement here, but I would like answers to my questions)

provider/source/src/time_zones/convert.rs
robertbastian: Yes

robertbastian:
This is meant to set the stage for one or two PRs that don't work with the panicking implementation.

This seems to impact some basic locales.

Issue #5653, opened 2024-10-04
Consider not using `Etc/Unknown` time zone


robertbastian:
The `Etc/Unknown` timezone is not a TZDB timezone. It was invented by CLDR to signal the absence of time zone information, and this is causing confusion when surfaced to users: https://www.google.com/search?q=etc%2Funknown, https://github.com/tc39/proposal-temporal/issues/2510. 

Using `Etc/Unknown` means that every API that handles a time zone input has to special-case this input, which is easy to miss. A more solid API would use `Option<TimeZoneBcp47Id>` with `None` instead of `Etc/Unknown`.

Our `TimeZoneMapper` already returns `Option<TimeZoneBcp47Id>`, and our formatter input already uses `Option<TimeZoneBcp47Id>` so we effectively have *two ways* of representing unknown time zones: `None` and `Some("unk")`. During formatting, `None` will fall back to the offset (good), and then to `{GMT+?}` (good), whereas `Some("unk")` will format as `Unknown City Daylight Time` or `heure: ville inconnue` under location formats *even when there is an offset that could be formatted*.[^1]

My concrete proposal is:
* Remove the `Etc/Unknown` key (and its alias, `Factory`) from the IANA->BCP47 mapping. 
  * Asking for the BCP47 ID for the `Etc/Unknown` IANA ID will fail, because `Etc/Unknown` is not a valid IANA ID
  * Asking for the BCP47 ID for the `Factory` IANA ID will fail because we say so (and document it)
* Remove the `unk` key from the BCP47->IANA mapping
  * It's incorrect to have it in there, because `Etc/Unknown` is not a canonical IANA ID
* Make the location formats fail if there's no exemplar city for a time zone, so that the next format can be tried. The location formats currently produce very awkward strings for the `Etc/Unknown` time zone, even when an offset could be formatted.

[^1]: I understand [UTS-35](https://unicode.org/reports/tr35/tr35-dates.html#Contents:~:text=falling%20back%20to%20the%20localized%20exemplar%20city%20for%20the%20unknown%20zone%20(Etc/Unknown)%2C%20for%20example%20%22Unknown%20City%22%20for%20English) wants it like this, but I don't see why.

sffc:
@justingrant

sffc:
We've discussed Etc/Unknown a number of times in the ECMAScript context, but I'm not finding the conclusion of those conversations. 

[...]

PR #5656, opened 2024-10-08
Embrace `Etc/Unknown` timezone

robertbastian:
#5653


sffc:
(we don't have consensus on this)

robertbastian:
(that's why it's draft)

components/timezone/src/ids.rs
```diff
@@ -209,27 +211,30 @@ impl TimeZoneIdMapperBorrowed<'_> {
...
+    pub fn normalize_iana<'s>(&self, iana_id: &'s str) -> (Cow<'s, str>, TimeZoneBcp47Id) {
+        self.iana_lookup_with_normalization(iana_id, |_| {})
+            .and_then(|(trie_value, string)| {
+                let Some(bcp47_id) = self.data.bcp47_ids.get(trie_value.index()) else {
+                    debug_assert!(false, "index should be in range");
+                    return None;
+                };
+                Some((string, bcp47_id))
+            })
+            .unwrap_or((Cow::Borrowed("Etc/Unknown"), TimeZoneBcp47Id::unknown()))
```
sffc, line 237: Issue: This function should return `None` if the time zone is not found. It should never map it to `Etc/Unknown`. Canonicalization can do that, but not normalization, please.

components/timezone/src/ids.rs
```diff
@@ -248,37 +253,38 @@ impl TimeZoneIdMapperBorrowed<'_> {
...
     ///
     /// // Borrows when able:
-    /// let result = mapper.canonicalize_iana("America/Chicago").unwrap();
+    /// let result = mapper.canonicalize_iana("America/Chicago");
     /// assert_eq!(result.0, "America/Chicago");
     /// assert!(matches!(result.0, Cow::Borrowed(_)));
     ///
     /// // Unknown IANA time zone ID:
-    /// assert_eq!(mapper.canonicalize_iana("America/San_Francisco"), None);
+    /// assert_eq!(mapper.canonicalize_iana("America/San_Francisco").1, TimeZoneBcp47Id::unknown());
```
sffc, line 268: Nit: Also show the equivalence to "Etc/Unknown"

sffc (changes requested):
I'm not super happy with descending from the type system to Etc/Unknown in the ID mapper. We definitely should not do it in the normalize function. In the canonicalize function, I can see an argument that we want it to be more transparent with the TimeZoneInfo struct.

You've sold me on _most_ of your time zone design things, but I still haven't heard a compelling argument about, what's so wrong with `None` and `Some(Etc/Unknown)` co-existing with slightly different meanings?

components/timezone/src/ids.rs
robertbastian: I've made this fallible again, because `Etc/Unknown` is not a canonical zone (it's not an IANA zone at all).

sffc (approved):

Issue #996, opened 2021-08-25
Add supplemental CLDR primaryZones.json to TimeZonesProvider


nordzilla:
CLDR [primaryZones.json](https://github.com/unicode-org/cldr-json/blob/39205fdc027c935e52a268afcf01cd240f3c442d/cldr-json/cldr-core/supplemental/primaryZones.json#L7) contains

* the primary zones for a given territory

This information is vital for performing region format, which occasionally uses the country itself name when formatting the primary zone for that country.

robertbastian:
This is a datagen implementation detail and does not need to be tracked as an issue.

PR #5657, opened 2024-10-08
Fix generic location format for single-tz countries

### LINKED ISSUE #987: Account for primary zones for generic location format in TimeZoneFormat
When performing generic location format, we need to check if the current zone is the primary zone for its country and use the country name instead of the exemplar city if so.

Depends on:
* #996

robertbastian:
It's meant to say "Germany Time" instead of "Berlin time" : [UTS 35]( https://unicode.org/reports/tr35/tr35-dates.html#:~:text=For%20the%20generic%20location%20format%3A)

Fixes #987 #996


provider/source/data/debug/time_zone/primary_zones@1.json
```diff
@@ -0,0 +1,228 @@
...
+  "aedxb": "AE",
+  "afkbl": "AF",
+  "aganu": "AG",
+  "aiaxa": "AI",
+  "altia": "AL",
+  "amevn": "AM",
+  "ancur": "CW",
+  "aolad": "AO",
+  "asppg": "AS",
+  "atvie": "AT",
```
sffc, line 12: Suggestion: you can eliminate almost all of these entries by filtering based on the first two letters in the BCP-47 ID. (I assume it's still possible for a time zone to change regions due to war so we probably need to keep this table)

components/datetime/src/format/neo.rs
```diff
@@ -1809,6 +1829,28 @@ impl<R: DateTimeNamesMarker> RawDateTimeNames<R> {
...
+        P: BoundDataProvider<tz::PrimaryZonesV1Marker> + ?Sized,
+    {
+        self.primary_zones
+            .load_put(provider, Default::default(), ())
+            .map_err(|e| {
+                MaybePayloadError2::into_single_load_error(
+                    e,
+                    fields::Field {
+                        symbol: FieldSymbol::TimeZone(fields::TimeZone::UpperV),
+                        length: FieldLength::Wide,
```
sffc, line 1846: Thought: we should probably plumb the NeoSkeletonLength through to here. (seems minor enough to make it an issue to be fixed later)

components/datetime/src/time_zone.rs
```diff
@@ -598,12 +600,23 @@ impl FormatTimeZone for GenericLocationFormat {
...
+        let location = if let Some(region) = primary_zones.0.get(&time_zone_id) {
+            exemplar_cities
+                .regions
+                .get(&region.into_tinystr().to_unvalidated())
+                .unwrap_or(region.as_str())
+        } else {
+            let Some(city) = exemplar_cities.cities.get(&time_zone_id) else {
+                return Ok(Err(FormatTimeZoneError::NameNotFound));
+            };
+            city
```
sffc, line 619: Suggestion: the mapping from time zone to city/region is deterministic based on the time zone identity, right? It seems more efficient, without having to change the data layout, to just set the strings in `exemplar_cities` to be regions when they need to be regions at datagen time.

The format `VVV` just prints out the city name, though. To keep that working, maybe add a new data struct for the time zones that have a city name but whose default location name is the region. That can be a back-burner issue because `VVV` is not reachable via semantic skeletons.

sffc:
Note: there is an issue for this already, https://github.com/unicode-org/icu4x/issues/987

I had been thinking for a while that the way to fix the issue would be to just change the strings in `exemplar_cities` via datagen, but I never posted it in the issue.

components/datetime/src/time_zone.rs
robertbastian: There are two things here:

1) I do want to use the region display names data struct in the future. If we inline the values, there'll be duplicated data. 
2) The `VVV` thing. We can split the `ExemplarCitiesV1` into a base key for this format, and an extended key for the `VVV` format, but as I've realised in this PR there is *so much boilerplate* required to add a datetime key that I don't want to do it right now.

sffc: I hear you about the duplicated data, but https://github.com/unicode-org/icu4x/issues/3260 isn't really compatible with that model. (Unless we wanted to change ExemplarCities to use a similar model??)

Base and extended ExemplarCitiesV1 is what I had in mind.

Here's the key thing: most users don't want `VVV` formats. Only `VVVV` is reachable from semantic skeletons and ECMA-402. So, we should optimize for `VVVV` by stripping out the strings that aren't reachable.

This is one of our biggest non-experimental data markers (maybe actually the biggest now), so I think it's important that we are careful with it.

robertbastian: I'm happy to do that, but this will break the `VVV` format, so I'd want to remove that then.

sffc: I'll make a PR removing `VVV`.

sffc: As far as the duplicated data goes... we're not really regressing with this PR. It seems like something we could fix in 3.0 or sooner if we have a clearer picture at that time of what to do.

sffc:
Thought: After this lands, we should strip data from `vvvv` that is equivalent to the data in `VVVV` since the two fall back between each other. Maybe even do a similar optimization with `zzzz`.

sffc (commented):
Send back when ready

provider/source/data/debug/time_zone/exemplar_cities@1/en-001.json
```diff
@@ -400,21 +400,21 @@
   "uswsq": "Petersburg, Indiana",
   "usxul": "Beulah, North Dakota",
   "usyak": "Yakutat",
-  "uymvd": "Montevideo",
+  "uymvd": "Uruguay",
   "uzskd": "Samarkand",
-  "uztas": "Tashkent",
-  "vavat": "Vatican",
```
robertbastian, line 406: funny that the city is `Vatican` and the country is `Vatican City`.

sffc (approved):

Issue #5658, opened 2024-10-08
Add back support for the exemplar city format


sffc:
The format was dropped in #5659 due to data optimization for the much more common location format. It can be added back as an extended data struct that sits on top of the regular exemplar cities data struct.

PR #5659, opened 2024-10-08
Remove VVV time zone format

sffc:
Unblocks #987
See #5658

### DIFF
--- a/components/datetime/src/time_zone.rs
+++ b/components/datetime/src/time_zone.rs
@@ -20,7 +20,6 @@ use writeable::Writeable;
 /// All time zone styles that this crate can format
 #[derive(Debug, Copy, Clone)]
 pub(crate) enum ResolvedNeoTimeZoneSkeleton {
-    City,
     Location,
     GenericShort,
     GenericLong,
@@ -44,6 +43,7 @@ pub(crate) enum ResolvedNeoTimeZoneSkeleton {
     IsoXXXXX,
     // TODO:
     // `VV` "America/Los_Angeles"
+    // `VVV` "Los Angeles"
     // Generic Partial Location: "Pacific Time (Los Angeles)"
 }
 
@@ -166,6 +166,7 @@ pub(super) enum TimeZoneFormatterUnit {
     LocalizedOffsetLong,
     LocalizedOffsetShort,
     Iso8601(Iso8601Format),
+    #[allow(dead_code)]
     ExemplarCity,
     Bcp47Id,
 }
--- a/components/datetime/src/time_zone.rs
+++ b/components/datetime/src/time_zone.rs
@@ -13,8 +13,7 @@ use core::fmt;
 use fixed_decimal::FixedDecimal;
 use icu_decimal::FixedDecimalFormatter;
 use icu_timezone::UtcOffset;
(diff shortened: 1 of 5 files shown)


components/datetime/src/time_zone.rs
```diff
@@ -44,6 +43,7 @@ pub(crate) enum ResolvedNeoTimeZoneSkeleton {
     IsoXXXXX,
     // TODO:
     // `VV` "America/Los_Angeles"
+    // `VVV` "Los Angeles"
     // Generic Partial Location: "Pacific Time (Los Angeles)"
```
robertbastian, line 47: nit: remove this, it's implemented

actually, do we need the full exemplar city names for it?

components/datetime/src/time_zone.rs
```diff
@@ -166,6 +166,7 @@ pub(super) enum TimeZoneFormatterUnit {
     LocalizedOffsetLong,
     LocalizedOffsetShort,
     Iso8601(Iso8601Format),
+    #[allow(dead_code)]
     ExemplarCity,
```
robertbastian, line 170: issue: remove this and `ExemplarCityFormat`, because that is what's actually blocking

components/datetime/src/time_zone.rs
sffc: hmm, that's annoying. Maybe not though because it seems Generic Partial Location is only used when you're not the primary zone?

sffc:
I'll let you merge this @robertbastian if you conclude that it is what we need for #5657

robertbastian (approved):

components/datetime/src/time_zone.rs
robertbastian: I think it's fine, because #5657 uses the new logic for the bracketed part, i.e. it will say `Central European Time (Germany)` and `Pacific Time (Los Angeles)`.

robertbastian: It would actually be nicer to say `Pacific Time (United States)`, but I'd have to check if the metazone invariants allow that.

sffc: `Pacific Time (United States)` sounds extremely weird.

robertbastian: Ah, `Mountain Time (United States)` doesn't work.

robertbastian: > Pacific Time (United States) sounds extremely weird.

Weirder than `Pacific Time (Los Angeles)`? It's the Pacific as used in the United States, following United States DST rules. They're not Los Angeles DST rules. Does `Pacific Time (Mexico)` sound weird to you as well?

sffc: I think we're safe there because we only do the region override if the time zone is the only one in the country, so the city and region are interchangeable. Right?

robertbastian: > I think we're safe there because we only do the region override if the time zone is the only one in the country, so the city and region are interchangeable. Right?

Yes, current behaviour is safe. The proposed behaviour of `Pacific Time (Canada)` needs more work, because we need to know whether a metazone+country combination is unique. `ampa`+`US` is, `ammo`+`US` isn't.

PR #5661, opened 2024-10-08
Add fixtures for incomplete timezone inputs

robertbastian:
This tests
* no offset, no zone
* no offset, `Etc/Unknown`
* offset, no zone
* offset, `Etc/Unknown`

I'm doing this to demonstrate that the `Unknown City Time` fallback is bad, and that the distinction between `Etc/Unknown` and `None` is not very useful.

### DIFF
--- a/components/datetime/tests/patterns/tests/time_zones.json
+++ b/components/datetime/tests/patterns/tests/time_zones.json
@@ -925,5 +925,157 @@
         "expected": ["GMT +০৫:৪৫"]
       }
     ]
+  },
+  {
+    "locale": "en",
+    "datetime": "2021-07-11T12:00:00.000-07:00[Etc/Unknown]",
+    "expectations": [
+      {
+        "patterns": [
+          "z",
+          "zz",
+          "zzz",
+          "ZZZZ",
+          "O"
+        ],
+        "configs": [],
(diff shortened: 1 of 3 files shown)


components/datetime/tests/patterns/tests/time_zones.json
```diff
@@ -925,5 +925,157 @@
...
+        "expected": ["-07:00"]
+      },
+      {
+        "patterns": [
+          "v",
+          "vvvv",
+          "VVVV"
+        ],
+        "configs": [],
+        "expected": ["Unknown City Time"]
```
robertbastian, line 975: this could have fallen back to the offset

components/datetime/tests/patterns/tests/time_zones.json
```diff
@@ -925,5 +925,157 @@
...
+        "expected": ["{GMT+?}"]
+      },
+      {
+        "patterns": [
+          "v",
+          "vvvv",
+          "VVVV"
+        ],
+        "configs": [],
+        "expected": ["Unknown City Time"]
```
robertbastian, line 1051: this should be an error value, and use our `{GMT+?}` unknown tz string

younies (approved):

PR #5664, opened 2024-10-09
Rework time zone fallback

robertbastian:
This PR improves error cases. For missing inputs, the field will be formatted, as for all other missing inputs (`{x}`).


sffc (commented):
I quite like the TimeZoneFormatterUnit design. I don't understand why this PR is an improvement. It might also make it harder to support https://github.com/unicode-org/icu4x/issues/5244.

robertbastian:
> I don't understand why this PR is an improvement

It's an improvement because we finally can format consistent error messages for missing time zone fields.

sffc (approved):
Much happier with this design than the previous one; thanks

PR #5667, opened 2024-10-10
Implement `SignedFixedDecimal` and `UnsignedFixedDecimal`

younies:
Related Issues: https://github.com/unicode-org/icu4x/issues/5065


sffc:
Good start; try to stick with the names we discussed in the thread. `FixedDecimal` and `Signed<FixedDecimal>`

younies:
@sffc , @robertbastian  and @Manishearth 

@sffc, @robertbastian, and @Manishearth,

This PR still needs testing fixes, but it is currently buildable.

I’m submitting it for review to confirm if we agree on the design and approach before moving forward with the next steps.

robertbastian (commented):
I've only reviewed `utils/fixed_decimal` so far.

My overall impression is that this is duplicating too much logic, this PR adds 5k lines of code. `signed_decimal.rs` adds 1k+ loc of logic (and 2k+ loc of tests), which seem to be copy-pasted from `decimal.rs`. All of this needs to be reviewed and maintained. The idea behind `Signed<T>` was that we don't have to duplicate a lot of code and API surface.

utils/fixed_decimal/src/signed_decimal.rs
```diff
@@ -0,0 +1,3364 @@
...
+    /// ```
+    /// use fixed_decimal::SignedFixedDecimal;
+    ///
+    /// let mut dec = SignedFixedDecimal::from(42);
+    /// assert_eq!("42", dec.to_string());
+    ///
+    /// dec.multiply_pow10(3);
+    /// assert_eq!("42000", dec.to_string());
+    /// ```
+    pub fn multiply_pow10(&mut self, delta: i16) {
```
sffc, line 268: You shouldn't need `multiply_pow10` since it doesn't depend on the sign.

utils/fixed_decimal/src/signed_decimal.rs
```diff
@@ -0,0 +1,3364 @@
...
+    /// dec.round(0);
+    /// assert_eq!("0", dec.to_string());
+    /// let mut dec = SignedFixedDecimal::from_str("0.6").unwrap();
+    /// dec.round(0);
+    /// assert_eq!("1", dec.to_string());
+    /// let mut dec = SignedFixedDecimal::from_str("1.5").unwrap();
+    /// dec.round(0);
+    /// assert_eq!("2", dec.to_string());
+    /// ```
+    pub fn round(&mut self, position: i16) {
```
sffc, line 440: Thought: Rounding modes are not consistently applicable to signed vs unsigned:

| Rounding Mode | Signed? | Unsigned? |
|---|---|---|
| (Half)Trunc | Yes | Yes |
| (Half)Expand | Yes | Yes |
| (Half)Ceil | Yes | No |
| (Half)Floor | Yes | No |
| HalfEven | Yes | Yes |

Maybe define the ones that work the same on the unsigned type on only the unsigned type and the ones that only work on the signed type to be on the signed type?

sffc (commented):
Doing a quick pass, I think I see where @younies drew the line, and it seems potentially reasonable, but I tend to favor less API surface, even if it comes at the expense of some ergonomics. If you need to change a call site from `decimal.foo()` to `decimal.absolute.foo()`, I think it is fine and expected. That's the point of composable types. I don't think we should copy the methods up and down just because we can.

utils/fixed_decimal/src/integer.rs
```diff
@@ -7,42 +7,42 @@ use core::fmt;
...
 /// );
 /// assert_eq!(
-///     FixedInteger::try_from(FixedDecimal::from_str("5.0").unwrap()),
+///     FixedInteger::try_from(SignedFixedDecimal::from_str("5.0").unwrap()),
 ///     Err(LimitError)
 /// );
 /// ```
 #[derive(Debug, Clone, PartialEq, Default)]
-pub struct FixedInteger(FixedDecimal);
+pub struct FixedInteger(SignedFixedDecimal);
```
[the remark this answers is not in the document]
younies: I have added this to the design doc : https://hackmd.io/IRsC6aBDQF2aO7b-fQMpNg#Fixed-Integer-Composition-Design

utils/fixed_decimal/src/variations.rs
```diff
@@ -0,0 +1,114 @@
...
+/// Specifies the precision of a floating point value when constructing a FixedDecimal.
+///
+/// IEEE 754 is a representation of a point on the number line. On the other hand, FixedDecimal
+/// specifies not only the point on the number line but also the precision of the number to a
+/// specific power of 10. This enum augments a floating-point value with the additional
+/// information required by FixedDecimal.
+#[non_exhaustive]
+#[cfg(feature = "ryu")]
+#[derive(Debug, Clone, Copy)]
+pub enum FloatPrecision {
```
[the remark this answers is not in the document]
younies: you are right

utils/fixed_decimal/src/decimal.rs
```diff
@@ -2396,20 +2051,36 @@ impl FixedDecimal {
...
             &input_str[1..]
         };
         if no_sign_str.is_empty() {
             return Err(ParseError::Syntax);
         }
+
+        Self::try_from_no_sign_utf8(no_sign_str)
+    }
+
+    /// This function is used to parse a [`UnsignedFixedDecimal`] from a string without a sign.
```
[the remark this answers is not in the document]
younies: yes, it is in the `SignedFixedDecimal` too

utils/fixed_decimal/src/compact.rs
```diff
@@ -19,13 +19,13 @@ use crate::ParseError;
 /// nor a sign in the exponent, and behaves differently in pluralization.
 #[derive(Debug, Clone, PartialEq)]
 pub struct CompactDecimal {
-    significand: FixedDecimal,
+    significand: SignedFixedDecimal,
```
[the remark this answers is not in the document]
younies: In this PR, we are only focusing on the composition, and all the current uses of `FixedDecimal` can be mapped to `SignedFixedDecimal` because `FixedDecimal` was signed.

In the next PRs, we can see if some places in the code can use `UnsignedFixedDecimal` instead of signed.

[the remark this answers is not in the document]
robertbastian: In `ScientificDecimal` you replaced `FixedDecimal` with `UnsignedFixedDecimal` though

utils/fixed_decimal/tests/signed_rounding.rs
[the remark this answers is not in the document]
robertbastian: There is like a 10-line diff out of 400+. This should not be copied.

younies:
@sffc, the testing is still failing. I would appreciate it if you could take a quick look to ensure the design aligns with what we discussed.

sffc:
I did a quick look at the errors and they look like regular coding / import / feature issues.

@robertbastian can you help?

robertbastian (commented):
If you want a thorough rereview, please address all comments first.

ffi/capi/src/fixed_decimal.rs
```diff
@@ -49,77 +49,94 @@ pub mod ffi {
...
         HalfTrunc,
         HalfEven,
     }
 
-    impl FixedDecimal {
-        /// Construct an [`FixedDecimal`] from an integer.
+    /// Mode used in a rounding operation for signed numbers.
+    #[diplomat::rust_link(fixed_decimal::SignedRoundingMode, Enum)]
+    #[diplomat::enum_convert(fixed_decimal::SignedRoundingMode, needs_wildcard)]
+    pub enum FixedDecimalRoundingMode {
```
younies, line 66: @robertbastian 

How to add the extension `Unsigned(FixedDecimalUnsignedRoundingMode)` in diplomat?

robertbastian: You need to flatten it. Did you see this comment? https://github.com/unicode-org/icu4x/issues/5065#issuecomment-2452995367

younies: I saw it, but we did not agree on it.

robertbastian: Please respond then so we can discuss it?

components/plurals/benches/operands.rs
```diff
@@ -27,7 +27,11 @@ fn operands(c: &mut Criterion) {
                     .expect("Failed to parse a number into an operands.");
             }
             for s in &data.fixed_decimals {
-                let f: FixedDecimal = FixedDecimal::from(s.value).multiplied_pow10(s.exponent);
+                let f: SignedFixedDecimal = {
+                    let mut dec = SignedFixedDecimal::from(s.value);
+                    dec.multiply_pow10(s.exponent);
```
sffc, line 32: Nit: `multiply_pow10` should be on `dec.absolute`, right?

ffi/capi/src/fixed_decimal.rs
sffc: I think you should have 2 enums, and the second enum contains all the variants of the first plus the four extra ones.

sffc (commented):
Praise: This looks much closer to the right design.

components/plurals/benches/operands.rs
younies: done

ffi/capi/src/fixed_decimal.rs
younies: done

utils/fixed_decimal/src/compact.rs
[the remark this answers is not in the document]
younies: done

utils/fixed_decimal/src/signed_decimal.rs
```diff
@@ -0,0 +1,3364 @@
...
+///
+/// - Integers, signed and unsigned
+/// - Strings representing an arbitrary-precision decimal
+/// - Floating point values (using the `ryu` feature)
+///
+/// To create a [`SignedFixedDecimal`] with fraction digits, either create it from an integer and then
+/// call [`SignedFixedDecimal::multiplied_pow10`], create it from a string, or (when the `ryu` feature is
+/// enabled) create it from a floating point value using [`SignedFixedDecimal::try_from_f64`].
+///
+/// # Magnitude and Position
```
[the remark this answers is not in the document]
younies: I agree with you, I removed starting from # Magnitude and Position.

components/datetime/src/format/datetime.rs
```diff
@@ -161,9 +161,12 @@ where
...
 
             // Always in latin digits according to spec
-            FixedDecimal::from(related_iso)
-                .padded_start(l.to_len() as i16)
-                .write_to(w)?;
+            {
+                let mut num = SignedFixedDecimal::from(related_iso);
+                num.pad_start(l.to_len() as i16);
+                num
+            }
```
sffc, line 168: Issue: Why did you change `padded_start` to `pad_start`?

However, you'll need the `{ ... }` block anyway if SignedFixedDecimal doesn't duplicate the functions of the inner type. Maybe we can just duplicate the `-ed` functions.

We discussed the `-ed` functions in June: https://github.com/unicode-org/icu4x/issues/2902#issuecomment-2153121175

tutorials/c-tiny/fixeddecimal/Cargo.lock
```diff
@@ -5,7 +5,7 @@ version = 3
 [[package]]
 name = "diplomat"
 version = "0.8.0"
-source = "git+https://github.com/rust-diplomat/diplomat?rev=8744ac97162341f347b63131969ad736e1047f6d#8744ac97162341f347b63131969ad736e1047f6d"
+source = "git+https://github.com/rust-diplomat/diplomat?rev=b49d63ec2a50a456b4b6e9d072d3650dca153ff8#b49d63ec2a50a456b4b6e9d072d3650dca153ff8"
```
sffc, line 8: Nit: Try not to include Cargo.lock changes in PRs with other functional changes. This Cargo.lock file might be stale, but it should be fixed in a standalone PR.

tutorials/c/Cargo.lock
```diff
@@ -47,7 +47,7 @@ dependencies = [
 [[package]]
 name = "diplomat"
 version = "0.8.0"
-source = "git+https://github.com/rust-diplomat/diplomat?rev=fe5201dd04ee6614942f817b291e90cb01440c19#fe5201dd04ee6614942f817b291e90cb01440c19"
```
sffc, line 50: Nit: Avoid lockfile diffs

utils/fixed_decimal/src/compact.rs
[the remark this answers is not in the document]
sffc: We should eventually make `Compact` composable, so that you can have an `UnsignedCompactDecimal`, for example. We can do that later in a different PR.

utils/fixed_decimal/src/decimal.rs
```diff
@@ -3167,10 +2527,6 @@ fn test_from_str_scientific() {
         pub output: &'static str,
     }
     let cases = [
-        TestCase {
-            input_str: "-5.4e10",
```
sffc, line 3171: Suggestion: Add back some of these cases as failure cases to demonstrate that UnsignedFixedDecimal will fail to parse strings with a minus sign

utils/fixed_decimal/src/decimal.rs
```diff
@@ -3155,6 +2508,13 @@ fn test_from_str() {
         assert_eq!(fd.nonzero_magnitude_end(), cas.magnitudes[2], "{cas:?}");
         let input_str_roundtrip = fd.to_string();
         let output_str = cas.output_str.unwrap_or(cas.input_str);
+
+        // Check if the output string starts with a plus sign and remove it,
+        // because we do not show sign for unsigned fixed decimal
```
sffc, line 2513: Suggestion: maybe just reject strings with a `+` sign in UnsignedRoundingMode. It's okay to be more strict.

sffc (commented):
Good work

sffc (dismissed):
LGTM; please sync to the latest main branch

sffc (approved):

PR #5671, opened 2024-10-11
Restructure time zone data structs

robertbastian:
* Exemplar cities should now be called locations (#5657)
* Region patterns can live in the locations struct
* Offset pattern can be split into signs and separator (#5638)


Manishearth (approved):
The offset pattern thing is the only thing here I'm iffy about: I'm worried a new offset pattern could be introduced in the future breaking these assumptions. But it does work.

robertbastian:
Shane allowed it :)

components/datetime/src/provider/time_zones.rs
```diff
@@ -42,13 +42,12 @@ pub(crate) mod tz {
...
-        feature = "serde",
-        serde(deserialize_with = "icu_provider::serde_borrow_de_utils::tuple_of_cow")
-    )]
-    pub hour_format: (Cow<'data, str>, Cow<'data, str>),
+    /// The sign for positive offsets
+    pub offset_positive_sign: char,
+    /// The sign for negative offsets
+    pub offset_negative_sign: char,
+    /// The separator sign
+    pub offset_separator: char,
```
sffc, line 50: I gave "permission" to flatten `offset_separator` to a char. The other two seem more likely to be non-chars (e.g. Arabic sometimes has a two-code-point plus sign); a `SinglePlaceholderPattern` would be more appropriate, though heavier on the stack size.

But, if it works with today's data, I'm not super concerned. Our stability policy gives us a pathway to fix this in the future if we need to.

provider/source/src/time_zones/convert.rs
```diff
@@ -51,38 +51,50 @@ impl DataProvider<TimeZoneEssentialsV1Marker> for SourceDataProvider {
...
+            .next()
+            .unwrap_or('-');
+        // The single character before `mm`
+        let offset_separator = offset_pattern_positive
+            .split_once('m')
+            .unwrap()
+            .0
+            .chars()
+            .next_back()
+            .unwrap_or(':');
```
sffc, line 81: Nit: I'd prefer this pattern parsing to be more strict. If there's data that violates one of our invariants, we should know about it.

sffc (commented):
Really nice improvements overall

provider/source/src/time_zones/convert.rs
robertbastian: #5680

PR #5674, opened 2024-10-11
Support localized offsets with seconds

### LINKED ISSUE #5638: Support short localized GMT format and GMT format with seconds

robertbastian:
Fixes #5638


Manishearth (dismissed):

sffc (approved):

components/datetime/src/time_zone.rs
```diff
@@ -414,6 +417,18 @@ impl FormatOffset for LocalizedOffsetFormat {
                             .write_to(sink)?;
                     }
 
+                    if self.offset.has_seconds() {
+                        sink.write_char(self.separator)?;
+                        self.fdf
+                            .format(
+                                &FixedDecimal::from(
+                                    (self.offset.offset_seconds() % 3600).unsigned_abs(),
```
sffc, line 425:
```suggestion
                                    (self.offset.offset_seconds() % 60).unsigned_abs(),
```

sffc (changes requested):
Actually wait, I think it is supposed to be `% 60`, right?

sffc (approved):

PR #5679, opened 2024-10-14
Remove internal traits DateSymbols, TimeSymbols, ZoneSymbols

sffc:
These were useful when code was shared between 1.x and 2.x formatting code paths, but now removing them reduces failure paths.

#1317


sffc:
Autosubmit = true

PR #5681, opened 2024-10-14
Move offset logic to `UtcOffset`

robertbastian:


components/timezone/src/types.rs
```diff
@@ -143,12 +143,12 @@ impl UtcOffset {
     }
 
     /// Returns the raw offset value in seconds.
-    pub fn offset_seconds(self) -> i32 {
+    pub fn into_seconds(self) -> i32 {
```
sffc, line 146: Nit: `into` is not really needed because it is a `Copy` type. It is a getter and it would take `&self` if it were not `Copy`

sffc (approved):

Manishearth (approved):

sffc (approved):

PR #5684, opened 2024-10-14
Use duration formatter patterns for localized offset

robertbastian:
#5638


sffc (approved):
Let's maybe check on the Serbian expectation

sffc (approved):

PR #5688, opened 2024-10-15
Start adding concrete format functions

sffc:
https://github.com/unicode-org/icu4x/issues/5269#issuecomment-2361895181

This approach we discussed doesn't completely work because it doesn't accomplish the type inference I wanted to see, because I need to accept a `Date<A: AsCalendar>`, not an actual concrete `Date<C>`. This voids most of the benefit of having concrete functions because then you can't `.into()` or `.parse()` into the type being formatted.

Thought: if I made `.format_date` take a concrete `Date<C>` (disallowing the use of `Ref`), I could keep `.format` as-is, and then these concrete format functions become nice-to-haves that can be added on later. I think this is most aligned with the design @Manishearth had preferred anyway.

Would be _nice_ to land this in 2.0-Alpha, but it's _fine_ if it changes in 2.0-Beta.


sffc:
Will re-do this.

PR #5691, opened 2024-10-15
`TimeZoneInfo`

robertbastian:
#5533


components/datetime/src/time_zone.rs
```diff
@@ -381,22 +464,26 @@ impl FormatTimeZone for SpecificNonLocationFormat {
...
-        let Some(metazone_id) = input.metazone_id else {
-            return Ok(Err(FormatTimeZoneError::MissingInputField("metazone")));
-        };
         let Some(time_zone_id) = input.time_zone_id else {
             return Ok(Err(FormatTimeZoneError::MissingInputField("time_zone_id")));
         };
+        let zone_variant = input.zone_variant_or_guess(
+            time_zone_id,
+            icu_timezone::provider::Baked::SINGLETON_ZONE_OFFSET_PERIOD_V1_MARKER,
+        );
```
sffc: Nit: I think we should refuse to format SpecificNonLocationFormat if we weren't able to compute the zone variant.

ffi/capi/src/timezone.rs
```diff
@@ -266,59 +240,22 @@ pub mod ffi {
         ///
         /// Returns null if the `zone_variant` field is empty.
         #[diplomat::rust_link(icu::timezone::ZoneVariant::daylight, FnInStruct)]
-        #[diplomat::rust_link(icu::timezone::CustomTimeZone::zone_variant, StructField, compact)]
+        #[diplomat::rust_link(icu::timezone::TimeZoneInfo::zone_variant, StructField, compact)]
         #[diplomat::attr(auto, getter)]
         pub fn is_daylight_time(&self) -> Option<bool> {
             Some(self.0.zone_variant? == icu_timezone::ZoneVariant::daylight())
         }
```
sffc, line 247: Nit: You probably need to add a way to set the `local_time`, right?

sffc (commented):
Praise: Good job getting all the boilerplate together

components/datetime/src/neo_marker.rs
```diff
@@ -3025,6 +3087,8 @@ impl DateTimeNamesMarker for NeoSkeleton {
     type ZoneGenericShort = datetime_marker_helper!(@names/zone/generic_short, yes);
     type ZoneSpecificLong = datetime_marker_helper!(@names/zone/specific_long, yes);
     type ZoneSpecificShort = datetime_marker_helper!(@names/zone/specific_short, yes);
+    type MetazoneLookup = datetime_marker_helper!(@names/zone/metazone_periods, yes);
+    type Offsets = datetime_marker_helper!(@names/zone/offsets, yes);
```
sffc, line 3091: You mentioned today that you were thinking about making ZoneVariant required on the input.

I'm ultimately fine with this, but I do think that the zone variant calculator is useful to our clients, both now and in the future. So, if we do this, I think we should keep the zone variant calculator where it is, external to icu_datetime.

components/datetime/src/neo_marker.rs
```diff
@@ -296,39 +295,40 @@
...
 //!
-//! // "uschi" - has metazone symbol data for generic_non_location_short
-//! let mut time_zone = "-0600".parse::<CustomTimeZone>().unwrap();
-//! time_zone.time_zone_id = mapper.as_borrowed().iana_to_bcp47("America/Chicago");
-//! time_zone.maybe_calculate_metazone(&mzc, &datetime);
+//! // "uschi" - has symbol data for generic_non_location_short
+//! let time_zone = TimeZoneInfo {
+//!     time_zone_id: mapper.as_borrowed().iana_to_bcp47("America/Chicago"),
+//!     offset: Some(UtcOffset::from_eighths_of_hour(-6 * 8)),
+//!     local_time: Some((datetime.date, datetime.time)),
```
sffc, line 302: Thought: internally we always convert this to a local timestamp. Maybe we should take a `LocalTimestamp` (new type) instead of the weird tuple.

robertbastian: A timestamp in local-time minutes is so odd, I don't want to expose that. This should stay a internal encoding of `(Date<Iso>, Time)`.

components/datetime/src/neo_marker.rs
robertbastian: > I do think that the zone variant calculator is useful to our clients

I don't think it is. It performs a very small subset of TZ calculations, I don't really see how that is something we should support, and the rest of the calculations isn't. I'm happy making zone variant required, and not exposing a zone calculator. We also don't expose an offset calculator, which is very similar (goes from ID + datetime to offset).

components/datetime/src/neo_marker.rs
sffc: If we exposed it, I would say to use an `i64` local-seconds timestamp, which is slightly less weird.

components/datetime/src/neo_marker.rs
sffc: > We also don't expose an offset calculator, which is very similar (goes from ID + datetime to offset).

These are two very different operations. Figuring out the zone variant requires a very similar shape of data as figuring out the metazone. Figuring out the actual offset requires full TZDB. Figuring out the zone variant involves looking at big ranges that last decades; figuring out the offset requires looking at the transitions each and every year.

My position is that ICU4X should support formatting of a standard full IXDTF string, which is time zone ID and offset.

You seem to think that the code you spent a few weeks writing for zone variant calculations is flawed, but I don't see how it is fundamentally flawed.

robertbastian: > My position is that ICU4X should support formatting of a standard full IXDTF string, which is time zone ID and offset.

I don't think IXDTF is the magical input format you want it to be.

components/datetime/src/neo_marker.rs
robertbastian: The local thing is the big weirdness, not the minutes.

sffc (dismissed):

components/timezone/src/ixdtf.rs
```diff
@@ -315,27 +334,30 @@ impl CustomZonedDateTime<AnyCalendar> {
...
-    /// assert_eq!(inconsistent_tz_from_both.zone.offset, Some(UtcOffset::try_from_seconds(-18000).unwrap()));
-    /// assert_eq!(inconsistent_tz_from_both.zone.zone_variant, None);
-    /// assert!(inconsistent_tz_from_both.zone.local_time.is_some());
+    /// // We know that America/Los_Angeles never used a -05:00 offset at any time of the year 2024
+    /// assert_eq!(
+    ///     CustomZonedDateTime::try_from_str("2024-08-08T12:08:19-05:00[America/Los_Angeles]").unwrap_err(), 
+    ///     ParseError::InvalidOffsetError
+    /// );
+    /// 
+    /// // We don't know that America/Los_Angeles didn't use DST (-08:00) in August
```
sffc, line 357:
```suggestion
    /// // We don't know that America/Los_Angeles didn't use standard time (-08:00) in August
```

sffc (approved):

sffc (approved):

PR #5700, opened 2024-10-17
Make TimeZoneInfo more strict

sffc:
Instead of `TimeZoneInfo` having optional fields, it now makes the fields required via the type system, which prevents callers from passing in a time zone with missing fields to a formatter that needs more of the fields.


Manishearth (approved):
I'm a bit worried about the UX of this but timezones are gnarly anyway.

components/timezone/src/ixdtf.rs
```diff
@@ -36,6 +34,8 @@ pub enum ParseError {
     InvalidOffsetError,
     /// There was an invalid IANA identifier provided.
     InvalidIanaIdentifier,
```
robertbastian, line 36: I don't think we ever want to raise `InvalidIanaIdentifier`, it might already be dead

components/timezone/src/ixdtf.rs
```diff
@@ -78,116 +78,214 @@ impl UtcOffset {
...
+        };
+        let offset = UtcOffset::try_from_utc_offset_record(offset)?;
+        let iso = DateTime::<Iso>::try_new_iso_datetime(
+            date.year,
+            date.month,
+            date.day,
+            time.hour,
+            time.minute,
+            time.second,
+        )?;
```
robertbastian, line 133: Nit: `UtcOffset` can construct itself from `OffsetRecord`, but `DateTime` cannot construct itself from `DateRecord`. Seems inconsistent.

Also, maybe create a `Date` and a `Time` instead of a `DateTime`, as that's what we need.

components/timezone/src/ixdtf.rs
```diff
@@ -78,116 +78,214 @@ impl UtcOffset {
...
+                tz:
+                    Some(TimeZoneAnnotation {
+                        tz: TimeZoneRecord::Offset(offset),
+                        ..
+                    }),
+                date: Some(date),
+                time: Some(time),
                 ..
-            } => Self::try_from_time_zone_record(&time_zone_annotation.tz, None, date, time),
+            } => (offset, date, time),
```
robertbastian, line 105: these cases can be collapsed

components/timezone/src/ixdtf.rs
```diff
@@ -78,116 +78,214 @@ impl UtcOffset {
...
-                                } else {
-                                    return Err(ParseError::InvalidOffsetError);
-                                }
-                            } else {
-                                debug_assert_eq!(time_zone_id.0.as_str(), "unk");
-                                ZoneVariant::standard()
-                            },
-                        );
+        let IxdtfParseRecord {
+            offset: None,
```
robertbastian, line 148: phew, not sure I want to make too much information an error. Imagine if you want to start sending more details across the network to change time zone format, if this errors you have to upgrade client and server at exactly the same time.

components/timezone/src/ixdtf.rs
```diff
@@ -78,116 +78,214 @@ impl UtcOffset {
...
+            time_zone_id,
+            offset: None,
+            zone_variant: (),
+            local_time: (iso.date, iso.time),
+        })
+    }
+}
+
+impl TimeZoneInfo<models::Full> {
+    fn try_full_from_ixdtf_record(ixdtf_record: &IxdtfParseRecord) -> Result<Self, ParseError> {
```
robertbastian, line 181: nit: keep constructors for the same type together

components/timezone/src/ixdtf.rs
```diff
@@ -78,116 +78,214 @@ impl UtcOffset {
...
+        let mapper = TimeZoneIdMapper::new();
+        let time_zone_id = mapper.as_borrowed().iana_bytes_to_bcp47(iana_identifier);
+        let iso = DateTime::<Iso>::try_new_iso_datetime(
+            date.year,
+            date.month,
+            date.day,
+            time.hour,
+            time.minute,
+            time.second,
+        )?;
```
robertbastian, line 206: this is duplicated code. you already have a `with_zone_variant` method to extend a location-only info, no?

components/timezone/src/ixdtf.rs
```diff
@@ -78,116 +78,214 @@ impl UtcOffset {
...
+        Self::try_location_only_iso_from_utf8(ixdtf_str.as_bytes())
+    }
+
+    /// Create a [`CustomZonedDateTime`] in ISO-8601 calendar from IXDTF syntax utf8 bytes.
+    ///
+    /// The string should have only a named time zone and no offset.
+    ///
+    /// ✨ *Enabled with the `compiled_data` and `ixdtf` Cargo features.*
+    ///
+    /// See [`Self::try_iso_from_str`].
```
robertbastian, line 276: I think we usually just say "See `Self::try_from_str`" to avoid overloading the docs pages with repeated information.

components/timezone/src/lib.rs
```diff
@@ -90,6 +90,7 @@ extern crate alloc;
 mod error;
 mod ids;
 pub mod provider;
+pub mod scaffold;
```
robertbastian, line 93: `#[doc(hidden)]`?

components/timezone/src/lib.rs
```diff
@@ -105,7 +106,9 @@ pub use ids::{
     TimeZoneIdMapperWithFastCanonicalizationBorrowed,
 };
 pub use provider::TimeZoneBcp47Id;
+pub use time_zone::models;
 pub use time_zone::TimeZoneInfo;
+pub use time_zone::TimeZoneModel;
```
robertbastian, line 111: when does a user need access to this trait? I don't think we need it reexported.

components/timezone/src/time_zone.rs
```diff
@@ -2,36 +2,92 @@
...
     ///
     /// This field is not enforced to be consistent with the time zone id.
     pub offset: Option<UtcOffset>,
     /// The time variant e.g. daylight or standard, if known.
-    pub zone_variant: Option<ZoneVariant>,
+    pub zone_variant: Model::ZoneVariant,
     /// The time at which to interpret the time zone.
     ///
     /// This can be set in order to get correct historical time zone names.
     /// If it's not set, the most recent data for the time zone will be used.
```
robertbastian, line 71: not anymore

components/timezone/src/time_zone.rs
```diff
@@ -40,8 +96,67 @@ impl TimeZoneInfo {
...
+        Self {
+            offset: Some(offset),
+            time_zone_id,
+            zone_variant: (),
+            local_time: (),
+        }
+    }
+
+    /// Sets a local time on this time zone.
+    pub const fn with_local_time(
```
robertbastian, line 115:
```suggestion
    pub const fn at_time(
```

components/timezone/src/time_zone.rs
```diff
@@ -40,8 +96,67 @@ impl TimeZoneInfo {
...
+            time_zone_id: self.time_zone_id,
+            zone_variant,
+            local_time: self.local_time,
+        }
+    }
+}
+
+impl TimeZoneInfo<models::Full> {
+    /// Creates a new [`TimeZoneInfo`] for the UTC time zone
+    /// with a reference point at the UNIX Epoch.
```
robertbastian, line 142: mmh kinda don't want to document this. The reference point for UTC is completely irrelevant, because its metazone doesn't change.

robertbastian (commented):
reviewed the timezone crate so far

components/timezone/src/ixdtf.rs
sffc: I think being strict about the expected fields is important. But, we can continue to iterate on exactly the shape of these functions and the strictness. In the mean time, I added more comments.

components/timezone/src/time_zone.rs
sffc: I'll delete this function since it has only 1 call site.

components/timezone/src/ixdtf.rs
robertbastian: Imagine if you're calling an API that only guarantees that it sends the offset, but doesn't guarantee that it *only* sends the offset.

components/timezone/src/time_zone.rs
robertbastian: I think it's still useful to have this as a public constructor

components/timezone/src/lib.rs
sffc: It shows up in a public API signature and I do not like to not export things that show up publicly. I'm happy to export it from the `scaffold` module..

components/timezone/src/lib.rs
sffc: But it is public, not just ICU4X-internal, and I really do not like `#[doc(hidden)]`. I'm happy to slap unstable warnings on the module and everything inside of it, similar to the provider modules.

components/timezone/src/ixdtf.rs
sffc: Hmm, maybe we should discuss that. I think critical information should be duplicated. For example, it is not useful to see a tooltip that says "see this other thing". More detailed docs and examples should be deduplicated, sure.

components/timezone/src/ixdtf.rs
sffc: Perhaps that use case should be another constructor / parse function that sets the time zone to `unk` if it is not present. It seems bad to _throw away_ the info, though.

robertbastian: There's currently not constructor here that does that.

sffc: (Do you want it now or can we add it later when we make a polished IXDTF parser?)

robertbastian: You're saying this is a workaround to the issue I'm describing. You can address this issue either by removing the error path here, or by adding a more lenient constructor.

sffc: This PR has

- `try_offset_only_from_str` = strict offset only parsing
- `try_location_only_from_str` = strict location only parsing
- `try_from_str` = strict offset+location parsing

I can add:

- `try_loose_from_str` = missing offset parses to `None`; missing location parses to `unk`

robertbastian: This does not address the issue here. If I only need location info, it should not be an error for there to be offset info. It should be an error for there not to be location info.

components/timezone/src/ixdtf.rs
sffc: it's because `DateTime` is in a different crate and we can't export from-ixdtf-record functions across crate boundaries because ixdtf is less stable than icu_calendar.

components/timezone/src/ixdtf.rs
sffc: no? they cover different cases which are in comments

components/timezone/src/ixdtf.rs
sffc: Yes, it should be an error. The constructor you choose asserts the shape of the string.

Can we please not make this block the PR.

components/timezone/src/ixdtf.rs
sffc: Right now all the `TimeZoneInfo` constructors are together and the `CustomZonedDateTime` constructors are together.

components/timezone/src/ixdtf.rs
robertbastian: No this is important. Look at serde json, it does not complain when an object contains a key it doesn't know, it only complains when it's missing a key.

sffc: ok can I make it `try_strict_from_str` and then `try_from_str` is the loose one, as it is now

robertbastian: yes

sffc: There are 4 functions which I think cover all of the use cases. This is how I have named them for now:

- try_offset_only_iso_from_str (strict, returns `UtcOffset`)
- try_location_only_iso_from_str (strict, returns `TimeZoneInfo<AtTime>`)
- try_loose_iso_from_str (loose, returns `TimeZoneInfo<AtTime>`)
- try_iso_from_str (strict, returns `TimeZoneInfo<Full>`)

There is only one "loose" function because it is only possible for a loose function to return `TimeZoneInfo<AtTime>`, not `TimeZoneInfo<Full>`.

I do not see a use case for a "loose" function returning `UtcOffset` because if you want things loose, just get a `TimeZoneInfo<AtTime>` which is more general and fully featured.

components/timezone/src/ixdtf.rs
sffc: Done

robertbastian (dismissed):

components/timezone/src/time_zone.rs
sffc: Done

components/timezone/src/time_zone.rs
sffc: Done

components/timezone/src/time_zone.rs
sffc: No actually I think this is a terrible function. For example, someone might make a `TimeZoneInfo::utc_full` as the only non-string constructor of `TimeZoneInfo<Full>` and then try to overwrite fields on it.

This function clearly is not that useful because there is only 1 call site and it is a weird call site in a match statement. There are no call sites in normal code.

sffc:
Ok I think this is ready now. Only thing I didn't do 100% from your feedback is the full deduplication of code in ixdtf.rs, but it is better than it was before.

robertbastian (approved):

robertbastian:
#5533

PR #5708, opened 2024-10-18
Rename things to DateTimeFormatter; break up most of neo_marker module

sffc:
#1317

Depends on #5707

Mostly search-and-replace. The breakup of neo_marker was largely manual. None of the public docs tests need to import things from the scaffold module, by design.


sffc:
The individual commits are reviewable.

robertbastian (approved):

PR #5712, opened 2024-10-18
Full transliterator datagen

robertbastian:
Well, almost full. Working around some CLDR data bugs.


utils/zerovec/src/varzerovec/databake.rs
```diff
@@ -2,10 +2,10 @@
...
 
-use crate::{ule::VarULE, VarZeroSlice, VarZeroVec};
+use crate::{ule::VarULE, varzerovec::VarZeroVecFormat, VarZeroSlice, VarZeroVec};
 use databake::*;
 
-impl<T: VarULE + ?Sized> Bake for VarZeroVec<'_, T> {
+impl<T: VarULE + ?Sized, F: VarZeroVecFormat> Bake for VarZeroVec<'_, T, F> {
     fn bake(&self, env: &CrateEnv) -> TokenStream {
         env.insert("zerovec");
         if self.is_empty() {
```
sffc, line 11: Thought: we may want to be explicit about the type parameter in the generated code in order to be safer. Can do later.

components/experimental/src/transliterate/provider.rs
```diff
@@ -65,7 +66,7 @@ impl<'de> serde::Deserialize<'de> for RuleBasedTransliterator<'de> {
             #[serde(borrow)]
             pub id_group_list: VarZeroVec<'a, VarZeroSlice<SimpleIdULE>>,
             #[serde(borrow)]
-            pub rule_group_list: VarZeroVec<'a, VarZeroSlice<RuleULE>>,
+            pub rule_group_list: VarZeroVec<'a, VarZeroSlice<RuleULE, Index32>, Index32>,
```
sffc, line 69: Nit: probably only 1 of these needs to be Index32.

It looks like the biggest are Hant to Hans, Hans to Hant, and Hani to Latn.

provider/source/src/transforms/mod.rs
```diff
@@ -39,15 +39,57 @@ impl CldrCache {
                         continue;
                     };
 
+                    if transform == "und-Ethi-t-und-latn-m0-beta_metsehaf-geminate" {
```
sffc, line 42: Nit: We should have issues for all these with a `TODO` in the code. One big issue is _probably_ fine.

components/experimental/src/transliterate/transliterator/mod.rs
```diff
@@ -232,15 +233,53 @@ impl Transliterator {
...
     /// let locale = "und-Arab-t-und-beng".parse().unwrap();
-    /// let t = Transliterator::try_new_unstable(locale, provider).unwrap();
+    /// let t = Transliterator::try_new(&locale).unwrap();
     /// let output = t.transliterate("অকার্যতানাযা".to_string());
     ///
     /// assert_eq!(output, "اكاريتانايا");
     /// ```
-    pub fn try_new_unstable<P>(locale: Locale, provider: &P) -> Result<Transliterator, DataError>
+    #[cfg(feature = "compiled_data")]
+    pub fn try_new(locale: &Locale) -> Result<Self, DataError> {
```
sffc, line 242: 🥳

sffc (dismissed):

utils/zerovec/src/varzerovec/databake.rs
robertbastian: agreed

sffc (approved):

FrankYFTang:
Could someone explain to me why Any-Lower is now changed to Any-XLower?

robertbastian:
Where do you see `Any-XLower`? https://raw.githubusercontent.com/unicode-org/icu4x/refs/heads/main/provider/data/experimental/data/transliterator_rules_v1.rs.data contains `any-lower` 53 times.

sffc:
In components/experimental/tests/transliterate/lower_ascii.rs

https://github.com/unicode-org/icu4x/pull/5712/files#diff-804ec3ecfd714a49b013fc85f542695e0488c1485d2f305b3c76d19c70ecef94L90

robertbastian:
That test defines an `Any-XLower` transliterator, because the `Any-Lower` built-in transliterator panics (during parsing already), and built-in transliterators cannot be redefined.

It's just a test identifier, there is no `Any-XLower` transliterator.

PR #5714, opened 2024-10-19
Pin Python 3.11 in bench-memory

sffc:
Maybe fixes CI. Need to merge to main or run the job on the PR to verify.


robertbastian (approved):

sffc:
I think it worked

https://github.com/unicode-org/icu4x/actions/runs/11451308031/job/31860307794

PR #5720, opened 2024-10-22
Fix GN CI

robertbastian:


Manishearth (approved):

sffc (approved):

sffc:
These pinned versions are to make things work, so changing the pin versions is fine if it fixes things

PR #5725, opened 2024-10-23
Fix FFI ZDT formatting

robertbastian:


robertbastian:
This actually works now, tested locally.

Manishearth (approved):

ffi/capi/src/zoned_formatter.rs
```diff
@@ -62,7 +63,9 @@ pub mod ffi {
             let zdt = icu_timezone::CustomZonedDateTime {
                 date: greg.date,
                 time: greg.time,
-                zone: icu_timezone::TimeZoneInfo::try_from(time_zone.0)?,
+                zone: icu_timezone::TimeZoneInfo::from(&time_zone.0)
+                    .at_time((datetime.0.date, datetime.0.time))
+                    .with_zone_variant(time_zone.0.zone_variant.unwrap_or(ZoneVariant::standard())),
```
sffc, line 68: ok in the short term since this fixes the WASM demo, but it is in general antithetical to the design of TimeZoneInfo that you would default it to `ZoneVariant::standard()`

We need a better FFI in 2.0

PR #5729, opened 2024-10-24
Update locale_core::preferences to stakeholder alignment

zbraniecki:
Address #419 feedback


zbraniecki:
@sffc @robertbastian - can I get a high level thumbs up on that direction?

robertbastian:
Looks like what we agreed on

zbraniecki:
Added list format - @sffc @robertbastian - PTAL.

If you're good with that, I'll document it and we're gtg

components/list/examples/and_list.rs
```diff
@@ -6,12 +6,15 @@
...
-use icu::list::{ListFormatter, ListLength};
+use icu::list::{ListFormatter, ListFormatterOptions, ListLength};
 use icu::locale::locale;
 
 fn main() {
-    let list_formatter =
-        ListFormatter::try_new_and_with_length(&locale!("es").into(), ListLength::Wide).unwrap();
+    let list_formatter = ListFormatter::try_new_and_with_length(
+        locale!("es").into(),
+        ListFormatterOptions::new().style(ListLength::Wide),
```
robertbastian, line 15: this should probably be `with_style`

robertbastian, line 15: I think we don't need a `new` if we have a `default`. Do we have a `default`?

components/list/src/list_formatter.rs
```diff
@@ -64,6 +81,16 @@ macro_rules! constructor {
     };
 }
 
+fn get_data_locale_from_prefs(prefs: ListFormatterPreferences) -> DataLocale {
```
robertbastian, line 84: Ideally this should be generated as well. The way to do this would be to add the marker as a macro argument and branch on `<$marker>::INFO.fallback_config.priority`

components/list/examples/and_list.rs
sffc: "style" is discouraged; should be `with_length`

(this is moot if we revert `ListFormatterOptions`)

components/list/src/list_formatter.rs
```diff
@@ -5,13 +5,28 @@
...
+);
+define_options!(
+    /// The options for list formatting.
+    ListFormatterOptions,
+    {
+        /// Style
+        style: ListLength
+    }
+);
+
```
sffc, line 29: Issue: I thought we agreed to not touch the `options` argument. It's fairly easy in this crate, but the options bags are more complicated in other crates like `datetime` and I really don't want to get in the business of generating them with a cross-crate macro. Preferences, yes, but not options.

(I have other comments regarding this, but they are moot if we revert `ListFormatterOptions`)

sffc (commented):
`ListFormatterPreferences` seems about right. As written in a comment, I don't really see switching around how we do options right now. I guess it generates `merge` functions, which is nice, but can't we add those on later?

I also don't agree with changing all constructors to use options bags just because they can. The ones that don't take options bags use positional arguments instead and are named `_with_foo` or `_with_foo_and_bar`. That seems fine to me.

components/list/examples/and_list.rs
zbraniecki: We do, I think having `new` and `default` is common and I prefer it.

components/list/src/list_formatter.rs
zbraniecki: Sounds good. I hand-rolled the options bag. I believe each constructor should take `prefs, options` pair.

components/list/src/list_formatter.rs
zbraniecki: Spun out as https://github.com/unicode-org/icu4x/issues/5764

components/list/src/options.rs
```diff
@@ -0,0 +1,49 @@
...
+}
+
+impl ListFormatterOptions {
+    /// Constructs a new [`ListFormatterOptions`] struct.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Auguments the struct with the set [`ListLength`].
+    pub fn with_length(mut self, length: ListLength) -> Self {
```
robertbastian, line 28: can also be const

components/list/src/options.rs
```diff
@@ -0,0 +1,49 @@
...
+/// use icu::list::{ListFormatterOptions, ListLength};
+///
+/// let options = ListFormatterOptions::new()
+///     .with_length(ListLength::Wide);
+/// ```
+#[derive(Default, Debug, Clone)]
+#[non_exhaustive]
+pub struct ListFormatterOptions {
+    /// The length variant should reflect available space for the list.
+    pub length: Option<ListLength>,
```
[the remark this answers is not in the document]
zbraniecki: which I think is reasonable, since we have a default.

components/locale_core/src/preferences/mod.rs
```diff
@@ -4,238 +4,626 @@
...
+                    ue_region: None,
+
+                    $(
+                        $key: None,
+                    )*
+                }
+            }
+        }
+
+        impl From<&icu_provider::DataLocale> for $name {
```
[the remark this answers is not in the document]
zbraniecki: DurationFormatter takes `DataLocale` and I need to get preferences out of it.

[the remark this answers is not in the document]
robertbastian: DurationFormatter should take DurationFormatterPreferences, no?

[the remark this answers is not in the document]
zbraniecki: Once we migrate it, yeah. But for now, I'm narrowing down the scope of this patch and would prefer to avoid scope creep. I'm landing an impl and a migration of list formatter, and that requires me to massage the duration formatter.

components/locale_core/src/extensions/unicode/subdivision.rs
```diff
@@ -131,6 +133,19 @@ impl SubdivisionId {
...
+    pub fn into_subtag(self) -> Subtag {
+        use writeable::Writeable;
+
+        // XXX: This can be optimized to concatenate two TinyAsciiStr.
+        let mut result = alloc::string::String::with_capacity(8);
+        let _ = self.write_to(&mut result);
+        #[allow(clippy::expect_used)]
+        let tinystr = TinyAsciiStr::try_from_str(&result)
+            .expect("Constructing 8 chars TinyAsciiStr from two 4 char ones");
+        Subtag::from_tinystr_unvalidated(tinystr)
```
sffc, line 147: https://github.com/unicode-org/icu4x/pull/5772

components/locale_core/src/preferences/extensions/unicode/keywords/numbering_system.rs
```diff
@@ -2,22 +2,209 @@
...
-            .ok_or(PreferencesParseError::InvalidKeywordValue)
-    },
-    |input: NumberingSystem| {
-        crate::extensions::unicode::Value::from_subtag(Some(input.0))
-    }
-);
+enum_keyword!(
+    /// A Unicode Number System Identifier defines a type of number system.
+    ///
+    /// The valid values are listed in [LDML](https://unicode.org/reports/tr35/#UnicodeNumberSystemIdentifier).
```
sffc, line 10: Issue: Why did you change this to a closed enum? This seems worse than an open enum because:

1. This list changes frequently and it's silly to update it here. We have to do that every time in ECMA-402 and it doesn't solve any real-world problems.
2. We have to go back and forth between a TinyAsciiStr and an enum discriminant. It would be nice to just reference the string, which we can do with `DataMarkerAttributes::try_from_str`.

zbraniecki: My logic was that numbering systems are actually quite stable and don't change frequently (like, currencies or timezones). I am surprised that you say that they change often - I'd expect the list to only grow, and not often (generating a new numbering system is non trivial).

My DX reason was that writing `NumberingSystem::Latn` is nicer than `NumberingSystem::from(value!("latn"))` and this is something I can see developers actually writing, or listing in Preferences.

My position is weak here and I can switch it if you prefer an open struct after my explanation.

sffc: Numbering systems have been added every year for the last 4 years: https://github.com/tc39/ecma402/issues/884

You shouldn't need to update ICU4X to be able to consume data that reflects the new numbering systems.

I don't buy the ergonomics argument. Just have `NumberingSystem::latin()` which you can write as a const fn or even `NumberingSystem::LATIN` which you can write as an associated constant.

components/locale_core/src/preferences/mod.rs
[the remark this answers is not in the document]
robertbastian: The macro shouldn't generate this for all preferences, the downstream crate can add a temporary implementation. There should also be a TODO because we should definitely not launch with this.

ffi/capi/src/list.rs
```diff
@@ -34,14 +35,15 @@ pub mod ffi {
             locale: &Locale,
             length: ListLength,
```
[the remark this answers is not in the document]
robertbastian: .

components/list/src/lib.rs
```diff
@@ -12,13 +12,14 @@
...
 //! # use icu::locale::locale;
 //! # use writeable::*;
 //! #
 //! let list_formatter = ListFormatter::try_new_and_with_length(
-//!     &locale!("es").into(),
-//!     ListLength::Wide,
+//!     locale!("es").into(),
+//!     ListFormatterOptions::new()
+//!       .with_length(ListLength::Wide)
 //! )
```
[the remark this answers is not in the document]
robertbastian: +1 

It would have been nice to make options changes in a separate PR, because those changes are most of the friction of this review. I'm aligned on the preferences stuff.

components/list/examples/and_list.rs
robertbastian: I don't think `new` is all that common on options bags (struct with only public optional fields). The `Default` implementation is clear, it returns a struct with all fields set to `None` (and that's why we should use `default()` in docs). This is not the expectation for `new`, so I think it adds friction to the API. It also adds another way of doing the same thing, so it makes people question if it actually does the same thing.

zbraniecki: I'm convinced. thank you!

ffi/capi/src/list.rs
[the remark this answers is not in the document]
zbraniecki: I'm converging on all options being optional (even if they have a default). The motivation is to distinguish between "developer specified value" from "developer did not explicitly specify a value". The former allows for selecting different value at runtime based on other signals (like an option who'se default value is different depending on a locale).

components/list/src/lib.rs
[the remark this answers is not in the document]
zbraniecki: I prefer to complete the migration of a single module to evaluate the DX.

components/locale_core/src/preferences/locale.rs
```diff
@@ -0,0 +1,139 @@
+// This file is part of ICU4X. For terms of use, please see the file
+// called LICENSE at the top level of the ICU4X source tree
+// (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
+use crate::subtags::{Language, Region, Script, Subtag, Variant, Variants};
+
+/// The structure storing locale subtags used in preferences.
+#[allow(clippy::exhaustive_structs)] // this type is stable
+#[derive(Debug, Clone)]
```
robertbastian, line 9: this can be `Copy`

components/locale_core/src/preferences/locale.rs
```diff
@@ -0,0 +1,139 @@
...
+            script: lid.script,
+            region: lid.region,
+            variant: lid.variants.iter().copied().next(),
+            subdivision: None,
+            ue_region: None,
+        }
+    }
+}
+
+impl From<&LocalePreferences> for crate::Locale {
```
robertbastian, line 72: this can be by-value once `LocalePreferences` is `Copy`

robertbastian (approved):

components/locale_core/src/preferences/locale.rs
```diff
@@ -0,0 +1,139 @@
+// This file is part of ICU4X. For terms of use, please see the file
+// called LICENSE at the top level of the ICU4X source tree
+// (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
+use crate::subtags::{Language, Region, Script, Subtag, Variant, Variants};
+
+/// The structure storing locale subtags used in preferences.
+#[allow(clippy::exhaustive_structs)] // this type is stable
+#[derive(Debug, Clone, Copy)]
+pub struct LocalePreferences {
```
sffc, line 10: Issue: Not a fan of this having public fields. But that can be fixed after the PR lands

components/list/src/options.rs
```diff
@@ -0,0 +1,55 @@
...
+    /// Constructs a new [`ListFormatterOptions`] struct.
+    pub const fn default() -> Self {
+        Self { length: None }
+    }
+
+    /// Auguments the struct with the set [`ListLength`].
+    pub const fn with_length(mut self, length: ListLength) -> Self {
+        self.length = Some(length);
+        self
+    }
```
sffc, line 37: Suggestion for a follow-up: we could add functions here `pub const fn long() -> Self` so that you don't need to import both `ListLength` and `ListFormatterOptions`. I did this in icu_datetime.

sffc (approved):

components/locale_core/src/preferences/locale.rs
zbraniecki: Filed as #5785

components/list/src/options.rs
zbraniecki: My concern is that it doesn't scale. I aspire to have our constructors be predictable over convenient. 
I realize that it's a preference and we may not be aligned on that.

sffc: I consider length enums to be special because they are so common. I wouldn't suggest this for other options, necessarily.

PR #5730, opened 2024-10-24
Borrowed structs cleanup

robertbastian:


components/locale/src/fallback/mod.rs
```diff
@@ -108,16 +108,12 @@ impl LocaleFallbacker {
     #[cfg(feature = "compiled_data")]
     #[allow(clippy::new_ret_no_self)] // keeping constructors together
     pub const fn new<'a>() -> LocaleFallbackerBorrowed<'a> {
```
sffc, line 110: Question: why aren't we just returning 'static?

robertbastian: Because the compiler cannot downcast this lifetime, and compiled data needed that somwhere.

sffc (approved):

Manishearth (approved):

PR #5736, opened 2024-10-25
Make an `IxdtfParser` type

robertbastian:
Based on #5735


sffc (approved):
Praise: Thanks for keeping the function names the same so it is easy to review.

robertbastian:
Hmm, how do I load all `AnyCalendar` data upfront?

sffc:
Just put in the bounds like usual:

https://unicode-org.github.io/icu4x/rustdoc/icu/calendar/enum.AnyCalendar.html#method.try_new_unstable

We do this in the DateTimeFormatter constructors, too (they are at the end of the list):

https://unicode-org.github.io/icu4x/rustdoc/icu/datetime/struct.DateTimeFormatter.html#method.try_new_unstable

sffc:
Oh, since it is cross-crate, you will need the `AnyCalendarLoader` thing.

sffc:
Oh, but you don't know the calendar until parsing time. Hmm.

Easy way out is to pass the AnyCalendar provider into the parse function, but that's annoying.

Maybe you can save a closure that calls the AnyCalendar constructor as a field of the parser? You'd need to put a lifetime on the parser object, I think.

sffc:
Or you could eagerly load all the markers into payloads and then make your own internal DataProvider that reads those payloads when calling `AnyCalendar::try_new_unstable`. Hmm. Maybe we want that to be a type in the `icu_calendar` crate, since it seems kind-of silly that `AnyCalendar::new` might need to load the same payload over and over, and you can't save it between loads.

sffc:
https://github.com/unicode-org/icu4x/pull/5740

sffc (approved):

PR #5744, opened 2024-10-29
Move pattern and skeleton modules to provider

sffc:
#1317

These are internal types used primarily for datagen and databake.

I need to move these now because I need both namespaces for new public types.

I also changed things from `pub` to `pub(crate)` where possible and deleted resulting dead code.


Manishearth (approved):
not looking too closely at the actual moved files

PR #5747, opened 2024-10-30
Improve timezone APIs

robertbastian:
* Fewer overlapping constructors for `TimeZoneInfo`
* Private fields for `TimeZoneInfo`
* Return type for `ZoneOffsetCalculator`
* docs


Manishearth (approved):

components/datetime/src/fieldset.rs
```diff
@@ -1013,7 +1007,7 @@ impl_zone_marker!(
     /// only a full time zone info can be formatted with this style.
     /// For example, [`TimeZoneInfo<AtTime>`] cannot be formatted.
     ///
-    /// ```compile_fail
+    /// ```compile_fail,E0271
```
sffc, line 1010: You can put the expected error number here? I've never seen that before but I love if it works.

sffc (commented):
I have comments on this but it's merged. I'll leave them after my meetings

sffc (commented):
OK, this looks good to me

PR #5748, opened 2024-10-30
Remove `icu::calendar` unix epoch APIs

robertbastian:
The Unix epoch is a physical time concept. It does not work on a civil time API. https://errorprone.info/docs/time


components/calendar/src/date.rs
```diff
@@ -320,12 +320,30 @@ impl<A: AsCalendar> Date<A> {
...
+    /// ```rust
+    /// use icu::calendar::Date;
+    /// use icu::calendar::DateDurationUnit;
+    ///
+    /// let a = Date::try_new_iso(1994, 12, 10).unwrap();
+    /// let b = Date::try_new_iso(2024, 10, 30).unwrap();
+    ///
+    /// assert_eq!(b.days_since(a), 10_917);
+    /// ```
+    #[doc(hidden)] // unstable
```
sffc, line 341: I think I would prefer embracing the replacement APIs but we can do that later

sffc (approved):
Yeah these functions were never very good and it's probably good to remove them, though it seems we should embrace a replacement.

components/calendar/src/date.rs
robertbastian: There is already a hidden `until` API which is more powerful than this, so I don't want to stabilise this. `until` is not fully implemented, I'd have used it if the max-unit argument were respected.

PR #5754, opened 2024-10-30
Fix some time zone examples

robertbastian:
There are no short English non-location names for `America/Sao_Paolo`, so the examples are currently just falling back to localized offset or location. If we use long patterns we get actual non-location names.


components/datetime/src/fieldset.rs
```diff
@@ -987,7 +987,7 @@ impl_zone_marker!(
     ///
     /// let fmt = FixedCalendarDateTimeFormatter::<Gregorian, _>::try_new(
     ///     &locale!("en").into(),
-    ///     Z::short(),
```
sffc, line 990: This test is specifically documented as demonstrating the behavior when the display name is not available.

sffc (changes requested):

PR #5756, opened 2024-10-31
Make TZ fixture more readable

robertbastian:


components/datetime/tests/patterns/tests/time_zones.json
```diff
@@ -2,1100 +2,207 @@
...
-        ],
-        "expected": ["+09"]
-      },
-      {
-        "patterns": [
-          "xx",
-          "XX",
-          "xxxx",
-          "XXXX"
-        ],
```
sffc, line 492: Seems okay although you lost the currently-unreachable ISO-8601 x/X patterns.

sffc (approved):

components/datetime/tests/patterns/tests/time_zones.json
robertbastian: Yep, they can easily be unit tested though.

PR #5757, opened 2024-10-31
Special-case `Z`, `Z[Etc/UTC]` and `Z[Etc/GMT]` in IXDTF parser

robertbastian:
These are the only `Z`s that don't require any calculation.

### DIFF
--- a/components/timezone/src/ixdtf.rs
+++ b/components/timezone/src/ixdtf.rs
@@ -213,7 +213,6 @@ impl<'a> Intermediate<'a> {
                     }),
                 ..
             } => (Some(*offset), true, None),
-            // Z[-0800]
             // Z[America/Los_Angeles]
             IxdtfParseRecord {
                 offset: Some(UtcOffsetRecordOrZ::Z),
--- a/components/timezone/src/ixdtf.rs
+++ b/components/timezone/src/ixdtf.rs
@@ -141,6 +141,7 @@ impl IxdtfParser {
 
 struct Intermediate<'a> {
     offset: Option<UtcOffsetRecord>,
+    is_z: bool,
     iana_identifier: Option<&'a [u8]>,
     date: DateRecord,
     time: TimeRecord,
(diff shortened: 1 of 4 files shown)


components/timezone/src/ixdtf.rs
```diff
@@ -189,14 +202,28 @@ impl<'a> Intermediate<'a> {
...
+            IxdtfParseRecord {
+                offset: Some(UtcOffsetRecordOrZ::Z),
+                tz:
+                    Some(TimeZoneAnnotation {
+                        tz: TimeZoneRecord::Offset(offset),
+                        ..
+                    }),
+                ..
+            } => (Some(*offset), true, None),
             // Z[-0800]
```
sffc, line 216:
```suggestion
```

components/timezone/src/ixdtf.rs
```diff
@@ -262,21 +298,42 @@ impl<'a> Intermediate<'a> {
             self.time.minute,
             self.time.second,
         )?;
```
sffc, line 300: Should we now be checking for `is_z` and failing with `MismatchedTimeZoneFields` in `location_only`?

components/timezone/src/ixdtf.rs
```diff
@@ -227,17 +254,26 @@ impl<'a> Intermediate<'a> {
...
     fn offset_only(self) -> Result<UtcOffset, ParseError> {
-        let Some(offset) = self.offset else {
+        let None = self.iana_identifier else {
             return Err(ParseError::MismatchedTimeZoneFields);
         };
-        let None = self.iana_identifier else {
+        if self.is_z {
+            if let Some(offset) = self.offset {
+                if offset != UtcOffsetRecord::zero() {
+                    return Err(ParseError::RequiresCalculation);
```
sffc, line 271: This is rejecting `Z[-0800]`, right? Seems like `RequiresCalculation` isn't quite the right error case?

sffc (dismissed):

components/timezone/src/ixdtf.rs
robertbastian: We need to subtract 8h from the timestamp, which is definitely a calculation. The only non-calculation offset is `Z[+0000]`.

components/timezone/src/ixdtf.rs
robertbastian: I think so.

Manishearth (approved):

sffc (approved):

PR #5758, opened 2024-10-31
Concept: finely sliced timezones

robertbastian:
This mainly exists as a baseline for #5759


sffc:
Is this the headline number?

```
time_zone/location@1, <lookup>, 387197B, 48313 identifiers
time_zone/location@1, <total>, 1362991B, 623319B, 32160 unique payloads
```

1.3 MB is nice, but I'm also glad that the root-based approach isn't _too much_ worse at 1.4 MB.

robertbastian:
The headline number is the 0.6MB in postcard.

sffc:
I wish we could realize the size reduction. Hmm.

Do you have data on which display names were successfully deduplicated in this PR that aren't deduplicated in the inherit-from-root PR?

PR #5759, opened 2024-10-31
Deduplicate tz locations against root

robertbastian:
A lot of location names are the same in many languages. 

For exemplar cities we have root data, however for territories we don't. For territories I'm resolving the `und` name by using a Latin-script endonym (i.e. for region `XY`, resolve `und-Latn-XY`), and if that fails, just use English (not perfect, but it's better than the raw region code for both display purposes and deduplication).


components/datetime/src/format/neo.rs
```diff
@@ -545,6 +547,8 @@ pub(crate) struct RawDateTimeNames<R: DateTimeNamesMarker> {
         >,
     zone_essentials:
         <R::ZoneEssentials as DateTimeNamesHolderTrait<tz::EssentialsV1Marker>>::Container<()>,
+    locations_root:
+        <R::ZoneLocations as DateTimeNamesHolderTrait<tz::LocationsV1Marker>>::Container<()>,
```
sffc, line 551: Thought: This makes `DateTimeFormatter<YMDHMSZ>` a little bit bigger in stack size.

provider/data/datetime/fingerprints.csv
```diff
@@ -35617,225 +35617,225 @@ time_zone/generic_short@1, yrl, -> kgp
 time_zone/generic_short@1, zh-Hant, 176B, 84B, 1944b724241800b8
 time_zone/generic_short@1, zh-SG, -> ta-MY
 time_zone/locations@1, <lookup>, 1097B, 218 identifiers
-time_zone/locations@1, <total>, 1979623B, 1951879B, 204 unique payloads
-time_zone/locations@1, af, 8161B, 8025B, 1cb532d1ca53b263
-time_zone/locations@1, ak, 8179B, 8043B, 61a5ec381137eb44
+time_zone/locations@1, <total>, 1435127B, 1407337B, 204 unique payloads
```
sffc, line 35620: Praise: 1'979'623B to 1'435'127B is a big enough improvement that I'm happy landing this despite my other comments.

sffc (approved):
I had mixed feelings about this, since mixing data from multiple CLDR versions could cause problems. But, I think the benefits outweigh the drawbacks. See #5763.

PR #5762, opened 2024-11-01
Split specific names map

robertbastian:
This is actually only a 2.5% data saving 🤷


Manishearth (approved):

sffc (changes requested):
It makes some locales bigger because now you have 4 ZeroMaps instead of 2 in the fields of the formatter.

Can you write this such that we don't have more ZeroMaps?

This directly impacts memory use over FFI.

robertbastian:
> Can you write this such that we don't have more ZeroMaps?

No, that's the purpose of this PR.

Don't we have size tests for neo formatter? They seem to be unimpacted.

robertbastian:
* This increases stack size, and heap size in FFI
* This makes adding more zone variants a data-breaking change in the future
* This slightly decreases data size

Conclusion: don't merge, maybe investigate other models in the future (@sffc, @robertbastian)

PR #5765, opened 2024-11-03
Add Chinese collator tests

### LINKED ISSUE #1964: Ensure that the provider performs correct alias mapping for Traditional Chinese locales
Ensure that if a specific (and existing) collation hasn't been specified with `-u-co-`, the following map to `zh-u-co-stroke`:
 * `zh-Hant` regardless of region.
 * `zh` without `Hans` but with any of `HK`, `MO`, `TW`.
 * `yue` without either `Hans` or `CN`.

robertbastian:
Fixes #1964

### DIFF
--- a/components/collator/tests/tests.rs
+++ b/components/collator/tests/tests.rs
@@ -921,12 +921,32 @@ fn test_vi_owned() {
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


hsivonen (approved):
LGTM. Thanks!

PR #5772, opened 2024-11-05
Add TinyAsciiStr::concat

sffc:
#5729

### DIFF
--- a/utils/tinystr/src/ascii.rs
+++ b/utils/tinystr/src/ascii.rs
@@ -281,6 +281,67 @@ impl<const N: usize> TinyAsciiStr<N> {
         unsafe { TinyAsciiStr::from_utf8_unchecked(bytes) }
     }
 
+    #[inline]
+    #[must_use]
+    /// Returns a `TinyAsciiStr<Q>` with the concatenation of this string,
+    /// `TinyAsciiStr<N>`, and another string, `TinyAsciiStr<M>`.
+    ///
+    /// If `Q < N + M`, the string gets truncated.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use tinystr::tinystr;
+    /// use tinystr::TinyAsciiStr;
+    ///
+    /// let abc = tinystr!(6, "abc");
+    /// let defg = tinystr!(6, "defg");
+    ///
+    /// // The concatenation is successful if Q is large enough...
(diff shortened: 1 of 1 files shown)


sffc:
I considered a few other signatures. I picked this one because:

1. If the input were a `&str`, we'd need to check each character again.
2. If you want to append a `&str`, you can first convert it to a `TinyAsciiStr` of the same size as the output.
3. I saw no good reason to enforce that `Q == M + N` because `resize` already is documented to do the truncation thing, and it saves the client from having to call `resize` themselves.

utils/tinystr/src/ascii.rs
```diff
@@ -281,6 +281,67 @@ impl<const N: usize> TinyAsciiStr<N> {
...
+    /// );
+    /// assert_eq!(
+    ///     abc.concat(defg),
+    ///     tinystr!(2, "ab")
+    /// );
+    /// ```
+    pub const fn concat<const M: usize, const Q: usize>(
+        self,
+        other: TinyAsciiStr<M>,
+    ) -> TinyAsciiStr<Q> {
```
zbraniecki, line 331: `nit`: I think you should debug assert that Q = N+M.

Manishearth (approved):

utils/tinystr/src/ascii.rs
sffc: Did you see my comment above about why I didn't add the assertion? It's fine if you disagree and want me to add the assertion, but I wanted to verify that you read my reasoning and are making this comment despite it

sffc: Please also note that the only use case for this function will involve `.concat().resize()` if I force `Q` to be `N + M`

sffc:
@zbraniecki You can submit this (to unblock your PR) if you agree with my logic about _not_ adding the assertion.

utils/tinystr/src/ascii.rs
zbraniecki: I don't think I understand your logic. My interpretation of your code is that I can now do:

```
let t: TinyAsciiStr<12> = tinystr!(4, "abcd").concat(tinystr!(4, "efgh"));
```

zbraniecki:
I'm fine merging it as is. No blockers.

robertbastian (approved):

PR #5773, opened 2024-11-05
Remove unsupported datetime fields

robertbastian:
https://github.com/unicode-org/icu4x/issues/4336

Blocked by #5774


Manishearth (approved):
Let's drop a link to this PR on the issue for resurrecting these so it's easier to find

Manishearth (dismissed):

Manishearth (dismissed):

sffc (commented):
This removal is more pervasive than I had been hoping, and I expect that we will be reverting this PR in the near future, since adding back week support was a shortcut I took in order to try to get semantic skeleta done for 2.0.

https://github.com/unicode-org/icu4x/issues/5643

It might be less work overall to just add support for the field instead of deleting over 1000 lines across 43 files that will just need to be manually added back again. In other words, it makes https://github.com/unicode-org/icu4x/issues/5643 a bigger task than it was supposed to be.

sffc (changes requested):
I'm going to put a soft block on this PR because I'm not convinced it is the best engineering project management decision. Removing an error variant from an enum that already has 5 or 6 error variants is not something most users care about.

PR #5774, opened 2024-11-05
Remove ResolvedNeoTimeZoneSkeleton

robertbastian:
This is blocking #5773


components/datetime/src/fields/symbols.rs
```diff
@@ -676,54 +675,36 @@ field_type!(
...
-        /// Field symbol for either the ISO8601 basic format or ISO8601 extended format.  This does not allow an
-        /// optional ISO8601 UTC indicator `Z`, whereas [`TimeZone::LowerX`] allows the optional `Z`.
-        'X' => UpperX = 6,
+        'V' => Location = 3,
+        /// Field symbol for either the ISO-8601 basic format or ISO-8601 extended format. This does not use an
+        /// optional ISO-8601 UTC indicator `Z`, whereas [`TimeZone::IsoWithZ`] produces `Z`.
+        'x' => Iso = 4,
+        /// Field symbol for either the ISO-8601 basic format or ISO-8601 extended format, with the ISO-8601 UTC indicator `Z`.
+        'X' => IsoWithZ = 5,
     };
```
robertbastian, line 694: I don't know what @sffc and @Manishearth discussed before, but after this PR, custom behaviour could be added like this:

```suggestion
    _ => WhateverChromeDoes = 7,
    };
```

This variant can be constructed as `TimeZone::WhateverChromeDoes`, but doesn't have an associated char for parsing.

components/datetime/src/fields/symbols.rs
```diff
@@ -676,54 +675,36 @@ field_type!(
...
-        'x' => LowerX = 5,
-        /// Field symbol for either the ISO8601 basic format or ISO8601 extended format.  This does not allow an
-        /// optional ISO8601 UTC indicator `Z`, whereas [`TimeZone::LowerX`] allows the optional `Z`.
-        'X' => UpperX = 6,
+        'V' => Location = 3,
+        /// Field symbol for either the ISO-8601 basic format or ISO-8601 extended format. This does not use an
+        /// optional ISO-8601 UTC indicator `Z`, whereas [`TimeZone::IsoWithZ`] produces `Z`.
+        'x' => Iso = 4,
+        /// Field symbol for either the ISO-8601 basic format or ISO-8601 extended format, with the ISO-8601 UTC indicator `Z`.
+        'X' => IsoWithZ = 5,
```
sffc, line 693: question: was this documentation backwards before, or is it backwards now? The polarity flipped.

sffc (approved):
Okay.

Please make sure that all time zone symbols that are reserved in CLDR are not accepted in ICU4X. For example, CLDR does not define `vv` or `vvv`, I think, and we shouldn't, either.

sffc:
Here is the post where I compared and contrasted the different approaches to the time zone intermediate type. There was a somewhat involved discussion with @Manishearth, @zbraniecki, and myself.

https://github.com/unicode-org/icu4x/pull/5024#issuecomment-2157351593

I had implemented option 2ii. This PR basically reverts us back to option 1.

components/datetime/src/fields/symbols.rs
robertbastian: It was backwards before, which is much easier to notice if the things aren't called `UpperX` and `LowerX`.

PR #5789, opened 2024-11-07
Audit datetime format function

robertbastian:
I've found several bugs:
* `FieldLength::TwoDigit` truncates the number, which is incorrect. Only the `yy` and `YY` patterns require truncation. I've moved this code to the year branch, and renamed the variant `Two`
* `c..cc` and `e..ee` require locale-sensitive day-of-week-number calculations, so they are now rejected during data loading
* `FieldLength::{Abbreviated, Wide, Narrow}` also apply to numeric fields, where they mean 3, 4, 5. I renamed them and updated documentation.
* Cyclic year names should be loaded from data (#3761)


sffc:
The names of the FieldLength enum variants predate my time working on this crate. I assume they were chosen to reflect the primary use case for those widths, even if they mean different things in different contexts

components/datetime/src/format/datetime.rs
```diff
@@ -152,42 +137,37 @@ where
             input!(year = input.year);
             input!(cyclic = year.cyclic());
 
-            // TODO(#3761): This is a hack, we should use actual data for cyclic years
-            let cyclics: &[&str; 60] = match input.any_calendar_kind {
-                Some(AnyCalendarKind::Dangi) => &[
```
sffc, line 163: Where did you move the cyclic year names?

sffc (dismissed):

components/datetime/src/format/datetime.rs
```diff
@@ -167,8 +160,10 @@ where
             input!(year = input.year);
             input!(related_iso = year.related_iso());
 
-            // TODO: This should use a latin FDF
-            try_write_number(w, fdf, related_iso.into(), l)?
+            FixedDecimal::from(related_iso)
```
sffc, line 163: Nit: Add comment that `impl Writeable for FixedDecimal` is what the spec says to use here

Manishearth (approved):

sffc (approved):

Manishearth (dismissed):

sffc (approved):

Issue #4478, opened 2023-12-20
Add eraDisplay support to NeoDateTimeFormatter


sffc:
Once open design questions are resolved in https://github.com/tc39/proposal-intl-eradisplay, we should implement it in ICU4X NeoDateTimeFormatter. I already have some stub code that needs to be filled in with the final algorithm.

sffc:
Re-opening and moving to a different milestone now that the core 2.0 work is landed

sffc:
New issue: https://github.com/unicode-org/icu4x/issues/6010

PR #5790, opened 2024-11-07
Add `YearAmbiguity`

robertbastian:
#4478

The input should declare how ambiguous the year is. There should not be any calendar-specific logic in `icu_datetime`.

### DIFF
--- a/components/calendar/src/types.rs
+++ b/components/calendar/src/types.rs
@@ -136,6 +136,7 @@ impl YearInfo {
 /// For example 2024 AD can be formatted as `2024`, or even `24`, but 1931 AD
 /// should not be formatted as `31`, and 2024 BC should not be formatted as `2024`.
 #[derive(Copy, Clone, Debug, PartialEq)]
+#[allow(clippy::exhaustive_enums)] // logically complete
 pub enum YearAmbiguity {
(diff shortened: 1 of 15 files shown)


sffc (approved):

components/calendar/src/types.rs
```diff
@@ -84,6 +84,14 @@ impl YearInfo {
         }
     }
 
+    /// Get the year ambiguity.
+    pub fn year_ambiguity(self) -> YearAmbiguity {
+        match self.kind {
+            YearKind::Cyclic(_) => YearAmbiguity::EraRequired,
+            YearKind::Era(e) => e.ambiguity,
+        }
+    }
```
sffc, line 93: Hi @robertbastian 

Can you explain why you set cyclic calendars to `YearAmbiguity::EraRequired`?

This assumption still exists on main, though it has moved to components/datetime/src/raw/neo.rs.

robertbastian: I think just so that it resolves to `PackedSkeletonVariant::Standard` like before this PR, when Dangi/Chinese was set to `YearDistance::Near`. `YearAmbiguity::Unambiguous` would also work, but the cyclic years really aren't unambiguous...

PR #5792, opened 2024-11-07
Fall back from specific non-location to specific location

robertbastian:
https://unicode-org.atlassian.net/browse/CLDR-18074


provider/data/datetime/fingerprints.csv
```diff
@@ -33657,116 +33657,118 @@ time_zone/locations@1, zh-MO, -> zh-HK
 time_zone/locations@1, zh-SG, 9163B, 9027B, cdda2647bc617a7a
 time_zone/locations@1, zu, 8939B, 8803B, 4ffe068623c9adb7
 time_zone/metazone_period@1, <singleton>, 11097B, 11009B, dcfa50fb39ebd549
-time_zone/specific_long@1, <lookup>, 948B, 190 identifiers
-time_zone/specific_long@1, <total>, 1539467B, 1522816B, 185 unique payloads
-time_zone/specific_long@1, af, 6263B, 6173B, f2cd1f83edbbc556
+time_zone/specific_long@1, <lookup>, 973B, 195 identifiers
+time_zone/specific_long@1, <total>, 1464039B, 1447028B, 189 unique payloads
```
sffc, line 33661: Thought: I expected a bigger data size reduction here.

sffc (approved):

provider/data/datetime/fingerprints.csv
robertbastian: The data is very incosistent. Ideally there would be more deduplication in future CLDR versions.

PR #5803, opened 2024-11-10
Migrate decimal and experimental to Preferences

zbraniecki:
Migrated experimental formatters to Preferences

### DIFF
--- a/components/experimental/src/compactdecimal/formatter.rs
+++ b/components/experimental/src/compactdecimal/formatter.rs
@@ -14,8 +14,10 @@ use crate::compactdecimal::{
 use alloc::borrow::Cow;
 use core::convert::TryFrom;
 use fixed_decimal::{CompactDecimal, FixedDecimal};
-use icu_decimal::FixedDecimalFormatter;
-use icu_locale_core::preferences::define_preferences;
+use icu_decimal::{FixedDecimalFormatter, FixedDecimalFormatterPreferences};
+use icu_locale_core::preferences::{
+    define_preferences, extensions::unicode::keywords::NumberingSystem,
+};
 use icu_plurals::PluralRules;
 use icu_provider::DataError;
 use icu_provider::{marker::ErasedMarker, prelude::*};
@@ -24,7 +26,9 @@ use zerovec::maps::ZeroMap2dCursor;
 define_preferences!(
     /// The preferences for compact decimal formatting.
     CompactDecimalFormatterPreferences,
-    {}
+    {
+        numbering_system: NumberingSystem
+    }
 );
 
 /// A formatter that renders locale-sensitive compact numbers.
@@ -99,7 +103,7 @@ impl CompactDecimalFormatter {
         );
         Ok(Self {
             fixed_decimal_formatter: FixedDecimalFormatter::try_new(
-                prefs.locale_prefs.into(),
+                (&prefs).into(),
                 options.fixed_decimal_formatter_options,
             )?,
             plural_rules: PluralRules::try_new_cardinal(prefs.locale_prefs.into())?,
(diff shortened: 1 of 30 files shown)


components/experimental/src/compactdecimal/formatter.rs
```diff
@@ -15,11 +15,18 @@ use alloc::borrow::Cow;
...
+use icu_locale_core::preferences::define_preferences;
 use icu_plurals::PluralRules;
 use icu_provider::DataError;
 use icu_provider::{marker::ErasedMarker, prelude::*};
 use zerovec::maps::ZeroMap2dCursor;
 
+define_preferences!(
+    /// The preferences for compact decimal formatting.
+    CompactDecimalFormatterPreferences,
+    {}
```
sffc, line 27: I think this should retain the numbering system

components/experimental/src/compactdecimal/formatter.rs
```diff
@@ -78,26 +85,28 @@ impl CompactDecimalFormatter {
...
         options: CompactDecimalFormatterOptions,
     ) -> Result<Self, DataError> {
-        let temp_loc = locale.clone().into_locale();
+        let locale = DataLocale::from_preferences_locale::<ShortCompactDecimalFormatDataV1Marker>(
+            prefs.locale_prefs,
+        );
         Ok(Self {
             fixed_decimal_formatter: FixedDecimalFormatter::try_new(
-                locale,
+                prefs.locale_prefs.into(),
```
sffc, line 102: And then down here, forward the numbering system into the FixedDecimalFormatter

components/locale_core/src/preferences/mod.rs
```diff
@@ -529,6 +529,18 @@ macro_rules! __define_preferences {
             }
         }
 
+        impl From<$crate::preferences::LocalePreferences> for $name {
+            fn from(locale_prefs: $crate::preferences::LocalePreferences) -> Self {
```
sffc, line 533: Issue: I don't know if I am onboard with this impl. It makes it easy to forget to forward the extensions from one preference bag to another preference bag. This is a source of bugs in ICU.

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -15,6 +16,12 @@ use super::CurrencyCode;
 
 extern crate alloc;
 
+define_preferences!(
+    /// The preferences for currency formatting.
+    CurrencyFormatterPreferences,
+    {}
```
sffc, line 22: Issue: This one should definitely take the numbering system.

components/experimental/src/dimension/percent/formatter.rs
```diff
@@ -17,6 +18,12 @@ use super::options::PercentFormatterOptions;
 
 extern crate alloc;
 
+define_preferences!(
+    /// The preferences for percent formatting.
+    PercentFormatterPreferences,
+    {}
+);
+
```
sffc, line 26: Issue: Numbering system

sffc (changes requested):
Yeah, so all the things that forward to FixedDecimalFormatter should retain the numbering system from the input locale. `impl From<LocalePreferences> for *Preferences` is a footgun because it drops those extensions.

components/locale_core/src/preferences/mod.rs
zbraniecki: that makes sense. I replaced it with a macro to generate conversions.

components/experimental/src/compactdecimal/formatter.rs
```diff
@@ -14,12 +14,30 @@ use crate::compactdecimal::{
...
+        numbering_system: NumberingSystem
+    }
+);
+
+prefs_convert!(
+    CompactDecimalFormatterPreferences,
+    FixedDecimalFormatterPreferences,
+    { numbering_system }
+);
+prefs_convert!(CompactDecimalFormatterPreferences, PluralRulesPreferences);
```
sffc, line 39: Praise: This `prefs_convert` thing seems like a good solution

sffc (approved):
Nice work!

PR #5805, opened 2024-11-12
Refactor size_test_macro and use it in icu_decimal

sffc:
Test for #4437


sffc:
I didn't implement `icu4x_shared` because I didn't want this PR to be controversial. It should merge before #5804 so we can measure the size improvement.

Manishearth (approved):

PR #5807, opened 2024-11-12
Delete old neo skeleton structs

sffc:
#1317

Depends on #5761
Depends on #5810


sffc:
cd61b8765626b76eaf0c3e6dbaaf1dd116029ce8 addresses https://github.com/unicode-org/icu4x/pull/5761#discussion_r1839004988

PR #5811, opened 2024-11-13
DateTime Options migration

sffc:
#1317

Removes the old options from `datetime::options` and moves the new options from `datetime::neo_skeleton` to `datetime::options`.

There is one change I am unsure about. Please take note of `datetime::options::components` moving to `provider`, despite it being returned by a pub fn in `DateTimePattern`.

Depends on #5807
Depends on #5814
Depends on #5817


Manishearth (approved):

sffc:
> There is one change I am unsure about. Please take note of `datetime::options::components` moving to `provider`, despite it being returned by a pub fn in `DateTimePattern`.

Working group discussion:

- @sffc You have to use a type from the `provider` module to get resolved components with my new PR. Is this OK? Since it is in `provider`, it is unstable, and we could change it later if needed.
- @nekevss Name the type or the fields DateTimePartial?

Conclusion: OK to put it in the `provider` module for now in 2.0

Manishearth (approved):

PR #5814, opened 2024-11-13
Tweaks on preference enum impls

sffc:
Changes:

- `TryFrom<Value>` to `TryFrom<&Value>`
- `derive(Hash)`


Manishearth (approved):

sffc:
@zbraniecki I'm adding these impls because I'm trying to switch from `icu::datetime::options::preferences::HourCycle` to `icu::locale::preferences::extensions::unicode::keywords::HourCycle`

There's one other impl that is missing. `HourCycle` is a field of `components::Bag` which is used in fixture tests and therefore implements Serde. Should I:

1. Implement Serde on the enums (and if so, how?)
2. Implement Serde on specifically the HourCycle enum
3. Keep using a separate `HourCycle` enum in `components::Bag`

sffc:
> 4. Stop using `components::Bag` in Serde directly and instead use a fixture-specific wrapper (or remove it entirely: I don't think these fixtures are used any more)

I'm doing this option

sffc:
@zbraniecki I'm going to merge this because I don't think it's super controversial but if you have feedback we can do it in a follow-up

PR #5826, opened 2024-11-15
Refactoring and tests for dateSkeleton semantic pattern generation

sffc:
Research and preliminary work for #5806 (https://unicode-org.atlassian.net/browse/CLDR-14993)


sffc:
I need a suggestion involving how to drive @gregtatum's (non-semantic) skeleton API.

Currently I use a `components::Bag` as the intermediate type. This lets me do things like change the width of the year or set a custom hour cycle when performing skeleton selection.

This has been working fine, but today I discovered that it seems to mess up weekday patterns in a few locales such as `fi` where there is a difference between `c` and `E`. It seems that `components::Bag` will always look up `E` and not `c`.

The type `Skeleton` supports `c` vs `E`, but it has a much more limited API, as essentially only a list of fields.

Do you have a suggestion on how to drive the skeleton selection better?

@zbraniecki

Manishearth (approved):

PR #5847, opened 2024-11-20
Rename datetime format functions and traits

sffc:
#5837

### DIFF
--- a/components/datetime/src/neo.rs
+++ b/components/datetime/src/neo.rs
@@ -400,7 +400,7 @@ where
     /// let datetime = DateTime::try_new_iso(2024, 5, 8, 0, 0, 0).unwrap();
     ///
     /// assert_try_writeable_eq!(
-    ///     formatter.convert_and_format(&datetime),
+    ///     formatter.format_converted(&datetime),
     ///     "30 Nisan 5784"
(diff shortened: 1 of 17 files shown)


components/datetime/src/neo.rs
```diff
@@ -563,7 +563,7 @@ where
         datetime: &I,
     ) -> Result<FormattedDateTime, crate::MismatchedCalendarError>
     where
-        I: ?Sized + IsInCalendar + AllInputMarkers<FSet>,
+        I: ?Sized + InCalendar + AllInputMarkers<FSet>,
```
markusicu, line 566: optional: how about InSameCalendar to (a) make it more meaningful and (b) make it more parallel with InFixedCalendar?

Manishearth (dismissed):

components/datetime/src/neo.rs
sffc: Done

Justification: the fully qualified names have good parallelism

- `DateTimeFormatter::format_same_calendar` => `InSameCalendar`
- `FixedCalendarDateTimeFormatter::format` => `InFixedCalendar`

Also, "InCalendar" is too short and could mean something else in the future.

PR #5848, opened 2024-11-20
FormattedDateTime should impl Writeable not TryWriteable

sffc:
#5494


sffc:
Messed up the git branch. I'll open a new PR.

PR #5852, opened 2024-11-22
Fix remaining preferences & options in constructors

### LINKED ISSUE #5839: Constructor Options Bag normalization
With introduction of Preferences we now have the final shape of the first argument to constructors and implemented the normalized model for how to converge Locale and Preferences.

The remaining question is how to handle the second argument which we currently have three approaches to:

1. Many constructors take a custom `OptionsBag`, similar to `PreferencesBag` but for developer-induced options
2. Some constructors take no second argument
3. Finally several take an explicit argument 

It looks like this:

```rust

// OptionsBag
struct FooOptions {
    type: FooType
}

impl Foo {
    pub fn try_new(prefs: FooPreferences, rule_type: FooOptions) -> Result<Self, Error>;
}

// No second argument
impl Foo {
    pub fn try_new(prefs: PluralRulesPreferences) -> Result<Self, Error>;
}

// Explicit argument
impl Foo {
    pub fn try_new(prefs: FooPreferences, type: FooType) -> Result<Self, Error>;
}
```

We have an opportunity to normalize it and conform all constructors to Option 1, if we want to, but it doesn't feel like a natural obvious choice, so I'd like to discuss it.

### How we modeled preferences, locales and options

The current architecture separates `Preferences` which represent *user*-provided values from `Options` which represent *developer*-provided values.
This is novel and different from how ECMA-402 handles the constructors. In ECMA-402 we have `(locale, options)` pair where Options is a bag combining keys set by user and developer and the user-ones get merged with locale extension keys.

In our model the Preferences and Locale get merged as they represent the user-driven keys, and the Options are split into two "types".

First type are options that are affecting data slicing. In that case we push them to the constructor itself to allow for DCE to remove unnecessary code. Second type is options that don't affect DCE. It looks like this:

```rust
enum FooType {
    Bar,
    Baz,
}

// Option as argument
impl Foo {
    pub fn try_new(prefs: FooPreferences, type: FooType) -> Result<Self, Error>;
}

// Option as part of the constructor
impl Foo {
    pub fn try_new_bar(prefs: FooPreferences) -> Result<Self, Error>;
}
impl Foo {
    pub fn try_new_baz(prefs: FooPreferences) -> Result<Self, Error>;
}
```

### Value proposition of normalization

If we were to normalize the second argument I see three main value props:
1. It simplifies the DX by providing consistent API shape for all constructors
2. It allows for introduction of patterns to normalize options merging, converging, converting and cascading (similar to Preferences)
3. It allows for an easy introduction of Option keys that are universal across all constructors - ECMA-402 has one like that called `localeMatcher`

### Why we shouldn't do this

There are several reasons for which we may not want to normalize it:
1. Not all constructors have options. Introducing an empty Bag just for the sake of normalization and asking developers to pass `Default::default()`, `None` or `options!{}` to each constructor just to have consistency may not be worth it.
2. As for ECMA-402 compatibility we'll have to introduce ECMA-402 OptionsBag anyway and write logic to convert from ECMA-402 behavior to ICU4X, so the value of normalized OptionsBag in ICU4X is reduced.
3. The only known universal-option across all components in ECMA-402 - `LocaleMatcher` has questionable value, is basically never used and while we may have to find a way to support it, it is not clear if we should design ICU4X component API around it just because `LocaleMatcher` exists.

Counters:
1. While not all constructors have options, it doesn't cost a lot to have two-argument constructors. We can even argue that the argument should be `Option<OptionsBag>` to allow for `None` if we think it's better DX than `Default::default()`.
2. Building conversion logic from ECMA-402 bag to ICU4X Preferences+Options can be abstracted to macros easier than if each constructor has it's own shape.
3. If we decide that we have to support `LocaleMatcher` (to pass test-262) then we have to find a way to pass this argument to each constructor somehow... Even if no other Options are needed by that constructor.

### Is it urgent?

I'm not sure if it is. It would be nice to have a consistent, ECMA-402 compatible, constructor story for 2.0 and the PR is going to be quite simple to write if we decide to go for it, but it may be that we have different opinions and should give ourselves time to think it through and plan this for 3.0.

zbraniecki:
Partially fixes #5839

### DIFF
--- a/components/calendar/src/week_of.rs
+++ b/components/calendar/src/week_of.rs
@@ -7,10 +7,18 @@ use crate::{
     provider::*,
     types::{DayOfMonth, DayOfYearInfo, IsoWeekday, WeekOfMonth},
 };
+use icu_locale_core::preferences::define_preferences;
 use icu_provider::prelude::*;
 
 /// Minimum number of days in a month unit required for using this module
 pub const MIN_UNIT_DAYS: u16 = 14;
+
+define_preferences!(
+    /// The preferences for list formatting.
+    [Copy]
+    WeekPreferences,
+    {}
+);
 
 /// Calculator for week-of-month and week-of-year based on locale-specific configurations.
 ///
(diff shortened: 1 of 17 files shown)


sffc:
@zbraniecki ok if I finish this and push to your branch?

Manishearth (dismissed):

sffc (approved):

Manishearth (approved):

PR #5855, opened 2024-11-22
Change AnyCalendar constructors

### LINKED ISSUE #5843: AnyCalendar locale constructor naming
https://github.com/unicode-org/icu4x/pull/5842 changed the argument of `AnyCalendar::new_for_locale` to take AnyCalendarPreferences instead of Locale, but it did not change the name of the function.

Please note that there is already another function `AnyCalendar::new` that takes an `AnyCalendarKind`.

Should we:

1. Leave things the way they are now
2. Change the `new` function to work with preferences and change the AnyCalendarKind constructor to be `new_for_kind`
3. Change `new_for_locale` to `new_for_preferences`
4. Add `new_for_preferences` and keep `new_for_locale` the way it was before
5. ?

I would rather not make this decision myself.

Needs input from: @zbraniecki @Manishearth and others

sffc:
Fixes #5843

I think this is better because it is more uniform, and most of the time you have a locale, not an AnyCalendarKind.

### DIFF
--- a/components/calendar/src/any_calendar.rs
+++ b/components/calendar/src/any_calendar.rs
@@ -774,7 +774,7 @@ impl AnyCalendar {
     icu_provider::gen_any_buffer_data_constructors!(
         (prefs: AnyCalendarPreferences) -> error: DataError,
         functions: [
-            new: skip,
+            try_new: skip,
             try_new_with_any_provider,
             try_new_with_buffer_provider,
             try_new_unstable,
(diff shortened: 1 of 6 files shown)


components/calendar/src/any_calendar.rs
```diff
@@ -762,24 +762,25 @@ impl AnyCalendar {
     ///
     /// [📚 Help choosing a constructor](icu_provider::constructors)
     #[cfg(feature = "compiled_data")]
-    pub fn new_for_locale(prefs: AnyCalendarPreferences) -> Self {
+    pub fn try_new(prefs: AnyCalendarPreferences) -> Result<Self, DataError> {
+        // NOTE: This returns DataError because this operation should become data-driven
```
sffc, line 766: Note: I made this return a `DataError` because we should change this to be data-driven instead of code-driven.

Manishearth (approved):

Manishearth (approved):

PR #5874, opened 2024-11-27
build: improve building with -Zminimal-versions

paolobarbolini:
This improves building with `-Zminimal-versions` by increasing the minimum allowed version of postcard to the first one that can be used to successfully compile this project.


Manishearth (approved):
Nice catch, thank you!!

We've wanted "lockfile CI" in the past, would be great to add.

https://github.com/unicode-org/icu4x/issues/2966#issuecomment-1414461599

PR #5456, opened 2024-08-27
Implement Long Currency Formatter

younies:

PR #5879, opened 2024-12-03
Add from_fixed_decimal method to construct CompactDecimal

younies:

### DIFF
--- a/utils/fixed_decimal/src/compact.rs
+++ b/utils/fixed_decimal/src/compact.rs
@@ -32,6 +32,20 @@ impl CompactDecimal {
         }
     }
 
+    /// Constructs a [`CompactDecimal`] from a [`SignedFixedDecimal`].
+    ///
+    /// Example:
+    /// ```
+    /// # use fixed_decimal::CompactDecimal;
+    /// # use fixed_decimal::SignedFixedDecimal;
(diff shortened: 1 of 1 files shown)


utils/fixed_decimal/src/compact.rs
```diff
@@ -32,6 +32,20 @@ impl CompactDecimal {
...
+    ///
+    /// Example:
+    /// ```
+    /// # use fixed_decimal::CompactDecimal;
+    /// # use fixed_decimal::SignedFixedDecimal;
+    /// # use std::str::FromStr;
+    /// #
+    /// assert_eq!(CompactDecimal::from_fixed_decimal(SignedFixedDecimal::from_str("+1.20").unwrap()), CompactDecimal::from_str("+1.20").unwrap());
+    /// ```
+    pub fn from_fixed_decimal(fixed_decimal: SignedFixedDecimal) -> Self {
```
sffc, line 45: Question: why do we want this function when we already have `from_significand_and_exponent`?

younies: Yes, we do not need it.

I used `format_fixed_decimal` to format `SignedFixedDecimal` in the `CompactFormatter`.

See: https://github.com/unicode-org/icu4x/pull/5456

PR #5888, opened 2024-12-11
Add `experimental` to datetime's `bench` features

Nerixyz:
Both benchmark suites in the `datetime` component use the `CompositeFieldSetSerde` which is under the `experimental` feature. As a result, running `cargo bench -F bench --manifest-path components/datetime/Cargo.toml` will not run any benchmark. This currently causes CI to fail on the main branch (e.g. https://github.com/unicode-org/icu4x/actions/runs/12279258426/job/34262931118).

This PR adds the `experimental` feature to the `bench` feature. Alternatively, the workflow could specify additional features to pass to a component's benchmarks. However, when running the benchmarks locally, one would always need to specify the `experimental` feature.

### DIFF
--- a/components/datetime/Cargo.toml
+++ b/components/datetime/Cargo.toml
@@ -97,7 +97,7 @@ datagen = [
 ]
 logging = ["icu_calendar/logging"]
 experimental = ["dep:litemap"]
-bench = ["serde", "datagen"]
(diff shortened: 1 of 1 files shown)


sffc (approved):

sffc:
Thank you!

PR #5890, opened 2024-12-11
Update to CLDR 46.1 beta1

sffc:
Mostly clean except there is an error in transliterator datagen.

The fixes to datetime patterns won't happen until #5826 lands.

PR #5891, opened 2024-12-11
Fix transforms datagen to work with both 46.0 and 46.1?

sffc:
Split from #5890


Manishearth (approved):

robertbastian:
I don't understand why this change is needed. 46.1 should not require any replacements, if it does then we should fix in 46.1.

sffc:
The 46.0 replacement was too aggressive and broke 46.1. This PR makes it so that the replacement targets 46.0 but not 46.1.

Issue #5782, opened 2024-11-06
Implement neo skeleton fractional second fallback correctly


sffc:
The time precisions called "Plus" should render every field in the time, including fractional second digits when available. It seems this is not yet implemented.

Not a 2.0 Beta blocker, but we should fix for 2.0 Final

sffc:
Related: https://github.com/unicode-org/icu4x/issues/5883

sffc:
New issue: #6008

PR #5898, opened 2024-12-13
DateTime + Decimal Parts Formatting

sffc:
We need this in order to support fractional second digits in #5845

Depends on #5896
Depends on #5897


sffc:
Stumbled across https://github.com/unicode-org/icu4x/issues/5782 again

Manishearth (approved):

PR #5909, opened 2024-12-17
DateTimeFormatter numbering_system getter

sffc:
Depends on #5907

Another part of #5900


sffc:
See the issue for another approach to do this.

PR #5959, opened 2025-01-07
Make `field` module be unstable and add stable enums for DateTimeNames

sffc:
The `field` module should be internal. I'm moving it to `provider`. It was used in the power-user `DateTimeNames` API which I am cleaning up in this PR.


sffc:
This is bigger than I was planning, but I think it's a positive change.

Manishearth (approved):
Praise: this is more types, but overall more comprehensible IMO

Issue #5901, opened 2024-12-14
Consider using script fallback for time zone names and maybe others


sffc:
The data size optimization landed in https://github.com/unicode-org/icu4x/pull/5751 reduces data size in English and other Latin-script languages, but it has minimal impact on other languages.

To reduce this bias, should we consider having a `und-Script` locale for time zone names. It can contain the names in the language according to likely subtags. This may help languages that share a script, such as sr/uk/ru or zh/zh-Hant/yue.

@robertbastian

robertbastian:
There might be political issues around region display names that need the region to be resolved (and time zone names are often region names).

robertbastian:
It looks like no script-sharing language overlaps on all names.

robertbastian:
Ah this makes sense of course, the payloads would have already been deduplicated before. I'm not sure I understand you proposition. We do fallback from non-location names to location names in the same language. What fallback exactly are you suggesting?

sffc:
Right now we load two payloads, one for the language and one for `und`. I was suggesting that we instead load two payloads, one for the language and one for the script: `und-Latn` for en, fr, vi, ..., `und-Hani` for zh, yue, zh-Hant, ..., `und-Cyrl` for sr, ru, uk, ...

sffc:
PR #5960 is not quite what I had in mind. What I had in mind was that, for example:

1. `und-Cyrl` contains everything from `ru`, chosen because `ru` is the likely subtag language for `und-Cyrl`
2. `ru` is empty and falls back to `und-Cyrl`
3. `uk` gets smaller and contains only those names that differ from `und-Cyrl` == `ru`
4. `sr` gets smaller and contains only those names that differ from `und-Cyrl` == `ru`

Just like how we made `und` contain English data and then `en` variants are smaller.

robertbastian:
Ah you mean the deduplication from #5759, not #5751

robertbastian:
I've merged deduplicating to a script representative, but I'm not convinced that that gives us minimal data size. If the script representative is the only language using a particular name, but many others share a name, this is not optimal.

PR #5966, opened 2025-01-09
Dedupe location names against a script representative

robertbastian:
Part of #5901 

I'm not introducing script fallback (because time zone names are not meant to fall back by script), and I'm not introducing `und-Xxxx` locales, as those would increase data size (due to the other fields in the struct). Instead, I store a `dedupe_locale` in each data struct, which contains the locale that was deduplicated against. So for `sr`, this would be `ru`, because `sr` -> `und-Cyrl` -> `ru`, and for `ru` it'd be `ru` as well, and `ru` is complete. This way, we also directly know which dedupe locale to load; if we didn't store it we'd have to load a fallbacker and step through it until we hit `und` (and then use the previous value).


sffc:
> Fixes #5901

Do we not do similar deduplication in the other time zone keys as well?

sffc:
So the problem with `dedupe_locale` is that the fallback engine has no knowledge that it needs to keep both payloads around. If you pass `LocaleFamily("sr")` into datagen, datagen is free to remove `ru`, but if we redefine the key to script fallback, it will know that it needs to retain `und-Cyrl`. (well, unless no-ancestor mode is used.)

sffc:
Hmm. An approach that would work with our framework would be to have two keys:

1. `location_base`
2. `location_override`

Every locale has an entry in each key, but `location_base` has one unique payload per script, and `location_override` contains smaller payloads, which are empty if the data exactly equals the base.

sffc (changes requested):
Requesting changes because this code won't work with data slicing.

robertbastian:
I do not want to change this to script fallback, because it should definitely do region fallback.

The two-key approach is pretty complex, the base keys would have to contain the intersection of all data structs that are dedupe against the locale, which is hard generate in our current framework (it's also not clear if grouping by script leads to the most deduping).

This currently has a data size improvement of <10%, I'm not sure it's worth pursuing at the moment.

sffc:
> I do not want to change this to script fallback, because it should definitely do region fallback.

Hm? Right now it is language fallback. Script fallback is identical except that it adds `und-Script` in the fallback chain.

> The two-key approach is pretty complex, the base keys would have to contain the intersection of all data structs that are deduped against the locale, which is hard to generate in our current framework (it's also not clear if grouping by script leads to the most deduping).

Hm? I don't understand this, either. Generating the base key is trivial. It is equal to the language data for the likely language of the locale's script. We could do something fancier but that's not what I suggested above

> This currently has a data size improvement of <10%, I'm not sure it's worth pursuing at the moment.

1'407'337 B to 1'297'285 B, or 110 kB.

It's about the raw data size improvement, yes, but also about language equity and not privileging English.

sffc:
@robertbastian Can you clarify your comment about why you don't want to use script fallback?

robertbastian:
> Hm? Right now it is language fallback. Script fallback is identical except that it adds und-Script in the fallback chain.

I guess that's right.

> Generating the base key is trivial. It is equal to the language data for the likely language of the locale's script.

It is not. `en` contains US-specific display names (like "CT") that `en-GB` should not fall back to.

provider/data/datetime/fingerprints.csv
```diff
@@ -27500,204 +27500,266 @@ time_zone/essentials@1, ur-IN, -> und
 time_zone/essentials@1, vec, -> eo
 time_zone/essentials@1, xnr, 173B, 81B, 331c125b1ab74aab
 time_zone/essentials@1, yo, 109B, 17B, 6f563aca2d7a093
-time_zone/exemplars@1, <lookup>, 969B, 196 identifiers
-time_zone/exemplars@1, <total>, 488149B, 480228B, 179 unique payloads
+time_zone/exemplars@1, <lookup>, 885B, 182 identifiers
+time_zone/exemplars@1, <total>, 244737B, 238270B, 146 unique payloads
```
sffc, line 27504: This size went down, and then we have

```
time_zone/exemplars_root@1, <total>, 167494B, 166036B, 33 unique payloads
```

So there is an overall size decrease, yes?

sffc (approved):
The resulting data looks good. Didn't do a thorough review of the datagen code.

provider/data/datetime/fingerprints.csv
robertbastian: Yes, about 10%. Actually the size decrease seems to be better for core languages, as many moderate languages fall back to und values instead of something in their script, which increases their data size.

PR #5982, opened 2025-01-14
Add calendar argument to IXDTF parsing

robertbastian:
Part of #5739

### DIFF
--- a/components/calendar/src/ixdtf.rs
+++ b/components/calendar/src/ixdtf.rs
@@ -4,7 +4,7 @@
 
 use core::str::FromStr;
 
-use crate::{AnyCalendar, Date, DateTime, Iso, RangeError, Time};
+use crate::{AnyCalendarKind, AsCalendar, Calendar, Date, DateTime, Iso, RangeError, Time};
 use ixdtf::parsers::records::IxdtfParseRecord;
 use ixdtf::parsers::IxdtfParser;
 use ixdtf::ParseError as IxdtfError;
@@ -23,6 +23,9 @@ pub enum ParseError {
     MissingFields,
     /// The IXDTF specifies an unknown calendar.
     UnknownCalendar,
+    /// Expected a different calendar.
+    #[displaydoc("Expected calendar {0} but found calendar {1}")]
+    MismatchedCalendar(AnyCalendarKind, AnyCalendarKind),
 }
 
 impl From<RangeError> for ParseError {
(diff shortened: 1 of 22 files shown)


components/datetime/src/pattern/formatter.rs
```diff
@@ -161,13 +161,11 @@ where
     /// use writeable::assert_try_writeable_eq;
     ///
     /// let mut london_winter = IxdtfParser::new()
-    ///     .try_from_str("2024-01-01T00:00:00+00:00[Europe/London]")
-    ///     .unwrap()
-    ///     .to_calendar(Gregorian);
+    ///     .try_from_str("2024-01-01T00:00:00+00:00[Europe/London]", Gregorian)
```
sffc, line 164: Issue: This string should fail to parse because the default calendar is Iso, not Gregorian. Right?

components/calendar/src/ixdtf.rs
```diff
@@ -276,30 +275,31 @@ impl DateTime<AnyCalendar> {
...
     /// Creates a [`DateTime`] in any calendar from an IXDTF syntax string with compiled data.
     ///
     /// See [`Self::try_from_str()`].
-    ///
-    /// ✨ *Enabled with the `compiled_data` and `ixdtf` Cargo features.*
-    #[cfg(feature = "compiled_data")]
-    pub fn try_from_utf8(ixdtf_str: &[u8]) -> Result<Self, ParseError> {
+    pub fn try_from_utf8(ixdtf_str: &[u8], calendar: A) -> Result<Self, ParseError> {
         let ixdtf_record = IxdtfParser::from_utf8(ixdtf_str).parse()?;
+        if let Some(ixdtf_calendar) = ixdtf_record.calendar {
```
sffc, line 287: Question/Issue: Why are you only checking the calendar if it is not None? I think we should be strict.

components/datetime/src/pattern/formatter.rs
robertbastian: It parses the date as ISO and converts to Gregorian.

components/calendar/src/ixdtf.rs
robertbastian: If there's no calendar then the string is ISO. This is the same thing as the `Z[Europe/Zurich]` notation, which is a UTC time that we need to convert to Zurich time (just that we can actually convert calendars).

PR #5985, opened 2025-01-14
Dumb down DateTime

robertbastian:
#5859


components/calendar/benches/datetime.rs
```diff
@@ -1,267 +0,0 @@
...
-
-mod fixtures;
-
-use criterion::{
-    black_box, criterion_group, criterion_main, measurement::WallTime, BenchmarkGroup, Criterion,
-};
-use fixtures::DateFixture;
-use icu_calendar::{AsCalendar, Calendar, DateDuration, DateTime, Time};
-
-fn bench_datetime<A: AsCalendar>(datetime: &mut DateTime<A>) {
```
sffc, line 13: Note: Looks like most/all of these are benched for Date in date.rs

sffc (approved):
Good work!

PR #5986, opened 2025-01-14
Remove bench features

robertbastian:
I was updating the `icu` documentation, which explains the `bench` feature as part of the features. This is unfortunate as it's not a user-facing feature, and we have enough features already. There's an easier way to run only overview benchmarks, I've updated benchmarking.md.


sffc (commented):
Fine with me

Manishearth (approved):
Great, fewer minor features is always good.

Manishearth (approved):

PR #5993, opened 2025-01-15
IXDTF parsers: Improve docs and add PartialEq

sffc:
#5739

This is mostly docs. The only non-docs is adding PartialEq bounds, including to DateTime and ZonedDateTime, which became reachable via one of my new docs tests. I think adding PartialEq to DateTime is not in contradiction to the thin DateTime effort.

### DIFF
--- a/components/timezone/src/ixdtf.rs
+++ b/components/timezone/src/ixdtf.rs
@@ -122,6 +122,12 @@ impl UtcOffset {
 }
 
 #[derive(Debug)]
+/// A parser for [`ZonedDateTime`] based on IXDTF strings.
+///
+/// Any function that takes a calendar argument  returns an error if the
+/// string has a calendar annotation that does not match the calendar
+/// argument.
+///
 /// ✨ *Enabled with the `ixdtf` Cargo feature.*
 pub struct IxdtfParser {
     mapper: TimeZoneIdMapper,
@@ -509,6 +515,11 @@ impl IxdtfParser {
     /// );
     /// assert_eq!(zoneddatetime.zone.zone_variant(), ZoneVariant::Daylight);
     /// let (_, _) = zoneddatetime.zone.local_time();
+    ///
+    /// // Calendar annotations are ignored:
+    /// assert_eq!(zoneddatetime, IxdtfParser::new()
+    ///     .try_iso_from_str("2024-08-08T12:08:19-05:00[America/Chicago][u-ca=hebrew]")
+    ///     .unwrap());
     /// ```
     ///
     /// For more information on date, time, and time zone parsing,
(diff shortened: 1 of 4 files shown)


Manishearth (approved):

components/timezone/src/ixdtf.rs
```diff
@@ -910,7 +923,12 @@ impl FromStr for DateTime<Iso> {
 }
 
 impl<A: AsCalendar> DateTime<A> {
-    /// Creates a [`DateTime`] in any calendar from an IXDTF syntax string with compiled data.
+    /// Creates a [`DateTime`] in any calendar from an IXDTF syntax string.
+    ///
+    /// Returns an error if the string has a calendar annotation that does not
+    /// match the calendar argument.
+    ///
+    /// ✨ *Enabled with the `ixdtf` Cargo feature.*
```
robertbastian, line 931: the type itself has this annotation, we don't put it on every method.

components/timezone/src/ixdtf.rs
```diff
@@ -676,11 +684,11 @@ impl IxdtfParser {
     /// DateTime string.
     ///
     /// ```
-    /// use icu_calendar::Gregorian;
+    /// use icu_calendar::Iso;
     /// use icu_timezone::{IxdtfParser, TimeZoneInfo, UtcOffset};
     ///
     /// let tz_from_offset = IxdtfParser::new()
-    ///     .try_offset_only_from_str("2024-08-08T12:08:19-05:00", Gregorian)
+    ///     .try_offset_only_from_str("2024-08-08T12:08:19-05:00", Iso)
```
robertbastian, line 691: I don't think using `Iso` as the calendar argument is a good example. First of all, you could be using `try_offset_only_from_iso_str`, and second `Iso` doesn't implement `CldrCalendar`, so it needs to be converted before formatting.

sffc: I don't want to overuse Gregorian in docs. It is an antipattern because it is a "strong" calendar in the sense that it overrides the locale calendar during formatting if using a FixedCalendarDTF, which exists for code size only. Better style is to get the calendar from somewhere and then pass it into the parse function, not hardcode something. ISO is slightly better in this example because it can't be used for formatting. I was concerned about this in your PR earlier but didn't want to stop it from landing.

components/timezone/src/ixdtf.rs
sffc: This is on DateTime which is not itself gated.

components/timezone/src/ixdtf.rs
robertbastian: maybe just a mix of calendars then?

robertbastian (approved):

PR #5996, opened 2025-01-15
Expand size of duration records to support ECMA402

jedel1043:
This updates `IsoDurationParser` to support parsing at least the maximum representable duration in the Temporal proposal; Number.MAX_SAFE_INTEGER * 10^9, which is exactly
90,071,992,547,409,910,000,000,000 nanoseconds.

From some quick calculations, changing the time-related types to `u64` is enough to fit the required range:

| Max nanos  |              90071992547409910000000000 |
|------------|----------------------------------------:|
| In seconds |                       90071992547409910 |
| In minutes |                        1501199875790165 |
| In hours   |                          25019997929836 |
| In days    |                            416999965497 |
| In weeks   |                             59571423642 |
| In months  |                             13899998849 |
| In years   |                              1158333237 |
| u64::MAX   |                    18446744073709551615 |

Years, months and weeks don't matter because the specification requires that they must be smaller than 2^32 - 1.

### DIFF
--- a/utils/ixdtf/src/parsers/duration.rs
+++ b/utils/ixdtf/src/parsers/duration.rs
@@ -65,14 +65,14 @@ pub(crate) fn parse_date_duration(cursor: &mut Cursor) -> ParserResult<DateDurat
     let mut previous_unit = DateUnit::None;
 
     while cursor.check_or(false, |ch| ch.is_ascii_digit()) {
-        let mut value: u32 = 0;
+        let mut value: u64 = 0;
         while cursor.check_or(false, |ch| ch.is_ascii_digit()) {
             let digit = cursor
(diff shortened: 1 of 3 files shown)


sffc (approved):
Thanks! 

I recall us making a change in Temporal such that  all the fields could be u64 or smaller. It seems excessive that more than one of them needs to be u64 at the same time. Haven't verified that this is required in the spec.

jedel1043:
@sffc FWIW, it's at least required to pass this test from test262:
https://github.com/tc39/test262/blob/main/test/built-ins/Temporal/Instant/prototype/add/argument-duration-max.js

nekevss (approved):

PR #5998, opened 2025-01-15
Remove ISO parsing specialisations

robertbastian:
Now that we parse `Date<A: AsCalendar>` instead of `Date<AnyCalendar>`, the `Date<Iso>` specializations are not needed anymore.

#5739


robertbastian:
> So, the MismatchedCalendars error variant wouldn't be reachable with this, right?

As it currently isn't. I don't think an `IsoParseError` pulls its weight, there are many other error variants that could happen.

> (acknowledging that this makes DateTime and ZonedDateTime call sites a bit uglier, but in the common call site, users shouldn't need to do this)

In the common call sites users will need this.

> parse_strict takes a calendar argument and requires that the annotation matches, even ISO

`parse_strict` doesn't currently exist, it can be added on top of this PR if we want to.

robertbastian:
Ok, but this does not contradict anything I'm doing here. I'm deduplicating code and API while keeping exactly the same behaviour. Can you not block this on vague API design ideas?

sffc (approved):
It's not the "exact same" behavior in the sense that we lose the ability to do a strict parse into Iso, but ok

PR #6000, opened 2025-01-15
ZonedDateTime FFI

robertbastian:
https://github.com/unicode-org/icu4x/issues/5859


Manishearth (approved):
Yay, fewer opaque types

PR #6004, opened 2025-01-15
Bug fix for subsecond part in `ixdtf`

nekevss:
Boa / `temporal_rs` had a subtraction overflow panic when running the test262 suite in debug. This PR addressed the bug and adds some unit tests for it.

Confirmed that it is correct when plugged into Boa and rerun through the Temporal test suite.

### DIFF
--- a/utils/ixdtf/src/parsers/tests.rs
+++ b/utils/ixdtf/src/parsers/tests.rs
@@ -975,3 +975,77 @@ fn test_zulu_offset() {
         })
     );
 }
+
+// Examples referenced from
+#[test]
+fn subsecond_string_tests() {
+    let subsecond_time = "2025-01-15T15:23:30.1";
+    let result = IxdtfParser::from_str(subsecond_time).parse();
+    assert_eq!(
+        result,
+        Ok(IxdtfParseRecord {
+            date: Some(DateRecord {
+                year: 2025,
+                month: 1,
+                day: 15,
+            }),
+            time: Some(TimeRecord {
+                hour: 15,
+                minute: 23,
+                second: 30,
+                nanosecond: 100_000_000,
+            }),
+            offset: None,
+            tz: None,
+            calendar: None,
+        })
+    );
+
+    let subsecond_time = "2025-01-15T15:23:30.12345678";
+    let result = IxdtfParser::from_str(subsecond_time).parse();
+    assert_eq!(
+        result,
+        Ok(IxdtfParseRecord {
+            date: Some(DateRecord {
+                year: 2025,
+                month: 1,
+                day: 15,
+            }),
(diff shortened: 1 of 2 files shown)


utils/ixdtf/src/parsers/tests.rs
```diff
@@ -975,3 +975,77 @@ fn test_zulu_offset() {
...
+            }),
+            offset: None,
+            tz: None,
+            calendar: None,
+        })
+    );
+
+    let subsecond_time = "1976-11-18T15:23:30.1234567890";
+    let err = IxdtfParser::from_str(subsecond_time).parse();
+    assert_eq!(err, Err(ParseError::FractionPart));
```
robertbastian, line 1050: is a trailing zero in the tenth place explicitly disallowed by the spec?

nekevss: The specification for a [`TemporalDecimalFraction`](https://tc39.es/ecma262/#prod-TemporalDecimalFraction) is specific to 9 digits in length. It's not so much that the digit is a trailing zero or not (it could be a 1, 2, etc.).

That being said, this could also be unbounded digit wise according to [RFC3339 Section 5.6](https://datatracker.ietf.org/doc/html/rfc3339#section-5.6).

I guess it depends how much precision we intend to have in `ixdtf`.

robertbastian: Even if we only store 9 digits, we should not fail if we're given more, but just ignore them.

This is not blocking.

robertbastian (approved):

utils/ixdtf/src/parsers/tests.rs
nekevss: I don't think ignoring them is an option. You'd still need to be able to signal the precision that was parsed, because this would need to be thrown on in [temporal](https://github.com/tc39/test262/blob/main/test/built-ins/Temporal/PlainDateTime/from/argument-string-subsecond.js). You do make a good point though. We could change the fraction parse to an enum to signal the precision, then this could be handled as an error downstream.

```rust
pub enum Fraction {
    Nanoseconds(u32),
    Picoseconds(u64),
    FemtoSeconds(u64),
    Truncated(u32), // Return truncated nanoseconds if precision is beyond 15
}
```

robertbastian: Yeah, the `ixdtf` crate should probably parse them if there are standards that require them, then we'd discard them in ICU4X.

nekevss: I'll work on adding that in a subsequent PR then. I'd prefer to merge this bug fix PR as is, unless you'd like to block on that change.

Issue #5928, opened 2024-12-19
Improve field set builders more


sffc:
Things to do:

- [ ] Add `.with_time_precision()` to date field sets, so the user can write `YMD::long().with_time_precision()`, getting a `YMDT` out the other side
- [ ] Rename `.zone_z()` and friends to `.with_zone_z()`
- [ ] Make these changes on `DateFieldSet` as well

robertbastian:
>  Rename .zone_z() and friends to .with_zone_z()

As mentioned in the other issue, I'd prefer if we said "time zone" instead of "zone". If you're not in the time zone mindset, it might not be clear what "zone" means.

sffc:
Do you prefer:

1. `.with_zone_specific`
2. `.with_specific_zone`
3. `.with_time_zone_specific`
4. `.with_specific_time_zone`
5. `.with_specific_non_location_time_zone`
6. `.with_z_time_zone`
7. `.with_time_zone_z`

Right now I believe the main branch is on 1.

People using these APIs are already in a datetime mindset and will understand what "zone" means. There will be people who are confused by "zone", but this is mitigated with docs.

To demonstrate the utility of "zone" as a self-contained term, please note the following enum and its variants.

https://unicode-org.github.io/icu4x/rustdoc/icu/datetime/fieldsets/enums/enum.CompositeFieldSet.html

PR #6014, opened 2025-01-17
Implement Serde and improve test coverage of FieldSetBuilder

sffc:
Toward #5928, #5940

### DIFF
--- a/components/datetime/src/builder.rs
+++ b/components/datetime/src/builder.rs
@@ -41,6 +41,21 @@
 //!     DateFieldSet::YMD(static_field_set),
 //! );
 //!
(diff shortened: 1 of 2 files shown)


components/datetime/src/builder.rs
```diff
@@ -13,6 +13,8 @@
 //! If the field set is known at compile time, use the static fieldset APIs instead of the
 //! builder exported in this module.
 //!
+//! A field set builder can be serialized with the right set of Cargo features.
```
sffc, line 16: (where the list of Cargo features can be found on the Deserialize impl.)

Manishearth (approved):

PR #6036, opened 2025-01-25
Update `ixdtf` to handle unbound fraction length

nekevss:
This PR updates `ixdtf`'s fraction handling to parse fraction values beyond 9 digits in length per feedback from #6004.

It introduces a new enum `Fraction` in place of the previous `nanosecond` field that allows parsing fraction values beyond 9 digits while preserving the precision in the enum.

It also adds a variety of tests to test the new behavior while adjusting tests from the old behavior to the new behavior


utils/ixdtf/src/parsers/records.rs
```diff
@@ -211,6 +211,26 @@ pub enum TimeDurationRecord {
...
+pub enum Fraction {
+    /// Parsed nanoseconds value (A fraction value from 1-9 digits length)
+    Nanoseconds(u64),
+    /// Parsed picoseconds value (A fraction value from 10-12 digits length)
+    Picoseconds(u64),
+    /// Parsed femtoseconds value (A fraction value from 12-15 digits length)
+    Femtoseconds(u64),
+    /// A parsed value truncated to nanoseconds (A fraction value of 15+ digits length)
+    Truncated(u64), // An unbound fraction value truncated to nanoseconds
+}
```
sffc, line 230: Suggestion: change this to

```rust
pub struct Fraction {
    fraction_digits: u8,
    value: u64,
}
```

and then give it functions such as

```rust
impl Fraction {
    pub fn try_to_nanoseconds(self) -> Option<u64> {
        if self.fraction_digits <= 9 {
            // compute the pow and return
        } else {
            None
        }
    }
    pub fn to_truncated_nanoseconds(self) -> u64 {
        // compute pow, either + or -
    }
}
```

Another future advantage of this model is that the Fraction can directly return a FixedDecimal, which could make fractional second formatting more efficient in icu4x.

components/timezone/src/ixdtf.rs
```diff
@@ -42,6 +42,8 @@ pub enum ParseError {
     InconsistentTimeZoneOffsets,
     /// There was an invalid Offset.
     InvalidOffsetError,
+    /// There was an invalid subsecond fraction.
+    InvalidFraction,
```
robertbastian, line 46: I think this is overkill. I'd rather document that sub-nanoseconds are truncated. The formatting code only supports nanosecond precision anyway.

components/timezone/src/ixdtf.rs
```diff
@@ -326,11 +328,14 @@ impl<'a> Intermediate<'a> {
...
+        let Some(nanosecond) = self.time.fraction.to_nanoseconds() else {
+            return Err(ParseError::InvalidFraction);
+        };
         let time = Time::try_new(
             self.time.hour,
             self.time.minute,
             self.time.second,
-            self.time.nanosecond,
+            nanosecond,
         )?;
```
robertbastian, line 339: Ideally there'd only be one code path that parses the IXDTF struct into `Time`, and it should be `Time::try_from_ixdtf_record`. Please try to use that (also for `Date` ideally).

utils/ixdtf/src/error.rs
```diff
@@ -46,6 +46,8 @@ pub enum ParseError {
     TimeSecond,
     #[displaydoc("Invalid character while parsing fraction part value.")]
     FractionPart,
+    #[displaydoc("Fraction part value exceeds a representable range.")]
+    InvalidFractionRange,
```
robertbastian, line 50: "Invalid" is not very descriptive. This is an error, that already tells me that something is not valid.
```suggestion
    ExcessiveSecondPrecision,
```

nekevss: I was double checking this and the error is specific to a Duration hour fraction and minute fraction potentially exceeding the range of `u64::MAX` when calculated. Maybe `DurationFractionExceededRange` (this would then be moved down into the Duration errors).

components/timezone/src/ixdtf.rs
nekevss: I updated `Time` with a `try_from_time_record` method to address this.

I'm not entirely clear what you had meant by also "for `Date` ideally." Unless the idea was to expose two methods: `Time::try_from_time_record` and `Date::try_from_date_record` on the referenced types. So if you were thinking of that or a bit of a different approach, let me know.

robertbastian: yeah I was thinking of that, but it's more complicated for `Date` because it's in a different crate, so what you have now is fine

utils/ixdtf/src/error.rs
```diff
@@ -104,6 +102,8 @@ pub enum ParseError {
     DateDurationPartOrder,
     #[displaydoc("Invalid time duration part order.")]
     TimeDurationPartOrder,
+    #[displaydoc("Duration part fraction exceeded a representable range.")]
+    DurationFractionExceededRange,
```
robertbastian, line 106:
```suggestion
    DurationFractionalDigitsExceededRange,
```

components/timezone/src/ixdtf.rs
nekevss: Yeah that's why I had paused on it. I wasn't the biggest fan of exposing a public `Date::try_from_date_record` on `Date`.

utils/ixdtf/src/parsers/records.rs
```diff
@@ -211,6 +214,46 @@ pub enum TimeDurationRecord {
...
+
+    /// Returns a `u64` representing the `Fraction` as it's computed
+    /// nanosecond value, truncating any value beyond 9 digits to
+    /// nanoseconds
+    pub fn to_truncated_nanoseconds(&self) -> u64 {
+        if self.digits <= 9 {
+            self.value
+        } else {
+            self.value / 10u64.pow(u32::from(self.digits - 9))
+        }
```
sffc, line 257: Issue: Something here doesn't look right. The code for the `self.digits <= 9` case should be the same in both functions I think. Please make sure this is tested.

nekevss: Huh, that's true. I think when I initially wrote that, it was with the idea of truncating on all cases, but that doesn't align with the function name.

I've updated the behavior and added some doc tests and tests for potentially malformed `Fraction`s since the fields are public.

sffc (approved):
Thought: kind-of wish the `Fraction` type didn't have so many weird invariants. It is also a bit sad that we need to do so much checked arithmetic to get the nanoseconds out of it. There's probably still room to iterate on its design.

Note: the TC decided to start using the language "subsecond" instead of "fractional second" in `icu_datetime`, but I don't know if that also applies to the `ixdtf` crate, and even if it does, it shouldn't block merging this PR.

nekevss:
Agreed on the amount of checking that's needed for arithmetic. I wasn't super happy with it. It could probably be avoided by making `Fraction` non-exhaustive or just removed in favor of GIGO alongside some `debug_assert`s as most of the checks are primarily for somebody constructing a `Fraction` of their own. Most of the arithmetic checks would not affect `Fraction`s output from parsing in `ixdtf`.

I'm more inclined to refer to this as `Fraction` as RFC3339 defines the fraction in ABNF as `time-secfrac` and refers to it elsewhere as "fractional second digits". At a quick glance, the language subsecond does not exist in either RFC3339 or RFC9557.

sffc:
Maybe make `Fraction` have private fields so we can mess with the implementation in the future.

And maybe keep `to_truncated_nanoseconds()` returning a non-Option.

components/time/src/ixdtf.rs
```diff
@@ -723,13 +710,26 @@ impl Time {
...
+        Self::try_from_time_record(&time_record)
+    }
+
+    fn try_from_time_record(time_record: &TimeRecord) -> Result<Self, ParseError> {
+        let nanosecond = time_record
+            .fraction
+            .map(|fraction| {
+                fraction
+                    .to_nanoseconds()
+                    .ok_or(ParseError::ExcessivePrecision)
```
nekevss, line 722: The `ExcessivePrecision` error could be remove here in favor of calling `to_truncated_nanoseconds` if that were preferred.

Would also remove the `transpose` below.

nekevss (commented):
Made some changes to `Fraction` by making the fields private (`pub(crate)`) and updating `digits` to be a `NonZeroU8`.

I think this is cleaner than the previous version.

General question: should there be getters for `digit` and `value`?

robertbastian (commented):
Code looks good, however I'm not happy with the `Fraction` naming. I think `Subsecond` would be clearer, and it's a term we're introducing in `icu_time` and `icu_datetime` right now as well (#5999).

sffc:
What terminology does the RFC use for fractional second / subsecond?

nekevss:
`Subsecond` isn't a totally accurate term when applied to the `TimeDurationRecord` (which actually now that I think about it needs to be adjusted for these changes even more than they currently are). Also, I mentioned it above, the language `Subsecond` does not exist in any of the RFCs, which uses "fractional second digits". I don't disagree with the change being made in reference to #5999, but I do think the context here is a tad different.

Essentially the above would mean that "T1.5H" is parsed as `"T" HOUR_VALUE "." SUBSECOND "H"`, which to me makes less sense than `"T" HOUR_VALUE "." FRACTIONAL_DIGITS "H"`

To be fair, the name could be changed to `FractionalDigits` to more align with the spec.

sffc:
Good point about this field being used for things other than seconds in the `ixdtf` crate.

sffc (approved):

PR #6037, opened 2025-01-27
Fetch tzdb from IANA

robertbastian:
This is preferrable to GitHub because it is the authorative source. They also have a `latest` link.


Manishearth (approved):
In favor of the approach. Mildly unhappy with the 3 new transitive deps, but they're all common / well known crates.

PR #6038, opened 2025-01-27
Fix the docs for HourCycle

### LINKED ISSUE #6030: HourCycle docs are wrong
https://unicode-org.github.io/icu4x/rustdoc/icu/locale/preferences/extensions/unicode/keywords/enum.HourCycle.html says:

> H11
> Hour system using 1–12; corresponds to ‘h’ in patterns
> H12
> Hour system using 0–23; corresponds to ‘H’ in patterns
> H23
> Hour system using 0–11; corresponds to ‘K’ in patterns
> H24
> Hour system using 1–24; corresponds to ‘k’ in pattern

The doc for H24 is correct, but the other three have a mismatch between the doc and the enum value.

Additionally, the docs should try to make an effort to minimize misuse. Specifically, the docs should say:

H24 isn't used anywhere and is _not_ what's usually understood as the "24-hour clock".
H11 is only used in Japan.
H23 is "24-hour clock".
H12 is "12-hour clock".

hsivonen:
Fixes #6030

### DIFF
--- a/components/locale_core/src/preferences/extensions/unicode/keywords/hour_cycle.rs
+++ b/components/locale_core/src/preferences/extensions/unicode/keywords/hour_cycle.rs
@@ -9,12 +9,12 @@ enum_keyword!(
     ///
     /// The valid values are listed in [LDML](https://unicode.org/reports/tr35/#UnicodeHourCycleIdentifier).
     HourCycle {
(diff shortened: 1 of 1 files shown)


nciric (approved):
Thanks!

PR #6042, opened 2025-01-28
Wire up `X` fieldset, remove unused data

robertbastian:
#5658 https://github.com/unicode-org/icu4x/pull/5936


sffc (dismissed):

sffc (approved):
Don't see the diff over the force-push but it still looks okay

PR #6043, opened 2025-01-28
Use `NonZeroU8` as metazone ID

robertbastian:
Metazone IDs don't appear on any APIs anymore, they are now a datetime implementation detail. While the `TinyAsciiStr<4>` representation is nicely readable, it wastes a lot of space: there are only 158 out of the possible 264M metazones in CLDR, and this is not likely to grow beyond 254 anytime soon. This means we can save 3 bytes per ID by just using a numeric ID, which saves 70kB (~10%) in total.


Manishearth (dismissed):

provider/source/src/time_zones/mod.rs
```diff
@@ -150,11 +150,16 @@ impl SourceDataProvider {
...
-                    meta_zone_ids.insert(meta_zone_id_data.long_id.to_string(), *meta_zone_id);
-                }
-                Ok(meta_zone_ids)
+                Ok(meta_zone_ids_resource
+                    .iter()
+                    .enumerate()
+                    .map(|(idx, (_short_id, alias))| {
+                        (
+                            alias.long_id.to_string(),
+                            MetazoneId::new(idx as u8 + 1).unwrap(),
```
sffc, line 159: Issue: is this where you convert from string to number? If we do this type of change, we should define the numbers as a fixed mapping, not dependent on cldr version. If you want to depend on version, add a hash of the list of strings to the data.

robertbastian:
There are multiple options:
* Add an extra bit (byte) to the metazone period struct to mark membership type
* Merge metazone and offset structs, the zone type can be inferred from the offsets
* Don't

components/datetime/src/pattern/names.rs
```diff
@@ -2027,6 +2027,22 @@ impl<FSet: DateTimeNamesMarker> RawDateTimeNames<FSet> {
             .map_err(|e| MaybePayloadError::into_load_error(e, error_field))?
             .map_err(|e| PatternLoadError::Data(e, error_field))?;
         self.load_mz_periods(mz_period_provider, error_field)?;
+        #[allow(clippy::unwrap_used)] // we just loaded them
```
sffc, line 2030: Thought: maybe the load functions should return a reference to the payload it just loaded so you don't need to unwrap

components/datetime/src/provider/time_zones.rs
```diff
@@ -252,9 +232,11 @@ impl<'a> zerovec::maps::ZeroMapKV<'a> for MetazoneId {
...
 #[yoke(prove_covariance_manually)]
-pub struct MetazonePeriodV1<'data>(
+pub struct MetazonePeriodV1<'data> {
     /// The default mapping between period and metazone id. The second level key is a wall-clock time represented as
     /// the number of minutes since the local [`EPOCH`](icu_timezone::provider::EPOCH). It represents when the metazone started to be used.
     #[cfg_attr(feature = "serde", serde(borrow))]
-    pub ZeroMap2d<'data, TimeZoneBcp47Id, IsoMinutesSinceEpoch, Option<MetazoneId>>,
-);
+    pub list: ZeroMap2d<'data, TimeZoneBcp47Id, IsoMinutesSinceEpoch, NichedOption<MetazoneId, 1>>,
+    /// An XxHash64 checksum of the full metazone names.
```
sffc, line 240: Thought: this increases stack size which is a bit sad, but ZM2D is already big on stack. Should consider fixing in the future.

provider/source/src/cldr_serde/time_zones/meta_zones.rs
```diff
@@ -20,7 +19,7 @@ pub(crate) struct MetazoneAliasData {
 }
 
 #[derive(Debug, Clone, Deserialize)]
-pub(crate) struct MetazoneIds(pub(crate) BTreeMap<MetazoneId, MetazoneAliasData>);
+pub(crate) struct MetazoneIds(pub(crate) BTreeMap<tinystr::TinyAsciiStr<5>, MetazoneAliasData>);
```
sffc, line 22: Question: why 5 and not 4

sffc (commented):
Partway through

provider/source/src/cldr_serde/time_zones/meta_zones.rs
robertbastian: typo

components/datetime/src/provider/time_zones.rs
robertbastian: Hmm, it would be nice if we could return things like checksums in the `DataResponseMetadata`, they only need to be checked once at load time, and are `Copy`.

robertbastian: Also, checksums are marker-specific, not payload specific. I will try to do this in a followup

sffc (approved):
This is a departure from how we've done things for years, but I can't think of a good reason why we shouldn't do this.

It makes it harder for people who want to override the name of a specific metazone at runtime, but they still _can_, so long as they have the mapping between numbers and metazones (for now, they can get it themselves, and if it is needed, we could add a standalone key containing this data), and this is a niche enough use case that I'm fine making it less convenient.

provider/source/src/time_zones/mod.rs
```diff
@@ -150,13 +155,26 @@ impl SourceDataProvider {
...
-                Ok(meta_zone_ids)
+                let mut hash = XxHash64::with_seed(0);
+                meta_zone_ids_resource.len().hash(&mut hash);
+
+                Ok((
+                    meta_zone_ids_resource
+                        .iter()
+                        .enumerate()
+                        .map(|(idx, (_short_id, alias))| {
+                            alias.long_id.hash(&mut hash);
```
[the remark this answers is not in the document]
robertbastian: using the long ID because iiuc we were the only users of the short one, so the long one has more coverage

provider/source/src/time_zones/mod.rs
```diff
@@ -150,13 +155,26 @@ impl SourceDataProvider {
...
-                for (meta_zone_id, meta_zone_id_data) in meta_zone_ids_resource.iter() {
-                    meta_zone_ids.insert(meta_zone_id_data.long_id.to_string(), *meta_zone_id);
-                }
-                Ok(meta_zone_ids)
+                let mut hash = XxHash64::with_seed(0);
+                meta_zone_ids_resource.len().hash(&mut hash);
+
+                Ok((
+                    meta_zone_ids_resource
+                        .iter()
```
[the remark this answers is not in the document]
robertbastian: I think it's sufficiently asserted by our CI, if this becomes non-deterministic all data generation checks will fail.

PR #6046, opened 2025-01-29
Add a checksum field to `DataResponseMetadata`, and use it for timezone consistency

robertbastian:
Based on #6043


components/datetime/src/pattern/names.rs
```diff
@@ -1997,22 +1997,6 @@ impl<FSet: DateTimeNamesMarker> RawDateTimeNames<FSet> {
         Ok(())
     }
 
-    fn load_mz_periods<P>(
```
sffc, line 2000: Nit: I prefer the previous style of separate functions for loading separate keys. One important decision here is that we use `Default` as the DataRequest, and I don't want to inline that assumption to a bunch of places.

Just keep this function and make it return the DataResponseMetadata

components/datetime/src/pattern/names.rs
```diff
@@ -2073,24 +2055,22 @@ impl<FSet: DateTimeNamesMarker> RawDateTimeNames<FSet> {
...
+            .mz_periods
+            .load_put(mz_period_provider, Default::default(), ())
+            .map_err(|e| MaybePayloadError::into_load_error(e, error_field))?
+            .map_err(|e| PatternLoadError::Data(e, error_field))?
+            .checksum;
+        if cs1.is_none() || cs1 != cs2 {
             return Err(PatternLoadError::Data(
                 DataErrorKind::InconsistentData(tz::MzPeriodV1Marker::INFO)
-                    .with_req(tz::MzGenericShortV1Marker::INFO, req),
+                    .with_req(tz::MzSpecificLongV1Marker::INFO, req),
```
sffc, line 2073: Observation: bug fix?

provider/blob/src/export/blob_exporter.rs
```diff
@@ -149,24 +157,32 @@ impl BlobExporter<'_> {
...
+                    resources.get(marker_path_hash),
+                    checksums.get(marker_path_hash),
+                )
+            })
+            .map(|(option_sub_map, checksum)| {
+                let mut sub_map = BTreeMap::new();
+                if let Some(sub_map_wrong) = option_sub_map {
+                    if let Some(&checksum) = checksum {
+                        // TODO how does this behave on 32-bit platforms?
+                        sub_map.insert(CHECKSUM_KEY, checksum as usize);
```
sffc, line 177: The checksum should be a `u64` not a `usize` right?

provider/blob/src/blob_schema.rs
```diff
@@ -160,7 +169,12 @@ impl<'data, LocaleVecFormat: VarZeroVecFormat> BlobSchemaV3<'data, LocaleVecForm
             .buffers
             .get(blob_index)
             .ok_or_else(|| DataError::custom("Invalid blob bytes").with_req(marker, req))?;
-        Ok(buffer)
+        Ok((
+            buffer,
+            ZeroTrieSimpleAscii::from_store(zerotrie)
+                .get(CHECKSUM_KEY)
+                .map(|cs| cs as u64),
```
sffc, line 176: Issue/Discussion: Not super happy that we run this all the time on every request even though most keys don't have a checksum.

sffc (commented):
Mostly looks good; let's get a design LGTM from someone else like @Manishearth since this is the first real use of DataResponseMetadata. Also have a few small questions

components/datetime/src/pattern/names.rs
robertbastian: typo

provider/blob/src/export/blob_exporter.rs
robertbastian: If I put a usize value into a `ZeroAsciiTrie` on a 64-bit system, and then access it on a 32-bit system, what will have happened to it? How many bits are stored on the wire.

robertbastian:
> this is the first real use of DataResponseMetadata

`DeserializingBufferProvider` wouldn't work without it.

provider/blob/src/export/blob_exporter.rs
sffc: It uses `usize` in the API because it doesn't really care what integer is stored; it stores a varint. Most use cases of ZeroTrie have involved storing indices into an array of data sitting on the side, so that's mainly what the API was optimized for.

provider/blob/src/blob_schema.rs
sffc: Well it's still a binary search.

If you want to get the very first item, you can use the constant-time `probe` function instead.

provider/blob/src/export/blob_exporter.rs
robertbastian: We need a way to get 64 bits out regardless of the usize length then.

provider/data/timezone/fingerprints.csv
```diff
@@ -1,4 +1,4 @@
-time_zone/bcp47_to_iana@1, <singleton>, 7630B, 7609B, 7efd2629d86e0cdc
-time_zone/iana_to_bcp47@3, <singleton>, 9551B, 9509B, ef97860b5f688e6d
+time_zone/bcp47_to_iana@1, <singleton>, 7622B, 7600B, 35e1d2b320b41234
+time_zone/iana_to_bcp47@3, <singleton>, 9543B, 9500B, 7218ae639c063764
```
sffc, line 2: Nit (optional): It's not clear whether these sizes include the checksum. I think they should, at least for postcard.

provider/blob/src/blob_schema.rs
sffc: Also, due to varint packing inefficiencies for large values, this probably ends up taking 10 bytes.

provider/blob/src/export/blob_exporter.rs
sffc: I suggest adding an API to ZeroTrie to get/set unsigned ints of specific sizes.

provider/fs/src/export/fs_exporter.rs
```diff
@@ -120,47 +137,62 @@ impl DataExporter for FilesystemExporter {
...
-            fs::File::create(&path_buf)?;
+        #[allow(clippy::unwrap_used)] // has parent by construction
+        let parent_dir = path_buf.parent().unwrap();
+
+        fs::create_dir_all(parent_dir)
+            .map_err(|e| DataError::from(e).with_path_context(parent_dir))?;
+
+        if let Some(checksum) = metadata.checksum {
+            write!(
+                &mut fs::File::create(format!("{}_checksum", path_buf.display()))?,
```
sffc, line 189: Thought/Suggestion: `{}.checksum`

provider/fs/src/export/fs_exporter.rs
```diff
@@ -120,47 +137,62 @@ impl DataExporter for FilesystemExporter {
...
+        let path_buf = self.root.join(path);
+
+        if !path_buf.exists() {
+            fs::create_dir_all(&path_buf)
+                .map_err(|e| DataError::from(e).with_path_context(&path_buf))?;
+            fs::File::create(path_buf.join(".empty"))?;
+        } else if let Some(checksum) = metadata.checksum {
+            write!(
+                &mut fs::File::create(path_buf.join(".checksum"))?,
+                "{checksum}"
```
sffc, line 163: Suggestion: write the checksum as a hexadecimal literal.

provider/fs/src/fs_data_provider.rs
```diff
@@ -76,20 +76,27 @@ impl FsDataProvider {
...
-                path.push(req.id.marker_attributes.as_str());
-            }
+        let checksum = if marker.is_singleton {
+            std::fs::read_to_string(format!("{}_checksum", path.display()))
+        } else {
+            std::fs::read_to_string(path.join(".checksum"))
         }
+        .ok()
+        .and_then(|s| s.parse().ok());
         if !marker.is_singleton {
```
sffc, line 86: Question/Issue: Why did you move the `is_singleton` condition above the `marker_attributes` condition

provider/data/timezone/fingerprints.csv
robertbastian: They don't, postcard doesn't track indexing overhead either, so I don't think we need to track checksizes

provider/fs/src/fs_data_provider.rs
robertbastian: this was a bug

Manishearth (approved):
approach lgtm. midly worried about this having to be passed through everywhere now, but should be fine. consider adding a data key assoc const / metadata bit the way we have for singletons that signals that a key has a checksum

Manishearth (approved):

Manishearth (approved):

sffc (approved):

sffc (approved):

PR #6051, opened 2025-01-31
Be less opinionated in HourCycle docs

sffc:
Just got to reviewing #6038. I think the changes are good, but I think stylistically they are a bit too opinionated. We normally write our docs to be more matter-of-fact, leaving editorials to tutorials and blog posts.

### DIFF
--- a/components/locale_core/src/preferences/extensions/unicode/keywords/hour_cycle.rs
+++ b/components/locale_core/src/preferences/extensions/unicode/keywords/hour_cycle.rs
@@ -13,8 +13,10 @@ enum_keyword!(
         ("h12" => H12),
         /// The 24-hour clock. Hour are numbered 0–23. Corresponds to 'H' in patterns.
         ("h23" => H23),
-        /// Variant of the 12-hour clock. Common in Japan. Hours are numbered 0–11. Corresponds to 'K' in patterns.
+        /// Variant of the 12-hour clock, sometimes used in Japan. Hours are numbered 0–11. Corresponds to 'K' in patterns.
         ("h11" => H11),
-        /// Variant of the 24-hour clock. Not in common use. Hours are numbered 1–24. Corresponds to 'k' in patterns.
+        /// Variant of the 24-hour clock, not in common use. Hours are numbered 1–24. Corresponds to 'k' in patterns.
+        ///
+        /// ❗ This is not the typical 24-hour clock! See `H23` above for the 24-hour clock used throughout most of the world.
         ("h24" => H24),
 }, "hc");
--- a/components/locale_core/src/preferences/extensions/unicode/keywords/hour_cycle.rs
+++ b/components/locale_core/src/preferences/extensions/unicode/keywords/hour_cycle.rs
@@ -9,12 +9,12 @@ enum_keyword!(
     ///
     /// The valid values are listed in [LDML](https://unicode.org/reports/tr35/#UnicodeHourCycleIdentifier).
     HourCycle {
-        /// The typical 12-hour clock. Hour system using 1–12; corresponds to 'h' in patterns.
+        /// The typical 12-hour clock. Hours are numbered 1–12. Corresponds to 'h' in patterns.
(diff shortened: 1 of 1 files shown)


components/locale_core/src/preferences/extensions/unicode/keywords/hour_cycle.rs
```diff
@@ -9,12 +9,12 @@ enum_keyword!(
...
+        /// The typical 12-hour clock. Hours are numbered 1–12. Corresponds to 'h' in patterns.
         ("h12" => H12),
-        /// The 24-hour clock. Hour system using 0–23; corresponds to 'H' in patterns.
+        /// The 24-hour clock. Hour are numbered 0–23. Corresponds to 'H' in patterns.
         ("h23" => H23),
-        /// Variant of the 12-hour clock only used in Japan. Hour system using 0–11; corresponds to 'K' in patterns.
+        /// Variant of the 12-hour clock. Common in Japan. Hours are numbered 0–11. Corresponds to 'K' in patterns.
         ("h11" => H11),
-        /// __Not actually in use!__ See `H23` above for the 24-hour clock! Included for theoretical completeness. Hour system using 1–24; corresponds to 'k' in pattern.
+        /// Variant of the 24-hour clock. Not in common use. Hours are numbered 1–24. Corresponds to 'k' in patterns.
```
hsivonen, line 18: I have seen an actual bug arise from this option existing, so I think it's important to give _very_ obvious "Do not use" signals in docs. When the 12-hour clock is `h12`, it's easy to make the mistake to assume that the 24-hour clock is `h24`, but that's actually not the case and, moreover, `h24` isn't attested to be in use in _any_ locale according to CLDR.

So I think the proposed change makes `h24` look way too legitimate to use relative to what the state of CLDR indicates.

Would you be OK with "__This is not the typical 24-hour clock; see `H23` above for the 24-hour clock!__ As of February 2025, this variant is not used in any region known to CLDR."

hsivonen (changes requested):

components/locale_core/src/preferences/extensions/unicode/keywords/hour_cycle.rs
```diff
@@ -9,12 +9,12 @@ enum_keyword!(
...
     /// The valid values are listed in [LDML](https://unicode.org/reports/tr35/#UnicodeHourCycleIdentifier).
     HourCycle {
-        /// The typical 12-hour clock. Hour system using 1–12; corresponds to 'h' in patterns.
+        /// The typical 12-hour clock. Hours are numbered 1–12. Corresponds to 'h' in patterns.
         ("h12" => H12),
-        /// The 24-hour clock. Hour system using 0–23; corresponds to 'H' in patterns.
+        /// The 24-hour clock. Hour are numbered 0–23. Corresponds to 'H' in patterns.
         ("h23" => H23),
-        /// Variant of the 12-hour clock only used in Japan. Hour system using 0–11; corresponds to 'K' in patterns.
+        /// Variant of the 12-hour clock. Common in Japan. Hours are numbered 0–11. Corresponds to 'K' in patterns.
```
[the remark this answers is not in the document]
sffc: My anecdotal experience suggests it is fairly common in Japan. But I'm okay removing the word "common". How's this?

components/locale_core/src/preferences/extensions/unicode/keywords/hour_cycle.rs
sffc: How's this?

We use the ❗ emoji elsewhere for this purpose; see https://unicode-org.github.io/icu4x/rustdoc/icu/datetime/pattern/index.html

I'm also open to the idea of renaming this to `LegacyH24` or something like that, if we want to take a harder stand. We'd want to pass that by the CLDR Design WG first.

sffc: https://unicode-org.atlassian.net/browse/CLDR-18303

PR #6090, opened 2025-02-10
Change long localized offset format to use +00:00

robertbastian:
[CLDR-18280](https://unicode-org.atlassian.net/browse/CLDR-18280)


sffc (dismissed):
Thought:

I think the "more semantic" way to handle this would be for this to be a combination of `LocalizedOffset` zone style and `Alignment::Column`.

Because if you make the skeleton `YMDT::short().alignment(Column).zone(LocalizedOffset)`, shouldn't that use a column-aligned time zone when able, as we are here?

sffc (dismissed):
My thought above doesn't block the PR

sffc (approved):

PR #6092, opened 2025-02-11
Download retries for `SourceDataProvider`

robertbastian:


provider/source/src/source.rs
```diff
@@ -175,13 +175,20 @@ impl AbstractFs {
...
-                        .into_reader(),
-                    &mut BufWriter::new(File::create(&root)?),
-                )?;
+                let mut retry = 5;
+                let mut response: Box<dyn Read + Send + Sync + 'static> = loop {
+                    match ureq::get(resource).call() {
+                        Ok(r) => break r.into_reader(),
+                        Err(e) if retry > 0 => {
+                            log::warn!("Download error {e:?}, retrying...");
+                            retry -= 1;
```
sffc, line 184: Suggestion: pause for a few seconds

robertbastian: done

sffc (approved):

PR #6100, opened 2025-02-12
fix: implement `Hash` and `Serde` traits for currency options

sabify:


sffc:
Seems harmless, but can you explain why you need these traits? We haven't generally implemented them on other types.

sffc (approved):
Thanks for the clarification, and thanks for gating these correctly on the "serde" Cargo feature.

PR #6105, opened 2025-02-12
Use `rearguard` version of TZDB

robertbastian:
This is better than the hack we have right now.


sffc (commented):
(from earlier)

sffc (approved):
Suggestion: Add documentation to ZoneVariant that it expects the rearguard version of tzif's `isDST` boolean

sffc:
https://cla-assistant.io/check/unicode-org/icu4x?pullRequest=6105

robertbastian:
https://icu4x-pr-artifacts.storage.googleapis.com/gha/349f8dd1ef3eab982798231a3b96e68a905c98dc/rustdoc/icu/timezone/struct.TimeZoneInfo.html

sffc (approved):
Good; docs can be tweaked later, but getting this content in is good

PR #6111, opened 2025-02-13
Variant-aware metazone name inclusion

robertbastian:
Shaves yet another few kB. I expected more, honestly.


sffc:
What's the key change here? Is it `with_timezone_horizon`?

robertbastian:
The key internal change is that time zone variants can be individually dropped. `with_timezone_horizon` then exercises that code.

PR #6112, opened 2025-02-13
Apply `Africa/Casablanca` rearguard fix

robertbastian:


sffc (approved):
I still lean toward running the real awk script somehow

robertbastian:
I would prefer that, but how? The `frawk` crate you found is a binary crate, and I don't see an alternative.

sffc:
Maybe the maintainer of `frawk` would accept a PR to add a feature for that?

robertbastian:
`frawk` has 30+ direct dependencies, and doesn't even work for me out of the box. It's not a solution.

sffc:
Most of the deps look optional-ish. But yes it would take some time that's probably better spent elsewhere. 

I mainly worry that we're not catching everything we should be catching with regards to rearguard.

PR #6124, opened 2025-02-14
Human readable time zone transition dates

robertbastian:
Now I can actually spot bugs.


components/calendar/src/iso.rs
```diff
@@ -266,8 +266,13 @@ impl Iso {
...
     // The fixed calculations algorithms are from the Calendrical Calculations book.
-    pub(crate) fn fixed_from_iso(date: IsoDateInner) -> RataDie {
-        calendrical_calculations::iso::fixed_from_iso(date.0.year, date.0.month, date.0.day)
+    #[doc(hidden)]
+    pub fn to_fixed(date: Date<Iso>) -> RataDie {
+        calendrical_calculations::iso::fixed_from_iso(
+            date.inner.0.year,
+            date.inner.0.month,
+            date.inner.0.day,
+        )
```
sffc, line 275: Nit: I think `fixed_from_iso` is a string from Reingold.

I suggest adding new public functions to `Date<Iso>` called `to_rata_die` and `from_rata_die`.

components/calendar/src/date.rs
```diff
@@ -344,7 +339,7 @@ impl Date<Iso> {
     /// ```
     #[doc(hidden)] // unstable
     pub fn days_since(&self, other: Date<Iso>) -> i32 {
-        (Iso::fixed_from_iso(*self.inner()) - Iso::fixed_from_iso(other.inner)) as i32
+        (Iso::to_fixed(*self) - Iso::to_fixed(other)) as i32
```
sffc, line 342: Nit: Too much of a diff.

components/time/src/provider.rs
```diff
@@ -114,12 +117,106 @@ impl<'a> zerovec::maps::ZeroMapKV<'a> for TimeZone {
...
+        if serializer.is_human_readable() {
+            let (date, time) = self.to_dt();
+            let year = date.year().extended_year;
+            let month = date.month().month_number();
+            let day = date.day_of_month().0;
+            let hour = time.hour.number();
+            let minute = time.minute.number();
+            serializer.serialize_str(&alloc::format!(
+                "{year:04}-{month:02}-{day:02} {hour:02}:{minute:02}"
+            ))
```
sffc, line 182: Suggestion: Use IXDTF syntax.

components/time/src/provider.rs
```diff
@@ -114,12 +117,106 @@ impl<'a> zerovec::maps::ZeroMapKV<'a> for TimeZone {
...
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        use serde::de::Error;
+        if deserializer.is_human_readable() {
+            let e0 = D::Error::custom("invalid");
+            let e1 = |_| D::Error::custom("invalid");
+            let e2 = |_| D::Error::custom("invalid");
+            let e3 = |_| D::Error::custom("invalid");
```
sffc, line 200: Suggestion: Parse the string with `ixdtf`.

sffc (commented):
I agree very much with the goal; just some nitpicks/suggestions about code.

components/time/src/provider.rs
robertbastian: If it wasn't behind a feature I would

components/calendar/src/iso.rs
robertbastian: and the method in `calendrical_calculations` will keep that name

robertbastian: let's not nitpick non-public names on this PR please, if you want to stabilise them please file an issue

sffc: ok, you know how I feel about `#[doc(hidden)]`. Can you at least reduce the diff and use the existing names please?

robertbastian: this goes from one doc-hidden API to two, because I need the other direction.

the diff is how it is because this crate previously had two `to_fixed` methods, now it has one

components/time/src/provider.rs
robertbastian: This is the syntax from CLDR metazones.xml: https://github.com/unicode-org/cldr/blob/2cd040906853ce18c3b87bcf270b3526537c66cc/common/supplemental/metaZones.xml#L23

components/calendar/src/iso.rs
sffc: I don't have time right now to review the big diff. I'll get to it soon, or sooner if the PR gets a smaller diff.

robertbastian: it's some tests that compile and pass

sffc (approved):

PR #6125, opened 2025-02-14
Look up location names from canonical alias only

robertbastian:
Previously it would look for data from all other aliases instead of deriving the name from the IANA ID.


Manishearth (approved):
This looks nicer but I don't fully understand the implications of this so I'd wait till Shane's approval too

sffc (approved):
So this is a bug fix, but there are no tests? (Mostly a rubber-stamp since this is all in datagen.)

robertbastian:
It's not a bug fix, it's just that I discovered while reading the code that it should behave differently. This doesn't make a difference for any language we test because they all have full data.

PR #6126, opened 2025-02-14
Remove YokeTraitHack from icu

robertbastian:
Should I also remove it from `yoke`, given that it's not required at its MSRV anymore?


Manishearth (dismissed):
Kill it with fire.

Please do also remove from Yoke

Manishearth (approved):

robertbastian:
Presumably it's not a required tests. I don't have permissions to edit those.

PR #6128, opened 2025-02-14
Use zone regions from CLDR instead of zone.tab

robertbastian:
https://github.com/unicode-org/cldr/pull/4516


provider/source/src/time_zones/mod.rs
```diff
@@ -272,28 +274,63 @@ impl SourceDataProvider {
...
+                for (&bcp47_tzid, bcp47_tzid_data) in bcp47_tzids_resource {
+                    regions_to_zones
+                        .entry(match bcp47_tzid.0.as_bytes() {
+                            // backfill since this data is not in 47 yet
+                            b"ancur" => region!("CW"),
+                            b"fimhq" => region!("AX"),
+                            b"gpmsb" => region!("MF"),
+                            b"gpsbh" => region!("BL"),
+                            b"gazastrp" | b"hebron" => region!("PS"),
+                            b"jeruslm" => region!("IL"),
```
sffc, line 299: But it is new in 47? How is this data missing?

sffc (approved):
No data diff, so I trust that the datagen code is correct

provider/source/src/time_zones/mod.rs
robertbastian: it's 48

PR #6132, opened 2025-02-15
Impl Extend for LiteMap and avoid quadratic behavior in from_iter and deserialize

arthurprs:
Adds `lm_extend` to the `StoreMut` trait. The Vec implementation is optimized to have O(N) performance on sorted inputs and an asymptotic worst case of O(NLog(N)), effectively avoiding quadratic worst cases.

This PR takes advantage of the new `extend` optimization and changes the `FromIterator` and `Deserialize` implementations to also avoid quadratic worst-case.

Before
```
litemap/from_iter_rand/small
                        time:   [916.60 ns 941.77 ns 965.82 ns]
litemap/from_iter_rand/large
                        time:   [2.3779 s 2.3863 s 2.3946 s]
litemap/from_iter_sorted/small
                        time:   [84.010 ns 84.084 ns 84.155 ns]
litemap/from_iter_sorted/large
                        time:   [725.31 µs 739.88 µs 755.20 µs]
```

After

```
litemap/from_iter_rand/small
                        time:   [354.62 ns 365.31 ns 376.02 ns]
litemap/from_iter_rand/large
                        time:   [25.129 ms 25.415 ms 25.704 ms]
litemap/from_iter_sorted/small
                        time:   [76.593 ns 76.767 ns 76.938 ns]
litemap/from_iter_sorted/large
                        time:   [662.70 µs 673.66 µs 686.83 µs]
litemap/extend_rand/large
                        time:   [66.221 ms 67.467 ms 68.814 ms]
litemap/extend_rand_dups/large
                        time:   [190.24 ms 196.05 ms 202.13 ms]
litemap/extend_from_litemap_rand/large
                        time:   [2.0401 s 2.0488 s 2.0575 s]
```

This is a followup of #6068

### DIFF
--- a/components/locale_core/src/shortvec/litemap.rs
+++ b/components/locale_core/src/shortvec/litemap.rs
@@ -12,6 +12,15 @@ impl<K, V> StoreConstEmpty<K, V> for ShortBoxSlice<(K, V)> {
     const EMPTY: ShortBoxSlice<(K, V)> = ShortBoxSlice::new();
 }
 
+impl<K, V> StoreSlice<K, V> for ShortBoxSlice<(K, V)> {
+    type Slice = [(K, V)];
+
+    #[inline]
+    fn lm_get_range(&self, range: core::ops::Range<usize>) -> Option<&Self::Slice> {
+        self.get(range)
+    }
+}
+
 impl<K, V> Store<K, V> for ShortBoxSlice<(K, V)> {
     #[inline]
     fn lm_len(&self) -> usize {
--- a/components/locale_core/src/shortvec/litemap.rs
+++ b/components/locale_core/src/shortvec/litemap.rs
@@ -83,12 +83,38 @@ impl<K, V> StoreMut<K, V> for ShortBoxSlice<(K, V)> {
     fn lm_clear(&mut self) {
         self.clear();
     }
+}
 
+#[cfg(feature = "alloc")]
+impl<K: Ord, V> StoreBulkMut<K, V> for ShortBoxSlice<(K, V)> {
     fn lm_retain<F>(&mut self, mut predicate: F)
     where
         F: FnMut(&K, &V) -> bool,
     {
         self.retain(|(k, v)| predicate(k, v))
+    }
+
+    fn lm_extend<I>(&mut self, other: I)
+    where
+        I: IntoIterator<Item = (K, V)>,
+    {
+        let other = other.into_iter();
+        let mut items = alloc::vec::Vec::with_capacity(self.len() + other.size_hint().0);
+        match core::mem::take(&mut self.0) {
+            ShortBoxSliceInner::ZeroOne(None) => (),
+            ShortBoxSliceInner::ZeroOne(Some(existing_item)) => {
+                items.push(existing_item);
+            }
+            ShortBoxSliceInner::Multi(existing_items) => {
+                items.extend(existing_items);
+            }
+        }
+        items.lm_extend(other);
+        if items.len() <= 1 {
+            self.0 = ShortBoxSliceInner::ZeroOne(items.pop());
+        } else {
+            self.0 = ShortBoxSliceInner::Multi(items.into_boxed_slice());
+        }
     }
 }
 
--- a/components/locale_core/src/shortvec/litemap.rs
+++ b/components/locale_core/src/shortvec/litemap.rs
@@ -107,21 +107,42 @@ impl<K: Ord, V> StoreBulkMut<K, V> for ShortBoxSlice<(K, V)> {
     where
         I: IntoIterator<Item = (K, V)>,
     {
-        let other = other.into_iter();
-        let mut items = alloc::vec::Vec::with_capacity(self.len() + other.size_hint().0);
+        let mut other = other.into_iter();
+        // Use an Option to hold the first item of the map and move it to
+        // items if there are more items. Meaning that if items is not
+        // empty, first is None.
+        let mut first = None;
+        let mut items = alloc::vec::Vec::new();
         match core::mem::take(&mut self.0) {
-            ShortBoxSliceInner::ZeroOne(None) => (),
-            ShortBoxSliceInner::ZeroOne(Some(existing_item)) => {
-                items.push(existing_item);
+            ShortBoxSliceInner::ZeroOne(zo) => {
+                first = zo;
+                // Attempt to avoid the items allocation by advancing the iterator
+                // up to two times. If we eventually find a second item, we can
+                // lm_extend the Vec and with the first, next (second) and the rest
+                // of the iterator.
+                while let Some(next) = other.next() {
+                    if let Some(first) = first.take() {
+                        // lm_extend will take care of sorting and deduplicating
+                        // first, next and the rest of the other iterator.
+                        items.lm_extend([first, next].into_iter().chain(other));
+                        break;
+                    }
+                    first = Some(next);
+                }
             }
             ShortBoxSliceInner::Multi(existing_items) => {
+                items.reserve_exact(existing_items.len() + other.size_hint().0);
+                // We use a plain extend with existing items, which are already valid and
+                // lm_extend will fold over rest of the iterator sorting and deduplicating as needed.
                 items.extend(existing_items);
+                items.lm_extend(other);
             }
         }
-        items.lm_extend(other);
-        if items.len() <= 1 {
-            self.0 = ShortBoxSliceInner::ZeroOne(items.pop());
+        if items.is_empty() {
+            debug_assert!(items.is_empty());
+            self.0 = ShortBoxSliceInner::ZeroOne(first);
         } else {
+            debug_assert!(first.is_none());
             self.0 = ShortBoxSliceInner::Multi(items.into_boxed_slice());
         }
     }
--- a/utils/litemap/Cargo.toml
+++ b/utils/litemap/Cargo.toml
@@ -28,6 +28,7 @@ serde = { workspace = true, optional = true, features = ["alloc"]}
 yoke = { workspace = true, features = ["derive"], optional = true }
 
 [dev-dependencies]
+rand = { workspace = true }
 bincode = { workspace = true }
 icu_benchmark_macros = { path = "../../tools/benchmark/macros" }
 icu_locale_core = { path = "../../components/locale_core" }
@@ -43,7 +44,7 @@ criterion = { workspace = true }
 default = ["alloc"]
 alloc = []
 databake = ["dep:databake"]
-serde = ["dep:serde"]
+serde = ["dep:serde", "alloc"]
 yoke = ["dep:yoke"]
 
 # Enables the `testing` module with tools for testing custom stores.
--- a/utils/litemap/src/serde.rs
+++ b/utils/litemap/src/serde.rs
@@ -4,6 +4,7 @@
 
 use super::LiteMap;
 use crate::store::*;
+use alloc::vec::Vec;
 use core::fmt;
 use core::marker::PhantomData;
 use serde::de::{MapAccess, SeqAccess, Visitor};
@@ -26,6 +27,8 @@ where
         // Many human-readable formats don't support values other
         // than numbers and strings as map keys. For them, we can serialize
         // as a vec of tuples instead
+        // Note from the future: while this is true for the provided stores
+        // (Vec, Slice, ..) it may not be true for all stores.
         if serializer.is_human_readable() {
             if let Some((ref k, _)) = self.values.lm_get(0) {
                 if !super::serde_helpers::is_num_or_string(k) {
@@ -35,6 +38,7 @@ where
             }
         }
         let mut map = serializer.serialize_map(Some(self.len()))?;
+        // TODO: We could require R to be StoreIter in order to avoid the manual loop here.
         let mut i = 0;
         while i < self.values.lm_len() {
             #[allow(clippy::unwrap_used)] // i is in range
@@ -64,7 +68,7 @@ impl<'de, K, V, R> Visitor<'de> for LiteMapVisitor<K, V, R>
 where
     K: Deserialize<'de> + Ord,
     V: Deserialize<'de>,
-    R: StoreMut<K, V> + StoreFromIterable<K, V>,
+    R: StoreMut<K, V>,
 {
     type Value = LiteMap<K, V, R>;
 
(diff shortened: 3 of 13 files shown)


utils/litemap/src/serde.rs
```diff
@@ -102,23 +102,27 @@ where
...
+            // the implementation an opportunity to avoid quadratic costs from calling
+            // insert() with out of order items.
+            // Handling ordered inputs first allows for arbitrary maps (e.g. from user JSON)
+            // to be deserialized into LiteMap without impacting performance in the case of
+            // deserializing a serialized map that came from another LiteMap.
             if let Some((key, value)) = map.try_append(key, value) {
-                // Note: this effectively selection sorts the map,
-                // which isn't efficient for large maps
-                map.insert(key, value);
+                out_of_order.push((key, value));
```
arthurprs, line 118: I considered making this based on `StoreMut::lm_push` and `Store::lm_into_iter` but that required changing the adding the `StoreIntoIterator` bound to R.

utils/litemap/Cargo.toml
```diff
@@ -43,7 +44,7 @@ criterion = { workspace = true }
 default = ["alloc"]
 alloc = []
 databake = ["dep:databake"]
-serde = ["dep:serde"]
+serde = ["dep:serde", "alloc"]
```
arthurprs, line 47: Note for reviewers: The addition of alloc was theoretically a breaking change for people using "no default features" so we may want to bump litemap to 0.8.

utils/litemap/src/store/vec_impl.rs
```diff
@@ -104,38 +104,116 @@ impl<K, V> StoreMut<K, V> for Vec<(K, V)> {
     {
         self.retain(|(k, v)| predicate(k, v))
     }
-}
 
-impl<K: Ord, V> StoreFromIterable<K, V> for Vec<(K, V)> {
-    fn lm_sort_from_iter<I: IntoIterator<Item = (K, V)>>(iter: I) -> Self {
+    /// Extends this store with items from an iterator.
+    ///
+    /// It uses a three-pass approach to avoid any potential quadratic costs.
```
robertbastian, line 110: issue: this implementation increases the capacity of the map by at least m/2, *and* also uses m auxiliary space in the worst case

I think this algorithm has equivalent runtime complexity, but only uses m storage:

1. append all the new items, O(m)
2. sort the appended items, O(m log(m)), Ω(m)
3. deduplicate the appended items against themselves, O(m)
4. if the first appended item is less than the last existing item
    - deduplicate the appended items against existing items, O(m log(n))
    - quicksort the whole thing, O((n+m) log(n+m))

The happy path here is still Ω(m), and the worst case is still O((n+m) log(n+m)), but less space is used.

utils/litemap/src/store/mod.rs
```diff
@@ -133,6 +133,24 @@ pub trait StoreMut<K, V>: Store<K, V> {
             }
         }
     }
+
+    /// Extends this store with items from an iterator.
+    fn lm_extend<I>(&mut self, other: I)
```
robertbastian, line 138: issue: having the quadratic algorithm in the default implementation is a footgun. same for `lm_retain`

utils/litemap/src/store/vec_impl.rs
arthurprs: Maybe I'm missing something, but the deduplication in place is O(n²) due to the memmove calls.

utils/litemap/src/store/mod.rs
arthurprs: To clarify, are you suggesting the default impl for lm_extend and lm_retain* to be removed?

* not part of the PR

utils/litemap/src/store/vec_impl.rs
robertbastian: You don't do a sequence of O(n) removals. Instead you iterate with two pointers, one pointing at the next element, and one pointing at the end of the deduplicated sequence (both starting at 0). If you encounter a duplicated element, you increment the next-element pointer (which will always be at least the other pointer), if you encounter a unique element you replace what's currently at the end of the deduplicated sequence (everything between the pointers is trash). When the next-element pointer reaches the end of the list, you remove the trash.

This requires some thought for non-Copy types, but the algorithm is definitely linear.

utils/litemap/src/store/mod.rs
robertbastian: yes

robertbastian: (if we're happy to make breaking changes)

utils/litemap/src/store/vec_impl.rs
arthurprs: I had to modify partition_dedup to keep the last duplicated entry instead of the first, but otherwise, it's similar to what you suggested. The only drawback is that it always reserves the lower bound length of the incoming iterator as it doesn't make sense to only reserve half.

The new benchmark numbers are below. Without dups it's slightly faster than my original solution. With dups it's roughly the same.

```
litemap/extend_rand/large
                        time:   [66.882 ms 67.852 ms 69.096 ms]
litemap/extend_rand_dups/large
                        time:   [174.22 ms 176.15 ms 178.28 ms]

File: utils/litemap/src/store/vec_impl.rs [hunk 5]
```diff
@@ -104,38 +104,191 @@ impl<K, V> StoreMut<K, V> for Vec<(K, V)> {
...
-                                core::mem::replace(container.lm_get_mut(found).unwrap().1, value);
-                        }
-                        Err(ins) => {
-                            container.insert(ins, (key, value));
-                        }
-                    }
-                } else {
-                    container.push((key, value))
-                }
+/// Moves all but the _last_ of consecutive elements to the end of the slice satisfying
```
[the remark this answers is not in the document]
arthurprs: It's at the bottom of the comment block

```
/// This is based from std::slice::partition_dedup_by (current unstable) but retains the
/// _last_ element of the duplicate run in the first slice (instead of first).

File: utils/litemap/src/store/vec_impl.rs [hunk 6]
```diff
@@ -104,38 +104,191 @@ impl<K, V> StoreMut<K, V> for Vec<(K, V)> {
...
+
+        // First pass: fast-path append or collect out-of-order elements
+        while let Some((key, value)) = iter.next() {
+            // The fast path checks if the incoming key is > the last key in self.
+            // This is the case when the incoming iterator is already sorted and deduplicated.
+            let key_gt_last = self.last().map_or(true, |(lk, _)| &key > lk);
+            self.push((key, value));
+            if !key_gt_last {
+                // extend self with the rest of the iterator and exit early
+                self.extend(iter);
```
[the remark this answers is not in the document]
arthurprs: Good point

utils/litemap/src/store/vec_impl.rs
```diff
@@ -104,38 +104,191 @@ impl<K, V> StoreMut<K, V> for Vec<(K, V)> {
...
+            .zip(new.first())
+            .is_some_and(|(a, b)| a.0 >= b.0)
+        {
+            // Sorted runs overlap, we need to merge and dedup them.
+            // sorted_runs_len_sum is the sum of the lengths of two non-overlaping subslices of self
+            #[allow(clippy::indexing_slicing)]
+            let both_runs = &mut self[..sorted_runs_len_sum];
+            // Use stable sort again as it's often significantly faster for already sorted data.
+            both_runs.sort_unstable_by(|a, b| a.0.cmp(&b.0));
+            let (dedup, _merged_dup) = partition_dedup_by(both_runs, |a, b| a.0 == b.0);
```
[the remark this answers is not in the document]
arthurprs: I missed that difference. I can probably try it later

utils/litemap/src/store/vec_impl.rs
arthurprs: I settled on a simplified version that is within +- 8% of the longer version depending on the benchmark.

utils/litemap/src/store/vec_impl.rs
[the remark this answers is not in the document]
arthurprs: This is a significant improvement depending on the input iterator.

utils/litemap/src/store/vec_impl.rs
```diff
@@ -104,38 +104,170 @@ impl<K, V> StoreMut<K, V> for Vec<(K, V)> {
...
+    // and `prev_ptr_write.offset(1)`.
+    //
+    // `next_write` is also incremented at most once per loop at most meaning
+    // no element is skipped when it may need to be swapped.
+    //
+    // `ptr_read` and `prev_ptr_write` never point to the same element. This
+    // is required for `&mut *ptr_read`, `&mut *prev_ptr_write` to be safe.
+    // The explanation is simply that `next_read >= next_write` is always true,
+    // thus `next_read > next_write - 1` is too.
+    unsafe {
```
sffc, line 246: Discussion: I think I'd prefer to not use `unsafe` here, because

1. We don't currently have any `unsafe` code in the `litemap` library crate. (I see some in integration tests but not in `src`.)
2. Everything else you're doing is already a big speedup; we don't need to hyper-optimize every last operation.
3. Bounds checks for this pattern, the "tortoise and hare" problem, are something that LLVM should eventually optimize. https://users.rust-lang.org/t/hare-and-tortoise-indexes-analysis/63985

utils/litemap/src/store/vec_impl.rs
```diff
@@ -104,38 +104,170 @@ impl<K, V> StoreMut<K, V> for Vec<(K, V)> {
...
+/// This is based on std::slice::partition_dedup_by (currently unstable) but retains the
+/// _last_ element of the duplicate run in the first slice (instead of first).
+#[inline]
+fn partition_dedup_by<T, F>(v: &mut [T], mut same_bucket: F) -> (&mut [T], &mut [T])
+where
+    F: FnMut(&mut T, &mut T) -> bool,
+{
+    // Although we have a mutable reference to `self`, we cannot make
+    // *arbitrary* changes. The `same_bucket` calls could panic, so we
+    // must ensure that the slice is in a valid state at all times.
```
sffc, line 167: Thought: I assume these docs are based on comments elsewhere, but we can guarantee that `same_bucket` won't panic.

utils/litemap/src/store/vec_impl.rs
```diff
@@ -104,38 +104,170 @@ impl<K, V> StoreMut<K, V> for Vec<(K, V)> {
...
+/// Moves all but the _last_ of consecutive elements to the end of the slice satisfying
+/// a given equality relation.
+///
+/// Returns two slices. The first contains no consecutive repeated elements.
+/// The second contains all the duplicates in no specified order.
+///
+/// This is based on std::slice::partition_dedup_by (currently unstable) but retains the
+/// _last_ element of the duplicate run in the first slice (instead of first).
+#[inline]
+fn partition_dedup_by<T, F>(v: &mut [T], mut same_bucket: F) -> (&mut [T], &mut [T])
```
sffc, line 161: Suggestion: I understand a desire to keep this code the same as upstream, but if we take out the `unsafe`, we're already diverging, so maybe just make the function bound on `T: PartialEq` and remove `same_bucket`.

sffc (commented):
Nice work!

I'm not too worried about making a breaking change right now; we're about to release ICU4X 2.0 so this is a decent time for that. If you could make things better, feel free to do so. Related: https://github.com/unicode-org/icu4x/pull/6192

utils/litemap/src/store/vec_impl.rs
arthurprs: In this case I think we need to make it work in terms of `(K, V)]` where K: Ord`, is that what you mean?

utils/litemap/src/store/vec_impl.rs
sffc: Good point

utils/litemap/src/store/vec_impl.rs
sffc: Yes.

arthurprs:
All concerns have been addressed.

Since we're doing breaking changes, I also included 
fb22056 to remove the `S: Serialize` bound but because of the higher-ranked-trait bounds, we hit a [borrow checker limitation in zerotrie](https://github.com/unicode-org/icu4x/actions/runs/13883257767/job/38844595550?pr=6132), which required 54c22cd. I suppose I could put back a manual loop if that's undesirable.

I'd appreciate if we could get this merged and a 0.8 release (not included)

components/locale_core/src/shortvec/litemap.rs
```diff
@@ -83,13 +92,39 @@ impl<K, V> StoreMut<K, V> for ShortBoxSlice<(K, V)> {
...
     {
         self.retain(|(k, v)| predicate(k, v))
     }
+
+    fn lm_extend<I>(&mut self, other: I)
+    where
+        I: IntoIterator<Item = (K, V)>,
+    {
+        let other = other.into_iter();
+        let mut items = alloc::vec::Vec::with_capacity(self.len() + other.size_hint().0);
```
robertbastian, line 111: nit: this needlessly allocates in some cases (self is empty, iterator is size 1 or 0, or self is size 1 and iterator is size 0)

utils/litemap/src/store/vec_impl.rs
```diff
@@ -104,38 +104,170 @@ impl<K, V> StoreMut<K, V> for Vec<(K, V)> {
...
+        #[allow(clippy::indexing_slicing)]
+        {
+            // Count new elements that are sorted and non-duplicated.
+            // Starting from the end of the existing sorted run, if any.
+            sorted_len += self[sorted_len.saturating_sub(1)..]
+                .windows(2)
+                .take_while(|w| w[0].0 < w[1].0)
+                .count()
+                + (sorted_len == 0) as usize;
+        }
```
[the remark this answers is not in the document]
arthurprs: Yes, it's an exclusive index so -1 is needed to include the existing last item, if any, in the slice.

sffc (dismissed):
Nice work! Thanks for making it fully safe.

arthurprs:
Gentle ping. It would be good to get this PR wrapped up and the crate version bumped to `0.8`

robertbastian:
Next time just click the rereview button (image), it's hard to keep track of which PRs need action if no review is requested.

robertbastian (approved):

PR #6133, opened 2025-02-16
Baked data: use VarULE to store data when specified

sffc:
#5230

I did it by creating a trait, `MaybeAsVarULE`, which is implemented on data structs, similar to how `Bake` and `Serialize` are implemented on data structs.

The trait has an associated type that I can leverage in the baked exporter without having to do `dyn Any` downcasting.

I tried to keep my understanding of @robertbastian's position in mind while writing this PR (don't add anything hyper-specific to bake in the provider crate), and I think I achieved this, since the thing I added is on the same level as the existing format-specific traits.

See the hello world data for an example. I didn't apply it to the other data markers yet.


provider/core/src/marker.rs
```diff
@@ -108,6 +108,45 @@ impl<M: DataMarker + Sized> DataMarkerExt for M {
...
+    type VarULE: VarULE + ?Sized;
+
+    /// If `self` can be wholy represented as [`Self::VarULE`],
+    /// returns it as a reference.
+    /// 
+    /// Otherwise, returns `None`.
+    fn maybe_as_varule(&self) -> Option<&Self::VarULE>;
+}
+
+/// Implements [`MaybeAsVarULE`] on a type that is NOT representable
```
Manishearth, line 126: thought: the reverse default might be better UX for ICU4X devs so they do not accidentally feel forced to structure all of their data structs to support VarULE. I know Younies has found this confusing before.

but I figure there's a reason behind this choice here. if it's just macroability I think the macro can be done differently.

utils/zerovec/src/ule/plain.rs
```diff
@@ -364,3 +364,36 @@ impl AsULE for bool {
...
 unsafe impl EqULE for bool {}
+
+// Safety (based on the safety checklist on the ULE trait):
+//  1. () does not include any uninitialized or padding bytes (it has no bytes)
+//  2. () is aligned to any number of bytes (it is zero-sized)
+//  3. The impl of validate_bytes() returns an error if any byte is not valid (any byte).
+//  4. The impl of validate_bytes() returns an error if there are leftover bytes (always).
+//  5. The other ULE methods use the default impl.
+//  6. () byte equality is semantic equality
+unsafe impl ULE for () {
```
Manishearth, line 375: This is potentially subtly unsound, ZSTs are weird.

https://github.com/rust-lang/unsafe-code-guidelines/issues/557

provider/core/src/marker.rs
```diff
@@ -108,6 +108,45 @@ impl<M: DataMarker + Sized> DataMarkerExt for M {
     }
 }
 
+/// This trait should be implemented on all data structs.
+///
+/// For most data structs, a default implementation available via
+/// [`does_not_deref_to_varule`] can be used.
+pub trait MaybeAsVarULE {
```
Manishearth, line 115: thought: not the best fan of this name, would _kind of_ prefer this talk about the optimizaiton it signals rather than whether or not the type can be converted to a VarULE.

We will have many types that can be VarULEd but I'm not convinced we should automatically apply the optimization there.

Manishearth (commented):
I think this design is

provider/core/src/marker.rs
sffc: Classic "what is is versus what it does" naming problem. Happy for naming suggestions. Here are a few:

- `StoreAsVarULE`
- `MaybeStoreAsVarULE`
- `OptimizeForVarZeroVec`
- `WithEfficientBakeStorage`

sffc:
Created #6135 to reduce the diff

utils/zerovec/src/ule/plain.rs
Manishearth: It is not unsound, but let's include a link to that issue.

sffc: Thanks for investigating.

Can I ask you to make a Suggested Edit? I'm not completely sure I like my unsafe docs, especially the thing about being align(1).

provider/core/src/hello_world.rs
```diff
@@ -41,6 +42,21 @@ impl Default for HelloWorld<'_> {
     }
 }
 
+impl MaybeAsVarULE for HelloWorld<'_> {
```
robertbastian, line 45: Issue: I think allowing implementation of this trait on arbitrary data structs is going to be super confusing (and it should definitely not use `From` to do the reverse).

`HelloWorldV1::DataStruct` should just be `Cow<'static, str>`, and this trait should be implemented for `Cow<'static, str>` in `icu_provider`.

components/list/src/provider/mod.rs
```diff
@@ -69,6 +69,8 @@ data_marker!(
     ListFormatterPatterns<'static>,
 );
 
+icu_provider::marker::does_not_deref_to_varule!(ListFormatterPatterns<'_>);
```
robertbastian, line 72: issue: this is an opt-out, but it should be opt-in

provider/core/macros/src/lib.rs
robertbastian, line 1: if we did the marker name cleanup first, we didn't have to keep maintaining this macro

components/list/src/provider/mod.rs
sffc: Without specialization, I need to implement the trait on everything, and use runtime `Option` to determine if it is actually implement or not.

provider/core/macros/src/lib.rs
sffc: To the contrary, I want to land this before marker cleanup so that we can make a decision on a case-by-case basis on whether to apply the optimization.

provider/core/src/hello_world.rs
sffc: I'm okay implementing it on `Cow`, but we still need the no-op version of the trait implement on other things. I don't see a reason to _not_ allow it to be implemented on arbitrary data structs.

provider/core/macros/src/lib.rs
robertbastian: Applying this optimization is not 2.0 blocking. This is scope creep.

sffc: > Applying this optimization is not 2.0 blocking. This is scope creep.

I consider it 2.0 to have a design for this. This is a specific design proposal. It involves changing the bounds of `ExportableDataPayload`, which seems like something that would be smoother to do in a major release.

components/list/src/provider/mod.rs
Manishearth: I don't see how specialization is relevant for opt out vs opt in? We're using a macro, the macro can default to generating a blank impl, and one can ask the macro to not do so.

This is my comment above about the reverse default.

provider/core/src/marker.rs
sffc: > the reverse default

I don't understand what you mean. Do you mean the name of the trait? The name of the macro? Something else?

components/list/src/provider/mod.rs
sffc: > I don't see how specialization is relevant for opt out vs opt in

Ideally I would have a blanket impl that returns None and then a small number of specific impls that return Some. Instead, everyone needs to decide how to implement the trait, which I think @robertbastian doesn't like.

sffc:
This PR is passing CI and ready from my perspective. I am unclear whether the open comments are bikeshed issues or fundamental design issues, and how you hope that they be resolved.

provider/core/src/marker.rs
Manishearth: This macro opts out of this system by default.

I'd rather have the data marker macro opt in by default, with a flag to opt out

components/list/src/provider/mod.rs
Manishearth: My objection at least is that people need to explicitly opt out: it's not an objection about the trait architecture, it's an objection about the macro architecture.

Manishearth:
@sffc From my side, the trait naming is a bikeshed that can be deferred, and the macro opt-in/opt-out is a concern I'd prefer solved here but is not a major design issue.

The overall design looks good.

provider/core/src/marker.rs
Manishearth: MaybeStoreAsVarULE is fine, or even better MaybeExportAsVarULE. WithEfficientBakeStorage isn't better because we may also use this in postcard.

provider/core/src/marker.rs
Manishearth: Not the name, just the default behavior, yes.

Manishearth: Basically: if I define a new data struct, the default behavior of the data struct/marker macros should not make me think I need to impl VarULE. The non-None impl of this trait should be opt in.

sffc: The macro does what I think you want it to do? It implements the trait to return None. If you want not None, there is no macro; implement the trait manually.

Manishearth (approved):
OK to land from my end, would prefer the trait be renamed, but can happen later.

robertbastian (dismissed):

provider/core/src/marker.rs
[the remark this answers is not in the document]
Manishearth: +1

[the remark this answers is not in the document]
sffc: Discussion with @robertbastian:

- The `maybe_as_var_ule` function can be gated, but not the whole trait, because we need the associated type
- Add `from_varule` to replace the `From` bound
- Use a never type (empty enum) for the non-implemented case; this lets the `from_varule` match on it (and be implementable)

provider/baked/src/zerotrie.rs
```diff
@@ -7,22 +7,28 @@
...
 pub use icu_provider::DynamicDataMarker;
+use icu_provider::{marker::MaybeAsVarULE, prelude::*};
 pub use zerotrie::ZeroTrieSimpleAscii;
+use zerovec::VarZeroSlice;
+
+#[cfg(feature = "export")]
+use zerovec::VarZeroVec;
+
+#[cfg(feature = "export")]
+pub(crate) enum BakedValue<'a> {
```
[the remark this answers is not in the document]
sffc: Is it ok if I make an issue to fix the binary search baked backend instead of deleting or updating it in this PR?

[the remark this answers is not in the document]
robertbastian: yes

provider/core/src/marker.rs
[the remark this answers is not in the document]
Manishearth: Not 100% sure how that gating would work, but will have a look when you have it ready. Would rather not have implementors of this forced to enable the export feature, i think

(yes,  that might mean not gating it, that's fine by me)

but no strong opinion, let's see how the gating works

provider/core/src/marker.rs
sffc: I applied @robertbastian's suggestion to create a new `icu_provider::marker::data_struct!` macro rules.

utils/zerovec/src/ule/plain.rs
sffc: I removed this impl but added a new impl and modified the safety constraint. PTAL.

provider/core/src/marker.rs
sffc: Changed it to `MaybeExportAsVarULE`

provider/core/src/hello_world.rs
sffc: (I am still allowing the trait to be implemented on any data struct)

components/list/src/provider/mod.rs
sffc: I did something like this.

provider/core/src/hello_world.rs
```diff
@@ -41,6 +41,24 @@ impl Default for HelloWorld<'_> {
...
+        }
+    }
+}
+
+crate::marker::data_struct!(
+    HelloWorld<'a>,
+    varule: str,
+    as_varule: HelloWorld::as_varule,
+    from_varule: HelloWorld::from_varule
+);
```
robertbastian, line 60:
```suggestion
crate::marker::data_struct!(
    HelloWorld<'a>,
    varule: str,
    as_varule: |v| &v.message,
    from_varule: |v| Self { message: Cow::Borrowed(v) }
);
```

provider/core/src/export/payload.rs
```diff
@@ -17,13 +18,20 @@ trait ExportableDataPayload {
...
     fn as_any(&self) -> &dyn Any;
     fn eq_dyn(&self, other: &dyn ExportableDataPayload) -> bool;
 }
 
 impl<M: DynamicDataMarker> ExportableDataPayload for DataPayload<M>
 where
-    for<'a> <M::DataStruct as Yokeable<'a>>::Output: Bake + BakeSize + serde::Serialize + PartialEq,
+    for<'a> <M::DataStruct as Yokeable<'a>>::Output:
+        Bake + BakeSize + serde::Serialize + MaybeExportAsVarULE<'a> + PartialEq,
+    for<'a> <<M::DataStruct as Yokeable<'a>>::Output as MaybeExportAsVarULE<'a>>::VarULE: VarULE,
```
robertbastian, line 30: I'm really not a fan of naming associated types by their bound. `...::VarULE: VarULE` does not really tell me what any of this *is*. Can you come up with clear names for these associated type, like we renamed `Yokeable` to `DataStruct`?

provider/baked/src/zerotrie.rs
```diff
@@ -111,17 +192,61 @@ impl<M: DataMarker> super::DataStore<M> for Data<M> {
...
+        Self { trie, values }
+    }
+}
+
+impl<M: DataMarker> super::DataStore<M> for DataForVarULEs<M>
+where
+    M::DataStruct: MaybeExportAsVarULE<'static>,
+    <M::DataStruct as MaybeExportAsVarULE<'static>>::VarULE: VarULE,
+    <M::DataStruct as MaybeExportAsVarULE<'static>>::EncodeAsVarULE:
+        From<&'static <M::DataStruct as MaybeExportAsVarULE<'static>>::VarULE>,
```
robertbastian, line 229: why is there still a `From` bound?

sffc: I need to transform `&str` (a `&Self::VarULE`) to `&str` (a `Self::EncodeAsVarULE`)

provider/baked/src/zerotrie.rs
```diff
@@ -111,17 +192,58 @@ impl<M: DataMarker> super::DataStore<M> for Data<M> {
...
+{
+    fn get(
+        &self,
+        id: DataIdentifierBorrowed,
+        attributes_prefix_match: bool,
+    ) -> Option<DataPayload<M>> {
+        get_index(self.trie, id, attributes_prefix_match)
+            // Safety: Allowed since `i` came from the trie and the field safety invariant
+            .map(|i| unsafe { self.values.get_unchecked(i) })
+            .map(From::from)
```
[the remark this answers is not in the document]
sffc: ^ would like to highlight this open question

provider/core/src/marker.rs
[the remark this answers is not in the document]
Manishearth: I prefer not feature gating. This is a trait impl and I'd rather keep it clean.

I'm happy to have the impl itself gated on datagen: the Encode impls could pull in additional deps, but I think the trait itself should always be available to avoid specifying transitive deps.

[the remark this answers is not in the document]
sffc: > the Encode impls could pull in additional deps

This is compelling to me.

I would prefer not to "split" the trait with a function enabled only on certain features enabled. That sounds like a bit of a disaster since the features need to be always 100% in sync. If the icu_provider trait enables its export feature, but a crate doesn't have its datagen feature enabled, then you break code.

So I would like to keep the trait unconditionally exported from icu_provider and conditionally implemented, in whole, in component crates.

provider/baked/src/zerotrie.rs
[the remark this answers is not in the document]
sffc: Given the proposed conclusion on https://github.com/unicode-org/icu4x/pull/6133#discussion_r1960648578, I want to double-down on my position that we should use two traits, one for datagen and one for runtime. I would like the datagen trait to be the `MaybeExportAsVarULE` and the runtime trait to be `ZeroFrom`. I am willing to use custom traits in both situations.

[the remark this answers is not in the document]
sffc: Technically this could be 3 traits, since there are 3 unique, disjoint things we need, in different situations:

1. `maybe_as_varule -> Option<(impl VarULE or impl EncodeAsVarULE)>`
    - Needed in datagen only
2. `pub type VarULE`
    - Needed in datagen for _all structs_ and at runtime _for structs that use it_
3. `from_varule -> Self`
    - Needed at runtime for structs that use it

A potential design for this would be 3 traits: (2) would be a base trait and (1) and (3) would extend it.

provider/core/src/marker.rs
```diff
@@ -108,6 +108,99 @@ impl<M: DataMarker + Sized> DataMarkerExt for M {
...
+/// 7. Byte equality is semantic equality
+unsafe impl VarULE for NeverVarULE {
+    fn validate_bytes(_bytes: &[u8]) -> Result<(), UleError> {
+        Err(UleError::parse::<Self>())
+    }
+    unsafe fn from_bytes_unchecked(bytes: &[u8]) -> &Self {
+        // Note: this returns a thin pointer instead of a fat pointer
+        &*(bytes.as_ptr() as *const Self)
+    }
+}
```
sffc, line 202: @Manishearth Flagging this code for a more detailed unsafe review.

utils/zerovec/src/ule/mod.rs
```diff
@@ -261,7 +261,7 @@ where
 /// Safety checklist for `VarULE`:
 ///
 /// 1. The type *must not* include any uninitialized or padding bytes.
-/// 2. The type must have an alignment of 1 byte.
+/// 2. The type must have an alignment of 1 byte, or it is a ZST or an empty enum.
```
sffc, line 264: @Manishearth Flagging this code for a more detailed unsafe review.

provider/core/src/marker.rs
```diff
@@ -108,6 +108,99 @@ impl<M: DataMarker + Sized> DataMarkerExt for M {
...
+                Some($as_varule(self))
+            }
+            fn from_varule(input: Self::EncodeAsVarULE) -> Self {
+                $from_varule(input)
+            }
+        }
+    };
+}
+#[doc(inline)]
+pub use __data_struct as data_struct;
```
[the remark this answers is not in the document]
robertbastian: no, most of what the old proc macro does is now done by `data_marker!`

components/list/src/provider/mod.rs
```diff
@@ -69,7 +69,7 @@ data_marker!(
     ListFormatterPatterns<'static>,
 );
 
-icu_provider::marker::does_not_deref_to_varule!(ListFormatterPatterns<'_>);
+icu_provider::marker::data_struct!(ListFormatterPatterns<'_>);
```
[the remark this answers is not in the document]
sffc: I can export this from `icu_provider`. Do you also have an opinion on where the trait and type (`MaybeExportAsVarULE`, `NeverVarULE`) should be exported?

provider/core/src/marker.rs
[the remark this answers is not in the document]
Manishearth: I'm aligned

utils/zerovec/src/ule/mod.rs
Manishearth: let's exclude empty enum

"or it is a ZST that is safe to construct" so that we disallow free construction of token ZSTs

provider/core/src/marker.rs
Manishearth: This shouldn't be an empty enum IMO, there are far too many tricky things there. Let's just use a ZST and have it always error.

sffc: The empty enum is required in order to implement @robertbastian's preferred solution (discussed in the meeting this morning) where `from_varule` is always implemented; the empty enum means that types that don't support VarULE can match on it. However, I would prefer a different set of traits.

sffc: Are you concerned about `from_bytes_unchecked` taking in a fat pointer and returning a thin pointer?

provider/baked/src/zerotrie.rs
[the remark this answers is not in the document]
Manishearth: Using ZeroFrom as the runtime trait sounds good to me.

provider/core/src/marker.rs
Manishearth: The unsafe review for zerovec becomes _much_ trickier when you start considering uninhabited types, since it's insta-UB to have an `Empty`, `&Empty`, or other such types floating around. We have not reviewed the zerovec crate for this, and I am unwilling to perform such a deep review and updation of invariants for this change.

Generally uninhabited types do not mix well with unsafe code. It's typical to assume they won't be introduced around this type of unsafe code when reviewing it, and opting uninhabited types into unsafe traits is a major red flag that requires reevaluation of this assumption.


`validate_bytes` always failing makes this _plausibly_ correct, but I'm not able to quickly verify that in a nonlocal way. We have a lot of code using these traits.

Manishearth: IMO just abort instead of matching on the empty enum.

sffc: Yes, I will remove the condition and impl for uninhabited types.

> Are you concerned about `from_bytes_unchecked` taking in a fat pointer and returning a thin pointer?

^ can you address this?

sffc: To clarify: I can do either of the following

1. Implement `ULE` on a ZST like `()`, and use `[()]` as the placeholder VarULE type (this is what I did in an earlier version of this PR)
2. Implement `VarULE` on the ZST and use `()` directly as the placeholder VarULE, but then it is a thin pointer.

Manishearth: Oh, I hadn't noticed that had changed as well. I am somewhat concerned about that, we do assume VarULE types are DSTs but I don't actually think we document it as a requirement on the trait, or have unsafe code relying on it.

So it should be fine. But the `[ZST]` approach is also good.

utils/zerovec/src/ule/plain.rs
Manishearth: keeping safety discussion in one place

Manishearth: I'll be sure to provide safety docs once we've settled on an approach

provider/core/src/export/payload.rs
sffc: It tells you what it literally _is_ (it is an associated VarULE type, which is 100% correct), just not what it _does_.

I don't have any other name that is as precise and correct as this. But if you want to brainstorm, and if you like the `DataStruct` themed naming:

- `DataStructAsVarULE`
- `OptionalVarULE`
- `CompactDataStruct`
- `DataStruct` (it is on a different trait so using the same name is possible)

provider/baked/src/zerotrie.rs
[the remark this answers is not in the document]
sffc: Currently exploring this constellation of traits:

```rust
pub trait MaybeAsVarULE {
    type VarULE: ?Sized + VarULE;
}

#[cfg(feature = "export")]
pub trait MaybeEncodeAsVarULE: MaybeAsVarULE {
    fn maybe_encode_as_varule(&self) -> Box<Self::VarULE>;
}

pub trait FromVarULE<'a>: MaybeAsVarULE {
    fn from_varule(varule: &'a Self::VarULE) -> Self;
}
```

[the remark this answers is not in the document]
sffc: 1. The trait makes it more explicit that it is the reverse operation as `MaybeEncodeAsVarULE`
2. A hand-written impl is sometimes required since there may be intermediate fields involved; for example, if the data struct is `Cow<str.>`, the auto-derived impl will work, but not if the data struct is `HelloWorld(Cow<str>)`
3. @robertbastian was opposed to using ZeroFrom

(none of these points are compelling to me and I am happy to switch to ZeroFrom)

provider/core/src/export/payload.rs
Manishearth: DataStructAsVarULE seems nice. but I don't think this part is overly important to get right

provider/baked/src/zerotrie.rs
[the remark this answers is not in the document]
Manishearth: > The trait makes it more explicit that it is the reverse operation as MaybeEncodeAsVarULE


ZeroFrom is supposed to be the reverse operation of EncodeAsVar, and we don't actually need a reverse of MaybeEncodeAsVar

Our trait universe here is already rather confusing: if we can make it work with ZeroFrom and EncodeAsVar, we should.  I think what you had previously with the single trait was better, except it should have used ZeroFrom. I'm not a fan of proliferating more traits, _I'm_ still struggling to understand how these all fit together.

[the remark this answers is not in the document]
sffc: > if we can make it work with ZeroFrom and EncodeAsVar, we should.

Do you have a suggestion on this? What should non-varule data structs do in their EncodeAsVarULE impls?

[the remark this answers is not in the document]
sffc: This comment is not relevant since we don't have specialization, but ideally what I'd like to do here is

```rust
impl default MaybeEncodeAsVarULE for T {
    fn maybe_encode_as_var_ule(&self) -> Option<Box<T>> {
        None
    }
}

impl MaybeEncodeAsVarULE for T where T: EncodeAsVarULE {
    fn maybe_encode_as_var_ule(&self) -> Option<Box<T>> {
        Some(zerovec::encode_varule_to_box(self))
    }
}
```

But since I can't do that, then I need to implement `MaybeEncodeAsVarULE` manually, which I think has fed the perception throughout this review process that it is an "opt out".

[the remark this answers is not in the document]
Manishearth: > Do you have a suggestion on this? What should non-varule data structs do in their EncodeAsVarULE impls?

Oh, I see the problem. We could have a blanket Encode impl for all T on the NeverVar ZST but if that doesn't work, then a separate trait is fine.

> which I think has fed the perception throughout this review process that it is an "opt out".



No, I understood that part: the "opt out" comment was specifically about the macro setup, as I have said before. That's already been clarified now. It was misleading to see non-VarULE optimized data structs in list and properties have to invoke an opt-out macro, but what was actually happening was subtler.

provider/baked/src/zerotrie.rs
```diff
@@ -39,15 +48,61 @@ pub(crate) fn bake(
...
+                BakedValue::VarULE(bytes) => &**bytes,
+                BakedValue::Struct(_) => {
+                    unreachable!(
+                        "All instances should equivalently return Some or None in MaybeEncodeAsVarULE"
+                    )
+                }
+            })
+            .collect::<Vec<&[u8]>>();
+        let vzv = VarZeroVec::<[u8]>::from(&byteses);
+        let vzv_bytes = vzv.as_bytes().bake(ctx);
```
robertbastian, line 87: can this not be more typed?

provider/core/macros/src/lib.rs
sffc: Discussion:

- @sffc wants this to land before the audit so that the macros can be wired up at the same time as we perform the data struct audit to save work overall
- When the audits are performed, no data structs will be opted in

provider/core/src/marker.rs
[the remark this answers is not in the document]
sffc: This thread is about the feature gating. I would like to do the feature gating after the proc macro is removed.

provider/baked/src/zerotrie.rs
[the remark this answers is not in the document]
sffc: Discussion:

- ZeroFrom is the VarULE universe's opposite of EncodeAsVarULE, so we should use it.

components/list/src/provider/mod.rs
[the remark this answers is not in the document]
sffc: I moved these to a new module called `icu_provider::ule`.

- @robertbastian The traits can live there, but the macro should live at the top level.
- @sffc I can't put the macro at the top level while the proc macro is there.
- @robertbastian Can you put the data struct at the top level and call it `data_struct_new`?
- @sffc Ok.

provider/core/src/export/payload.rs
sffc: Other suggestions:

- `EncodedStruct`
- `VarULEEncoded`

Agreement: `EncodedStruct`

provider/baked/src/zerotrie.rs
sffc: The `From` bound is gone now. Will be replaced by ZeroFrom.

Manishearth:
Current contentions: how the traits look.
 - @robertbastian: I don't want people to have to think of `ule` here. This should  be the toplevel data struct so we should just have one macro, once we gfinish the audit
 - @manishearth: And that's planned for 2.0, yes?
 - @robertbastian: Yes. We can fix once the audit is done
 - @sffc: Can I resolve the opt-in/opt-out issue? 
 - (room) : Yes
 - (arbitrary data structs)
 - @robertbastian: We agreed to special case cow, yes?
 - @manishearth: kinda prefer trait based
 - @sffc tried a bunch of things on the plane ride. Works fine `Cow<str>`, but doesn't generalize.
 - @manishearth: I understood the original agreement to hardcode types in icu_baked_provider, and it just knows how to codegen.
 - @sffc: second opt-in/opt-out issue?
 - @robertbastian: resolved by naming of module 
 - @sffc. will revisit
 - (scope creep issue)
 - @robertbastian: scope creep to apply to existing data structs. blocks data makrer audit
 - @sffc: Wiring up the new macro happens in the audit anyway. would like to get this in before the audit
 - @robertbastian: I don't wish to opt anyone in to this now
 - @robertbastian We want this for 2.0 for the breaking changes in export/provider, we don't want to actually *use* this yet since we've got other priorities and we should measure this.
 - @sffc (disagrees on whether or not we agreed upon doing this previously)
 - @robertbastian - I do also want to see benchmarks before we land this, how much this improves compile times, and how it affects runtime performance
 - @sffc In previous quarters, I posted evidence in the issue about the impact on compile times.
 - @robertbastian: If we do this I want to do this right
 - @sffc But you said you didn't want scope creep
 - @robertbastian: Yes, but the whole PR is scope creep IMO, this is me illustrating that
 - @manishearth: I am perfectly happy to block 2.0 on this, this is an existential issue for some components. displaynames?
 - @robertbastian: I'd like to see numbers.
 - @sffc For the purpose of this thread: can we agree to land this and add macro calls during the audit?
 - @robertbastian: yes, but not opt anyone in
 - discussion on export feature
 - @sffc: Tried this, then everyone needs to enabvle export to implement this
 - @sffc Proposal: land without feature gating, then try after the audit
 - @robertbastian: Sounds like more work than doing the audit now and then adding it to every data struct later. Not that many data structs
 - @sffc: (repeats proposal)
 - @robertbastian: it seems like it would be more work, but fine. 
 - @robertbastian: You seem to have implemented feature gating for the new data struct macro, just not the old one. 
 - Discussion on specific traits
 - @manishearth ZeroFrom is the VarULE->normal path.
 - @robertbastian: Weird choice from the varule universe but if that's how it does it then fine
 - (agreed on ZeroFrom)
 - @manishearth Would like to try killing the encode trait
 - Discussion on modules
 - @sffc: icu_provider::ule
 - @robertbastian: the two traits go there, the macro should not
 - @manishearth that gets resolved after the audit
 - @robertbastian: can you put the macro at the top and call it data_struct_new? easier to rename instead of fixing imports
 - @manishearth i like that
 - @sffc i can do that
 - @robertbastian: Macro should take expr not path for (reasons)
 - @sffc can do that
 - Discussion on assoc type name
 - @sffc can do DataStructAsVarULE?
 - @robertbastian: EncodedStruct?
 - @manishearth, @sffc: +1




```rust
// current:
trait MaybeAsVarULE {
    pub type EncodedStruct;
}
trait MaybeEncodeAsVarULE: MaybeAsVarULE {
    fn maybe_as_var_ule(&self) -> &Self::EncodedStruct
}
```

On DataPayload ExportMarker:

```rust
// Current:
impl DataPayload<ExportMarker> {
    pub fn tokenize(&self, env: &CrateEnv) -> TokenStream {
        self.get().payload.bake_yoke(env)
    }
    pub fn maybe_as_varule_bytes(&self) -> Option<Box<[u8]>> {
        self.get().payload.maybe_as_varule_bytes()
    }
}

// Rob's proposal:
impl DataPayload<ExportMarker> {
    pub fn tokenize(&self, env: &CrateEnv) -> TokenStream {
        self.get().payload.bake_yoke(env)
    }
    pub fn tokenize_encoded_seq(payloads: &[DataPayload<ExportMarker>], env: &CrateEnv) -> TokenStream {
        // may need &self to call object safe stuff
        // build nicely typed VZV from payloads. this will need some Any magic.
        // It will go through ExportableDataPayload to do it.
        vzv.bake(env)
    }
}
```

Manishearth (dismissed):
I'm happy to land this without safety comments and I can add them myself, in the interest of reducing complexity in landing

sffc:
I believe I have addressed all feedback, and I caught up with the bitrot caused by some audit PRs landing in the last few days.

Follow-ups are in #5230 and elsewhere.

@Manishearth plans to review the diff to the safety docs and suggest changes in a follow-up PR.

Autosubmit = true

components/list/src/provider/mod.rs
```diff
@@ -69,6 +69,11 @@ data_marker!(
     ListFormatterPatterns<'static>,
 );
 
+icu_provider::data_struct_new!(
+    ListFormatterPatterns<'_>,
+    #[cfg(feature = "datagen")]
```
robertbastian, line 74: why is this setting a feature, but the other invocations are not?

provider/baked/src/export.rs
```diff
@@ -565,20 +565,20 @@ impl DataExporter for BakedExporter {
                     .any(|(_, ids)| ids.iter().any(|id| !id.locale.is_default()));
 
             let mut baked_values = deduplicated_values
-                .into_iter()
+                .iter()
                 .map(|(payload, ids)| {
                     stats.structs_count += 1;
                     stats.identifiers_count += ids.len();
                     stats.structs_total_size += payload.baked_size();
```
robertbastian, line 572: issue: I don't think the size calculations are correct anymore

can be fixed in a follow-up

provider/core/src/export/payload.rs
```diff
@@ -4,26 +4,37 @@
...
 use databake::{Bake, BakeSize, CrateEnv, TokenStream};
 use yoke::*;
+use zerovec::VarZeroVec;
+
+#[cfg(doc)]
+use zerovec::ule::VarULE;
 
 trait ExportableDataPayload {
-    fn bake_yoke(&self, env: &CrateEnv) -> TokenStream;
+    fn bake_yoke(&self, ctx: &CrateEnv) -> TokenStream;
```
robertbastian, line 19: why?

provider/core/src/hello_world.rs
```diff
@@ -41,6 +41,21 @@ impl Default for HelloWorld<'_> {
     }
 }
 
+impl<'a> ZeroFrom<'a, str> for HelloWorld<'a> {
```
robertbastian, line 44: nb: I still want it on the record that I disapprove of implicitly using `ZeroFrom` to do the reverse transform. It's not much better than `From`, it can just borrow better.

provider/core/src/varule_traits.rs
```diff
@@ -0,0 +1,121 @@
...
+use zerovec::{maps::ZeroMapKV, ZeroMap2d};
+
+/// A trait that associates a [`VarULE`] type with a data struct.
+///
+/// Some data structs can be represented compactly as a single [`VarULE`],
+/// such as `str` or a packed pattern. This trait allows for data providers
+/// to use storage optimizations for such types.
+///
+/// ❗ Not all data structs benefit from this optimization. It is best when the
+/// data struct is multiplied across a large number of data marker attributes.
```
robertbastian, line 17: this sounds like multiplying across a large number of data marker attributes is an alternative to this optimisation
```suggestion
/// ❗ Not all data structs benefit from this optimization. It works best when the
/// data struct is multiplied across a large number of data marker attributes.
```

provider/core/src/varule_traits.rs
```diff
@@ -0,0 +1,121 @@
...
+use zerovec::ule::VarULE;
+
+#[cfg(feature = "alloc")]
+use zerovec::{maps::ZeroMapKV, ZeroMap2d};
+
+/// A trait that associates a [`VarULE`] type with a data struct.
+///
+/// Some data structs can be represented compactly as a single [`VarULE`],
+/// such as `str` or a packed pattern. This trait allows for data providers
+/// to use storage optimizations for such types.
```
robertbastian, line 14: issue: you frame this as "storage optimizations", but that's not why we're doing this.

provider/core/src/varule_traits.rs
```diff
@@ -0,0 +1,121 @@
...
+impl<'a, K0, K1, V> MaybeAsVarULE for ZeroMap2d<'a, K0, K1, V>
+where
+    K0: ZeroMapKV<'a>,
+    K1: ZeroMapKV<'a>,
+    V: ZeroMapKV<'a>,
+    K0: ?Sized,
+    K1: ?Sized,
+    V: ?Sized,
+{
+    type EncodedStruct = [()];
```
robertbastian, line 95: why?

sffc: Why what?

Why I am using `[()]` here is documented in the docs of the associated type:

> The `VarULE` type for this data struct, or `[()]` if it cannot be represented as `VarULE`.

provider/core/src/export/payload.rs
sffc: Making the parameter named the same thing everywhere to make it easier to refactor code, and because `env` shadows a macro of the same name... I do complain about unrelated renames/refactorings making it into your PRs often, but this is like only a few lines of code beyond the ones I was already changing.

provider/baked/src/export.rs
sffc: Added a TODO

components/list/src/provider/mod.rs
sffc: It can be applied anywhere such a feature exists. Fixed.

provider/core/src/varule_traits.rs
robertbastian: no, this is a compile time improvement, at the expense of runtime performance

I have not seen any data that this reduces storage size

provider/core/src/varule_traits.rs
```diff
@@ -0,0 +1,121 @@
...
+/// such as `str` or a packed pattern. This trait allows for data providers
+/// to use storage optimizations for such types.
+///
+/// ❗ Not all data structs benefit from this optimization. It is best when the
+/// data struct is multiplied across a large number of data marker attributes.
+///
+/// Both [`MaybeAsVarULE`] and [`MaybeEncodeAsVarULE`] should be implemented
+/// on all data structs. The [`data_struct!`] macro provides an impl.
+pub trait MaybeAsVarULE {
+    /// The [`VarULE`] type for this data struct, or `[()]`
```
robertbastian, line 22: ok, let's have this conversation here then:

why use `[()]`?

provider/core/src/varule_traits.rs
sffc: It is a storage optimization that reduces compile times and binary size.

sffc: I removed the word "storage". The point of this remark is that it can be used for optimizations at the discretion of the data provider.

sffc: As we discussed in the WG meeting, I previously shared what I consider sufficient preliminary data in the main issue #5230, and more data can be gathered once this lands and is integrated into places such as datetime.

provider/core/src/export/payload.rs
sffc: Including this change,

25 hits in 6 files for `ctx: &CrateEnv`

13 hits in 7 files for `env: &CrateEnv`

So we were already inconsistent and we are still inconsistent. This file was internally inconsistent. Let's not bikeshed this here.

robertbastian: ok, pointless diff but I'm the bad guy for mentioning this

provider/core/src/varule_traits.rs
Manishearth: Right. I'm fine with a zero-sized VarULE type, and I'm fine with `[()]` or `[DummyZST]` or something, the former concerns me a _little bit_ more but I already went through our code and I think it's not risky.


Uninhabited types are a whole other ballgame and we should not be introducing those.

provider/core/src/hello_world.rs
sffc: #6183

Manishearth (approved):
Glad to see this landing!

sffc:
> can you post the semver diffs that this locks us in to?

Changed APIs:

1. `impl<M> UpcastDataPayload<M> for ExportMarker` gains a bound on `MaybeEncodeAsVarULE`

New APIs: 

1. `DataPayload<ExportMarker>` gains a `pub fn tokenize_encoded_seq`
2. `impl<'a> ZeroFrom<'a, str> for HelloWorld<'a>`
3. New traits `icu_provider::ule::MaybeAsVarULE`, `icu_provider::ule::MaybeEncodeAsVarULE`
4. New struct `icu_provider_baked::zerotrie::DataForVarULEs`
5. New macro `icu_provider::data_struct_new` (I think we should have this macro regardless)
6. `unsafe impl ULE for ()` (something we can/should have regardless)

Manishearth:
> unsafe impl ULE for () (something we can/should have regardless)

I actually disagree with this one and have a mild preference for using a custom ZST here, but didn't think it that big a deal. ZSTs are weird and there are multiple ways one might expect a `()` ZST to behave here.

Happy to make that change if people agree.

robertbastian:
> impl<M> UpcastDataPayload<M> for ExportMarker gains a bound on MaybeEncodeAsVarULE

This has given me some confusing errors while migrating this. The place where this will fail if the macro hasn't been called is in `impl_dynamic_data_provider` inside a registry callback in `icu_provider_source`. If we add the bound on `DataMarker::DataStruct`, it should give much clearer error messages, and it's the natural place for this, if we say `data_struct!` needs to be called on each data struct.

sffc:
> > unsafe impl ULE for () (something we can/should have regardless)
> 
> I actually disagree with this one and have a mild preference for using a custom ZST here, but didn't think it that big a deal. ZSTs are weird and there are multiple ways one might expect a `()` ZST to behave here.
> 
> Happy to make that change if people agree.

What are the multiple ways one might expect `impl ULE for ()` to behave?

The one I implemented was "Ok if the bytes are empty, Err if they are nonempty". I guess one could also implement "Always Err"?

Manishearth:
Yes, effectively. It's the difference between whether people plan to actually use it or just have it be a sentinel.

Manishearth:
Even `&()` is mostly "valid" in the sense that it's a thing that makes sense in some generic contexts: it would be silly to _exclude_ `()` from generic contexts just because its references are strange. But the ULEverse already excludes lots of types, its driven by explicit inclusion.

I don't really see a purpose for `(): ULE` beyond as a placeholder in some generic contexts, and there I see two ways for it to work, as a "never" type or as a stand in similar to how `Vec<()>` is basically just a `usize`. For that, I'd prefer having an explicit type that makes it clear what's going on. When I see `(): ULE` I don't have much clarity at all.

We ended up releasing this in zerovec anyway (I didn't want to further block landing this) so there's not much to do about undoing it without a breaking release.

But basically: I definitely do not think this impl is "something we can/should have regardless", and in the medium term I'd like to deprecate it.

PR #6150, opened 2025-02-18
Rename `fdf` to `formatter` in decimal formatting examples and benchmarks

younies:
# Description

Consistently rename the variable name from `fdf` (fixed decimal formatter) to `formatter` (decimal formatter) across documentation, benchmarks, and tests to match recent type renaming efforts.

### DIFF
--- a/components/decimal/src/lib.rs
+++ b/components/decimal/src/lib.rs
@@ -26,9 +26,9 @@
 //! )
 //! .expect("locale should be present");
 //!
-//! let fixed_decimal = Decimal::from(1000007);
+//! let decimal = Decimal::from(1000007);
 //!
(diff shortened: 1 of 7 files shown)


components/decimal/src/lib.rs
```diff
@@ -62,15 +62,15 @@
 //! use icu::locale::locale;
 //! use writeable::assert_writeable_eq;
 //!
-//! let fdf = DecimalFormatter::try_new(
+//! let df = DecimalFormatter::try_new(
```
robertbastian, line 65: could you rename all `let df =` to `let formatter =`? I think we usually call these things `formatter`

younies: done

robertbastian (approved):

Issue #5610, opened 2024-09-27
Should TimeZoneIdMapper support efficient storage and retrieval of non-canonical IANA IDs?


sffc:
See https://github.com/unicode-org/icu4x/issues/5533#issuecomment-2380240259

Currently, TimeZoneIdMapper has two data payloads:

1. Trie: IANA string to BCP-47 ID (does NOT support random access: only key-to-value lookup)
2. Map: BCP-47 ID to canonical IANA string (random access supported)

These data structures allow efficient implementation of the following operations:

- IANA to BCP-47
- BCP-47 to IANA
- Normalize IANA
- Canonicalize IANA

However, it does _not_ support efficient storage or retrieval of a non-canonical IANA name. For example, it supports mapping from "europe/kiev" to any of the following:

- "Europe/Kiev" returned as an owned String
- "uaiev" returned as a TinyStr
- "Europe/Kyiv" returned as `&str`

But we don't have a representation that allows the user to return "Europe/Kiev" later: only "Europe/Kyiv". The client would need to store the string on their own.

A few ways to enable this behavior with data:

1. Change the map from BCP-47 ID to IANA string to also include non-canonical aliases, and return an opaque integer of some sort to select the correct alias
2. Switch from using a Trie to using a Map so that random access is supported, and then let the client save the index of their non-canonical time zone in the payload
3. Add a third payload where this is its only job

One concern I have is that any of these approaches involves returning an index of some sort into a structure. However, that structure is ephemeral and could change across ICU4X versions, so I don't want people storing these indices. We currently always return BCP-47 IDs because those have a stability policy.

An alternative solution might be to intern the strings, which could be done externally to ICU4X. If you need to save the non-canonical ID, just stuff it into your own static string interning structure that deduplicates. The total size of the interned string collection is bounded. An example utility for this is [elsa::FrozenIndexSet](https://docs.rs/elsa/latest/elsa/index_set/struct.FrozenIndexSet.html).

Thoughts @justingrant?

justingrant:
Thanks @sffc for starting this conversation.  I don't yet have an opinion about the best solution here, but I've got some initial questions/comments below. 

The main guidance I'd have (based on my experience with time zone canonicalization in ECMAScript) is that a time zone API should enable an efficient way to answer the following questions/usecases: 

1. Is this string a valid IANA ID?
[...]

PR #6152, opened 2025-02-18
Restructure `icu_time`

robertbastian:
This fixes some docs as canonical locations are better.

Also audits data markers (#4991) and implements new IANA parser API (#5610).


ffi/capi/src/iana_parser.rs
```diff
@@ -47,149 +44,19 @@ pub mod ffi {
...
-        hidden
-    )]
-    #[diplomat::rust_link(icu::time::zone::iana::IanaParserExtended::inner, FnInStruct, hidden)]
-    #[diplomat::rust_link(icu::time::zone::iana::IanaParserExtendedBorrowed, Struct, hidden)]
-    #[diplomat::rust_link(
-        icu::time::zone::iana::IanaParserExtendedBorrowed::inner,
-        FnInStruct,
-        hidden
-    )]
-    pub struct IanaParserExtended(
```
robertbastian, line 132: I had to remove `IanaParserExtended` for now because we cannot return `(TimeZone, &str, &str)` from Diplomat. Need some API design.

PR #6158, opened 2025-02-19
Parse system locales in `env_preferences`

Finchiedev:
As per the discussion in #6028, I have created a POSIX locale parser/converter, currently hidden in the private `env_preferences::parse` module. This is meant to change at some point while the PR is being drafted, especially once I add support for other platforms. Once more platforms are supported, I would also like to implement universal and platform-specific APIs, as per [this comment](https://github.com/unicode-org/icu4x/discussions/6028#discussioncomment-12037139) by @zbraniecki.

My current thinking on code structure is to have some distinction between platform `fetch` and `parse` code (either using modules or files), but please let me know if all platform logic should just be kept in the same file.

Of course, feedback on the code itself would be very much appreciated!

### DIFF
--- a/utils/env_preferences/Cargo.toml
+++ b/utils/env_preferences/Cargo.toml
@@ -18,6 +18,8 @@ include.workspace = true
 
 [dependencies]
 core-foundation-sys = "0.8.6"
+icu_locale = { workspace = true }
+icu_provider = { workspace = true }
 libc = "0.2.155"
 
 [dependencies.windows]
@@ -34,4 +36,8 @@ features = [
 ]
 
 [dev-dependencies]
-icu_locale = { path = "../../components/locale" }
+icu_locale = { workspace = true }
+
+[features]
+default = ["logging"]
+logging = ["icu_provider/logging"]
--- a/utils/env_preferences/src/parse/mod.rs
+++ b/utils/env_preferences/src/parse/mod.rs
@@ -1,0 +1,5 @@
+// This file is part of ICU4X. For terms of use, please see the file
+// called LICENSE at the top level of the ICU4X source tree
+// (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
+mod posix;
--- a/utils/env_preferences/src/parse/posix.rs
+++ b/utils/env_preferences/src/parse/posix.rs
@@ -1,0 +1,315 @@
+// This file is part of ICU4X. For terms of use, please see the file
+// called LICENSE at the top level of the ICU4X source tree
+// (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
+use icu_locale::extensions::unicode::{key, value};
+use icu_locale::extensions::Extensions;
+use icu_locale::subtags::{language, script, variant, Language, Region, Variants};
+use icu_locale::{LanguageIdentifier, Locale};
+
+// An incomplete list of POSIX language names that need to be converted to a BCP-47 equivalent
+// Some of these are approximate and not exact (e.g. "C"->"und")
+// This is based on real-world data only, so will hopefully cover most of the more common edge cases
+// Ideally, this would be part of the CLDR dataset, see discussion for more details:
+// https://github.com/unicode-org/icu4x/discussions/6028#discussioncomment-12227372
+const POSIX_LANGUAGE_NAMES: [(&str, &str); 39] = [
+    // Convert the default "C"/"POSIX" locale to `Locale::unknown()`
+    ("c", "und"),
+    ("posix", "und"),
+    // Native language names
+    ("bokmal", "nb"),   // Norwegian Bokmål
+    ("dansk", "da"),    // Danish
+    ("deutsch", "de"),  // German
+    ("eesti", "et"),    // Estonian
+    ("galego", "gl"),   // Galician
+    ("hrvatski", "hr"), // Croatian
+    ("nynorsk", "nn"),  // Norwegian Nynorsk
+    ("slovene", "sl"),  // Slovenian
+    // English language names
+    ("catalan", "ca"),
+    ("croatian", "hr"),
+    ("czech", "cs"),
+    ("danish", "da"),
+    ("dutch", "nl"),
+    ("estonian", "et"),
+    ("finnish", "fi"),
+    ("french", "fr"),
+    ("galician", "gl"),
+    ("german", "de"),
+    ("greek", "el"),
+    ("hebrew", "he"),
+    ("hungarian", "hu"),
+    ("icelandic", "is"),
+    ("italian", "it"),
+    ("japanese", "ja"),
+    ("korean", "ko"),
+    ("lithuanian", "lt"),
+    ("norwegian", "no"),
+    ("polish", "pl"),
+    ("portuguese", "pt"),
+    ("romanian", "ro"),
+    ("russian", "ru"),
+    ("slovak", "sk"),
+    ("slovenian", "sl"),
+    ("spanish", "es"),
+    ("swedish", "sv"),
+    ("thai", "th"),
+    ("turkish", "tr"),
+];
+
+#[derive(Debug, PartialEq)]
+pub enum ParseError {
+    EmptyLocale,
+    EmptySection {
+        offset: usize,
+    },
+    InvalidCharacter {
+        offset: usize,
+    },
+    InvalidLocale,
+    RepeatedDelimiter {
+        first_offset: usize,
+        second_offset: usize,
+    },
+    UnorderedDelimiter {
+        first_offset: usize,
+        second_offset: usize,
+    },
+}
+
+#[derive(Debug, PartialEq)]
+pub enum ConversionError {
+    InvalidLanguage {
+        start_offset: usize,
+        end_offset: usize,
+    },
+    InvalidRegion {
+        start_offset: usize,
+        end_offset: usize,
+    },
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, PartialOrd)]
+enum Delimiter {
+    Territory,
+    Codeset,
+    Modifier,
+}
+
+impl Delimiter {
+    /// Find any optional sections, returning an error if the delimiters are invalid
+    pub fn try_find_sections(src: &str) -> Result<Vec<(usize, Self)>, ParseError> {
+        // Find the offset and delimiter of each optional section
+        let optional_sections = src
+            .chars()
+            .enumerate()
+            .flat_map(|(index, character)| match character {
+                '_' => Some((index, Self::Territory)),
+                '.' => Some((index, Self::Codeset)),
+                '@' => Some((index, Self::Modifier)),
+                _ => None,
+            })
+            .collect::<Vec<_>>();
+
+        // Find any errors in the arrangement of delimiters
+        for (index, (first_offset, first_delimiter)) in optional_sections.iter().enumerate() {
+            // Find any repeated delimiters
+            if let Some((second_offset, _second_delimiter)) = optional_sections
+                .iter()
+                // Check all delimiters past this index
+                .skip(index + 1)
+                .find(|(_second_offset, second_delimiter)| first_delimiter == second_delimiter)
+            {
+                return Err(ParseError::RepeatedDelimiter {
+                    first_offset: *first_offset,
+                    second_offset: *second_offset,
+                });
+            }
+
+            // Find any delimiters that have been invalidated by a delimiter that should appear after it
+            // For example "en.utf8_US" is invalid because codeset appears before territory
+            if let Some((second_offset, second_delimiter)) = optional_sections.get(index + 1) {
+                if first_delimiter > second_delimiter {
+                    return Err(ParseError::UnorderedDelimiter {
+                        first_offset: *first_offset,
+                        second_offset: *second_offset,
+                    });
+                }
+            }
+        }
+
+        Ok(optional_sections)
+    }
+}
+
+#[derive(Debug)]
+pub struct PosixLocale<'src> {
+    language: &'src str,
+    territory: Option<&'src str>,
+    codeset: Option<&'src str>,
+    // TODO: is it possible to have multiple modifiers?
+    modifier: Option<&'src str>,
+}
+
+impl<'src> PosixLocale<'src> {
+    /// Attempt to parse a POSIX locale.
+    ///
+    /// Locales are expected to be in the format `language[_territory][.codeset][@modifier]`;
+    /// only the language section is mandatory, all other sections are optional.
+    /// Example with all sections: `en_US.utf8@euro`
+    ///
+    /// See section 8.2 of the POSIX spec for more details:
+    /// <https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap08.html#tag_08_02>
+    pub fn try_from_str(src: &'src str) -> Result<Self, ParseError> {
+        // These cases are implementation-defined and can be ignored:
+        // - Empty locales
+        if src.is_empty() {
+            return Err(ParseError::EmptyLocale);
+        }
+        // - Any locale containing '/'
+        if let Some(offset) = src.find('/') {
+            return Err(ParseError::InvalidCharacter { offset });
+        }
+        // - Locales consisting of "." or ".."
+        if src == "." || src == ".." {
+            return Err(ParseError::InvalidLocale);
+        }
+
+        // Find any optional sections, and return any delimiter-related errors
+        let optional_sections = Delimiter::try_find_sections(src)?;
+
+        // The language field continues until the start of the first optional section, if one exists
+        let language = match optional_sections.first() {
+            Some((offset, _delimiter)) => &src[..*offset],
+            None => src,
+        };
+
+        // Make sure the language itself is non-empty
+        if language.is_empty() {
+            return Err(ParseError::EmptySection { offset: 0 });
+        }
+
+        let mut locale = Self {
+            language,
+            territory: None,
+            codeset: None,
+            modifier: None,
+        };
+
+        for (index, (start_offset, delimiter)) in optional_sections.iter().enumerate() {
+            // Find the offset of the next section, or end of the string if none exist
+            let end_offset = optional_sections
+                .get(index + 1)
+                .map(|(next_offset, _next_delimiter)| *next_offset)
+                .unwrap_or(src.len());
+
+            // Make sure this section is non-empty (more characters than just the delimiter)
+            if start_offset + 1 >= end_offset {
+                return Err(ParseError::EmptySection {
+                    offset: *start_offset,
+                });
+            }
+
+            // Write the section to the appropriate field
+            let section_value = Some(&src[start_offset + 1..end_offset]);
+            match delimiter {
+                Delimiter::Territory => locale.territory = section_value,
+                Delimiter::Codeset => locale.codeset = section_value,
+                Delimiter::Modifier => locale.modifier = section_value,
+            }
+        }
+
+        Ok(locale)
+    }
+
+    /// Attempt to convert a POSIX locale into a valid BCP-47 identifier.
+    ///
+    /// This is a best-effort conversion process, and there are valid
+    /// POSIX locales that will return an error or silently ignore data.
+    /// In particular, the codeset section is always ignored, and only some common modifiers are handled
+    /// (unknown modifiers will be silently ignored).
+    ///
+    /// If the "logging" feature is enabled, this ignored data will be logged as a warning.
+    pub fn try_convert_lossy(&self) -> Result<Locale, ConversionError> {
+        // Check if the language needs to be mapped to a BCP-47 language subtag
+        let language_code = POSIX_LANGUAGE_NAMES
+            .iter()
+            .find(|(posix_name, _subtag)| self.language.eq_ignore_ascii_case(posix_name))
+            .map(|(_posix_name, subtag)| *subtag)
+            .unwrap_or(self.language);
+
+        let mut language = match Language::try_from_str(language_code) {
+            Ok(language) => language,
+            Err(_) => {
+                return Err(ConversionError::InvalidLanguage {
+                    start_offset: 0,
+                    end_offset: self.language.len(),
+                })
+            }
+        };
+
+        let region = if let Some(territory) = self.territory {
+            match Region::try_from_str(territory) {
+                Ok(region) => Some(region),
+                Err(_) => {
+                    return Err(ConversionError::InvalidRegion {
+                        // Add 1 to skip the delimiter
+                        start_offset: self.language.len() + 1,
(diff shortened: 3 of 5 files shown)


utils/env_preferences/src/parse/posix.rs
```diff
@@ -0,0 +1,622 @@
...
+                        "devanagari" => script!("Deva"),
+                        "latin" => script!("Latn"),
+                        // Already matched against one of the known strings above
+                        _ => unreachable!(),
+                    });
+                }
+                // Saaho seems to be the only "legacy variant" that appears as a modifier:
+                // https://www.unicode.org/reports/tr35/#table-legacy-variant-mappings
+                "saaho" => language = language!("ssy"),
+                "valencia" => variants.push(variant!("valencia")),
```
robertbastian, line 296: leave a comment that this insertion keeps the vec sorted

utils/env_preferences/src/parse/posix.rs
```diff
@@ -0,0 +1,622 @@
...
+        // Script
+        expect_success("uz_UZ@cyrillic", "uz-Cyrl-UZ-posix");
+        expect_success("sd_IN@devanagari", "sd-Deva-IN-posix");
+        expect_success("sr_RS@latin", "sr-Latn-RS-posix");
+
+        // Language
+        expect_success("aa_ER@saaho", "ssy-ER-posix");
+
+        // Variant
+        expect_success("ca_ES@valencia", "ca-ES-valencia-posix");
```
robertbastian, line 367:
```suggestion
        expect_success("ca_ES@valencia", "ca-ES-posix-valencia");
```

robertbastian (commented):
code looks great, @zbraniecki can you also take a look

robertbastian:
(whoops wrong button)

utils/env_preferences/src/parse/posix.rs
```diff
@@ -0,0 +1,622 @@
...
+        let mut script = None;
+        let mut variants = vec![variant!("posix")];
+
+        if let Some(modifier) = self.modifier {
+            match modifier.to_ascii_lowercase().as_str() {
+                "euro" => {
+                    // `.set()` returns the previous value if overwritten
+                    let first_value = extensions.unicode.keywords.set(key!("cu"), value!("eur"));
+                    // Make sure to not accidentally override any other currency values
+                    debug_assert!(first_value.is_none());
```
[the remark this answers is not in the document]
Finchiedev: It was attempting to guard against future modifications, but was overly conservative so I have removed it!

utils/env_preferences/src/parse/posix.rs
Finchiedev: Good catch, thank you!

Finchiedev:
Thanks very much for your time and feedback @robertbastian, I have implemented all the suggested changes and marked them as resolved.

utils/env_preferences/src/parse/posix.rs
```diff
@@ -0,0 +1,621 @@
...
+// (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
+use icu_locale::extensions::unicode::{key, value};
+use icu_locale::extensions::Extensions;
+use icu_locale::subtags::{language, script, variant, Language, Region, Variants};
+use icu_locale::{LanguageIdentifier, Locale};
+
+// An incomplete list of POSIX language names that need to be converted to a BCP-47 equivalent
+// Some of these are approximate and not exact (e.g. "C"->"und")
+// This is based on real-world data only, so will hopefully cover most of the more common edge cases
```
zbraniecki, line 12: `nit`: please, provide source of the keys for this list. In the references discussion you list a gist, which is documenting a single generation from arch distro. Let's peel this onion. Where does that program take the data from? Who maintains the database that is used to generate it?
Please, find that and link it here.

utils/env_preferences/src/parse/posix.rs
```diff
@@ -0,0 +1,621 @@
...
+// Ideally, this would be part of the CLDR dataset, see discussion for more details:
```
zbraniecki, line 13: `opinion`: I disagree that this should be in CLDR. CLDR is a Unicode database of locale information, not a POSIX compatibility maintenance db. This file, in the crate you're building may be the best source of such mapping.

utils/env_preferences/src/parse/posix.rs
```diff
@@ -0,0 +1,621 @@
...
+    ("romanian", language!("ro")),
+    ("russian", language!("ru")),
+    ("slovak", language!("sk")),
+    ("slovene", language!("sl")),
+    ("slovenian", language!("sl")),
+    ("spanish", language!("es")),
+    ("swedish", language!("sv")),
+    ("thai", language!("th")),
+    ("turkish", language!("tr")),
+];
```
zbraniecki, line 55: `nit`: the previous leads me to belief that it is worth exposing this mapping - place it in a separate file (posix_data.rs or something) and expose a getter - `pub fn get_bcp47_language_from_posix(input: &str) -> Option<Language>`.
This may be useful independent of this crate and we may eventually want to codegen it from some db file like json.

utils/env_preferences/src/parse/posix.rs
```diff
@@ -0,0 +1,621 @@
...
+
+#[derive(Debug, PartialEq)]
+pub enum ParseError {
```
zbraniecki, line 58: `nit`: please, use `displaydoc` for errors, like in here https://github.com/unicode-org/icu4x/blob/6f62fe16a0b0fe100a0f722732a8367d25dc88a2/components/datetime/src/provider/fields/symbols.rs#L19

utils/env_preferences/src/parse/posix.rs
```diff
@@ -0,0 +1,621 @@
...
+
+impl<'src> PosixLocale<'src> {
+    /// Attempt to parse a POSIX locale.
+    ///
+    /// Locales are expected to be in the format `language[_territory][.codeset][@modifier]`;
+    /// only the language section is mandatory, all other sections are optional.
+    /// Example with all sections: `en_US.utf8@euro`
+    ///
+    /// See section 8.2 of the POSIX spec for more details:
+    /// <https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap08.html#tag_08_02>
```
zbraniecki, line 159: `praise`: thank you for linking to the spec!

utils/env_preferences/src/parse/posix.rs
```diff
@@ -0,0 +1,621 @@
...
+            Err(_) => match Language::try_from_str(self.language) {
+                Ok(language) => language,
+                Err(_) => {
+                    return Err(ConversionError::InvalidLanguage {
+                        start_offset: 0,
+                        end_offset: self.language.len(),
+                    })
+                }
+            },
+        };
```
zbraniecki, line 250: `nit`: you can make this code a bit cleaner by using chaining:

```rust
let language = get_bcp47_language_from_posix(&self.language)
    .or(|| Language::try_from_str(self.language).ok())
    .ok_or(ConversionError::InvalidLanguage { start_offset: 0, end_offset: self.language.len()})?;
```

utils/env_preferences/src/parse/posix.rs
```diff
@@ -0,0 +1,621 @@
...
+                        // Add 1 to skip the delimiter
+                        start_offset: self.language.len() + 1,
+                        // Add 1 to include the final character
+                        end_offset: self.language.len() + territory.len() + 1,
+                    });
+                }
+            }
+        } else {
+            None
+        };
```
zbraniecki, line 266: `nit`: you can make it easier to read with:

```rust
let region = self.territory
    .map(|territory| {
        Region::try_from_str(&territory).map_err(|_| ConversionError::InvalidRegion {
            start_offset: self.language.len() + 1,
            end_offset: self.language.len() + territory.len() + 1,
        })
    })
    .transpose()?;
```

utils/env_preferences/src/parse/posix.rs
```diff
@@ -0,0 +1,621 @@
...
+
+        if let Some(codeset) = self.codeset {
+            icu_provider::log::warn!("Ignoring codeset: `{codeset}`");
+        }
```
zbraniecki, line 270: `question`: I'm confused about the use of provider logging here. This is a library, not a program. My understanding is that we don't do UX (logging is a form of ux) in libraries. What I'd expect is that if you need this you could add information to the return value, or allow for outparam that you populate with such information as "omitted fields".

But frankly, I don't think this is needed at all. You're trying to construct a Unicode BCP47 Locale from POSIX Locale. It's lossy, best attempt effort. No need to document what was omitted I think.

utils/env_preferences/src/parse/posix.rs
```diff
@@ -0,0 +1,621 @@
...
+                "cyrillic" => script = Some(script!("Cyrl")),
+                "devanagari" => script = Some(script!("Deva")),
+                "latin" => script = Some(script!("Latn")),
+                // Saaho seems to be the only "legacy variant" that appears as a modifier:
+                // https://www.unicode.org/reports/tr35/#table-legacy-variant-mappings
+                "saaho" => {
+                    // The only known locale with this modifier is `aa_ER`
+                    if self.language != "aa" || self.territory != Some("ER") {
+                        icu_provider::log::warn!("Overriding language: `{}`->`ssy`", self.language);
+                    }
```
zbraniecki, line 291: `question`: the LDML you referenced indicates that `saaho` should result in language being set to `ssy`, irrelevant of the original `language`. value, why do you narrow it to `aa`?

zbraniecki: Also, similar concern about logging as above.

utils/env_preferences/src/parse/posix.rs
```diff
@@ -0,0 +1,621 @@
...
+            None
+        };
+
+        if let Some(codeset) = self.codeset {
+            icu_provider::log::warn!("Ignoring codeset: `{codeset}`");
+        }
+
+        let mut extensions = Extensions::new();
+        let mut script = None;
+        let mut variants = vec![variant!("posix")];
```
zbraniecki, line 274: `question`:
1. Why are you using Variant subtag instead of `-u-va-posix`?
2. Why are you automatically setting posix variant for that Locale? I don't fully understand the value of maintenance of this bit of information during conversion. If I start with `en_US` in POSIX and I use your code to convert it to ICU4X Locale, why do I end up with `en-US-posix` and not `en-US`? What's the impact of that difference? Can we avoid it?

utils/env_preferences/src/parse/posix.rs
```diff
@@ -0,0 +1,621 @@
...
+                // https://www.unicode.org/reports/tr35/#table-legacy-variant-mappings
+                "saaho" => {
+                    // The only known locale with this modifier is `aa_ER`
+                    if self.language != "aa" || self.territory != Some("ER") {
+                        icu_provider::log::warn!("Overriding language: `{}`->`ssy`", self.language);
+                    }
+                    language = language!("ssy")
+                }
+                // This keeps `variants` sorted; "-posix" comes before "-valencia"
+                "valencia" => variants.push(variant!("valencia")),
```
zbraniecki, line 295: `question`: why are you adding a variant here, and not in `-u-va`?

zbraniecki (commented):
This looks great! Thank you for picking it up. I left comments. Feel free to extract them to new discussion topics if you think it's big enough, or we can continue in this PR for smaller back-and-forths.

utils/env_preferences/src/parse/posix.rs
Finchiedev: This is a great point, especially since the `intl/locale.alias` file linked above explicitly mentions these are deprecated aliases and probably won't change. I was looking for a canonical list of _all_ language endonyms and exonyms but it seems like only the data from `intl/locale.alias` is actually needed.

utils/env_preferences/src/parse/posix.rs
Finchiedev: Much cleaner, thank you :)

utils/env_preferences/src/parse/posix.rs
Finchiedev: I am using the variant subtag based on [advice](https://github.com/unicode-org/icu4x/discussions/6028#discussioncomment-12200364) from @robertbastian:

> Regarding `-u-va-posix`, this could be expressed as a `Variant` as well, i.e. `de-posix`. We actually have data in this format (`en-US-posix` in collator), and we don't have any data using `-u-va-posix`, so I think we should parse to a variant.

I have no opinion either way, so happy to do whatever is best. Same with removing the `-posix` variant altogether, it seems valuable to have `Locale`s be consistent across platforms but I don't have enough experience in this area to say if that's the right call.

Might be worth noting that as a user of this library I am piping the output into [fluent-langneg-rs](https://github.com/projectfluent/fluent-langneg-rs), which I'd expect to drop the `-posix` variant anyways during language negotiation.

utils/env_preferences/src/parse/posix.rs
Finchiedev: For context, the [advice](https://github.com/unicode-org/icu4x/discussions/6028#discussioncomment-12200364) I received from @robertbastian:

> Similar with `@valencia`, I've seen this as both a variant and a subdivision tag (`-u-sd-valencia`). We don't currently have data keyed by either.

I could have misinterpreted the meaning of "variant" in the text above, happy to change it to `-u-va-valencia` or `-u-sd-valnecia` if that would be more appropriate!

utils/env_preferences/src/parse/posix.rs
zbraniecki: Thanks. @robertbastian - do you have any guidance on the addition of `posix` variant value? At most I would expect that to be optional, but I can't come up with any use case.

Finchiedev:
I've pushed some commits that re-arrange the module structure, and created an MVP of what I'd expect `env_preferences` to look like API-wise. Very happy to bikeshed these changes, or remove them entirely if they'd be better suited for another PR (or shouldn't be merged at all). I'd like to extend this API to use some kind of `LocaleOptions` so that users have a cross-platform method to select the category, for example `LocaleOptions::Time` would query `LC_TIME` on Linux and `GlobalizationPreferences.Clocks` on Windows.

utils/env_preferences/src/parse/posix.rs
robertbastian: @hsivonen the collator seems to be the only component that has data for a `-posix` variant, can you weigh in here?

zbraniecki:
I'm fine with that, but:

1. Watch out for cross-OS. POSIX  categories may not translate well to others OSes.
2. `options` term in ICU4X refers to a bag of options used to customize behavior of a component. Such category is just one of those options.

You may solve it in several ways:
a) Have cross-OS behavior be "singular" locale list, but per-OS allow for retrieving per-category so that `env_perferences::posix` can have per-category
b) Have the categories be actually ICU4X components. "Get Locales for DateTimeFormat" is universal and may result in different behavior per-OS.

Finchiedev:
Thanks @zbraniecki, should have clarified I was leaning towards the solution b) you suggested, where the categories would correspond to ICU4X components. Also, will make sure to use `options` idiomatically, thanks for the heads up :)

I've already drafted a table of the differences between categories on different operating systems, and am in the process of testing using real-world setups - I expect that the MacOS section in particular is wrong and needs correcting. Once I'm confident in the categories I'll start investigating what can map back to ICU4X components.

| Linux                      | Windows              | MacOS                   |
|----------------------------|----------------------|-------------------------|
| LC_ALL<br>LANG<br>LANGUAGE | Languages            | Preferred Languages     |
| LC_ADDRESS                 |                      |                         |
| LC_COLLATE                 |                      | List sort order (?)     |
| LC_CTYPE                   |                      |                         |
| LC_IDENTIFICATION          |                      |                         |
| LC_MEASUREMENT             |                      | Measurement system      |
| LC_MESSAGES                |                      |                         |
| LC_MONETARY                | Currencies           | Region                  |
| LC_NAME                    |                      | List sort order (?)     |
| LC_NUMERIC                 |                      | Number format<br>Region |
| LC_PAPER                   |                      |                         |
| LC_TELEPHONE               |                      |                         |
| LC_TIME                    | Clocks               | Date format<br>Region   |
|                            | Calendars            | Calendar                |
|                            | HomeGeographicRegion |                         |
|                            | WeekStartsOn         | First day of week       |
|                            |                      | Temperature             |

Table references:
https://man.archlinux.org/man/locale.7
https://support.apple.com/en-au/guide/mac-help/intl163/mac (https://stackoverflow.com/questions/45511458/get-user-preferred-temperature-setting-in-macos)
https://learn.microsoft.com/en-us/uwp/api/windows.system.userprofile.globalizationpreferences?view=winrt-26100#properties

Finchiedev:
I've had some time to investigate MacOS locale preferences, and it seems like there is a lot more customization than POSIX/Windows - for example you can customize the exact date format or which characters to use for number separators. Assuming this is information that should be exposed to the user, and it should be done so in a cross-platform way, I can think of a few different ways to design this API:

1.  Some kind of `FromEnv` trait that is implemented for each relevant ICU component configuration (e.g. `icu_datetime::CompositeFieldSet`, `icu_collator::CollatorOptions`), with all code inside `env_preferences`, likely with each dependency/implementation behind a feature flag
2.  Adding a `try_from_env` or similar method directly to the aforementioned components (I don't think this is something worth doing, just adding for completeness)
3.  A different `env_preferences` function for each ICU component, such as `try_get_short_date_format() -> icu_datetime::CompositeFieldSet` (with similar feature flags for each component)

I am leaning towards option 3, but would very much appreciate feedback on if that's the right design, or if it's not worth the complexity to implement in the first place.

zbraniecki:
I suggest separating this out for later. (so, effectively option 3, but don't block landing of this PR on it).

Acquiring *locale* is unfortunately limited to what Locale Unicode Extensions support. Not only MacOS, Windows also allows for customization of date/time format patterns and we don't have a way to encode it as a locale (there's no `en-Latn-US-u-dtf-YYYY/mm/dd` or equivalent).

In anticipation of the reality that OSes have regional preferences not encodable in Locale I constructed Preferneces API in ICU4X to allow for encoding information that doesn't come from Unicode Extensions. That's *specifically* for such case!

So I'm aligned - we want to extract Locale from the OS for a given ICU4X component. This is what you're improving now. And it is limited to what we can encode in a Locale.

Separately, `env_preferences` should allow extracting *more* information that is *not* encodable in a Locale. I suggest doing it in a separate PR later.

For the design conversation on how it may look like - 

I specifically added tests for icu_preferences to showcase how Preferences can integrate with `env_preferences`, but those were unfortunately removed when `icu_preferences` got merged into `locale_core`. I am now sad that I didn't oppose that removal back then. But you can find them here: https://github.com/zbraniecki/icu4x/blob/preferences/utils/preferences/tests/dtf_os_prefs_tests.rs

What I imagine (handwaving) you end up with is that `env_preferences` handle two use cases:
1) You can get a Locale out of environment.
2) You can get Preferences out of environment.

Both APIs should allow for fine-tuning for a given use case (`get_foo(options { component: datetime | numbers, ... })`).

Retrieving a locale is common, but for a good integration you need to retrieve Preferences, and those may include `DateTimePatternShort(Pattern)` entry.

Finchiedev:
Thanks again @zbraniecki, I really appreciate the feedback! In that case I'll start finalizing this PR and investigate your suggestions later :)

zbraniecki (approved):
You didn't coalesce the errors, but I think that's fine. @robertbastian - pending your review.

utils/env_preferences/src/error.rs
```diff
@@ -49,3 +56,31 @@ impl From<FromVecWithNulError> for RetrievalError {
...
+    #[displaydoc("Unable to retrieve locales: {0:?}")]
+    Retrieval(RetrievalError),
+    #[displaydoc("Unable to parse locale: {0}")]
+    Parse(ParseError),
+}
+
+#[cfg(any(feature = "parse_posix", target_os = "linux"))]
+impl From<crate::parse::posix::PosixParseError> for LocaleError {
+    fn from(value: crate::parse::posix::PosixParseError) -> Self {
+        Self::Retrieval(RetrievalError::Posix(value))
```
robertbastian, line 72: why does a `PosixParseError` become a `RetrievalError` instead of a `ParseError`?

utils/env_preferences/Cargo.toml
```diff
@@ -30,8 +32,14 @@ features = [
...
-    "Win32_Globalization"
+    "Win32_Globalization",
 ]
 
 [dev-dependencies]
-icu_locale = { path = "../../components/locale" }
+icu_locale = { workspace = true }
+windows-core = "0.60.1"
+
+[features]
```
robertbastian, line 42: I'm not sure we need features for this. They don't pull in any additional dependencies, and the crate is pretty useless with them turned off. `cfg`ing by target should be enough

utils/env_preferences/src/error.rs
```diff
@@ -28,6 +31,10 @@ pub enum RetrievalError {
     #[cfg(target_os = "windows")]
```
robertbastian, line 31: ^`UnknownCategory` can also be linux-only?

utils/env_preferences/src/parse/mod.rs
```diff
@@ -0,0 +1,16 @@
+// This file is part of ICU4X. For terms of use, please see the file
+// called LICENSE at the top level of the ICU4X source tree
+// (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
+//! Parsing functionality for various popular operating systems
+
+// Re-export all alias functions
+mod aliases;
+pub use aliases::*;
```
robertbastian, line 9: why are we making these public?

robertbastian (commented):
looks good, couple small questions

utils/env_preferences/src/error.rs
Finchiedev: Good point, this was not very clean and has been resolved :)

utils/env_preferences/src/error.rs
Finchiedev: Good point!

utils/env_preferences/src/parse/mod.rs
Finchiedev: Based on [this comment](https://github.com/unicode-org/icu4x/pull/6158#discussion_r1966270208) by @zbraniecki, but happy to change it if that's preferred!

robertbastian (dismissed):

Finchiedev:
Apologies for invalidating your review @robertbastian, I added some small fixes to hopefully fix the errors caught by CI.

robertbastian (approved):

utils/env_preferences/src/parse/posix.rs
hsivonen: Sorry about missing this ping in February. I filed https://github.com/unicode-org/icu4x/issues/6573

PR #6163, opened 2025-02-19
Fix doc path to fields module

sffc:
Might fix nightly CI


robertbastian (approved):
> Might fix nightly CI

You can verify this by running `cargo +nightly make ci-job-doc`

PR #6165, opened 2025-02-20
Fix typo in build_libs.dart

mosuem:


mosuem:
I don't know what the test failure means - any help would be appreciated. Would it be possible to use Github CI directly for this? Out of curiosity, what is the reason for using the detour through the `ffi.toml` and duckscript?

.github/workflows/build-test.yml
```diff
@@ -376,9 +376,21 @@ jobs:
...
+    - name: Analyze Dart code
+      working-directory: ffi/dart
+      run: dart analyze --fatal-infos
+
+    - name: Test Dart code
+      working-directory: ffi/dart
+      run: dart --enable-experiment=native-assets test
+
+    - name: Smoke test by executing build_libs.dart
+      run: dart ffi/dart/tool/build_libs.dart --file bin/linux_x64 --os linux --architecture x64 --compile_type dynamic --cargo_features icu_collator,icu_datetime,icu_list,icu_decimal,icu_plurals,experimental_components,compiled_data
```
robertbastian, line 392: this fails on my machine (mac)

.github/workflows/build-test.yml
```diff
@@ -376,9 +376,21 @@ jobs:
         sdk: 3.7.0
 
     # Actual job
-    - name: Run `cargo make ci-job-test-dart`
-      run: cargo make ci-job-test-dart
```
robertbastian, line 380: add tests to `ci-job-test-dart` so that CI is reproducible locally

robertbastian:
> what is the reason for using the detour through the ffi.toml and duckscript?

So that you can run CI locally.

.github/workflows/build-test.yml
mosuem: Yes, this works on linux only. I will remove that test.

robertbastian (approved):

PR #6171, opened 2025-02-20
Simplify Dart scripts

robertbastian:


Manishearth (approved):

ffi/dart/tool/build_libs.dart
```diff
@@ -39,96 +41,41 @@ Future<void> main(List<String> args) async {
...
-    if (folder.parent == folder) {
-      throw ArgumentError(
-        'Running in the wrong directory, as no Cargo.toml exists above $startingPoint',
-      );
-    }
-  }
-  return folder;
-}
-
-Future<Uri> buildLibraryFromInput(
```
mosuem, line 86: Why was this deleted? I do plan on using this once the package is published (which ideally would happen now). It should probably be moved to a `lib/` folder to be part of the public API instead.

robertbastian: I was careful to keep the public API and CLI the same :)

if you want this package to be published, then build and links hook should happen in this package, so it wouldn't need to be public.

I'm still not convinced this should be published before native assets stabilise, it's already broken on 3.8.0-beta again.

PR #6213, opened 2025-02-28
Fix for offset parsing leap seconds in `ixdtf`

nekevss:
This removes the support for leap second in UTC offsets along with a couple adjustments to the logic for UTC offsets.


Manishearth (approved):

Manishearth (approved):

Manishearth (approved):

robertbastian:
Wait, why? `icu::time::Time` supports values up to `23:59:60.999999999`, so why reject them during parsing?

nekevss:
I was approaching the inclusion of leap seconds in the offset as a bug / implementation oversight, specifically from the [Temporal support side](https://tc39.es/proposal-temporal/#prod-UTCOffset), which is where the sub minute precision offsets in `ixdtf` come from.

Is it preferred for `icu::time::Time` to included leap seconds in the offset? If it is preferred for `icu::time::Time` to support leap second offsets, we can be keep the leap seconds and I can handle it downstream.

robertbastian:
More context in https://github.com/unicode-org/icu4x/pull/6002

nekevss:
I'm confused. Are these in conflict with one another? 

This PR removes parsing leap seconds on sub-minute precision offsets and not on time component. Leap seconds on the time component are still preserved.

Valid strings:
  - 23:59:60.999999999
  - 23:59:60+23:59:59.999999999
  
Invalid string
  - 23:59:00+23:59:60
  - 23:59:00+23:59:60.99999999
  
Hmmmm, admittedly, the error is a little bit vague now that I look at it. Maybe that caused the confusion? I can adjust it specifically to an `InvalidLeapSecond` error.

Manishearth (approved):

nekevss:
I will admit. It's a bit hard because I don't have a direct copy of ISO-8601 myself. And this PR is in the space of ISO8601/Temporal compatibility. 

But this is going off the grammar for [`UtcOffset`](https://tc39.es/proposal-temporal/#prod-UTCOffset) in Temporal that lists a [`MinuteSecond`](https://tc39.es/ecma262/#prod-MinuteSecond) as 00..=59 and a [`TimeSecond`](https://tc39.es/proposal-temporal/#prod-TimeSecond) as 00..=60. Linked for reference to the nodes in the Temporal specification.

robertbastian:
Ooh this is for offsets only, sorry my bad. We only support seconds in offsets.

robertbastian (approved):

PR #6216, opened 2025-03-03
Test CLI

younies:


younies:
@robertbastian & @Manishearth 

test-dart is already failing

PR #6244, opened 2025-03-06
Internal refactor of DateTimeFormatter

sffc:
Organizes the inner workings without impacting stack size. I've wanted to do this for a while and #6243 was the impetus.


Manishearth (approved):
praise: thanks for splitting up the commits so nicely

Manishearth (approved):

PR #6259, opened 2025-03-09
Add transliteration test_katakana_hiragana

sffc:
CC @FrankYFTang


Manishearth (approved):

robertbastian:
I don't think this necessarily needs to be merged, it doesn't exercise any code paths that are untested otherwise, and it's potentially brittle under CLDR updates. It demonstrates to Frank how to load this transliterator, and has fulfilled its purpose already.

sffc:
Yeah I'll wait for feedback from @FrankYFTang on whether this works for their use case.

PR #6280, opened 2025-03-12
Improve datetime checksum validation

sffc:
Split from #5940
Part of #6063

Both of the following conditions are OK for the checksum matching:

1. All checksums are `Some(x)` where `x` is the same
2. All checksums are `None`: this means they all had already been previously loaded into their slot on DateTimeNames

The cases that are not OK are:

3. Two checksums are `Some(x)` and `Some(y)` where `x != y` (something is definitely wrong)
4. One checksum is `Some(x)`  and another is `None` (one payload was already loaded: might be OK but we aren't sure)

I believe I am changing the condition so that it correctly allows cases 1 and 2 but rejects cases 3 and 4.

### DIFF
--- a/components/datetime/src/pattern/names.rs
+++ b/components/datetime/src/pattern/names.rs
@@ -2624,7 +2624,7 @@ impl<FSet: DateTimeNamesMarker> RawDateTimeNames<FSet> {
             .map_err(|e| MaybePayloadError::into_load_error(e, error_field))?
             .map_err(|e| PatternLoadError::Data(e, error_field))?
             .checksum;
-        if cs1.is_none() || cs1 != cs2 || cs1 != cs3 {
+        if cs1 != cs2 || cs1 != cs3 {
             return Err(PatternLoadError::Data(
                 DataErrorKind::InconsistentData(tz::MzPeriodV1::INFO)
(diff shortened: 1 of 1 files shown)


sffc:
I added a test to demonstrate code that should work. The test fails on main with the following error:

```
---- components/datetime/src/pattern/names.rs - pattern::names::FixedCalendarDateTimeNames (line 554) stdout ----
Test executable failed (exit status: 101).

stderr:
ICU4X data error: The data for two markers is not consistent: MetazonePeriodV1 (were they generated in different datagen invocations?) (marker: MetazoneGenericNamesShortV1, request: uk)

thread 'main' panicked at components/datetime/src/pattern/names.rs:21:47:
called `Result::unwrap()` on an `Err` value: Data(DataError { kind: InconsistentData(MetazonePeriodV1), marker: Some(DataMarkerId { debug: "MetazoneGenericNamesShortV1", hash: [116, 100, 109, 104, 124, 150, 182, 205] }), str_context: None, silent: false }, ErrorField(Field { symbol: TimeZone(GenericNonLocation), length: One }))
```

Manishearth (approved):
Had discussion with Shane, he explained how this lifecycle works.

This seems correct to me. It feels like there may also be use cases for case 4 but we don't have to handle them now.

PR #6290, opened 2025-03-14
Add Vertical_Orientation support to icu_properties.

makotokato:


components/properties/src/props.rs
```diff
@@ -1600,6 +1600,64 @@ make_enumerated_property! {
     /// ```
 }
 
+/// Property Vertical_Orientation
+/// See UTR #50:
+/// <https://www.unicode.org/reports/tr50/#vo>
+///
+/// The numeric value is compatible with `UVerticalOrientation` in ICU4C.
```
robertbastian, line 1607: there is no public field, and `to_icu4c_value`/`from_icu4c_value` have their own docs
```suggestion
```

robertbastian (dismissed):

robertbastian (approved):

PR #6301, opened 2025-03-17
Support aliases in preferences

robertbastian:
#6220
#6227


Manishearth (approved):

zbraniecki:
@robertbastian - no concerns about the PR, thank you! 

But I'd like to ask to keep documentation value when extending macros - this argument is hard to interpret without any doc string and reading marcos is hard. I put a lot of effort to provide documentation for Preferences and I'd like us to make an effort to not regress places where we do have good API docs when we extend the functionality.

PR #6309, opened 2025-03-18
Change LocalTime associated type to a type called LocalTime

### LINKED ISSUE #6235: TimeZoneInfo should use DateTime
It still uses `(Date<Iso>, Time)`.

@robertbastian

sffc:
Fixes #6235

No conclusion was written down there. I think this concept is weird enough that we should make a type for it. Currently it is just a thin wrapper around `DateTime<Iso>` but this could change in the future (e.g. merge it with MinutesSinceLocalEpoch).


sffc:
Replaced by #6322

PR #6313, opened 2025-03-18
Fix C++ doc

robertbastian:


sffc (approved):
Rubber stamp two-line change

PR #6316, opened 2025-03-18
Rename "latest tested" to "tested" in `icu_provider_source`

robertbastian:
This is because it's the *only* tested data.

### DIFF
--- a/provider/export/README.md
+++ b/provider/export/README.md
@@ -16,7 +16,7 @@ use icu_provider_export::prelude::*;
 use icu_provider_source::SourceDataProvider;
 use std::fs::File;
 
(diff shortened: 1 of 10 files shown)


sffc (dismissed):

sffc:
Don't really like either `new_latest_tested` or `new_tested`. @robertbastian suggested `new` which maybe is fine.

sffc (approved):
`SourceDataProvider::new()` seems too "easy", since you are making a specific decision on your data provenance, and since it will cause the provider to download data from the network, but I can't think of a better name, and I can't think of any other configuration that would be better for the `::new()` function.

PR #6319, opened 2025-03-18
Hide week-of-month, non-iso week-of-year

robertbastian:
#6157


components/calendar/src/types.rs
```diff
@@ -353,37 +353,25 @@ impl MonthInfo {
...
-/// A week number in a year. Usually 1-based.
-#[derive(Clone, Copy, Debug, PartialEq)]
-#[allow(clippy::exhaustive_structs)] // this is a newtype
-pub struct WeekOfYear(pub u8);
+pub struct WeekOfYear {
+    /// The 1-based ISO week number
+    pub week_number: u8,
+    /// The ISO year
+    pub iso_year: i32,
+}
```
sffc: Issue: I thought we decided to rename this IsoWeekOfYear. https://github.com/unicode-org/icu4x/pull/6268#discussion_r1989655919

Alternatively, keep it named WeekOfYear but then please change the `iso_year` field to be of type `YearInfo`.

robertbastian: you're commenting on an outdated diff

PR #6322, opened 2025-03-18
Change LocalTime associated type to MinutesSinceEpoch

### LINKED ISSUE #6235: TimeZoneInfo should use DateTime

sffc:
Fixes #6235

It keeps `at_datetime` taking a `DateTime<Iso>`.

This was mostly clean, except FFI wants to convert back and forth, so it needs a getter and setter to the inner type. Would be nice to not have to do this. I can add a getter that returns `DateTime<Iso>`, but it seems wasteful since it will just be converted back again.


components/datetime/src/scaffold/get_field.rs
```diff
@@ -4,9 +4,10 @@
 
 use icu_calendar::{
     types::{DayOfMonth, DayOfYearInfo, MonthInfo, Weekday, YearInfo},
-    AsCalendar, Calendar, Date, Iso,
+    AsCalendar, Calendar, Date,
 };
 use icu_time::{
+    provider::MinutesSinceEpoch,
```
sffc, line 10: Not sure if you care about this or not but note that I am using `MinutesSinceEpoch`, which is defined in `provider`, in a non-provider module.

robertbastian: Maybe move it into `scaffold`? I don't really want this to live in `zone`.

components/time/src/provider/mod.rs
```diff
@@ -168,12 +168,15 @@ pub enum TimeZoneVariant {
 
 /// Storage type for storing UTC offsets as eights of an hour.
 pub type EighthsOfHourOffset = i8;
-/// Storage type for storing `(Date<Iso>, Time)`.
+
+/// Storage type for a time zone local time.
```
robertbastian, line 172:
```suggestion
/// Storage type for a datetime.
```

components/time/src/zone/mod.rs
```diff
@@ -245,8 +247,20 @@ impl TimeZoneInfo<models::Base> {
...
 
-    /// Sets a local time on this time zone.
-    pub const fn at_time(self, local_time: (Date<Iso>, Time)) -> TimeZoneInfo<models::AtTime> {
+    /// Sets the reference time for which time zone names should be resolved.
+    pub fn at_datetime(self, datetime: DateTime<Iso>) -> TimeZoneInfo<models::AtTime> {
+        self.at_inner(datetime.into())
+    }
+
+    /// Sets the reference time for which time zone names should be resolved.
+    pub fn at_minutes_since_local_epoch(
```
robertbastian, line 256: the type is called `MinutesSinceEpoch`, so this should be `at_minutes_since_epoch`.

ffi/capi/src/timezone.rs
```diff
@@ -194,7 +191,7 @@ impl From<icu_time::TimeZoneInfo<icu_time::zone::models::AtTime>> for TimeZoneIn
             time_zone_id: other.time_zone_id(),
             offset: other.offset(),
             zone_variant: None,
-            local_time: Some(other.local_time()),
+            local_time: Some(other.local_reference_time()),
```
robertbastian, line 194: this field name should match the method name

sffc:
I am replacing this PR with #6513

PR #6323, opened 2025-03-18
Re-enable ZonedDTF in demogen

sffc:
Should have been unhidden after https://github.com/unicode-org/icu4x/commit/a92720cb5de20dc5fde0baf7145557300dc2de3d (#6286)


sffc:
This is obsolete. ZonedDateTime FFI is running:

https://unicode-org.github.io/icu4x/wasm-demo/template.html?func=ZonedDateTimeFormatter.formatIso

PR #6329, opened 2025-03-21
Stop formatting askama codegen'd files

sffc:
Short-term fix for #6327 and the beta CI failure


sffc:
Autosubmit = true if CI passes

tools/make/codegen/templates/zoned_date_formatter.rs.jinja
```diff
@@ -2,9 +2,9 @@
 // called LICENSE at the top level of the ICU4X source tree
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
-// This file is automatically @generated by: tools/make/codegen/src/capi_datetime.rs
-// It is not intended for manual editing.
+// {{ crate::capi_datetime::GENERATED_BY }}
 
+#[rustfmt::skip]
```
robertbastian, line 7: you should need either `// @generated` or `#[rustfmt::skip]`. If `// @generated by` doesn't work, make it
```rust
// @generated
// by 
```

PR #6334, opened 2025-03-22
Add ZonedTimeFormatter, TimeZoneFormatter FFI

sffc:
#5940


ffi/capi/src/zoned_time_formatter.rs
```diff
@@ -0,0 +1,500 @@
...
+
+    impl ZonedTimeFormatter {
+        #[diplomat::attr(all(supports = fallible_constructors, supports = named_constructors), named_constructor = "specific_long")]
+        #[cfg(feature = "compiled_data")]
+        pub fn create_specific_long(
+            locale: &Locale,
+            length: Option<DateTimeLength>,
+            time_precision: Option<TimePrecision>,
+            alignment: Option<DateTimeAlignment>,
+        ) -> Result<Box<Self>, DateTimeFormatterLoadError> {
```
sffc, line 51: Thought: It's nice that inlining the options here avoids the extra box/clone, but it is then misleading that `length` is a prominent argument but applies only to the day period, a setting that rarely has any effect.

ffi/capi/src/zoned_time_formatter.rs
```diff
@@ -0,0 +1,500 @@
...
+
+    #[cfg(feature = "buffer_provider")]
+    use crate::provider::ffi::DataProvider;
+
+    #[diplomat::opaque]
+    #[diplomat::rust_link(icu::datetime::FixedCalendarDateTimeFormatter, Typedef)]
+    pub struct ZonedTimeFormatter(
+        pub  icu_datetime::FixedCalendarDateTimeFormatter<
+            (),
+            icu_datetime::fieldsets::enums::CompositeFieldSet,
```
sffc, line 39: Thought: This could be `Combo<T, ZoneFieldSet>` to reduce memory size a bit

Manishearth (approved):
easy to understand codegen changes

PR #6350, opened 2025-03-25
DTF calendar handling

robertbastian:


components/datetime/src/neo.rs
```diff
@@ -555,22 +555,66 @@ where
...
+                | CalendarAlgorithm::Persian
+                | CalendarAlgorithm::Roc,
+            ) => {}
+            // Calendar not supported by DateTimeFormatter
+            // Currently this is CalendarAlgorithm::Iso8601
+            // Let AnyCalendarKind constructor select an appropriate fallback
+            _ => prefs.calendar_algorithm = None,
+        }
+        let kind = AnyCalendarKind::new((&prefs).into());
+        // AnyCalendarKind constructor should not select an unsupported fallback
```
robertbastian, line 587: this is unfortunate

components/calendar/src/any_calendar.rs
```diff
@@ -560,14 +560,11 @@ impl Calendar for AnyCalendar {
...
     ///
-    /// As this requires a valid [`AnyCalendarKind`] to work, it does not do any kind of locale-based
-    /// fallbacking. If this is desired, use [`Self::try_new()`].
-    ///
     /// ✨ *Enabled with the `compiled_data` Cargo feature.*
     ///
     /// [📚 Help choosing a constructor](icu_provider::constructors)
     #[cfg(feature = "compiled_data")]
-    pub const fn new_for_kind(kind: AnyCalendarKind) -> Self {
+    pub const fn new(kind: AnyCalendarKind) -> Self {
```
sffc, line 567: Nit: Keep the method named `new_for_kind` and we just don't have a `new`. I still stand by my previous position that `new` and `try_new` should take Preferences by convention.

components/calendar/src/any_calendar.rs
```diff
@@ -560,14 +560,11 @@ impl Calendar for AnyCalendar {
 impl AnyCalendar {
     /// Constructs an AnyCalendar for a given calendar kind from compiled data.
     ///
-    /// As this requires a valid [`AnyCalendarKind`] to work, it does not do any kind of locale-based
-    /// fallbacking. If this is desired, use [`Self::try_new()`].
```
sffc, line 564: Nit: Restore these docs

components/datetime/src/neo.rs
sffc: Yeah, a way to fix this would be for `icu_calendar` to export an enum `LocaleDefaultCalendarKind` which we can exhaustively match over here. But this is fine. I do prefer for this to be a `debug_assert` instead of an `assert`.

components/datetime/src/scaffold/calendar.rs
```diff
@@ -335,10 +337,7 @@ where
             JapaneseExtended => H::JapaneseExtended::bind(p).load_bound(req),
             Persian => H::Persian::bind(p).load_bound(req),
             Roc => H::Roc::bind(p).load_bound(req),
-            _ => Err(
-                DataError::custom("Don't know how to load data for specified calendar")
-                    .with_debug_context(&self.kind),
-            ),
+            _ => unreachable!(),
```
sffc, line 340: Issue: I won't approve this with a panic. It is not a panic that the compiler can easily eliminate.

components/datetime/src/scaffold/calendar.rs
```diff
@@ -277,12 +277,14 @@ where
...
     provider: P,
+    // This is one of Buddhist, Chinese, Coptic, Dangi, Ethiopian, EthiopianAmeteAlem, Gregorian, Hebrew, Indian,
+    // HijriCivil, HijriObservationalMecca, HijriTabular, HijriUmmAlQura, Japanese, JapaneseExtended, Persian, Roc
     kind: AnyCalendarKind,
     _helper: PhantomData<H>,
 }
 
 impl<H, P> AnyCalendarProvider<H, P> {
-    pub(crate) fn new(provider: P, kind: AnyCalendarKind) -> Self {
+    pub(crate) fn new_unchecked(provider: P, kind: AnyCalendarKind) -> Self {
```
sffc, line 287: Suggestion: you could make this `try_new -> Option<Self>` and do the match in here. Less coding at a distance. I don't really like having a match in the middle of `neo.rs` which is supposed to be a thin API wrapper file.

sffc (changes requested):
Best so far

components/calendar/src/any_calendar.rs
Manishearth: I'm okay with that, though I think just `new` is fine here.

sffc:
I made a new PR incorporating this one: #6352

components/calendar/src/any_calendar.rs
sffc: eh, it is nice that `try_new_from_buffer_provider` doesn't need to be `try_new_for_kind_with_buffer_provider`

Let's keep it as-is

components/calendar/src/any_calendar.rs
sffc: I guess you don't need to

components/datetime/src/scaffold/calendar.rs
sffc: I did this in a different way in #6352

components/datetime/src/scaffold/calendar.rs
sffc: I fixed this in #6352

components/datetime/src/pattern/names.rs
```diff
@@ -591,6 +591,8 @@ pub struct FixedCalendarDateTimeNames<C, FSet: DateTimeNamesMarker = CompositeDa
 #[derive(Debug, Clone)]
 pub struct DateTimeNames<FSet: DateTimeNamesMarker> {
     inner: FixedCalendarDateTimeNames<(), FSet>,
+    // calendar.kind() is one of Buddhist, Chinese, Coptic, Dangi, Ethiopian, EthiopianAmeteAlem, Gregorian, Hebrew, Indian,
```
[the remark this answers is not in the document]
sffc: I fixed this in #6352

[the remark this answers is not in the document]
robertbastian: I specifically don't want `icu_calendar` to define any sets for `icu_datetime`

PR #6364, opened 2025-03-27
Add DateTimeFieldSetBuilder API and corresponding formatter constructors

sffc:
Needs discussion.

This PR adds constructors from a DateTimeFieldSetBuilder on each of the formatter types, allowing them to be constructed from a dynamic field set. The constructors are strict: they will reject builders with extra or missing or otherwise invalid options.

I recall last time I tried to make this FFI, there was contention surrounding the strictness of the builder, with a desire for the builder to build a formatter capable of supporting _any valid field set_.

We don't really have such a formatter, though. We have `ZonedDateTimeFormatter`, which currently supports only zoned datetimes, and semantically it requires a field set that is compatible with zoned datetimes.

Some things we could do:

1. Add a new type `CompositeDateTimeFormatter` for this purpose. It supports all valid field sets. It could even have format functions that take different subsets of date, time, and zone, failing dynamically if not all of the required input fields were specified.
2. Change `ZonedDateTimeFormatter` to support this. It will have a constructor from the builder and will accept field sets besides just ZonedDateTime field sets.
3. Ignore this problem and say that you need to select the correct formatter for your field set builder.


Manishearth (approved):

Manishearth (approved):

sffc:
- @sffc What I realized is that we don't have such a type to build a date time from any valid field set. We have a ZonedDateTimeFormatter which formats a zoned date time, but we don't have a specific date time formatter. There are 3 options listed in the issue description. One tradeoff is that the genericized from-field-set builder doesn't lend itself to dead code optimization.
- @nekevss You'll probably want the Composite version for special cases. I don't see an JS Intl equivalent.
- @sffc That's a good point. JS Intl has a all-in-one builder.
- @nekevss It depends how you want to approach it. If you want to support that functionality by creating a wrapper function, then that works, too. It all depends on how you approach it.
- @Manishearth My thoughts are... is there a good reason for all of our types not to do this? DateTimeFormatter should format dates and times. Having separate methods might be good, maybe? I'm kind-of wondering, with field sets, does it make sense in all of those cases... is a CompositeDateTimeFormatter really that much different than ZonedDateTimeFormatter? To express a preference, it seems like we shouldn't add a whole new type. Or ask people to use an enum.
- @sffc First of all, the convention of DateTimeFormatter that it formats different types is because of Rust conventions because it's in a crate called `datetime`. The reason we have 7 different constructors is because they all take different arguments. This design that we have landed gets us most of the benefits of the Rust type system that is flatter. The question is really only about, should the Composite formatter exist, or is it a part of the ZonedDateTimeFormatter? If we do the latter, it's probably fine, but I worry about the discoverability. It's not clear to non-i18n people why ZonedDateTimeFormatter is more special, and it's only special because it requires the superset of fields of all the others. So having a separate type called CompositeDateTimeFormatter allows us to be less surprising to users, and to tailor the docs. The other concern is that tacking it on to ZonedDateTimeFormatter would make it serve more than 1 purpose, and because that code is generated from templates, we would have to modify the templates.
- @hsivonen When you introduced this topic, the Composite made sense based on how you introduced it. It makes sense based on what I know.
- @sffc Do we think we need this in 2.0, or is something we can add after 2.0 after doing more research?
- @robertbastian Yeah, that is what I was thinking.
- @sffc Okay, maybe that's a good conclusion. I don't want to spend more time talking about this instead of thinking about 2.0 work.

Conclusion:

- Using `CompositeDateTimeFormatter` or `ZonedDateTimeFormatter` are both valid options
- Can be added post-2.0 since all datetime formatting behavior is _accessible_ via FFI, just only via the compile-time options

LGTM: @sffc @Manishearth @hsivonen @robertbastian

sffc:
Based on the discussion, I don't know if we want to move forward with the design in this PR. I split the good parts of this PR into #6373 and I'll rebase this one on top with only the narrow FFI changes.

sffc:
Follow-up in https://github.com/unicode-org/icu4x/issues/6445

Code archived in https://github.com/sffc/omnicu/tree/archive/DateTimeFieldSetBuilder

PR #6368, opened 2025-03-27
Unify day-of-week calculation

robertbastian:
Weekdays are not calender-specific, and the current code is overly complex.


sffc (dismissed):
I approve but would like @Manishearth to also approve

Manishearth (approved):

PR #6382, opened 2025-03-28
Remove prev_year info

robertbastian:
This became dead code in #6319


sffc (commented):
Good to remove dead code. I'll let @Manishearth approve as more familiar with the packed data stuff.

Manishearth (approved):
Yeah we only needed this for week calculations, I always disliked the need to carry this around

PR #6405, opened 2025-04-02
Update era codes

robertbastian:
https://github.com/unicode-org/cldr/pull/4519


Manishearth (approved):
LG with the understanding that today's meeting may change some of these. But happy to land this and make smaller incremental changes as needed.

We'll probably end up renaming saka to shaka.

Manishearth (approved):

sffc (commented):
Some of these era codes are already different than what is in this PR, and I'm not currently of the opinion that we should be on the vanguard here before CLDR adopts them. Nothing has been approved by anyone yet. I think keeping the test is more valuable at this instant in time. This might be different in 5-10 days from now but I don't know at this moment in time

sffc:
> LG with the understanding that today's meeting may change some of these. But happy to land this and make smaller incremental changes as needed.
> 
> We'll probably end up renaming saka to shaka.

I'm not convinced that future changes will be smaller and incremental. We're already likely backtracking on some of my proposed changes to Minguo and Coptic.

robertbastian:
While what's currently on main aligns with what's currently in CLDR (https://github.com/unicode-org/icu4x/pull/6248, https://github.com/unicode-org/icu4x/pull/6252), it is sufficiently in flux that I don't want to release the current code. If we release 2.0 with calendars that return eras like `buddhist` and `gregory-inverse`, `islamic-tbla`, then that behaviour will be out there, and it will be a bigger semver hazard to do the CLDR alignment in the future.

What's currently proposed in CLDR, and what this PR implements, is a much smaller diff to the 1.5 behaviour.

provider/source/src/calendar/eras.rs
```diff
@@ -504,25 +504,26 @@ fn test_calendar_eras() {
             let in_era = in_era_iso.to_calendar(cal);
             let not_in_era = not_in_era_iso.to_calendar(cal);
 
+            // TODO: reenable with CLDR-48
```
sffc, line 507: issue?

sffc (approved):

provider/source/src/calendar/eras.rs
robertbastian: filed https://github.com/unicode-org/icu4x/issues/6437

Issue #4048, opened 2023-09-18
Tracking caller on `DataError` construction


robertbastian:
It would be very useful to store a `core::panic::Location` inside the `DataError`. Currently panicking is much more ergonomic than using `DataError`, because with `DataError` you only get the location where the error is unwrapped and have to chase down the error message, which often is not a pure literal.

This is currently blocked on https://github.com/rust-lang/rust/issues/76156, as our error constructors are `const` and `core::panic::Location::caller()` is not yet.

sffc:
Losing the caller location when bubbling up errors is I think a known issue with the Rust error propagation model, not specific to DataError. CC @yaahc, thoughts on the recommended pattern here for library code?

sffc:
See https://github.com/unicode-org/icu4x/pull/6414 for a possible approach.

yaahc:
Pardon the late reply, but the issue is around how you construct your errors. If you construct errors in any of the std functions that take a closure and they don't have `track_caller` applied, then you will end up getting a location in std instead of the location in your own code.

The trick right now is to do everything in From impls, because we already allow track_caller to propagate through FromResidual for Result

https://doc.rust-lang.org/stable/src/core/result.rs.html#2017

Here's an example of what I mean

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=8205bcd02b380d9fd02b1f1153ac9c4e

Beyond that, though, I don't think there's any reason why we wouldn't want to apply `#[track_caller]` to all the common places where std ends up eating the user's error locations like `map_err`, since we already do it in `FromResidual` when people use `?`. I can go ahead and open a PR for that specific API using this issue as the motivation, but if y'all end up starting to track error locations and run into any other footguns where std is frequently eating your error locations in an unintuitive way lmk via this issue so we can try to tackle those as well (depending of course on how the `map_err` change is received).

Manishearth:
We should check: is this still something we want to do? We have a PR open for this (#7637)  but I've also not really experienced problems around this in a while.


I have a mild worry about this bloating baseline binary sizes due to all the `Location`s. We should measure.

robertbastian:
The last time I'd have liked to have this was today.

We should put it behind the `logging` feature.

Manishearth:
Fair enough!

sffc:
Doesn't js ffi enable the logging feature?

I guess the marginal binary size of adding this versus adding the other debug impls that get linked with the logging feature is sufficiently small.

[...]

PR #6414, opened 2025-04-03
Fix DataError struct location initialization

mdfarid01:
This PR addresses issue [#4048] ) by adding `core::panic::Location::caller()` to `DataError.`

- Previously, DataError lacked precise error locations, making debugging difficult.

- Now, DataError stores the caller's location, improving error traceability.

Changes Made:
✅ Added location: `Some(core::panic::Location::caller())` to `DataError`.
✅ Updated relevant error constructors.


sffc (commented):
Hmm, do you think we should log out the caller instead of storing it in the data model of DataError, more like the Debug or Display context frontends?

We should also be sure that we are logging the correct caller. For example, `with_path_context` calls `with_str_context`, but it is only `with_str_context` that uses `Location::caller()`, which is not the intent: it should return the caller of `with_path_context`.

@robertbastian

mdfarid01:
1. I see your point about logging the caller instead of storing it in DataError. That makes sense to keep the data model cleaner 
2.  also understand the concern about capturing the correct caller. Right now, `Location::caller()` is inside `with_str_context`, which might not be returning the actual caller of `with_path_context `

robertbastian:
You need to add `#[track_caller]` annotations to all methods in the call stack that you want to skip. These should be all the constructors for `DataError`. More info on `#[track_caller]`: https://rustc-dev-guide.rust-lang.org/backend/implicit-caller-location.html

> Hmm, do you think we should log out the caller instead of storing it in the data model of DataError, more like the Debug or Display context frontends?

It depends how important the size of the `DataError` is to us. The ergonomics of log-and-forget are *much worse*, as especially in multi-threaded programs (i.e. datagen) the logging can be hundreds of output lines away from the unwrap panic. It's also much less useful for tests, as they don't usually have a logger set up. I'd prefer if the `DataError` just stored the location, so that unwrapping the `Result` would show the error location.

robertbastian (commented):
.

PR #6443, opened 2025-04-11
Simplify EraYear

robertbastian:
This removes the duplication between `standard_era` and `formatting_era`.

Effectively, whenever `formatting_era` is `FormattingEra::Code`, it is the same code as `standard_era`, so it could just be `Option<(u8, TinyStr16)>`. But the fallback era name isn't really needed, as it's always the same as code up to capitalization, and we don't really care about capitalization in gigo paths.

The new `era_index` field can be documented without deferring to whatever datagen does.

### DIFF
--- a/components/calendar/src/types.rs
+++ b/components/calendar/src/types.rs
@@ -95,8 +95,8 @@ pub struct EraYear {
     ///
     /// In this context, chronological ordering of eras is obtained by ordering by their start date (or in the case of
     /// negative eras, their end date) first, and for eras sharing a date, put the negative one first. For example,
-    /// bce < ce, and mundi < incar for Ethiopian.
-    pub era_ordinal: Option<u8>,
+    /// bce < ce.
+    pub era_index: Option<u8>,
     /// The ambiguity of the era/year combination
     pub ambiguity: YearAmbiguity,
 }
--- a/components/calendar/src/types.rs
+++ b/components/calendar/src/types.rs
@@ -13,24 +13,6 @@ use tinystr::{TinyStr16, TinyStr4};
 use zerovec::maps::ZeroMapKV;
 use zerovec::ule::AsULE;
 
-/// The era of a particular date
-///
-/// Different calendars use different era codes, see their documentation
-/// for details.
-///
-/// Era codes are shared with Temporal, [see Temporal proposal][era-proposal].
-///
-/// [era-proposal]: https://tc39.es/proposal-intl-era-monthcode/
-#[derive(Copy, Clone, Debug, PartialEq)]
-#[allow(clippy::exhaustive_structs)] // this is a newtype
-pub struct Era(pub TinyStr16);
-
-impl From<TinyStr16> for Era {
-    fn from(o: TinyStr16) -> Self {
-        Self(o)
-    }
-}
-
 /// The type of year: Calendars like Chinese don't have an era and instead format with cyclic years.
 #[derive(Copy, Clone, Debug, PartialEq)]
 #[non_exhaustive]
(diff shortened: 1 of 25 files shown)


sffc (dismissed):
I was never really fond of the standard_era/formatting_era split... people just really didn't like the standard era codes we were using, so much so that they wanted to stuff them away in the standard_era field. This goes back to something more like we had in earlier 1.x versions which is simpler.

components/calendar/src/types.rs
```diff
@@ -101,56 +83,21 @@ pub enum YearAmbiguity {
...
     pub year: i32,
-    /// The ambiguity when formatting this year
+    /// The era code as defined by CLDR, expect for cases where CLDR does not define a code.
+    pub era: TinyStr16,
+    /// An Era Index, for calendars with a small, fixed set of eras. The eras are indexed chronologically.
+    ///
+    /// In this context, chronological ordering of eras is obtained by ordering by their start date (or in the case of
+    /// negative eras, their end date) first, and for eras sharing a date, put the negative one first. For example,
+    /// bce < ce.
+    pub era_index: Option<u8>,
```
sffc, line 99: Nit: Make it super clear that the era index is NOT stable.

Consider naming it something more scary like `formatting_era_index`?

For example, we may be changing it next release for the Hijri and Coptic eras.

components/calendar/src/types.rs
```diff
@@ -101,56 +83,21 @@ pub enum YearAmbiguity {
...
-    ///
-    /// It will typically be a valid era alias.
-    ///
-    /// <https://tc39.es/proposal-intl-era-monthcode/#table-eras>
-    pub formatting_era: FormattingEra,
-    /// The era code as expected by Temporal/CLDR. This era code is unique for the calendar
-    /// and follows a particular scheme.
-    ///
-    /// <https://tc39.es/proposal-intl-era-monthcode/#table-eras>
-    pub standard_era: Era,
```
sffc, line 150: Issue: I just remembered that this struct is parallel with the MonthInfo struct in terms of having both a "standard" code and a "formatting" code. Should we think about this holistically?

components/calendar/src/types.rs
robertbastian: But it's not formatting-only, it's the index of the era after you sort all eras.

sffc: It seems risky that there would be drift here. I fully expect that Islamic and Coptic formatting will break after we change the index of an existing era.

Why do we have all this language about chronological ordering? I feel like we should make index 0 be the extended year (primary) era and index 1 be the alternate (secondary) era. There is no calendar with more than 2 eras except Japanese.

sffc: Idea: change the `Option<u8>` to `Option<FormattingEraKind>` with

```rust
pub enum FormattingEraKind {
    Primary,
    Secondary,
}
```

components/calendar/src/types.rs
robertbastian: looks like month can be solved with `code: MonthCode, is_leap: bool`?

components/calendar/src/types.rs
robertbastian: primary era is not a concept that exists yeat

components/calendar/src/types.rs
sffc: No, the reason they are different is because the Hebrew month Adar has display name "Adar" in normal years and "Adar II" in leap years, but it is the same month code. We weasel around this by making Adar II be a different _formatting_ month code.

components/calendar/src/types.rs
sffc: Era index is also an ICU4X concept. The concept coincidentally happens to mostly correspond to the separate but strikingly similar concept in UTS 35 called era type, but the two are not derived from each other.

sffc: The reason I like something more similar to `FormattingEraKind` is that every calendar always has a primary era, and a secondary era is optional. In this way, we can look up the primary era name at index 0 and the secondary era name at index 1.

robertbastian: But it's not. Sorting all the eras chronologically to interpret the index can be done by with just CLDR data. Figuring out what the "primary era" is requires ICU4X specific data.

robertbastian: What you're proposing is a data model change that can be discussed, but not in this PR. This PR is just moving around fields that already exist while not changing behaviour.

I would prefer this field to be

```rust
/// For calendars with just a primary/secondary era, whether this era is the primary one
is_primary: Option<bool>
```

I wouldn't want to do this before "primary" is a concept in CLDR.

sffc: (My CLDR PR adds primary era annotations.)

The indexes are also basically hard-coded with ICU4X specific data. Actually it's quite dangerous and broken that icu_calendar hardcodes these indexes when they actually should be data-driven.

components/calendar/src/types.rs
robertbastian: we have `Date::is_in_leap_year`

components/calendar/src/types.rs
robertbastian: That's what I added the eras test for, it also checks that the indices match.

components/calendar/src/types.rs
sffc: I'm not sure that works. The data layout is a VZV with 24 entries

```
<M00>  <M01>  <M02>  <M03>  <M04>  <M05>  <M06>  <M07>  <M08>  <M09>  <M10>  <M11>  <M12>
<M00L> <M01L> <M02L> <M03L> <M04L> <M05L> <M06L> <M07L> <M08L> <M09L> <M10L> <M11L> <M12L>
```

We have "Adar" stored in M06 and "Adar II" stored in M06L. However, we have "Shebat" in M05 and "Adar I" in M05L, which _both_ exist in a leap year.

components/calendar/src/types.rs
sffc: yeah it's an orthogonal concern. #6444

sffc (approved):
IMO, this is an improvement on the status quo with the possible exception of making `era_index` look like something more friendly to use. I suggest a remedy of renaming it to `formatting_era_index`. But renaming is easy to do in a follow-up.

I'm approving this as an LGTM, but @Manishearth is the codeowner. Please proceed accordingly.

components/calendar/src/types.rs
robertbastian: The code in `Hebrew::month` works with just the month ordinal and `year_is_leap`, so we can probably just move that into datetime. Might want to add `year_is_leap` to `MonthInfo` so that we don't need a newtype to pass it around in datetime.

sffc: The code in `Hebrew::month` is very Hebrew specific and doesn't generalize. I think we've re-written that code 3 or 4 times because we kept finding edge cases.

sffc:
@Manishearth If this is landable, I'd like to land it so I avoid conflicts when working on related issues

Manishearth (approved):
Yeah, this split is in part because the standard kept changing, and we also switched how we optimized this a couple times.

PR #6447, opened 2025-04-11
Remove unused data markers from datetime crate

sffc:
Toward #5613
Unblocks #5837

This moves more of the legacy stuff into the icu_provider_source crate.


Manishearth (approved):
praise: all the buildup has led to clean code deletion

PR #6479, opened 2025-04-21
Add PluralElements [try_]for_each, [try_]for_each_mut

sffc:
Depends on #6478

I was able to write #6477 without using these functions, but they improve ergonomics a bit.

Three things we could do:

1. Close this Pull Request
2. Land this Pull Request as-is
3. Rename the functions to `.inspect` and `.update` to align with the standard library and itertools

### DIFF
--- a/components/plurals/src/lib.rs
+++ b/components/plurals/src/lib.rs
@@ -89,6 +89,7 @@ mod raw_operands;
 pub use raw_operands::RawPluralOperands;
 
 use core::cmp::{Ord, PartialOrd};
+use core::convert::Infallible;
 use icu_locale_core::preferences::define_preferences;
 use icu_provider::marker::ErasedMarker;
 use icu_provider::prelude::*;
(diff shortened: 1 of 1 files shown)


Manishearth (approved):
I think for_each is fine, inspect and update are specifically for iterators. People otherwise use for_each a fair amount, and it's better if these APIs have similar names IMO.

Manishearth (approved):

sffc:
I'd like a second opinion from @zbraniecki or @hsivonen regarding function naming

sffc:
I really want to use this so I am going to merge it

sffc:
- @Manishearth (reiterates comments from above)
- @Manishearth We could add iterator functions if you want to avoid these names.
- @sffc I don't see a use case for iterators. There's nothing useful collection target for them.
- @nekevss I think the names you proposed make sense.

Conclusion: Keep the names in this PR.

PR #6493, opened 2025-04-25
Improve checksum validation in DateTimeNames

sffc:
#6063


Manishearth (approved):
Wonderful, thanks! The issue with from_formatter is disappointing, but this is a niche use case of a power user api, so I'm happy with the filed issue.

PR #6494, opened 2025-04-25
Split collator benchmarking loops to never-inlined functions to avoid collator changes affecting the baseline

hsivonen:


hsivonen:
Unfortunately, the baseline case fluctuates a lot even with this change. The baseline workload seems to be too small.

Manishearth (approved):
I generally find that this type of tuning doesn't really tend to bear much fruit, but it's still an improvement to the code organization

PR #6495, opened 2025-04-25
Add benchmark simulating sorting a directory of filenames containing photos

hsivonen:
This is a realistic case with identical prefixes.

Intended to apply on top of #6494.


Manishearth (approved):
This seems fine to merge separately.

PR #6499, opened 2025-04-25
Remove dead type DateTimeSkeletons and its data marker

sffc:
Toward #5634


Manishearth (approved):
Surprised this doesn't lint!

sffc:
They were public types

Issue #6489, opened 2025-04-23
ZonedDateTime parse functions are a pain to use


sffc:
`try_from_str` calculates the variant, but it is strict about what fields should be there.

`try_loose_from_str` is more lenient, but it doesn't calculate the variant.

My input string is `"2000-01-01T00:00Z[Etc/GMT]"` and `try_from_str` doesn't like it.

@robertbastian

sffc:
Maybe this is being caused by the `Z`.

sffc:
CC @nekevss

sffc:
The change i want to implement is a 5-line change to make the string in the OP work correctly.

nekevss:
I've been looking at this. I think I'm still confused about the wording on the base issue above.

From what I understand, "2000-01-01T00:00Z[Etc/GMT]" should work with `try_from_str`, but then the same string should fail with the `try_lenient_from_str`, at least according to current behavior.

My general thoughts are:

If it is not working with `try_from_str`, then yes that should definitely be fixed, because that feels like a bug.

If it's not working with `try_from_lenient_str` due to the `RequiresCalculation` error, then I think it would be reasonable to have a check to exempt `Etc/GMT` and `Etc/UTC` from requiring calculation. My reasoning is mostly due to the special treatment of those identifiers in ECMA402's [`AvailableNamedTimeZoneIdentifiers()`](https://tc39.es/ecma402/#sup-availablenamedtimezoneidentifiers) in step 5.c., which seems to align with the thought process already at play in [`Intermediate::lenient`](https://github.com/unicode-org/icu4x/blob/main/components/time/src/ixdtf.rs#L314).

But again, I might be misunderstanding the issue above.

sffc:
I realized that I think the problem is that we unconditionally treat `Z` as "requires calculation", even if the time zone is `Etc/GMT` or `Etc/UTC`. My proposal is to allow `Z` if the time zone is one of those fixed zones.

nekevss:
Agreed. The method should be able to handle `Etc/UTC`, `Etc/GMT`, `GMT`, or `UTC` without returning the requires calculation error.

Manishearth:
@robertbastian are you interested in fixing this?

robertbastian:
I think this is obsolete, variants are deprecated and not needed by the formatter anymore

sffc:
The case stated in the OP is still failing on main.

[...]

Issue #6354, opened 2025-03-26
Feedback on `icu_time` API


robertbastian:
https://docs.google.com/document/d/1gKZCaiTG2eeyxrIKUF2tf6PIsaPCIBaUVRXD-JDh6dI/edit?tab=t.0#heading=h.zfxbicahzeqi

sffc:
API Breaking:

- [x] Return the TimeZone BCP 47 as a Subtag
- [ ] Revisit naming convention of `is_default` on Locale types (split to #6505)
    - Note: The und language subtag is short for "undetermined" in the registry
    - Note: Maybe have `is_unknown()` on all of these: Locale, TimeZone, etc. It is a useful predicate to have. This is one of the more successful things we've had in III.
- [x] IanaParserExtendedBorrowed::parse should return a struct with named fields. Markus: That would definitely be nicer.
- [x] Revisit naming of subsecond field to make sure it is clear that it is in nanoseconds. At least explain clearly in docs.
    - Markus says: Subsecond doesn't say where it is millisecond or nanosecond. Call sites that call .subsecond() are unclear. (again later in ZonedDateTime::try_from_str)
- [x] Don't have the midnight() constructor. Call it zero() or something. I expect midnight() to be the end of the day. Rob: or start_of_day().
- [x] ZonedDateTime: Markus says: How about "lenient" instead of "loose". Or "generic" or something. "loose" is not a good opposite of "strict".
- [x] ZonedDateTime: try_from_str should be more explicit what it needs. Rob: try_full_from_str?

High Impact Non-Breaking:

- [ ] Better docs to explain how to make an offset-only time zone (use unknown and add an offset)
- [ ] Better docs to explain what TimeZoneInfo is and what it does and why it's different from a TimeZone
    - Markus said: I see why you decouple time zone ID from offset, to reduce data dependencies, but you should explain that more. It is counterintuitive compared to how most i18n libraries work.
- [ ] You should talk about where people should get the offset and variant. Links to the chrono and jiff examples should be fine.
- [ ] TimeZoneInfo: Improve docs regarding the multiple different reckonings
    - Markus says: Why are the functions in different sections? Are the ones on top (time_zone_id, offset) the main functions?
[...]

PR #6506, opened 2025-04-29
Rename ZonedDateTime::try_from_str to try_full_from_str

sffc:
Part of #6354

Related to #6489

The function's signature and behavior are unintuitive and confusing. If we want to add back `try_from_str`, we can do it later. Let's not waste the name on the current function.


sffc:
@gemini-code-assist Summarize the changes since you posted your first review.

Manishearth (approved):
praise: I like this idea for sidestepping the issue!

PR #6518, opened 2025-05-02
Add tests for era index

sffc:
Shows us where we're at regarding #6444, so that if someone wants to fix it, they know what to fix.

### DIFF
--- a/components/calendar/src/any_calendar.rs
+++ b/components/calendar/src/any_calendar.rs
@@ -1571,19 +1571,20 @@ mod tests {
     #[track_caller]
     fn single_test_roundtrip(
         calendar: Ref<AnyCalendar>,
-        era: Option<&str>,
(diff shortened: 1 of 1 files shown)


sffc:
> * **Clarity of test parameters**: The `single_test_roundtrip` and `single_test_error` functions now accept a tuple for the `era` parameter, which includes both the era code and an optional era index. While this change is functional, it might be beneficial to clarify the purpose of the optional era index in the function documentation or comments.

This function already had no documentation. We should add documentation for it, but I don't want to hold up this PR on that, since the PR is still a net improvement.

> * **Consistency in test cases**: The test cases have been updated to include era indices. Ensure that all relevant test cases are updated consistently to reflect the new parameter structure.

The build should fail if anything hasn't been updated.

Manishearth (approved):

PR #6520, opened 2025-05-04
Improve docs on DateTimeWriteError

### LINKED ISSUE #6269: Reduce variants from DateTimeWriteError
As of #6233, we use the same error enum in 3 places:

1. FormattedDateTimePattern
2. FormattedDateTimeUnchecked
3. FFI ZonedDateTimeFormatter

However, the error cases are not the same across all 3.

sffc:
Fixes #6269

The error conditions are not exactly the same, especially if we suppress errors involving bad data, but they're _close enough_, and this is a power-user API that I don't really want to add different enums for the different cases.


sffc:
I will make a new PR and incorporate some of the docs from this one.

PR #6529, opened 2025-05-06
Remove `Language`/`LanguageIdentifier`/`Locale` `Default` impls

robertbastian:
#6519


sffc (approved):
I think this is a very positive change. Turns out we weren't using this in very many places. `Locale::UNKNOWN` is more clear; `Locale::default()` is very confusing (looks like the system locale). If we find that clients really need the `Default` impl, it's trivial to add back. Now is the time to remove it.

PR #6532, opened 2025-05-06
Change VariantOffsetsCalculator to take TimeZoneInfo

sffc:
See #6238

VariantOffsetsCalculator might need to know the offset in order to correctly determine the variant offsets. Rather than rely solely on ZoneNameTimestamp for this, I realized that it should take a TimeZoneInfo, which can contain an offset.

### DIFF
--- a/components/datetime/examples/timezone_picker.rs
+++ b/components/datetime/examples/timezone_picker.rs
@@ -8,7 +8,6 @@ use icu::calendar::Date;
 use icu::datetime::{fieldsets, NoCalendarFormatter};
 use icu::locale::locale;
 use icu::time::{DateTime, Time};
-use icu_time::zone::ZoneNameTimestamp;
 use icu_time::TimeZone;
 
 fn main() {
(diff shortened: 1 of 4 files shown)


sffc:
@gemini-code-assist

> * **TODO comment**: The TODO comment in `compute_offsets_from_time_zone` indicates that the offset is not yet being used. This should be addressed to fully realize the benefits of the change.

You missed that the TODO has an issue number for follow-up work.

> * **Redundant `with_offset` call**: In `ffi/capi/src/variant_offset.rs`, the `with_offset` call seems redundant as the `time_zone` already contains the offset information.

Wrong, `time_zone` is only the time zone identifier, without the offset.

sffc:
Around a metazone transition, you may need to know the offset in order to know which metazone to use. For example, if the metazone changes at 02:00 local time, and it is 01:30, you don't know whether you are before or after the transition.

Manishearth (approved):

sffc:
Replaced by #6540

PR #6536, opened 2025-05-06
Add `Time::noon()`

### LINKED ISSUE #6530: Add Time::middle_of_day()
We have start_of_day. We should also have middle_of_day. It is the thing that returns the "average" over the course of a day, useful for ZoneNameTimestamp.

robertbastian:
Fixes #6530 

Not calling this `middle_of_day` because that feels more ambiguous than noon, e.g. it's different for 23h/25h days.

### DIFF
--- a/components/time/src/types.rs
+++ b/components/time/src/types.rs
@@ -150,13 +150,23 @@ impl Time {
         }
     }
 
-    /// Construct a new [`Time`] representing the start of the day (00:00.000)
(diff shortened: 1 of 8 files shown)


Manishearth (approved):

sffc (commented):
I think I prefer the parallelism between `start_of_day` and `middle_of_day` but 🤷‍♂️

PR #3972, opened 2023-08-30
LocaleFallbackPriority for Transliterator fallback

skius:
This adds simple language-ID-only fallback that will be used for transliterator constructor fallback. Fallback is only performed on language, script, and region, as [the examples in UTS #35](https://unicode.org/reports/tr35/tr35-general.html#Inheritance) show.

PR #6542, opened 2025-05-06
Use script fallback in transliterator

robertbastian:
Replaces #3972


components/experimental/src/transliterate/transliterator/mod.rs
```diff
@@ -382,13 +396,15 @@ impl Transliterator {
...
 
+    #[allow(clippy::too_many_arguments)]
     fn load_rbt<PT, PN, PC, F>(
         marker_attributes: &DataMarkerAttributes,
         lookup: Option<&F>,
         transliterator_provider: &PT,
         normalizer_provider: &PN,
         casemap_provider: &PC,
         allow_internal: bool,
+        fallbacker: impl Fn() -> Result<LocaleFallbacker, DataError>,
```
sffc, line 407: Thought: You can't use the fallbacker built into databake because you're falling back within data marker attributes? Maybe worth adding a comment explaining this, since it is unusual ICU4X style to explicitly take a fallbacker in a constructor.

components/experimental/src/transliterate/transliterator/mod.rs
```diff
@@ -402,11 +418,74 @@ impl Transliterator {
...
+                m
+            },
         };
-        let transliterator = transliterator_provider.load(req)?.payload;
+        let transliterator = if let Some(t) = transliterator_provider
+            .load(req)
+            .allow_identifier_not_found()?
+        {
+            t.payload
+        } else if let Ok(locale) = marker_attributes.parse::<Locale>() {
```
sffc, line 436: Question: is the request `und/de-t-fr-ch-d0-ascii` or `de/fr-ch-d0-ascii` ?

components/experimental/src/transliterate/transliterator/mod.rs
```diff
@@ -402,11 +418,74 @@ impl Transliterator {
...
+                'source: loop {
+                    if source_iterator.get().is_unknown() {
+                        break 'source;
+                    }
+                    let mut candidate = target_iterator.get().into_locale();
+                    candidate.extensions.transform = transform_extensions.clone();
+                    candidate.extensions.transform.lang =
+                        Some(source_iterator.get().into_locale().id);
+                    if let Ok(t) = Self::load_rbt(
+                        #[allow(clippy::unwrap_used)] // infallible
```
sffc, line 464: Nit: It isn't an infallible unwrap; the unwrap is safe because the invariants of being ASCII-only are upheld

components/experimental/src/transliterate/transliterator/mod.rs
```diff
@@ -402,11 +418,74 @@ impl Transliterator {
...
+                }
+                'source: loop {
+                    if source_iterator.get().is_unknown() {
+                        break 'source;
+                    }
+                    let mut candidate = target_iterator.get().into_locale();
+                    candidate.extensions.transform = transform_extensions.clone();
+                    candidate.extensions.transform.lang =
+                        Some(source_iterator.get().into_locale().id);
+                    if let Ok(t) = Self::load_rbt(
```
sffc, line 463: Question/Suggestion: why the recursive `load_rbt`? Should we also be checking the user `lookup` function? Or should you be calling directly into `transliterator_provider.load` like before?

sffc (commented):
Mostly looks good; main comment is about why you're using a recursive load.

PR #6554, opened 2025-05-07
Update release checklist

robertbastian:


Manishearth (approved):
We no longer do tutorials?

sffc (changes requested):
We still have `test-tutorials-cratesio` and it's still our very bad semver test

PR #6558, opened 2025-05-07
Allow typical accented characters at identical prefix boundary

hsivonen:
We don't currently have a bench workload where the prefix boundary would fall next to a precomposed accented character, but we should expect that situation to arise with various European languages.


echeran (approved):
LGTM. You can ignore or apply optional nits as you see fit.

As a followup at some point in the future, it would be useful to have a diagram of the various data structures used in the collator code, just because the optimizations require the different pieces in precise and specific ways.

robertbastian (approved):

robertbastian:
Merging to pick this up for the release.

Issue #6573, opened 2025-05-09
Map C/POSIX to en-US-posix and don't append -posix to other locales


hsivonen:
https://unicode-org.github.io/icu4x/rustdoc/env_preferences/parse/posix/struct.PosixLocale.html says that all locales that are parsed from a POSIX-formatted locale string get `-posix` appended to the BCP47 locale id and it says the `C` locale is mapped to `und-posix`.

This seems incorrect and inconsistent with ICU4C. According to https://bugzilla.mozilla.org/show_bug.cgi?id=1685075#c4 ICU4C maps the POSIX-reported `C` locale to `en-US-posix` in the BCP47 value space and [on StackExchange](https://superuser.com/a/1644097) @srl295 said "`en_US_POSIX` was invented to refer to the `C` (or 'null') locale code [used in POSIX libraries](https://unix.stackexchange.com/questions/87745/what-does-lc-all-c-do)."

Therefore, I'd expect us to map the `C` or `POSIX` locale to `en-US-posix` but not to append `-posix` to other locales parsed from a POSIX source. That is, I'd expect POSIX `fi_FI.UTF-8` to turn in to BCP47 `fi-FI` (_not_ `fi-FI-posix`).

See also #6511 about making it easy to get rid of the `en-US-posix` collation data consistent with Chrome and [Firefox](https://phabricator.services.mozilla.com/D117976).

hsivonen:
CC @Finchiedev @robertbastian

Finchiedev:
This seems like a good change! I was liberal with what information was added and conservative with what was removed, but I don't see a situation where the user would ever *expect* the `posix` variant of a locale - especially since the function is specifically `_lossy`.

I've opened #6575 which should hopefully address this issue. Thank you for including the extra context, it is very much appreciated as I haven't been able to find much documentation in this area!

PR #6575, opened 2025-05-10
Improve correctness of POSIX locale conversion

Finchiedev:
Addressing #6573, this PR implements the requested changes to how `env_preferences` converts POSIX locales to ICU4X locales:
- The default `C`/`POSIX` locale now maps to `en-US-posix` instead of `und-posix`
- All other locales no longer have the `-posix` variant added

Thanks to @hsivonen for clarifying how it should be done! Please let me know if there's anything I can do to help in the review process :)


utils/env_preferences/src/parse/posix.rs
```diff
@@ -233,49 +233,60 @@ impl<'src> PosixLocale<'src> {
...
-    ///     locale!("ssy-ER-posix")
+    ///     locale!("ssy-ER")
     /// );
     /// # Ok(())
     /// # }
     /// ```
     pub fn try_convert_lossy(&self) -> Result<Locale, ParseError> {
+        let mut extensions = Extensions::new();
+        let mut script = None;
+        let mut variants = Vec::new();
```
robertbastian, line 278: the comment in line 310 is obsolete

utils/env_preferences/src/parse/aliases.rs
```diff
@@ -4,59 +4,59 @@
...
 //! Platform-specific conversion from locale strings to BCP-47 identifiers.
 
-/// Find a BCP-47 language/region from a list of POSIX locale aliases.
+/// Find a BCP-47 language/region from a list of POSIX 'locale name' aliases.
 ///
-/// Some of these conversions are approximate and not exact (e.g. "C"->"und").
 /// This is based on GNU libc's `intl/locale.alias` file, with some manual processing
-/// to remove duplicates. The original file is available at:
+/// to remove duplicates. The notable exception is that the default `C`/`POSIX` locales
+/// map to `en-US`. The original file is available at:
```
robertbastian, line 11:
```suggestion
/// map to `en-US-posix`. The original file is available at:
```

utils/env_preferences/src/parse/aliases.rs
```diff
@@ -4,59 +4,59 @@
 
 //! Platform-specific conversion from locale strings to BCP-47 identifiers.
 
-/// Find a BCP-47 language/region from a list of POSIX locale aliases.
+/// Find a BCP-47 language/region from a list of POSIX 'locale name' aliases.
 ///
-/// Some of these conversions are approximate and not exact (e.g. "C"->"und").
 /// This is based on GNU libc's `intl/locale.alias` file, with some manual processing
-/// to remove duplicates. The original file is available at:
+/// to remove duplicates. The notable exception is that the default `C`/`POSIX` locales
```
robertbastian, line 10:
```suggestion
/// to remove duplicates. The notable addition is that the default `C`/`POSIX` locales
```

utils/env_preferences/src/parse/aliases.rs
```diff
@@ -4,59 +4,59 @@
...
-        "romanian" => Some((language!("ro"), Some(region!("RO")))),
-        "russian" => Some((language!("ru"), Some(region!("RU")))),
-        "slovak" => Some((language!("sk"), Some(region!("SK")))),
-        "slovene" => Some((language!("sl"), Some(region!("SI")))),
-        "slovenian" => Some((language!("sl"), Some(region!("SI")))),
-        "spanish" => Some((language!("es"), Some(region!("ES")))),
-        "swedish" => Some((language!("sv"), Some(region!("SE")))),
-        "thai" => Some((language!("th"), Some(region!("TH")))),
-        "turkish" => Some((language!("tr"), Some(region!("TR")))),
+    match posix_locale {
```
hsivonen, line 21: Please remove this whole `match` block to remove the language name checks and unify the "C" and "POSIX" case handling with the block below where you currently append `-posix`.

The X11 (see https://www.x.org/releases/X11R6.9.0/src-single/X11R6.9.0-src.tar.bz2 ) language name aliases are obsolete, so we shouldn't carry the data or spend time matching them. They resolved to legacy-encoding locales and present-day distros generate UTF-8 locales, so even on the libc-level the aliases don't resolve to an available locale and result in `C`.

hsivonen (changes requested):
Thanks, but while at it, let's remove the obsolete aliases.

utils/env_preferences/src/parse/aliases.rs
sffc: ICU4X WG discussion:

- @hsivonen These aliases don't work any more on modern systems. So it's not clear we need them.
- @robertbastian It seems harmless to include them; they come from X11. But we've already dropped a lot of support for legacy things.
- @hsivonen Both in glibc and X11, the aliases are for non-UTF-8 locales. The locales in active use from Linux distributions... RedHat switched over in ~2002. Ubuntu didn't have anything other than UTF-8. Debian switched in ~2007.
- @sffc If these locales are only used in software of that legacy, then I agree we shouldn't include them; ICU4X aims to not carry too much legacy code.
- @hsivonen And we can just ask people to go get the aliases from X11 if they need them. And we can reconsider if we get a concrete bug report.
- @hsivonen I tested on Ubuntu with UTF-8 locales installed that the aliases fall back to C, because they are aliases for non-UTF-8 locales, so the installed UTF-8 don't match.

utils/env_preferences/src/parse/aliases.rs
Finchiedev: Resolved as the alias code has been removed

utils/env_preferences/src/parse/aliases.rs
Finchiedev: Resolved as the alias code has been removed

utils/env_preferences/src/parse/aliases.rs
Finchiedev: The latest commit implements these changes, thanks all for the extra context :)

hsivonen (approved):
LGTM. Thank you!

PR #6582, opened 2025-05-12
Fix single-locale bakes

### LINKED ISSUE #6581: Single locale baked data does not include fallback if locale deduplicates to `und`
Generating with `--locale ja` produces something like

```rust

impl $provider {
            const DATA_DECIMAL_SYMBOLS_V1: icu_provider::baked::zerotrie::Data<icu::decimal::provider::DecimalSymbolsV1> = {
                const TRIE: icu_provider::baked::zerotrie::ZeroTrieSimpleAscii<&'static [u8]> = icu_provider::baked::zerotrie::ZeroTrieSimpleAscii { store: b"und\x80" };
                const VALUES: &'static [<icu::decimal::provider::DecimalSymbolsV1 as icu_provider::baked::zerotrie::DynamicDataMarker>::DataStruct] = &[icu::decimal::provider::DecimalSymbols { strings: unsafe { zerovec::VarZeroCow::from_bytes_unchecked(b"\x01\x01\x02\x02\x03\x04-+.,latn") }, grouping_sizes: icu::decimal::provider::GroupingSizes { primary: 3u8, secondary: 3u8, min_grouping: 1u8 } }];
                unsafe { icu_provider::baked::zerotrie::Data::from_trie_and_values_unchecked(TRIE, VALUES) }
            };
        }

        impl icu_provider::DataProvider<icu::decimal::provider::DecimalSymbolsV1> for $provider {
            fn load(&self, req: icu_provider::DataRequest) -> Result<icu_provider::DataResponse<icu::decimal::provider::DecimalSymbolsV1>, icu_provider::DataError> {
                let metadata = icu_provider::DataResponseMetadata::default();
                let Some(payload) = icu_provider::baked::DataStore::get(&Self::DATA_DECIMAL_SYMBOLS_V1, req.id, req.metadata.attributes_prefix_match) else { return Err(icu_provider::DataErrorKind::IdentifierNotFound.with_req(<icu::decimal::provider::DecimalSymbolsV1 as icu_provider::DataMarker>::INFO, req)) };
                Ok(icu_provider::DataResponse { payload, metadata })
            }
        }
```

Here, the `ja` values gets deduplicated to `und`, but because all values are `und`, we remove fallback, which is incorrect. We still don't need fallback, but we need to call `DataStore::get` with `und`.

robertbastian:
Fixes #6581

This will need a patch release for `icu_provider_baked`. Patching `icu_decimal_data`, `icu_segmenter_data` and `icu_experimental_data` is not strictly necessary, as we only call these implementations manually with `und`.


sffc (approved):
I think it makes sense

PR #6585, opened 2025-05-13
In zerovec, let the zerofrom dev-dependency be workspace rather than path

musicinmybrain:
This allows the dev-dependency (and particularly, the dependency on the `derive` feature) to appear in the normalized `Cargo.toml` in the published crate.

```toml
[dev-dependencies.zerofrom]
version = "0.1.3"
features = ["derive"]
default-features = false
```

This is helpful when packaging for Fedora Linux, because the `dev-dependency` on `zerofrom` is the one that carries `features = ["derive"]`, needed for running various tests included in the published `zerovec` crate.

This PR is very similar to https://github.com/unicode-org/icu4x/pull/5537.

### DIFF
--- a/utils/zerovec/Cargo.toml
+++ b/utils/zerovec/Cargo.toml
@@ -47,7 +47,7 @@ serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
 potential_utf = { path = "../../utils/potential_utf", features = ["zerovec"] }
 yoke = { workspace = true, features = ["derive"] }
-zerofrom = { path = "../../utils/zerofrom", features = ["derive"] }
+zerofrom = { workspace = true, features = ["derive"] }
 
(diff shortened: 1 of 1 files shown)


musicinmybrain:
Hmm. I thought this would be the explanation for why I couldn’t compile a few of the tests in the distribution package due to `error[E0277]: the trait bound `Cow<'a, str>: ZeroFrom<'a, str>` is not satisfied`; https://github.com/unicode-org/icu4x/pull/5537 did fix a similar problem at the time. However, while it looks like this PR does do what it claims to, it doesn’t appear to help with my testing problem.

musicinmybrain:
I’m closing this since it doesn’t seem to help me after all. The change might still be viable, but if it doesn’t have a benefit, it’s not necessarily worth reviewer attention.

PR #6592, opened 2025-05-14
Move all examples to top-level `examples` directory and refresh them

robertbastian:


echeran (commented):
This PR will break links being used in the tutorials to individual files. Those tutorial files' links should updated accordingly.

Also, a very related but separate issue, there's a lot of churn going on in the `icu4x-docs` repo. The converted versions of the affected tutorial files were [manually deleted from the repo](https://github.com/unicode-org/icu4x-docs/pull/31#issuecomment-2881446823). Similarly, there's a lot of rearrangement of files that have already happened there a couple of days ago, and that [has broken the tutorial MD conversion tool](https://github.com/unicode-org/icu4x-docs/pull/32#issuecomment-2881403172). The script should be fixed accordingly, and my advice is to avoid rearranging files at this juncture in time because it's high risk low benefit if you're trying to urgently update/fix a release.

sffc:
I didn't consider the impact on broken links.

robertbastian:
> This PR will break links being used in the tutorials to individual files. Those tutorial files' links should updated accordingly.

I believe I fixed all the links.

> my advice is to avoid rearranging files at this juncture in time because it's high risk low benefit if you're trying to urgently update/fix a release.

I'm trying to complete the website, which requires rearranging files. I don't see how it's high risk low benefit

sffc (commented):
I don't understand the motivation for this change. I feel like we've moved these files multiple times and at this point it's not clear that this layout is better than the previous layout. I understood the motivation for merging rust examples into the metacrate but can you explain what you're trying to accomplish please? 

/Gemini summarize

sffc:
/gemini review

@gemini-code-assist Does this pull request retain all the same content as before, just in different locations, or are there things being deleted, and if so, what?

sffc (approved):
ok. I don't think this PR is harmful but I don't understand the full picture you're going for, but i trust that you have a good vision in mind. But let's stop moving these files around after we land this, please :)

PR #6594, opened 2025-05-14
Update wasm demo

robertbastian:
Include the diplomat change that supports enum and primitive return values, and write a manual page for `LocaleCanonicalizer`.


ffi/capi/src/bidi.rs
```diff
@@ -22,6 +22,7 @@ pub mod ffi {
     #[diplomat::opaque]
     /// An ICU4X Bidi object, containing loaded bidi data
     #[diplomat::rust_link(icu::properties::props::BidiClass, Struct)]
+    #[diplomat::attr(demo_gen, disable)]
```
sffc, line 25: Issue: Why disable?

If a type already doesn't have a signature compatible with demo_gen, we don't need to explicitly disable anything.

Please write comments at least.

ffi/capi/src/calendar.rs
```diff
@@ -101,6 +101,7 @@ pub mod ffi {
         /// Returns the kind of this calendar
         #[diplomat::rust_link(icu::calendar::AnyCalendar::kind, FnInEnum)]
         #[diplomat::attr(auto, getter)]
+        #[diplomat::attr(demo_gen, disable)] // trivial
```
sffc, line 104: Praise: Thanks for adding a comment on why you are disabling.

Please do this everywhere. I encourage more than 1 word.

ffi/capi/src/casemap.rs
```diff
@@ -115,6 +115,7 @@ pub mod ffi {
             hidden
         )]
         #[cfg(feature = "compiled_data")]
+        #[diplomat::attr(demo_gen, disable)]
```
sffc, line 118: I won't leave comments everywhere, but please add comments explaining why you're disabling these. In this case I understand; these static functions duplicate functionality available on methods.

ffi/capi/src/date_formatter.rs
```diff
@@ -735,17 +735,18 @@ pub mod ffi {
...
-            let date = date.0;
+            let date = iso_date.0;
             let value = date;
             let _infallible = self.0.format(&value).write_to(write);
         }
         
         #[diplomat::rust_link(icu::datetime::DateTimeFormatter::format_same_calendar, FnInStruct)]
         #[diplomat::rust_link(icu::datetime::FormattedDateTime, Struct, hidden)]
         #[diplomat::rust_link(icu::datetime::FormattedDateTime::to_string, FnInStruct, hidden)]
+        #[diplomat::attr(demo_gen, disable)] // confusing
```
sffc, line 749: 🤷

ffi/capi/bindings/cpp/icu4x/DecomposingNormalizer.d.hpp
```diff
@@ -89,7 +89,7 @@ class DecomposingNormalizer {
    *
    * See the [Rust documentation for `is_normalized_utf16`](https://docs.rs/icu/latest/icu/normalizer/struct.DecomposingNormalizerBorrowed.html#method.is_normalized_utf16) for more information.
    */
-  inline bool is_normalized_utf16(std::u16string_view s) const;
+  inline bool is_normalized16(std::u16string_view s) const;
```
sffc, line 92: Question/Issue: Why did you change the function names? This seems like a breaking change

tutorials/web-demo/custom/LocaleCanonicalizer.mjs
```diff
@@ -0,0 +1,23 @@
+import { lib } from "./index.mjs";
+
+export default {
+	"LocaleCanonicalizer.canonicalize": {
```
sffc, line 4: Praise: Thanks for adding a custom function for LocaleCanonicalizer

ffi/capi/bindings/cpp/icu4x/DecomposingNormalizer.d.hpp
robertbastian: `ComposingNormalizer` had the correct `diplomat:attr` regarding UTF-8/16 on its methods, while `DecomposingNormalizer` didn't. This fixes that.

This is a bindings-only change which I *hope* should be fine as a patch release before anyone is using 2.0.0 bindings.

ffi/capi/src/bidi.rs
robertbastian: https://github.com/rust-diplomat/diplomat/pull/867 made a lot more methods "compatible", however many of them are not useful to expose in the demo. This PR does not disable any methods that were previously generated

ffi/capi/bindings/cpp/icu4x/DecomposingNormalizer.d.hpp
sffc: ok because you are following precedent from ComposingNormalizer

ffi/capi/src/units_converter.rs
```diff
@@ -73,6 +73,7 @@ pub mod ffi {
     ///
     /// You can create an instance of this object using [`UnitsConverterFactory`] by calling the `converter` method.
     #[diplomat::rust_link(icu::experimental::units::converter::UnitsConverter, Struct)]
+    #[diplomat::attr(demo_gen, disable)]
```
sffc, line 76: (missed one)

sffc (approved):
Praise: The comments are very helpful; thank you

PR #6598, opened 2025-05-15
Add UnitsIdsV1 to support the fundamental units' indices

younies:
# Description

This commit introduces the UnitsIdsV1 module, including its data provider implementation and associated data files.


provider/source/src/units/ids.rs
```diff
@@ -0,0 +1,74 @@
...
+        let indices_map: BTreeMap<String, usize> = units_data
+            .supplemental
+            .convert_units
+            .convert_units
+            .keys()
+            .enumerate()
+            .map(|(index, unit_name)| (unit_name.clone(), index))
+            .collect();
+
+        let index = indices_map
```
robertbastian, line 39: it's probably faster to use `.find` on the iterator than to create a map for a single lookup

provider/source/src/units/ids.rs
```diff
@@ -0,0 +1,74 @@
...
+            .read_and_parse("supplemental/units.json")?;
+
+        let unit = req.id.marker_attributes.as_str();
+
+        let indices_map: BTreeMap<String, usize> = units_data
+            .supplemental
+            .convert_units
+            .convert_units
+            .keys()
+            .enumerate()
```
robertbastian, line 35: I think using `enumerate` to get the ID is error prone, as you need to use the same code in all places that need IDs. Prefer adding a method on the `units_data` struct.

younies: done

components/experimental/src/measure/provider/ids.rs
```diff
@@ -0,0 +1,34 @@
...
+/// 🚧 This code is considered unstable; it may change at any time, in breaking or non-breaking ways,
+/// including in SemVer minor releases. While the serde representation of data structs is guaranteed
+/// to be stable, their Rust representation might not be. Use with caution.
+/// </div>
+#[zerovec::make_ule(IndexULE)]
+#[derive(Copy, Clone, Debug, PartialOrd, Ord, PartialEq, Eq, Default, yoke::Yokeable)]
+#[cfg_attr(feature = "datagen", derive(serde::Serialize, databake::Bake))]
+#[cfg_attr(feature = "datagen", databake(path = icu_experimental::measure::provider::ids))]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize))]
+pub struct Index(pub u16);
```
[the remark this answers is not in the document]
younies: can you illustrate this more?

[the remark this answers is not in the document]
younies: done

provider/source/src/cldr_serde/units/info.rs
```diff
@@ -80,3 +81,29 @@ pub(crate) struct Supplemental {
...
+            .enumerate()
+            .find(|(_, (name, _))| name.as_str() == unit_name)
+            .map(|(index, _)| index)
+            .ok_or_else(|| DataError::custom("Unit not found"))
+            .and_then(|index| {
+                u16::try_from(index).map_err(|_| DataError::custom("Index out of range for u16"))
+            })
+    }
+
+    pub fn get_unit_ids_map(&self) -> BTreeMap<String, usize> {
```
robertbastian, line 100: this should return `u16`s as well

robertbastian (commented):
Looks good

provider/source/src/cldr_serde/units/info.rs
robertbastian: and also be used in the other places where you use ids

younies: I prefer to return `usize` because we do not return an error with it, if you insist, then we need to return `Result`

robertbastian: yes it should return a `Result`

robertbastian (commented):
.

components/experimental/src/measure/provider/ids.rs
```diff
@@ -0,0 +1,18 @@
...
+
+#[cfg(feature = "compiled_data")]
+/// Baked data
+///
+/// <div class="stab unstable">
+/// 🚧 This code is considered unstable; it may change at any time, in breaking or non-breaking ways,
+/// including in SemVer minor releases. In particular, the `DataProvider` implementations are only
+/// guaranteed to match with this version's `*_unstable` providers. Use with caution.
+/// </div>
+pub use crate::provider::Baked;
```
Manishearth, line 13: issue: I don't think we need this.

I also think the macro below can be at the top level of provider/mod.rs, we don't need a separate module here.

provider/core/src/varule_traits.rs
```diff
@@ -121,3 +121,14 @@ impl<T, const N: usize> MaybeEncodeAsVarULE for [T; N] {
         None
     }
 }
+
+#[cfg(feature = "export")]
+impl MaybeEncodeAsVarULE for u16 {
```
Manishearth, line 126: question: why do we end up needing this?

sffc: Issue: I think this impl is used in the efficient baked data struct encoder. This should definitely not return empty, or else the packed encoder won't work

robertbastian: `u16` should not use a packed encoding. By default a type should use the normal encoding, certainly the primitives should.

sffc: Yeah I agree that this should use the normal encoding.

Also, this trait is in `icu_provider`. I regret naming this `MaybeEncodeAsVarULE` instead of `MaybeEncodeForEfficientDataBake` or something. I confused myself :)

So this is not a problem.

provider/core/src/varule_traits.rs
```diff
@@ -121,3 +121,14 @@ impl<T, const N: usize> MaybeEncodeAsVarULE for [T; N] {
...
+    fn maybe_encode_as_varule(&self) -> Option<&Self::EncodedStruct> {
+        None
+    }
+}
+
+impl MaybeAsVarULE for u16 {
```
sffc, line 132: Nit: Keep the impls in the same order as above. `MaybeAsVarULE` should go above `MaybeEncodeAsVarULE`.

provider/core/src/varule_traits.rs
Manishearth: Right, we shouldn't have this impl; u16 shouldn't be handled differently here than most data structs.

robertbastian: We need this impl if we want to use `u16` as a data struct, because all data structs need to implement this.

provider/source/src/cldr_serde/units/info.rs
younies: done

provider/core/src/varule_traits.rs
younies: done

components/experimental/src/measure/provider/ids.rs
younies: done

components/experimental/src/lib.rs
```diff
@@ -115,6 +116,7 @@ pub mod provider {
         super::displaynames::provider::RegionDisplayNamesV1::INFO,
         super::displaynames::provider::ScriptDisplayNamesV1::INFO,
         super::displaynames::provider::VariantDisplayNamesV1::INFO,
+        super::measure::provider::UnitIdsV1::INFO,
```
robertbastian, line 119: keep sorted

provider/registry/src/lib.rs
```diff
@@ -267,6 +267,7 @@ macro_rules! registry(
             icu::experimental::displaynames::provider::LocaleDisplayNamesV1: LocaleDisplayNamesV1,
             icu::experimental::displaynames::provider::ScriptDisplayNamesV1: ScriptDisplayNamesV1,
             icu::experimental::displaynames::provider::VariantDisplayNamesV1: VariantDisplayNamesV1,
+            icu::experimental::measure::provider::UnitIdsV1: UnitIdsV1,
```
robertbastian, line 270: keep sorted

provider/source/src/cldr_serde/units/info.rs
```diff
@@ -80,3 +81,54 @@ pub(crate) struct Supplemental {
...
+    /// and returns its position as a `u16` identifier.
+    ///
+    /// # Arguments
+    ///
+    /// * `unit_name` - A string slice that holds the name of the unit to search for.
+    ///
+    /// # Returns
+    ///
+    /// * `Ok(u16)` - The unique identifier for the unit if found.
+    /// * `Err(DataError)` - An error if the unit is not found or if the index is out of range for `u16`.
```
robertbastian, line 98: observation: this is very verbose for an internal function

provider/source/src/cldr_serde/units/info.rs
```diff
@@ -80,3 +81,54 @@ pub(crate) struct Supplemental {
...
+            .ok_or_else(|| DataError::custom("Unit not found"))
+            .and_then(|index| {
+                u16::try_from(index).map_err(|_| DataError::custom("Index out of range for u16"))
+            })
+    }
+
+    /// Constructs a map of unit names to their corresponding unique identifiers.
+    ///
+    /// This function iterates over the units and assigns each a unique identifier
+    /// based on its position in the list. The identifiers are represented as `u16`.
```
robertbastian, line 116: you don't need to describe implementation details in the docs

provider/source/src/cldr_serde/units/info.rs
younies: skip this for now as discussed online:  https://github.com/unicode-org/icu4x/pull/6620

provider/source/src/cldr_serde/units/info.rs
younies: done

robertbastian (approved):

sffc:
This PR landed, and it adds a new data key, so I think no more fast-forwarding the release/2.0 branch. @robertbastian

PR #6601, opened 2025-05-15
In locale_core, let litemap dev-dep. be workspace rather than path

musicinmybrain:
This allows the dev-dependency (and particularly, the dependency on the `testing` feature) to appear in the normalized `Cargo.toml` in the published crate.

This is helpful when packaging for Fedora Linux, because the `dev-dependency` on `zerofrom` is the one that carries `features = ["derive"]`, needed for running various tests included in the published `zerovec` crate.

This PR is very similar to https://github.com/unicode-org/icu4x/pull/5537 and https://github.com/unicode-org/icu4x/pull/6585, only, unlike https://github.com/unicode-org/icu4x/pull/6585, it actually does help me with running tests from the crate in the downstream package.

### DIFF
--- a/components/locale_core/Cargo.toml
+++ b/components/locale_core/Cargo.toml
@@ -33,7 +33,7 @@ zerovec = { workspace = true, optional = true }
 iai = { workspace = true }
 icu = { path = "../../components/icu", default-features = false }
 icu_provider = { path = "../../provider/core" }
-litemap = { path = "../../utils/litemap", features = ["testing"] }
+litemap = { workspace = true, features = ["testing"] }
 postcard = { workspace = true, features = ["use-std"] }
 potential_utf = { workspace = true }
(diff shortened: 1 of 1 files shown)


Manishearth (approved):
I think we typically do this when `cargo publish` would otherwise whine about nonexistant deps (https://github.com/rust-lang/cargo/issues/4242, see https://github.com/rust-lang/cargo/issues/4242#issuecomment-2578428685 about this being the "standard solution")

So I would not be in favor of accepting PRs like this one _in general_ (we'd have to revert them on publish). However in this case specifically this is about a component crate depending on a util crate, and that's fine since util crates are all published before component crates. Other PRs with that same property should also be fine.

musicinmybrain:
> So I would not be in favor of accepting PRs like this one _in general_ (we'd have to revert them on publish). However in this case specifically …

Thanks for explaining that.

PR #6602, opened 2025-05-15
In locale_core, skip check_sizes() test on non-64-bit arches

musicinmybrain:
Since the expected sizes are based on 64-bit architectures, we need to either maintain additional expected sizes for 32-bit architectures, or (as implemented in this commit) just skip this test on architectures that are not 64-bit.

I didn’t file an issue for this, but it fixes failures like this:

```
---- locale::test_sizes stdout ----

thread 'locale::test_sizes' panicked at src/locale.rs:107:5:
  left: 12
 right: 16
```

This is basically the same problem and fix as https://github.com/unicode-org/icu4x/pull/5606.

### DIFF
--- a/components/locale_core/src/locale.rs
+++ b/components/locale_core/src/locale.rs
@@ -99,6 +99,8 @@ pub struct Locale {
 }
 
 #[test]
+// Expected sizes are based on a 64-bit architecture
+#[cfg(target_pointer_width = "64")]
 fn test_sizes() {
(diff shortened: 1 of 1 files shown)


sffc (approved):
A stretch goal would be to add size assertions for 32-bit as well :) but that would require adding it to CI, etc. This is good

sffc:
@robertbastian I hit the merge button but this is test-only so it shouldn't impact 2.0.1

PR #6608, opened 2025-05-16
Generate versioned rust links, validate targets

### LINKED ISSUE #6276: Diplomat generated Dart docs should point to right version
The docs currently point to `latest`, for example `https://docs.rs/fixed_decimal/latest/fixed_decimal/struct.FixedDecimal.html#method.from_str`, not actually used version.

robertbastian:
The DDOS invocation to validate rust links is:

```shell
grep -r ffi/capi/bindings/ -hoe 'https://docs.rs/[^)]*' | sort | uniq | xargs -I{} -S1024 bash -c "wget --spider --quiet '{}' || echo '{}'"
```

Fixes #6276


sffc (approved):
Surprised our tooling didn't catch these

PR #6610, opened 2025-05-16
Fix TimeZoneVariant constructor

robertbastian:
This shouldn't take a `&mut self`


sffc (approved):
Bug fix. Should land despite changing the FFI fn signature

PR #6611, opened 2025-05-19
Fix artifacts-build.yml

robertbastian:


sffc (approved):
Yes because it is installed with `-g`

PR #6620, opened 2025-05-20
Add a unit map retrieval method to Resource struct

younies:

### DIFF
--- a/provider/source/src/cldr_serde/units/info.rs
+++ b/provider/source/src/cldr_serde/units/info.rs
@@ -7,6 +7,7 @@
 //! The file:
 //! <https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-core/supplemental/units.json>
 
+use icu_provider::DataError;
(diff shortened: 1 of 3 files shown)


younies:
offline discussion:

```rust

let parser =... ;
parser.try_from_str("square-meter-per-second");
```

@robertbastian 
Alternative with no intermediate parser type
```rust
impl MeasureUnit {
    pub fn try_new_unstable(s: &str, provider: impl DataProvider<UnitsIdV1>) -> Result<Self> {

       for part in s.split('-') {
        let id = *provider.load(request with part)?.payload.get();
           // ...
       }

    }

    pub fn try_new(s: &str) -> Result<Self> {
        Self::try_new_unstable(&crate::provider::Baked)
        }

    }

    #[cfg(feature = "serde")]
    pub fn try_new_with_buffer_provider(s: &str, provider: impl BufferProvider) -> Result<Self> {
    Self::try_new_unstable(&provider.as_deserializing())
    }

}
```

Issue #6572, opened 2025-05-09
Windows radstr should map to CLDR unihan


hsivonen:
`radstr` stands for radical-stroke, and `unihan` in CLDR means radical-stroke.

hsivonen:
CC @Finchiedev @robertbastian

Finchiedev:
Thank you very much! I will draft a PR, just to clarify:
- The Windows `_radstr` suffix should map to `-u-co-unihan` (e.g. `zh-TW_radstr` -> `zh-TW-u-co-unihan`)?
- Should the Windows `_stroke` suffix also map to `unihan`? For example, what would be the correct representation of the Windows LCID `zh-CN_stroke`?

hsivonen:
> Thank you very much! I will draft a PR, just to clarify:
> 
>     * The Windows `_radstr` suffix should map to `-u-co-unihan` (e.g. `zh-TW_radstr` -> `zh-TW-u-co-unihan`)?

Yes.

>     * Should the Windows `_stroke` suffix also map to `unihan`? For example, what would be the correct representation of the Windows LCID `zh-CN_stroke`?

`zh-CN-u-co-stroke`

I didn't test right now, but I believe we correctly let `-u-co` take precedence over the region, so it should be OK to map to `zh-CN-u-co-stroke` and let `CN` not have an effect. Likewise, `TW` is not supposed to have an effect in `zh-TW-u-co-unihan`.

PR #6621, opened 2025-05-21
Map Windows `_radstr` to CLDR `unihan` in env_preferences

Finchiedev:
Fixes issue #6572, where the `_radstr` suffix was being incorrectly mapped to `-u-co-stroke`, instead of `-u-co-unihan`.

CC @hsivonen, thanks for pointing out the mistake :)

### DIFF
--- a/utils/env_preferences/tests/parse/windows.rs
+++ b/utils/env_preferences/tests/parse/windows.rs
@@ -22,16 +22,16 @@ fn collation() {
     const CASES: [(&str, &str); 12] = [
         ("de-DE_phoneb", "de-DE-u-co-phonebk"),
         ("es-ES_tradnl", "es-ES-u-co-trad"),
-        ("ja-JP_radstr", "ja-JP-u-co-stroke"),
+        ("ja-JP_radstr", "ja-JP-u-co-unihan"),
         ("zh-CN_phoneb", "zh-CN-u-co-phonebk"),
         ("zh-CN_stroke", "zh-CN-u-co-stroke"),
-        ("zh-HK_radstr", "zh-HK-u-co-stroke"),
-        ("zh-MO_radstr", "zh-MO-u-co-stroke"),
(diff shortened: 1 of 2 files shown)


utils/env_preferences/tests/parse/windows.rs
```diff
@@ -22,16 +22,16 @@ fn collation() {
...
-        ("ja-JP_radstr", "ja-JP-u-co-stroke"),
+        ("ja-JP_radstr", "ja-JP-u-co-unihan"),
         ("zh-CN_phoneb", "zh-CN-u-co-phonebk"),
         ("zh-CN_stroke", "zh-CN-u-co-stroke"),
-        ("zh-HK_radstr", "zh-HK-u-co-stroke"),
-        ("zh-MO_radstr", "zh-MO-u-co-stroke"),
+        ("zh-HK_radstr", "zh-HK-u-co-unihan"),
+        ("zh-MO_radstr", "zh-MO-u-co-unihan"),
         ("zh-MO_stroke", "zh-MO-u-co-stroke"),
         ("zh-SG_phoneb", "zh-SG-u-co-phonebk"),
```
hsivonen, line 31: Side note: We should find out what Singapore phonebook order really is. (Not for this changeset.)

hsivonen: Or is `zh-SG_phoneb` just a test case and not something that Windows actually does?

Finchiedev: It might be worth mentioning thata all of those test cases are extracted directly from the [MS-LCID](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-lcid/70feba9f-294e-491e-b6eb-56532684c37f) document (specifically final table in section 2.2). There's a comment in `tests/dataset/mod.rs` that provides attribution, but I must have forgot to do it here.

PR #6632, opened 2025-05-26
Refactor inner single unit handling with `SingleUnitVec`

younies:
# Description 

This commit introduces the followings:
 -  The `SingleUnitVec` enum to represent collections of SingleUnit instances, allowing for zero, one, two, or multiple units.
 - The MeasureUnit struct has been updated to use `SingleUnitVec`, and related methods have been modified accordingly. 
 - Additionally, the ConverterFactory has been adjusted to work with the new `SingleUnitVec` type, enhancing the handling of unit conversions.
 - The Cargo.toml files have been updated to include the new alloc feature where necessary.

### DIFF
--- a/components/experimental/src/measure/single_unit_vec.rs
+++ b/components/experimental/src/measure/single_unit_vec.rs
@@ -20,7 +20,7 @@ pub(crate) enum SingleUnitVec {
 impl SingleUnitVec {
     /// Returns a slice of the [`SingleUnit`] instances contained
     /// within the [`SingleUnitVec`].
-    pub fn as_slice(&self) -> &[SingleUnit] {
+    pub(crate) fn as_slice(&self) -> &[SingleUnit] {
         match self {
             SingleUnitVec::Zero => &[],
             SingleUnitVec::One(unit) => core::slice::from_ref(unit),
@@ -40,27 +40,5 @@ impl SingleUnitVec {
             }
             SingleUnitVec::Multi(current_units) => current_units.push(input_unit),
         }
-    }
-}
-
-impl FromIterator<SingleUnit> for SingleUnitVec {
-    fn from_iter<I: IntoIterator<Item = SingleUnit>>(iter: I) -> Self {
-        let mut iter = iter.into_iter();
-
-        if let Some(unit1) = iter.next() {
-            if let Some(unit2) = iter.next() {
-                if let Some(unit3) = iter.next() {
-                    let mut units = Vec::with_capacity(3);
-                    units.push(unit1);
-                    units.push(unit2);
-                    units.push(unit3);
-                    units.extend(iter);
-                    return SingleUnitVec::Multi(units);
-                }
-                return SingleUnitVec::Two([unit1, unit2]);
-            }
-            return SingleUnitVec::One(unit1);
-        }
-        SingleUnitVec::Zero
     }
 }


components/experimental/src/measure/single_unit_vec.rs
```diff
@@ -0,0 +1,35 @@
...
+    One(SingleUnit),
+    Two(SingleUnit, SingleUnit),
+
+    #[cfg(feature = "alloc")]
+    Multi(Vec<SingleUnit>),
+}
+
+impl SingleUnitVec {
+    #[cfg(feature = "alloc")]
+    pub fn iter(&self) -> Box<dyn Iterator<Item = &SingleUnit> + '_> {
```
sffc, line 25: Suggestion: I would instead make a function `.as_slice()` (as suggested below) and then when you need an iterator, you just call `.as_slice().iter()`

components/experimental/src/units/converter_factory.rs
```diff
@@ -98,12 +102,19 @@ impl ConverterFactory {
...
-            && input_unit.single_units[0].si_prefix.power == 0
-            && output_unit.single_units[0].si_prefix.power == 0)
+        let input_unit = match input_unit.single_units {
+            SingleUnitVec::One(input_single_unit) => input_single_unit,
+            _ => return Some(IcuRatio::zero()),
+        };
+        let output_unit = match output_unit.single_units {
+            SingleUnitVec::One(output_single_unit) => output_single_unit,
+            _ => return Some(IcuRatio::zero()),
+        };
```
sffc, line 112: Praise: This code is much better than before

components/experimental/src/units/converter_factory.rs
```diff
@@ -168,10 +183,22 @@ impl ConverterFactory {
...
+        ) -> Result<(), InvalidUnitError> {
+            let units: &[&SingleUnit] = match units {
+                SingleUnitVec::Zero => &[],
+                SingleUnitVec::One(unit) => &[unit],
+                SingleUnitVec::Two(unit1, unit2) => &[unit1, unit2],
+                #[cfg(feature = "alloc")]
+                SingleUnitVec::Multi(units) => &units.iter().collect::<Vec<_>>(),
+                #[cfg(not(feature = "alloc"))]
+                _ => return Err(InvalidUnitError),
+            };
```
sffc, line 198: Suggestion: This should be a helper fn on SingleUnitVec. `pub fn as_slice(&self) -> &[SingleUnit]`

To make that fn work, you will need to change the Two variant to store `([SingleUnit; 2])` instead of `(SingleUnit, SingleUnit)`.

components/experimental/src/units/converter_factory.rs
```diff
@@ -271,21 +303,44 @@ impl ConverterFactory {
...
+            Ok(is_reciprocal) => is_reciprocal,
+            Err(InvalidUnitError) => return None,
+        };
 
         // Determine the sign of the powers of the units from root to unit2.
         let root_to_unit2_direction_sign = if is_reciprocal { 1 } else { -1 };
 
         let mut conversion_rate = IcuRatio::one();
-        for input_item in input_unit.single_units.iter() {
+        let input_items: &[&SingleUnit] = match &input_unit.single_units {
```
sffc, line 315: Nit: Use `.as_slice()`

components/experimental/src/units/converter_factory.rs
```diff
@@ -271,21 +303,44 @@ impl ConverterFactory {
...
+            SingleUnitVec::Zero => &[],
+            SingleUnitVec::One(input_item) => &[input_item],
+            SingleUnitVec::Two(input_item1, input_item2) => &[input_item1, input_item2],
+            #[cfg(feature = "alloc")]
+            SingleUnitVec::Multi(input_items) => &input_items.iter().collect::<Vec<_>>(),
+            #[cfg(not(feature = "alloc"))]
+            _ => return None,
+        };
+
+        let output_items: &[&SingleUnit] = match &output_unit.single_units {
```
sffc, line 325: Nit: Use `.as_slice()`

provider/source/Cargo.toml
```diff
@@ -84,7 +84,7 @@ simple_logger = { workspace = true }
 icu = { path = "../../components/icu", default-features = false, features = ["experimental"] }
 
 [features]
-default = ["use_wasm", "networking"]
+default = ["use_wasm", "networking", "experimental"]
```
sffc, line 87: Issue: Don't add "experimental" as a default feature

provider/source/Cargo.toml
```diff
@@ -70,7 +70,7 @@ zip = { workspace = true, features = ["deflate"] }
 ureq = { workspace = true, optional = true}
 
 # `experimental` feature 
-icu_experimental = { workspace = true, features = ["datagen"], optional = true }
+icu_experimental = { workspace = true, features = ["datagen", "alloc"], optional = true }
```
sffc, line 73: Nit: You won't need the "alloc" feature if you implement `.as_slice()`

components/experimental/src/measure/parser.rs
```diff
@@ -222,6 +223,17 @@ impl MeasureUnitParser {
...
+        let single_units = match single_units.len() {
+            0 => SingleUnitVec::Zero,
+            1 => SingleUnitVec::One(single_units.remove(0)),
+            2 => SingleUnitVec::Two(single_units.remove(0), single_units.remove(0)),
+            #[cfg(feature = "alloc")]
+            _ => SingleUnitVec::Multi(single_units.into_vec()),
+
+            #[cfg(not(feature = "alloc"))]
+            _ => return Err(InvalidUnitError),
+        };
```
sffc, line 235: Suggestion: this looks like something that should be a helper fn on SingleUnitVec

sffc (changes requested):
Nice work overall; some suggestions

components/experimental/src/units/converter_factory.rs
younies: done

components/experimental/src/measure/parser.rs
younies: done

components/experimental/src/units/converter_factory.rs
younies: Here, I could not implement `as_slice`, because in the case of `One` and `Two`, I could not return `&[SingleUnit]` because there is no locale container holds the value

do you have suggestions ?

provider/source/Cargo.toml
younies: done

components/experimental/src/measure/single_unit_vec.rs
younies: done

younies: Here, I could not implement `as_slice`, because in the case of `One` and `Two`, I could not return `&[SingleUnit]` because there is no locale container holds the value

do you have suggestions ?

components/experimental/src/units/converter_factory.rs
younies: thanks for the info, done :)

components/experimental/src/measure/parser.rs
```diff
@@ -217,13 +218,14 @@ impl MeasureUnitParser {
...
 
+        // TODO: shall we allow units without any single units?
         // There is no unit without any valid single units.
         if single_units.is_empty() {
             return Err(InvalidUnitError);
         }
 
         Ok(MeasureUnit {
-            single_units,
+            single_units: SingleUnitVec::from_iter(single_units),
```
robertbastian, line 228: the `single_units` variable here is still a `SmallVec`; we should construct a `SingleUnitVec` directly. You'll have to implement `push` instead of `from_iter`

components/experimental/src/measure/measureunit.rs
```diff
@@ -34,9 +32,10 @@ pub struct MeasureUnit {
 }
 
 impl MeasureUnit {
-    /// Returns a reference to the single units contained within this measure unit.
-    pub fn get_single_units(&self) -> &SmallVec<[SingleUnit; 8]> {
-        &self.single_units
+    /// Returns a slice of references to the single units contained within this measure unit.
+    #[cfg(feature = "alloc")]
+    pub fn get_single_units(&self) -> &[SingleUnit] {
```
robertbastian, line 37: nit: rust getters are not usually prefixed with `get_`
```suggestion
    pub fn single_units(&self) -> &[SingleUnit] {
```

components/experimental/src/measure/single_unit_vec.rs
```diff
@@ -0,0 +1,58 @@
...
+    Two([SingleUnit; 2]),
+
+    #[cfg(feature = "alloc")]
+    Multi(Vec<SingleUnit>),
+}
+
+impl SingleUnitVec {
+    /// Returns a slice of the [`SingleUnit`] instances contained
+    /// within the [`SingleUnitVec`].
+    #[cfg(feature = "alloc")]
```
robertbastian, line 31: remove cfg

components/experimental/src/units/converter_factory.rs
```diff
@@ -98,12 +99,19 @@ impl ConverterFactory {
...
-            && input_unit.single_units[0].si_prefix.power == 0
-            && output_unit.single_units[0].si_prefix.power == 0)
+        let input_unit = match input_unit.single_units {
+            SingleUnitVec::One(input_single_unit) => input_single_unit,
+            _ => return Some(IcuRatio::zero()),
+        };
+        let output_unit = match output_unit.single_units {
+            SingleUnitVec::One(output_single_unit) => output_single_unit,
+            _ => return Some(IcuRatio::zero()),
+        };
```
robertbastian, line 109: suggestion: use the `as_slice` path to not make this depend on `SingleUnitVec` internal representation
```suggestion
        let &[input_unit] = match input_unit.get_single_units() else {
            return Some(IcuRatio::zero());
        };
        let &[output_unit] = match output_unit.get_single_units() else {
            return Some(IcuRatio::zero());
        };
```

components/experimental/src/units/converter_factory.rs
```diff
@@ -168,11 +180,11 @@ impl ConverterFactory {
         ///     For example, `newton` has the basic units:  `gram`, `meter`, and `second` (each one has it is own power and si prefix).
         fn insert_non_basic_units(
             factory: &ConverterFactory,
-            units: &[SingleUnit],
+            units: &SingleUnitVec,
```
robertbastian, line 183: suggestion: pass a `&[SingleUnit]` and abstract away the representation
```suggestion
            units: &[SingleUnit],
```

components/experimental/src/units/converter_factory.rs
```diff
@@ -271,21 +288,25 @@ impl ConverterFactory {
...
+        for input_item in input_unit.get_single_units().iter() {
             conversion_rate *= Self::compute_conversion_term(self, input_item, 1)?;
         }
 
         if input_unit.constant_denominator != 0 {
             conversion_rate /= IcuRatio::from_integer(input_unit.constant_denominator);
         }
 
-        for output_item in output_unit.single_units.iter() {
+        for output_item in output_unit.get_single_units().iter() {
```
robertbastian, line 309: same

provider/source/src/units/helpers.rs
```diff
@@ -164,7 +164,7 @@ pub(crate) fn extract_conversion_info<'data>(
         .map_err(|_| DataError::custom("the base unit is not valid"))?;
 
     Ok(ConversionInfo {
-        basic_units: ZeroVec::from_iter(base_unit.get_single_units().iter().copied()),
+        basic_units: ZeroVec::from_iter(base_unit.get_single_units().into_iter().copied()),
```
robertbastian: nit: don't call `slice.into_iter()`, it's the same as `.iter()`.
```suggestion
        basic_units: ZeroVec::from_iter(base_unit.get_single_units().iter().copied()),
```

components/experimental/src/units/converter_factory.rs
```diff
@@ -271,21 +288,25 @@ impl ConverterFactory {
...
+            Err(InvalidUnitError) => return None,
+        };
 
         // Determine the sign of the powers of the units from root to unit2.
         let root_to_unit2_direction_sign = if is_reciprocal { 1 } else { -1 };
 
         let mut conversion_rate = IcuRatio::one();
-        for input_item in input_unit.single_units.iter() {
+
+        for input_item in input_unit.get_single_units().iter() {
```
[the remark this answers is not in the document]
younies: done

components/experimental/src/units/converter_factory.rs
younies: done

provider/source/src/units/helpers.rs
younies: done

components/experimental/src/measure/measureunit.rs
younies: done

components/experimental/src/measure/single_unit_vec.rs
younies: done

components/experimental/src/units/converter_factory.rs
younies: We need to return zero if the single unit is not equal to 1. For example, if it is zero, two, or any other number, we return zero. There is no offset in these cases.

robertbastian: this code does that, it returns zero if the slice doesn't have length 1, as opposed to the enum variant being 1

components/experimental/src/measure/measureunit.rs
```diff
@@ -34,9 +32,10 @@ pub struct MeasureUnit {
 }
 
 impl MeasureUnit {
-    /// Returns a reference to the single units contained within this measure unit.
-    pub fn get_single_units(&self) -> &SmallVec<[SingleUnit; 8]> {
-        &self.single_units
+    /// Returns a slice of references to the single units contained within this measure unit.
+    #[cfg(feature = "alloc")]
```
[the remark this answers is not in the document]
younies: done

components/experimental/Cargo.toml
```diff
@@ -75,6 +75,7 @@ compiled_data = ["dep:icu_experimental_data", "icu_decimal/compiled_data", "icu_
 datagen = ["serde", "dep:databake", "zerovec/databake", "zerotrie/databake", "tinystr/databake", "icu_collections/databake", "log", "icu_pattern/databake", "icu_plurals/datagen", "icu_pattern/alloc", "icu_provider/export"]
 ryu = ["fixed_decimal/ryu"]
 serde = ["dep:serde", "zerovec/serde", "potential_utf/serde", "tinystr/serde", "icu_collections/serde", "icu_decimal/serde", "icu_list/serde", "icu_pattern/serde", "icu_plurals/serde", "icu_provider/alloc", "icu_provider/serde", "zerotrie/serde", "icu_normalizer/serde", "icu_casemap/serde"]
+alloc = []
```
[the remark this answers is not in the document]
robertbastian: no. what exactly are you doing here?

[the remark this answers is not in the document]
younies: I removed the dependence of `alloc` as you suggested

components/experimental/src/measure/measureunit.rs
```diff
@@ -34,9 +32,9 @@ pub struct MeasureUnit {
 }
 
 impl MeasureUnit {
-    /// Returns a reference to the single units contained within this measure unit.
-    pub fn get_single_units(&self) -> &SmallVec<[SingleUnit; 8]> {
-        &self.single_units
+    /// Returns a slice of references to the single units contained within this measure unit.
```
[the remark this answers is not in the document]
younies: done

components/experimental/src/measure/parser.rs
younies: done

components/experimental/src/measure/single_unit_vec.rs
```diff
@@ -28,6 +28,21 @@ impl SingleUnitVec {
             SingleUnitVec::Multi(units) => &units[..],
         }
     }
+
+    pub(crate) fn push(&mut self, unit: SingleUnit) {
+        match self {
+            SingleUnitVec::Zero => *self = SingleUnitVec::One(unit),
+            SingleUnitVec::One(unit) => *self = SingleUnitVec::Two([*unit, *unit]),
```
robertbastian, line 35: you're duplicating the first value because you're shadowing the argument. this is probably why there's a data diff

components/experimental/src/measure/single_unit_vec.rs
```diff
@@ -28,6 +28,21 @@ impl SingleUnitVec {
...
+            SingleUnitVec::Two(units) => {
+                *self = SingleUnitVec::Multi(vec![units[0], units[1], unit]);
+            }
+            SingleUnitVec::Multi(units) => {
+                let mut units = core::mem::take(units);
+                units.push(unit);
+                *self = SingleUnitVec::Multi(units);
```
robertbastian, line 42: I don't think the take and set is necessary, `units` should be a `&mut Vec<_>` that you can just push to

sffc (commented):
Looks better; thank! I defer to @robertbastian for the detailed review and approval. The issue he found with parameter shadowing seems to be a real problem to fix.

components/experimental/src/measure/single_unit_vec.rs
```diff
@@ -0,0 +1,66 @@
...
+            }
+            SingleUnitVec::Two(current_units) => {
+                *self = SingleUnitVec::Multi(vec![current_units[0], current_units[1], input_unit]);
+            }
+            SingleUnitVec::Multi(current_units) => current_units.push(input_unit),
+        }
+    }
+}
+
+impl FromIterator<SingleUnit> for SingleUnitVec {
```
robertbastian: this is unused now

components/experimental/src/measure/single_unit_vec.rs
```diff
@@ -0,0 +1,66 @@
...
+    Zero,
+    One(SingleUnit),
+    Two([SingleUnit; 2]),
+    Multi(Vec<SingleUnit>),
+}
+
+impl SingleUnitVec {
+    /// Returns a slice of the [`SingleUnit`] instances contained
+    /// within the [`SingleUnitVec`].
+    pub fn as_slice(&self) -> &[SingleUnit] {
```
robertbastian: nit: `pub(crate)`

younies: done

components/experimental/src/measure/single_unit_vec.rs
younies: done

robertbastian (approved):

PR #6636, opened 2025-05-27
Add `set_unicode_extension` to FFI`

robertbastian:
We probably want to patch-release this, as it's blocking Dart upgrading to 2.0.


Manishearth (approved):
In favor of a patch release.

zbraniecki (commented):
How does the FFI handle when you try to set to invalid key/value?

zbraniecki:
(I'm ok with the PR, just curious how does the error handling path look like in FFI)

robertbastian:
It returns an error if either the key or the value don't parse, and doesn't modify the locale.

PR #6643, opened 2025-05-28
Remove crates.io CI

robertbastian:
We now have a semver CI, and we branch tutorials to the website on releases, so we don't really need to test against crates.io anymore.


sffc:
> We now have a semver CI

When did this happen? https://github.com/unicode-org/icu4x/pull/6619?

PR #6644, opened 2025-05-28
Remove manual clang installation in CI

robertbastian:
It's now in the actions runner image.


robertbastian:
huh looks like they didn't actually add it to the runners

sffc:
Well it is probably trying to use `clang-19` which is itself `clang` on that version

PR #6651, opened 2025-06-04
Run clippy in ignore-mode on nightly CI

robertbastian:


Manishearth (commented):
Won't this trigger the notify-slack job?

PR #6656, opened 2025-06-10
Minor collation key improvements

rs-sac:
- Minor style cleanups to make the file consistent
- Make `CollationKeySink::finish` take mutable receiver — I envisioned that the function might write to its output, but I mistakenly didn't give it a mutable reference.
- Return full collation key length on too small error — This is the polite thing to do when given a too small buffer so the caller doesn't have to guess a new size.


hsivonen (approved):
Thank you!

PR #6673, opened 2025-06-16
Add `working_directory` to dart build script

mosuem:
... and update the dependencies while at it.

Also move the hook helpers to the `src/` folder, and add an export. That necessitates another nesting layer for the symlink to the bindings.


.github/workflows/artifacts-build.yml
```diff
@@ -237,7 +237,7 @@ jobs:
     - name: Install Dart
       uses: dart-lang/setup-dart@0a8a0fc875eb934c15d08629302413c671d3f672 # v1.6.5
       with:
-        sdk: 3.7.0
+        sdk: 3.8.0
```
robertbastian: the build scripts works with 3.7.0 but not 3.8.0, which is why I haven't updated this yet.

mosuem: I updated to `3.9.0-23.0.dev` instead.

Manishearth (approved):

robertbastian (approved):

mosuem:
@robertbastian @Manishearth I don't have merge rights - feel free to merge.

PR #6674, opened 2025-06-16
Add Latin1 comparisons to the collator

hsivonen:
The motivation is for SpiderMonkey to be able to compare strings that are represented as Latin1 without converting them into a UTF-16 buffer. There is deliberately no sort key generation and no FFI, since those are not needed for the motivating use case.

The macro shape here could easily be extended to support the CPython use case from #545.

### DIFF
--- a/components/collator/Cargo.toml
+++ b/components/collator/Cargo.toml
@@ -53,6 +53,7 @@ default = ["compiled_data"]
 serde = ["dep:serde", "zerovec/serde", "icu_properties/serde", "icu_normalizer/serde", "icu_collections/serde", "icu_provider/serde"]
 datagen = ["serde", "dep:databake", "zerovec/databake", "icu_properties/datagen", "icu_normalizer/datagen", "icu_collections/databake", "icu_provider/export"]
 compiled_data = ["dep:icu_collator_data", "icu_normalizer/compiled_data", "dep:icu_locale", "icu_locale?/compiled_data", "icu_provider/baked"]
+latin1 = []
 
 [[bench]]
 name = "bench"
--- a/components/collator/src/comparison.rs
+++ b/components/collator/src/comparison.rs
@@ -158,6 +158,68 @@ fn in_inclusive_range16(i: u16, start: u16, end: u16) -> bool {
     i.wrapping_sub(start) <= (end - start)
 }
 
+/// Helper trait for getting a `char` iterator from Latin1 data.
+///
+/// ✨ *Enabled with the `latin1` Cargo feature.*
+#[cfg(feature = "latin1")]
+trait Latin1Chars {
+    fn latin1_chars(&self) -> impl DoubleEndedIterator<Item = char>;
+}
+
+#[cfg(feature = "latin1")]
+impl Latin1Chars for [u8] {
+    fn latin1_chars(&self) -> impl DoubleEndedIterator<Item = char> {
+        self.iter().map(|b| char::from(*b))
+    }
+}
+
+/// Finds the identical prefix of `left` and `right` containing
+/// Latin1.
+///
+/// Returns the identical prefix, the part of `left` after the
+/// prefix, and the part of `right` after the prefix.
+///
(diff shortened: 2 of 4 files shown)


sffc (approved):
I like this design, even though it creates many little functions. I know I said on the issue that I think we should use traits here, but it is more consistent with FFI if we just name the functions individually in Rust.

hsivonen:
Thanks!

sffc:
Henri asked in ICU4X WG whether we could cut a release for this. Notes:

- @sffc What are you hoping to see in the dot release?
- @hsivonen The latin1 stuff is what I need. It seems also polite to release the search key stuff at the same time.
- @robertbastian Are we planning a 2.1 release?
- @sffc I was thinking that 2.1 would be the CLDR 48 release.
- @robertbastian Yeah, that's what I've been assuming.
- @hsivonen Could we have a 2.0.1 release?
- @robertbastian On the release branch, sure.
- @hsivonen Can the collator component be released on its own?
- @robertbastian Seems safer to do the patch release.
- @sffc Can you use a tag, or do you need crates.io?
- @hsivonen `cargo vendor` has a limitation where it doesn't allow single crates from a workspaces to be overridden from git. There might be workarounds. But it's cleaner to have a crates.io release.
- @sffc We've typically used patch releases for bug fixes and regressions only. For shipping new features, it's nicer to do it in a normal minor release.
- @sffc Could we release a prerelease version like "2.1-dev" from the main branch?
- @robertbastian Are there breaking changes to the data structs?
- @hsivonen No, but there was new data added
- @robertbastian So we also need to release the datagen and data crates. But the compiled data might be enough for them.
- @Manishearth I'd rather not do prereleases.
- @hsivonen Is it a principle that data we put on crates.io should be generatable from datagen?
- @sffc Yes, I think we should retain that principle.

Options:

1. Cherry-pick the latin1 change to the maintenance branch and release 2.0.x of icu_collator
2. Release 2.1.0-dev1 of icu_collator and icu_collator_data from the main branch
3. Release 2.1 in the next few weeks
4. Don't release these features now; wait until October

More discussion:

- @Manishearth When it comes to patch releases... using the logic of how Rust does backports, I think it is good to keep them minimal. I don't think that means we restrict them to only fixes. But fundamentally the mechanics of a patch release are that a pull request goes to the maintenance branch and push. In other words, I don't think the politeness argument applies.
- @hsivonen The contributor hasn't specifically asked for a release at this time. They just seem eager to use it. We could ask if they need it.
- @robertbastian We have a release schedule. They can use it from the git tag if they need it. Eventually it will be on crates.io. When I contribute to another project, I don't have an expectation for the feature to be immediately released.
- @sffc I mostly agree with @robertbastian; I do think we should look at doing more frequent releases, but maybe we can set a schedule for that in 2026.
- @sffc We have a process in place for option 1; we don't have one for option 2. Maybe we want to make a process for option 2, but that is a can of worms that could have unintended side-effects. Like, maybe we release icu_collator 2.1.0-dev1 and then we discover that we need a zerovec prerelease, too. I don't want to sign up for that kind of thing. So I prefer to use our existing patch release mechanism, if we are going to do a release.
- @Manishearth Can the client just use an in-tree patch file? Run `patch` after `cargo vendor`.
- @hsivonen `cargo vendor` deliberately makes it hard to do this.
- @Manishearth I didn't know we had a release schedule of twice per year. We have never actually followed a release schedule. I wasn't aware that we were planning to follow the CLDR schedule. I would suggest, why not do a 2.1 now? I know the timezone stuff is in flux.
- @sffc Let's not scope-creep this discussion into one about release schedules in general. I was only making a remark that we will likely have a release in October _anyway_. I do not believe that what we currently have on main is worthy of a cross-component release at this time.

Conclusion:

Henri may cherry-pick the latin1 change to the maintenance branch and release 2.0.x of icu_collator if he still needs it. He may also investigate other methods of patching.

LGTM: @manishearth @hsivonen @sffc

PR #6675, opened 2025-06-16
WASM demo: expose functions under `ICU4X` namespace in the global scope

lionel-rowe:
This exposes the ICU4X module as `ICU4X` in the global scope of the WASM demo, allowing the classes to be used directly from the browser console.

It also adds code highlighting to the new snippets.

image

### DIFF
--- a/tools/web-demo/gen/rendering/rendering.mjs
+++ b/tools/web-demo/gen/rendering/rendering.mjs
@@ -290,7 +290,7 @@ export class TerminusRender extends HTMLElement {
     #parameters;
     #output;
     #code;
-    constructor(library, evaluateExternal, terminus, exprCallback = ((element) => {})) {
+    constructor(library, evaluateExternal, terminus, renderExpr = (el, expr) => el.textContent = expr) {
         super();
         generateTemplate(TerminusRender, "template", "template#terminus");
         let clone = TerminusRender.template.cloneNode(true);
@@ -329,12 +329,10 @@ export class TerminusRender extends HTMLElement {
         const shadowRoot = this.attachShadow({ mode: "open" });
         shadowRoot.appendChild(clone);
 
-        this.#code.textContent = this.#expr(...this.#parameters.exprs);
-        exprCallback(this.#code);
+        renderExpr(this.#code, this.#expr(...this.#parameters.exprs));
         for (let param of this.#parameters.children) {
             param.addEventListener('parameter-input', () => {
-                this.#code.textContent = this.#expr(...this.#parameters.exprs);
-                exprCallback(this.#code);
+                renderExpr(this.#code, this.#expr(...this.#parameters.exprs));
                 this.#output.textContent = "";
                 this.#output.classList = "";
                 if (this.#parameters.values.every((e) => e != undefined)) {
--- a/tools/web-demo/gen/rendering/rendering.mjs
+++ b/tools/web-demo/gen/rendering/rendering.mjs
@@ -314,9 +314,13 @@ export class TerminusRender extends HTMLElement {
         this.#parameters.slot = "parameters";
         this.appendChild(this.#parameters);
 
+        const pre = document.createElement("pre");
+        pre.classList.add("language-js");
+        pre.slot = "code";
+        this.appendChild(pre);
+
         this.#code = document.createElement("code");
-        this.#code.slot = "code";
-        this.appendChild(this.#code);
+        pre.appendChild(this.#code);
 
         this.#output = document.createElement("span");
         this.#output.slot = "output";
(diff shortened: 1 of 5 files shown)


tools/web-demo/public/index.html
```diff
@@ -7,7 +7,8 @@
 		<link rel="stylesheet" href="dist/index.css"/>
 		<meta name="viewport" content="width=device-width, initial-scale=1">
 	</head>
-	<body class="container">
+	<body class="container my-2 mx-0">
+        <div class="loading">Loading...</div>
 		<template id="terminus">
```
robertbastian: could you fix the indentation of the first template while you're here

tools/web-demo/src/js/index.mjs
```diff
@@ -13,5 +18,24 @@ Object.values(RenderInfo.termini).toSorted((a, b) => a.funcName < b.funcName ? -
...
+	}
+	o[segments.at(-1)] = terminus.func;
+
+	container.appendChild(details);
 });
+
+container.querySelector(".loading").remove();
+container.insertAdjacentHTML(
+	'afterbegin',
+	`<p>
```
robertbastian: please put this in the HTML

tools/web-demo/src/js/index.mjs
```diff
@@ -13,5 +18,24 @@ Object.values(RenderInfo.termini).toSorted((a, b) => a.funcName < b.funcName ? -
...
+		Functions are also available in the browser console under the global
+		<code>ICU4X</code> namespace, e.g.
+		<code>ICU4X.CaseMapper.fold("Σσς")</code>.
```
robertbastian: the issue here is that the actual ICU4X JS API is `new CaseMapper().fold("Σσς")`. This method is an implementation detail of the WASM demo, which we shouldn't expose because it's subject to change, and it's confusingly similar to our actual API. It abstract away a lot of things, like creating intermediate objects from strings, which shouldn't be abstracted away.

What I've been thinking for a while is that the WASM demo could display to the user that the `CaseMapper.fold` example runs the code `new CaseMapper().fold(s).toString()`, which the user can then run on the console (we would need to expose `CaseMapper`).

Manishearth: +1

tools/web-demo/public/index.html
lionel-rowe: Fixed, I also changed the JS indentation from tabs to spaces (I'm assuming 4 spaces is correct going by most other files, but it's not 100% consistent).

tools/web-demo/src/js/index.mjs
lionel-rowe: I've just exposed the whole module and made the example a bit more involved so it shows how to use various classes/methods together.

Manishearth: I think https://github.com/rust-diplomat/diplomat/pull/888 is the correct fix here

robertbastian:
I stomped all over this in https://github.com/unicode-org/icu4x/pull/6682, but I still think this is useful to merge. It would be nice to get the syntax highlighting to work on all code examples, and to expose the library on `globalThis`.

lionel-rowe:
> I stomped all over this in https://github.com/unicode-org/icu4x/pull/6682, but I still think this is useful to merge. It would be nice to get the syntax highlighting to work on all code examples, and to expose the library on globalThis.

If anyone else picks this up before me, just make sure to expose on [`window.top`](https://developer.mozilla.org/en-US/docs/Web/API/Window/top) as well as `window` (or `globalThis.top` as well as `globalThis`, which is the same thing). This is necessary because the docs render the demo within an iframe.

lionel-rowe:
Redone on top of the new snippet logic (which I also added syntax highlighting for). Example/explainer at the top now just shows how to destructure from the module object, as a full example isn't really necessary now the snippets are available.

tools/web-demo/gen/rendering/rendering.mjs
```diff
@@ -290,7 +290,7 @@ export class TerminusRender extends HTMLElement {
     #parameters;
     #output;
     #code;
-    constructor(library, evaluateExternal, terminus, exprCallback = ((element) => {})) {
+    constructor(library, evaluateExternal, terminus, renderExpr = (el, expr) => el.textContent = expr) {
```
robertbastian: this file is generated, so any changes to this file need a change to the Diplomat tool. can you try to make this work without touching this file? I think passing the string in `textContent` is fine, we just need to figure out the pre/code nesting, probably in HTML?

lionel-rowe: Ahh I did wonder given `/gen/` in the path. The `textContent` thing is hopefully fairly minor, the main issue is I'm not sure Prism will properly highlight if the `<code>` is within a `<slot>` in `<pre>` rather than directly in `<pre>`. I'll see if I can get it working.

lionel-rowe: Fixed, styles now look like this:

image

PR #6676, opened 2025-06-16
Introduce `duration` category for measure units

younies:
# Description

- Introducing a new module for duration measures, including methods for milliseconds, seconds, minutes, hours, days, and weeks.
- Updating the category module to include the new duration category and its associated traits.
- Adding unit tests to validate that the added units are the same as the parsed ones.
-


younies:
nice ping !

PR #6680, opened 2025-06-17
Update tzdb to 2025b

robertbastian:
I've been wanting to wait for CLDR 48, because without a short name we cannot add the new timezone. However I have a need for this data, we're already three months late, and I'm confident the CLDR short name is final.


Manishearth (approved):
we should maybe keep the future_zones function around even after this transition

robertbastian:
yes I'm planning to use it for quick tzdb updates

provider/source/src/time_zones/mod.rs
```diff
@@ -168,6 +181,7 @@ impl SourceDataProvider {
                         );
                     }
                 }
+                bcp47_tzids.extend(self.future_zones()?);
```
sffc, line 184: Suggestion: it seems like this should check for duplicates or conflicts

provider/source/src/time_zones/mod.rs
```diff
@@ -213,6 +227,7 @@ impl SourceDataProvider {
                         );
                     }
                 }
+                canonical_tzids.extend(self.future_zones()?.map(|(i, t)| (t, i)));
```
sffc, line 230: Suggestion: it seems like this should check for duplicates or conflicts

sffc (changes requested):

PR #6683, opened 2025-06-18
Refactor measure unit parser by using fixed unit ids

younies:
# Description

- Move `power` and `si_prefix` modules from the measure module to parser module.
- Use the `MeasureUnitParser` without depending on an external data.
- Remove the trie for unit indices.
- Updated tests in area, length, mass, and volume categories to directly use `MeasureUnitParser::try_from_str` instead of creating a default parser instance.

**NOTE**
We need to add a future possibility for adding new units, but for now we can have a fixed place for unit names and ids.


components/experimental/src/measure/parser/ids.rs
```diff
@@ -0,0 +1,170 @@
+// This file is part of ICU4X. For terms of use, please see the file
+// called LICENSE at the top level of the ICU4X source tree
+// (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
+use zerotrie::ZeroTrieSimpleAscii;
+
+/// This constant maps simple unit names to their unique identifiers.
+/// Identifiers are immutable; any new unit must be assigned a distinct identifier.
+/// NOTE: These identifiers are tied to the units data generation. After adding a new unit,
```
robertbastian, line 9: is there a test to enforce this?

components/experimental/src/measure/parser/ids.rs
```diff
@@ -0,0 +1,170 @@
...
+/// remember to regenerate the units data to reflect changes.
+pub const DECIMAL_PREFIXES_TRIE: ZeroTrieSimpleAscii<[u8; 1332]> =
+    ZeroTrieSimpleAscii::from_sorted_str_tuples(&[
+        // NOTE: `100-kilometer` is must be removed from CLDR, see https://unicode-org.atlassian.net/browse/CLDR-18736
```
robertbastian:
```suggestion
        // NOTE: `100-kilometer` must be removed from CLDR, see https://unicode-org.atlassian.net/browse/CLDR-18736
```

components/experimental/src/measure/parser/mod.rs
```diff
@@ -2,79 +2,37 @@
...
-    fn default() -> Self {
-        Self {
-            payload: DataPayload::from_static_ref(crate::provider::Baked::SINGLETON_UNITS_TRIE_V1),
-        }
-    }
-}
+/// A parser for the CLDR unit identifier (e.g. `meter-per-square-second`)
+pub struct MeasureUnitParser;
 
 impl MeasureUnitParser {
```
robertbastian: the type is no longer needed

```suggestion
impl MeasureUnit {
```

ffi/capi/src/measure_unit_parser.rs
```diff
@@ -27,42 +22,12 @@ pub mod ffi {
...
-        }
-
         #[diplomat::rust_link(
             icu::experimental::measure::parser::MeasureUnitParser::parse,
             FnInStruct
         )]
-        pub fn parse(&self, unit_id: &DiplomatStr) -> Option<Box<MeasureUnit>> {
-            self.0
-                .try_from_utf8(unit_id)
+        pub fn parse(unit_id: &DiplomatStr) -> Option<Box<MeasureUnit>> {
```
robertbastian:
```suggestion
        #[diplomat::attr(*, constructor)]
        pub fn create_from_string(unit_id: &DiplomatStr) -> Option<Box<MeasureUnit>> {
```

provider/source/src/cldr_serde/units/info.rs
```diff
@@ -91,17 +92,24 @@ impl Resource {
...
+            DECIMAL_PREFIXES_TRIE.cursor(),
+            |mut cursor, &byte| {
+                cursor.step(byte);
+                if cursor.is_empty() {
+                    Err(DataError::custom("Unit not found"))
+                } else {
+                    Ok(cursor)
+                }
+            },
+        )?;
```
robertbastian: you can use the `Write` trait to write into a cursor so you don't have to do this

components/experimental/src/measure/parser/ids.rs
younies: done

components/experimental/src/measure/parser/ids.rs
younies: added

components/experimental/src/measure/parser/mod.rs
younies: I prefer to have the MeasureUnitParser separate from MeasureUnit. Users will rarely need to use the parser, and in my opinion, separating it (along with the data) from MeasureUnit results in better organization.

ffi/capi/src/measure_unit_parser.rs
younies: I think this change will be needed if we removed the `MeasureUnitParser`

provider/source/src/cldr_serde/units/info.rs
younies: done

components/experimental/src/measure/parser/ids.rs
younies: done

components/experimental/src/measure/parser/ids.rs
```diff
@@ -168,3 +168,110 @@ pub const DECIMAL_PREFIXES_TRIE: ZeroTrieSimpleAscii<[u8; 1332]> =
         ("year", 153_usize),
         ("year-person", 154_usize),
     ]);
+
+#[cfg(test)]
+mod tests {
+    use crate::measure::parser::CLDR_IDS_TRIE;
+
+    // This test is to ensure that that the IDs is fixed and not changed.
+    #[test]
```
robertbastian: this test is just noise because it repeats code from above. is this linked to anything in datagen, or is this file the source of truth?

provider/source/src/cldr_serde/units/info.rs
```diff
@@ -92,24 +92,16 @@ impl Resource {
...
-            },
-        )?;
-
-        cursor.take_value().map_or_else(
-            || Err(DataError::custom("Unit not found")),
-            |value| {
+        let mut cursor = CLDR_IDS_TRIE.cursor();
+        cursor
+            .write_str(unit_name)
+            .map_err(|_| DataError::custom("Unit not found"))?;
```
robertbastian: nit: this is infallible if the id is ascii, maybe a better error message?

components/experimental/src/measure/parser/ids.rs
robertbastian: there's another instance of this

components/experimental/src/measure/parser/mod.rs
robertbastian: I don't think it makes sense to have a parser type that doesn't have any state and is not constructible. We don't do this for any other types that have a `try_from_str` method.

ffi/capi/src/measure_unit_parser.rs
younies: done

components/experimental/src/measure/parser/ids.rs
younies: right now, this file is the source of truth, it is the file the states the ids of each unit

components/experimental/src/measure/parser/ids.rs
younies: done

components/experimental/src/measure/parser/mod.rs
younies: done

younies: by the way, it will be constructable in the future when will we allow the users to add their own units :)

ffi/capi/src/measure_unit.rs
```diff
@@ -19,7 +19,8 @@ pub mod ffi {
             icu::experimental::measure::measureunit::MeasureUnit::try_from_str,
             FnInStruct
         )]
-        #[diplomat::attr(*, constructor)]
+        #[diplomat::attr(dart, rename = "from_str")]
```
robertbastian:
```suggestion
        #[diplomat::attr(dart, rename = "from_string")]
```

components/experimental/Cargo.toml
```diff
@@ -5,7 +5,7 @@
 [package]
 name = "icu_experimental"
 description = "ICU4X pre-release components under active development; all code in this crate is unstable."
-version = "0.4.0-dev"
+version = "0.5.0-dev"
```
robertbastian: don't bump dev versions

robertbastian: you can make this 0.4.0-dev.2

provider/source/src/cldr_serde/units/info.rs
younies: now, the error message is correct.

components/experimental/Cargo.toml
younies: ### done

components/experimental/src/measure/parser/ids.rs
robertbastian: please remove this test. it's not stopping anyone from changing an ID because they can just change the test as well

provider/source/src/units/helpers.rs
```diff
@@ -141,10 +141,10 @@ pub(crate) fn process_factor(
 
 /// Extracts the conversion info from a base unit, factor and offset.
 pub(crate) fn extract_conversion_info<'data>(
+    unit_id: u16,
```
robertbastian: suggestion: define `pub type UnitId = u16` somewhere and use that consistently

components/experimental/src/measure/parser/ids.rs
sffc: A good test would be one that reads the units from CLDR and checks for parity with the hard-coded list in this file. I agree that a test that simply repeats the code above it is not a useful test.

components/experimental/src/measure/parser/mod.rs
```diff
@@ -228,11 +132,69 @@ impl MeasureUnitParser {
             constant_denominator,
         })
     }
+
+    /// Retrieves the unit identifier from the given byte slice.
+    ///
+    /// # Returns
+    /// - `Ok((unit_id, remaining_part))`: If the unit id is successfully found, where `unit_id` is the identifier and `remaining_part` is the slice without the unit name and any leading `-` if present.
+    /// - `Err(InvalidUnitError)`: If the unit id is not found in the provided slice.
+    fn unit_id(part: &[u8]) -> Result<(u16, &[u8]), InvalidUnitError> {
```
robertbastian, line 141:
```suggestion
    fn unit_id(part: &[u8]) -> Result<(UnitId, &[u8]), InvalidUnitError> {
```

components/experimental/src/units/converter_factory.rs
```diff
@@ -181,21 +179,20 @@ impl ConverterFactory {
             single_units: &[SingleUnit],
             sign: i16,
             map: &mut LiteMap<u16, PowersInfo>,
```
robertbastian, line 181:
```suggestion
            map: &mut LiteMap<UnitId, PowersInfo>,
```

components/experimental/src/units/provider.rs
```diff
@@ -40,12 +40,33 @@ icu_provider::data_marker!(UnitsInfoV1, UnitsInfo<'static>, is_singleton = true)
...
+    ///
+    /// # Parameters
+    ///
+    /// * `unit_id` - A unique identifier representing the unit to be located.
+    ///
+    /// # Returns
+    ///
+    /// * `Some(&ConversionInfoULE)` - A reference to the conversion information if the unit_id is found.
+    /// * `None` - If the unit_id is not found.
+    pub fn conversion_info_by_unit_id(&self, unit_id: u16) -> Option<&ConversionInfoULE> {
```
robertbastian:
```suggestion
    pub fn conversion_info_by_unit_id(&self, unit_id: UnitId) -> Option<&ConversionInfoULE> {
```

components/experimental/src/units/provider.rs
```diff
@@ -66,6 +87,10 @@ icu_provider::data_struct!(UnitsInfo<'_>, #[cfg(feature = "datagen")]);
 )]
 #[zerovec::derive(Debug)]
 pub struct ConversionInfo<'data> {
+    /// Represents the unique identifier for the unit that is being converted.
+    /// For example, when converting from `square-meter`, `unit_id` corresponds to the identifier of `meter`.
+    pub unit_id: u16,
```
robertbastian:
```suggestion
    pub unit_id: UnitId,
```

provider/source/src/units/info.rs
```diff
@@ -27,18 +23,16 @@ impl DataProvider<UnitsInfoV1> for SourceDataProvider {
             .read_and_parse("supplemental/units.json")?;
 
         struct ConversionInfoPreProcessing<'a> {
+            unit_id: u16,
```
robertbastian:
```suggestion
            unit_id: UnitId,
```

components/experimental/src/measure/parser/ids.rs
younies: Okay, I can also parse the CLDR IDs and ensure that the unit ID is a specific unit ID. However, at the same time, I will need to enumerate all the units to ensure that if any of them change, a test will fail.

provider/source/src/units/helpers.rs
younies: done

components/experimental/src/measure/parser/ids.rs
robertbastian: > I will need to enumerate all the units to ensure that if any of them change, a test will fail.

such "stability tests" are pointless, as explained above.

components/experimental/src/measure/provider/single_unit.rs
```diff
@@ -21,5 +21,8 @@ pub struct SingleUnit {
     pub si_prefix: SiPrefix,
 
     /// The id of the unit.
-    pub unit_id: u16,
+    pub unit_id: UnitID,
 }
+
+/// Represents the unique identifier for a unit.
+pub type UnitID = u16;
```
robertbastian, line 52:
```suggestion
pub type UnitId = u16;
```

younies: done

provider/source/src/units/info.rs
younies: done

robertbastian (dismissed):

robertbastian (approved):

PR #6685, opened 2025-06-19
Add short representation generator for MeasureUnit

younies:
# Desctiption

- Implemented `generate_short_representation` method in `MeasureUnit` to provide a concise string representation of measure units, including handling of constant denominators and scientific notation.
- Added `ToString` implementation for `SiPrefix` and `SingleUnit` to facilitate their string representations, following a defined format for SI prefixes and unit IDs.
- Updated documentation with examples for clarity on usage and expected outputs.


components/experimental/src/measure/measureunit.rs
```diff
@@ -44,4 +46,176 @@ impl MeasureUnit {
...
+            let zeros_count = result.chars().rev().take_while(|&c| c == '0').count();
+
+            if zeros_count > 3 {
+                let significant_digits = &result.split_at(result.len() - zeros_count).0;
+                return [
+                    significant_digits.to_string(),
+                    "E".to_string(),
+                    zeros_count.to_string(),
+                ]
+                .concat();
```
robertbastian: this is very inefficient, it allocates 4 strings. Instead, pass around a `&mut String` and append to it

components/experimental/src/measure/measureunit.rs
```diff
@@ -44,4 +46,176 @@ impl MeasureUnit {
...
+            }
+
+            result
+        }
+
+        // Split the single units into two groups based on their SI prefix power:
+        // 1. Single units with a non-negative SI prefix power (positive or zero).
+        // 2. Single units with a negative SI prefix power.
+        // Since single units with non-negative powers are expected to appear first,
```
robertbastian: please state this invariant on the `single_units` field

components/experimental/src/measure/measureunit.rs
```diff
@@ -44,4 +46,176 @@ impl MeasureUnit {
...
+                        acc + &unit.to_string() + "#"
+                    } else {
+                        acc + &unit.to_string()
+                    }
+                });
+
+        // add constant part to the negative part
+        if self.constant_denominator > 0 {
+            if !negative_part.is_empty() {
+                negative_part.insert(0, '#');
```
robertbastian: do you need the `#`? `C` and `R` should be unambiguous without them

components/experimental/src/measure/provider/si_prefix.rs
```diff
@@ -39,3 +39,26 @@ pub struct SiPrefix {
...
+///
+/// let si_prefix = SiPrefix { power: 3, base: Base::Decimal };
+/// let string_representation = si_prefix.to_string();
+/// assert_eq!(string_representation, "D3");
+///
+/// let si_prefix = SiPrefix { power: -3, base: Base::Binary };
+/// let string_representation = si_prefix.to_string();
+/// assert_eq!(string_representation, "B-3");
+/// ```
+impl core::fmt::Display for SiPrefix {
```
robertbastian: issue: this encoding detail should not use the `Display` trait. Just define a private method that writes to a `&mut String`.

components/experimental/src/measure/provider/single_unit.rs
```diff
@@ -23,3 +23,58 @@ pub struct SingleUnit {
...
+///
+/// let full_unit = "square-millimeter";
+/// let measure_unit = parser.try_from_str(full_unit).unwrap();
+/// let single_unit = measure_unit.single_units().last().unwrap();
+/// let string_representation = single_unit.to_string();
+/// assert_eq!(string_representation, "P2D-3I85", "{}", full_unit);
+/// assert_eq!(full_unit.len() - string_representation.len(), 9, "{}", full_unit);
+///
+/// ```
+impl core::fmt::Display for SingleUnit {
```
robertbastian: ditto

components/experimental/src/measure/measureunit.rs
```diff
@@ -44,4 +46,176 @@ impl MeasureUnit {
...
+    /// use icu_experimental::measure::measureunit::MeasureUnit;
+    ///
+    ///
+    /// let parser = MeasureUnitParser::new();
+    ///
+    /// let full_unit = "meter";
+    /// let measure_unit = parser.try_from_str(full_unit).unwrap();
+    /// let short_representation = measure_unit.generate_short_representation();
+    /// assert_eq!(short_representation, "I85", "{}", full_unit);
+    /// assert_eq!(full_unit.len() - short_representation.len(), 2, "{}", full_unit);
```
robertbastian: these length difference tests are pointless because they cannot fail unless the test itself changes.

younies: I removed them from the doc test and I added them to the `mod test`, they show how much data we have saved.

components/experimental/src/measure/measureunit.rs
younies: good idea, done

components/experimental/src/measure/measureunit.rs
younies: I made a new approach, so , we will not need this

components/experimental/src/measure/measureunit.rs
younies: Sure, I was thinking about that. We only need C. In fact, we could even eliminate C if we declare the constant at the beginning, but let’s keep it for readability for now.

components/experimental/src/measure/provider/si_prefix.rs
younies: done

components/experimental/src/measure/provider/single_unit.rs
younies: done

components/experimental/src/measure/measureunit.rs
```diff
@@ -62,160 +61,90 @@ impl MeasureUnit {
...
-    ///
-    /// let full_unit = "millimeter-per-square-microsecond";
-    /// let measure_unit = parser.try_from_str(full_unit).unwrap();
-    /// let short_representation = measure_unit.generate_short_representation();
-    /// assert_eq!(short_representation, "D-3I85#R#P-2D-6I127", "{}", full_unit);
-    /// assert_eq!(full_unit.len() - short_representation.len(), 14, "{}", full_unit);
-    ///
+    /// assert_eq!(short_representation, "C100I82P-1D3I85", "{}", "liter-per-100-kilometer");
     /// ```
     pub fn generate_short_representation(&self) -> String {
```
robertbastian, line 91: is this actually intended to be a public API?

components/experimental/src/measure/measureunit.rs
```diff
@@ -62,160 +61,90 @@ impl MeasureUnit {
...
         // Convert the constant to scientific notation if it is a power of 10 with more than 3 trailing zeros
-        fn power_of_10_to_scientific(n: u64) -> String {
+        fn append_power_of_10_to_scientific(n: u64, append_to: &mut String) {
             if n < 1000 {
-                return n.to_string();
+                append_to.push_str(&n.to_string());
```
robertbastian: this allocates an intermediate string.
```suggestion
                let _infallible = write!(append_to, "{n}");
```

components/experimental/src/measure/measureunit.rs
```diff
@@ -62,160 +61,90 @@ impl MeasureUnit {
...
                 let significant_digits = &result.split_at(result.len() - zeros_count).0;
-                return [
-                    significant_digits.to_string(),
-                    "E".to_string(),
-                    zeros_count.to_string(),
-                ]
-                .concat();
+                append_to.push_str(significant_digits);
+                append_to.push('E');
+                append_to.push_str(&zeros_count.to_string());
```
robertbastian:
```suggestion
                let _infallible = write!(append_to, "{significant_digits}E{zeros_count}");
```

components/experimental/src/measure/measureunit.rs
```diff
@@ -62,160 +61,90 @@ impl MeasureUnit {
...
         // Convert the constant to scientific notation if it is a power of 10 with more than 3 trailing zeros
-        fn power_of_10_to_scientific(n: u64) -> String {
+        fn append_power_of_10_to_scientific(n: u64, append_to: &mut String) {
             if n < 1000 {
-                return n.to_string();
+                append_to.push_str(&n.to_string());
+                return;
             }
 
             let result = n.to_string();
```
robertbastian: intermediate allocation

components/experimental/src/measure/measureunit.rs
```diff
@@ -62,160 +61,90 @@ impl MeasureUnit {
...
             let zeros_count = result.chars().rev().take_while(|&c| c == '0').count();
```
robertbastian: use arithmetic instead of counting characters in a string

components/experimental/src/measure/measureunit.rs
```diff
@@ -62,160 +61,90 @@ impl MeasureUnit {
...
-    /// let short_representation = measure_unit.generate_short_representation();
-    /// assert_eq!(short_representation, "D-3I85#R#P-2D-6I127", "{}", full_unit);
-    /// assert_eq!(full_unit.len() - short_representation.len(), 14, "{}", full_unit);
-    ///
+    /// assert_eq!(short_representation, "C100I82P-1D3I85", "{}", "liter-per-100-kilometer");
     /// ```
     pub fn generate_short_representation(&self) -> String {
         // Convert the constant to scientific notation if it is a power of 10 with more than 3 trailing zeros
-        fn power_of_10_to_scientific(n: u64) -> String {
+        fn append_power_of_10_to_scientific(n: u64, append_to: &mut String) {
```
robertbastian: `buf` would be a better name for `append_to`

components/experimental/src/measure/provider/si_prefix.rs
```diff
@@ -40,25 +43,27 @@ pub struct SiPrefix {
...
+    /// si_prefix.append_short_representation(&mut short_representation);
+    /// assert_eq!(short_representation, "D3");
+    ///
+    /// let mut short_representation = String::new();
+    /// let si_prefix = SiPrefix { power: -3, base: Base::Binary };
+    /// si_prefix.append_short_representation(&mut short_representation);
+    /// assert_eq!(short_representation, "B-3");
+    /// ```
+    pub fn append_short_representation(&self, append_to: &mut String) {
+        append_to.push(if self.base == Base::Decimal { 'D' } else { 'B' });
```
robertbastian: nit: use match for exhaustive matching in case we add more bases in the future

components/experimental/src/measure/provider/si_prefix.rs
```diff
@@ -40,25 +43,27 @@ pub struct SiPrefix {
...
+        append_to.push_str(&self.power.to_string());
```
robertbastian: intermediate allocation

components/experimental/src/measure/provider/single_unit.rs
```diff
@@ -24,57 +26,41 @@ pub struct SingleUnit {
...
+    ///     unit_id: 85,
+    /// };
+    /// single_unit.append_short_representation(&mut short_representation);
+    /// assert_eq!(short_representation, "P3D2I85");
+    /// ```
+    pub fn append_short_representation(&self, append_to: &mut String) {
         if self.power != 1 {
-            write!(f, "P{}", self.power)?;
+            append_to.push('P');
+            append_to.push_str(&self.power.to_string());
```
robertbastian: intermediate allocations

components/experimental/src/measure/measureunit.rs
```diff
@@ -62,160 +61,90 @@ impl MeasureUnit {
...
+        for (full_unit, expected_short, expected_diff) in test_cases {
+            let measure_unit = parser.try_from_str(full_unit).unwrap();
+            let short_representation = measure_unit.generate_short_representation();
+            assert_eq!(short_representation, expected_short, "{}", full_unit);
+            assert_eq!(
+                full_unit.len() - short_representation.len(),
+                expected_diff,
+                "{}",
+                full_unit
+            );
```
robertbastian: still, this is pointless assertion. it's at most useful for a code reviewer to demonstrate that the strings are shorter, but I can see that. there's no point in calculating the diff numbers when new test cases are added, or actually executing this code

younies: okay, as you like :)

components/experimental/src/measure/provider/si_prefix.rs
younies: done

components/experimental/src/measure/provider/si_prefix.rs
younies: done

younies: However, using Write! is returning an error, I am not sure If I need to bubble it up ?

components/experimental/src/measure/measureunit.rs
younies: For now, yes. Once we start using it internally, it will be private.

components/experimental/src/measure/measureunit.rs
younies: done

components/experimental/src/measure/measureunit.rs
younies: done

components/experimental/src/measure/measureunit.rs
younies: done

components/experimental/src/measure/measureunit.rs
younies: We need to have the string version either way, i.e. the first part, and I would like to make the code simpler

components/experimental/src/measure/measureunit.rs
robertbastian: what requires this to be public now? I'm worried that this will be forgotten and accidentally made public

components/experimental/src/measure/measureunit.rs
```diff
@@ -77,9 +78,9 @@ impl MeasureUnit {
     /// ```
     pub fn generate_short_representation(&self) -> String {
         // Convert the constant to scientific notation if it is a power of 10 with more than 3 trailing zeros
-        fn append_power_of_10_to_scientific(n: u64, append_to: &mut String) {
+        fn append_power_of_10_to_scientific(n: u64, buff: &mut String) {
             if n < 1000 {
-                append_to.push_str(&n.to_string());
+                write!(buff, "{}", n).unwrap();
```
robertbastian: this is infallible even though the API is fallible, which is why my suggestion was
```suggestion
                let _infallible = write!(buff, "{}", n);
```
this way no panicking code gets linked, and Rust doesn't complain about not using a `Result`.

components/experimental/src/measure/measureunit.rs
```diff
@@ -88,13 +89,11 @@ impl MeasureUnit {
 
             if zeros_count > 3 {
                 let significant_digits = &result.split_at(result.len() - zeros_count).0;
-                append_to.push_str(significant_digits);
-                append_to.push('E');
-                append_to.push_str(&zeros_count.to_string());
+                write!(buff, "{}E{}", significant_digits, zeros_count).unwrap();
```
robertbastian: nit: write variables in the format string so that it's clearer what goes where
```suggestion
                write!(buff, "{significant_digits}E{zeros_count}").unwrap();
```

components/experimental/src/measure/measureunit.rs
robertbastian: we don't need the string in an extra allocation, we *might* write it to `buff` (but also maybe not, if we use scientific notation), so this is an allocation that is not necessary.

components/experimental/src/measure/measureunit.rs
younies: We will needed in the data gen

components/experimental/src/measure/measureunit.rs
younies: done

components/experimental/src/measure/provider/si_prefix.rs
younies: `let _infaillable`

components/experimental/src/measure/measureunit.rs
```diff
@@ -77,23 +76,52 @@ impl MeasureUnit {
...
+            while n % 10_000 == 0 {
+                n /= 10_000;
+                zeros_count += 4;
+            }
+
+            // Divide by 10^2
+            while n % 100 == 0 {
+                n /= 100;
+                zeros_count += 2;
+            }
```
robertbastian: what's the point of these? repeatedly dividing by 10 is sufficient

younies: for reducing the time  complexity

robertbastian: the time complexity is constant because the input is fixed-length. I don't think this is worth the additional code complexity

components/experimental/src/measure/provider/si_prefix.rs
```diff
@@ -39,3 +42,31 @@ pub struct SiPrefix {
...
+    /// let si_prefix = SiPrefix { power: 3, base: Base::Decimal };
+    /// si_prefix.append_short_representation(&mut short_representation);
+    /// assert_eq!(short_representation, "D3");
+    ///
+    /// let mut short_representation = String::new();
+    /// let si_prefix = SiPrefix { power: -3, base: Base::Binary };
+    /// si_prefix.append_short_representation(&mut short_representation);
+    /// assert_eq!(short_representation, "B-3");
+    /// ```
+    pub fn append_short_representation(&self, buff: &mut String) {
```
robertbastian: private?

components/experimental/src/measure/provider/single_unit.rs
```diff
@@ -23,3 +25,42 @@ pub struct SingleUnit {
...
+    /// let mut short_representation = String::new();
+    /// let single_unit = SingleUnit {
+    ///     power: 3,
+    ///     si_prefix: SiPrefix { power: 2, base: Base::Decimal },
+    ///     unit_id: 85,
+    /// };
+    /// single_unit.append_short_representation(&mut short_representation);
+    /// assert_eq!(short_representation, "P3D2I85");
+    /// ```
+    pub fn append_short_representation(&self, buff: &mut String) {
```
robertbastian: private?

components/experimental/src/measure/measureunit.rs
robertbastian: how will you make it private if it's used in datagen? should it be in `provider`?

younies: we should have a field for the short unit id, but it is okay, the `fn` can be public for now

components/experimental/src/measure/provider/si_prefix.rs
younies: `pub(crate)`

components/experimental/src/measure/provider/single_unit.rs
younies: `pub(crate)`

components/experimental/src/measure/provider/si_prefix.rs
```diff
@@ -39,3 +42,14 @@ pub struct SiPrefix {
...
 }
+
+impl SiPrefix {
+    /// Appends the short representation of the si prefix to the given string.
+    pub(crate) fn append_short_representation(&self, buff: &mut String) {
+        buff.push(match self.base {
+            Base::Decimal => 'D',
+            Base::Binary => 'B',
+        });
+        write!(buff, "{}", self.power).unwrap();
```
robertbastian: don't unwrap, it adds panicking code

younies: good catch

robertbastian: also in other places

robertbastian (commented):
.

components/experimental/src/measure/provider/si_prefix.rs
younies: done

robertbastian (approved):

PR #6689, opened 2025-06-20
Update diplomat

robertbastian:
Uses a select element ("enum") for the segmenter type


robertbastian:
admin-merging, no changes that could break ci-job-test

Issue #6696, opened 2025-06-23
zerotrie: several test failures related on 32-bit platforms


musicinmybrain:
Working in a 32-bit chroot:

```
# git clone https://github.com/unicode-org/icu4x.git
# cd icu4x/utils/zerotrie
# cargo test --all-features
[…]
     Running unittests src/lib.rs (/builddir/icu4x/target/debug/deps/zerotrie-e67f800abbeaa387)
[…]
failures:

---- varint::tests::test_max stdout ----

thread 'varint::tests::test_max' panicked at utils/zerotrie/src/varint.rs:437:9:
assertion `left == right` failed
  left: [47, 223, 223, 223, 95]
 right: [33, 223, 223, 223, 223, 223, 223, 223, 223, 95]
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- varint::tests::text_extended_max stdout ----

thread 'varint::tests::text_extended_max' panicked at utils/zerotrie/src/varint.rs:462:9:
assertion `left == right` failed
  left: [31, 239, 239, 239, 111]
  right: [17, 239, 239, 239, 239, 239, 239, 239, 239, 111]


failures:
    varint::tests::test_max
    varint::tests::text_extended_max

test result: FAILED$<2>. 21 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.27s

error: test failed, to rerun pass `--lib`
# cargo test --all-features --test locale_aux_test
[…]
failures:

---- test_combined stdout ----

thread 'test_combined' panicked at utils/zerotrie/tests/locale_aux_test.rs:31:5:
assertion `left == right` failed
  left: 14427
 right: 18635
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- test_aux_split stdout ----

thread 'test_aux_split' panicked at utils/zerotrie/tests/locale_aux_test.rs:142:5:
assertion `left == right` failed
  left: 6742
 right: 8386


failures:
    test_aux_split
    test_combined

test result: FAILED$<2>. 0 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s

error: test failed, to rerun pass `--test locale_aux_test`
```

musicinmybrain:
[...]

PR #6697, opened 2025-06-23
Fix test failures in zerotrie on 32-bit platforms

musicinmybrain:
- Provide 32-bit cases for `MAX_VARINT` tests
- Make `test_combined` and `test_aux_split` 64-bit-only

Fixes https://github.com/unicode-org/icu4x/issues/6696; see the discussion in that issue for rationale.

### DIFF
--- a/utils/zerotrie/src/varint.rs
+++ b/utils/zerotrie/src/varint.rs
@@ -434,6 +434,7 @@ mod tests {
         );
         assert!(remainder.is_empty());
         assert_eq!(recovered_value, MAX_VARINT);
+        #[cfg(target_pointer_width = "64")]
         assert_eq!(
             write_bytes.as_slice(),
             &[
@@ -443,6 +444,17 @@ mod tests {
                 0b11011111, //
                 0b11011111, //
                 0b11011111, //
+                0b11011111, //
+                0b11011111, //
+                0b11011111, //
+                0b01011111, //
+            ]
+        );
+        #[cfg(target_pointer_width = "32")]
+        assert_eq!(
+            write_bytes.as_slice(),
+            &[
+                0b00101111, //
                 0b11011111, //
                 0b11011111, //
                 0b11011111, //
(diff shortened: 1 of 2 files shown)


Manishearth (approved):
Quite reasonable: We do not want there to ever be things that are represented differently on 32bit vs 64bit (since we want to be able to zero-copy deserialize on both), but 64bit supporting more possible values where 32bit errors is okay.

Testing that with target_pointer_width is fine. Since we don't run them in CI these tests may break every now and then but that should be ok, it should be easy to fix.

musicinmybrain:
> Quite reasonable: We do not want there to ever be things that are represented differently on 32bit vs 64bit (since we want to be able to zero-copy deserialize on both), but 64bit supporting more possible values where 32bit errors is okay.

With this reminder that the representation should be architecture-independent, I think I better understand the problem with the size assertions in `utils/zerotrie/tests/locale_aux_test.rs`. For example:

https://github.com/unicode-org/icu4x/blob/0d2e6b2cf75d7e141c0d22edeef45189579a912d/utils/zerotrie/tests/locale_aux_test.rs#L31-L35

On 32-bit, this fails with:

```
---- test_combined stdout ----

thread 'test_combined' panicked at utils/zerotrie/tests/locale_aux_test.rs:31:5:
  left: 14427
 right: 18635
```

Based on your comment, this assertion should not use `core::mem::size_of::<usize>()`, but a hard-coded `8`, because the representation should not depend on the pointer size.

musicinmybrain:
> Based on your comment, this assertion should not use `core::mem::size_of::<usize>()`, but a hard-coded `8`, because the representation should not depend on the pointer size.

Hmm, no, that’s not correct either. But I’m having a hard time understanding whether the discrepancy represents a real problem or not.

robertbastian:
I think there are definitely issues with the API, even though the encoding might work because it uses varints. Afaik you can serialize `usize::MAX` values on 64-bit platforms, and then access them as `usize`s on 32-bit platforms, but I don't know what happens if the value uses more than 32 bits.

sffc:
I designed the API with `usize` because the encoding in the zerotrie stores all values as varints and is therefore not opinionated about how many bytes you get out. If your integer type is smaller than the one you used to build the zerotrie, everything is well-behaved, but you just get fewer bits (the high bits are lost).

PR #6712, opened 2025-06-25
Delete dead code in icu_provider_source legacy datetime module

robertbastian:
#5613

### DIFF
--- a/provider/source/src/datetime/legacy.rs
+++ b/provider/source/src/datetime/legacy.rs
@@ -5,19 +5,12 @@
 //! TODO(#5613): Even though these markers are no longer exported, we need them in order to export
 //! semantic skeleton data markers. This should be refactored to skip the intermediate data struct.
 
-use super::DatagenCalendar;
-use crate::cldr_serde::eras::EraData;
-use crate::cldr_serde::{self, ca};
-use crate::IterableDataProviderCached;
-use crate::SourceDataProvider;
 use alloc::borrow::Cow;
 use icu::datetime::provider::pattern::runtime;
 use icu::datetime::provider::skeleton::*;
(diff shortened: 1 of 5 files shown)


sffc:
Please don't send PRs with a 3-word title and no description and no issue, especially when an issue exists: #5613

provider/source/src/datetime/legacy.rs
```diff
@@ -324,1050 +317,367 @@ symbols!(
...
-                STADLN_NARW_STR => payload
-                    .day_periods
+    impl ca::Contexts<cldr_serde::ca::DaySymbols> {
+        fn get(&self, ctx: &()) -> weekdays::Contexts<'static> {
+            weekdays::Contexts {
+                format: weekdays::FormatWidths {
+                    abbreviated: self.format.abbreviated.get(ctx),
+                    narrow: self.format.narrow.get(ctx),
+                    short: self.format.short.as_ref().map(|width| width.get(ctx)),
+                    wide: self.format.wide.get(ctx),
```
sffc, line 476: Observation: why is `short` different from the others? It doesn't seem to be that way in the old code

robertbastian:
Dead code deletion is a pretty clear description, I don't see what more context you'd want. I didn't know there was an issue for this.

provider/source/src/datetime/legacy.rs
sffc: ^ this looks fishy to me, although it is in test code

sffc:
I changed the title for you.

The people who have to read these commit messages are people who do things like prepare the changelog. It is useful to know what the change is related to.

robertbastian:
you want to add dead code deletion commits to the changelog?

PR #6713, opened 2025-06-25
Add `TimeZoneVariant::Sundown` and use it for Morocco

robertbastian:
This is option 4 from https://unicode-org.atlassian.net/browse/CLDR-18795


robertbastian:
This PR is not ready to be reviewed. You can leave comments on the CLDR issue

sffc:
Belated notes from 2025-07-10 ICU4X WG:

- @robertbastian Also see my time zone explorer: https://robertbastian.dev/tz.html?hl=en#3.04/51.01/3.06
- @sffc This seems like a reasonable approach. But I'm concerned about how this interacts with Rearguard isdst. What we shipped in 2.0, by my mental model, is just a named Rearguard isdst.
- @robertbastian I've found rearguard isdst to be pretty unusable anyway, no implementation I've found so far can guarantee rearguard semantics. We've always kept the door open for more variants in the future, this is it
- @sffc Potentially someone could use `#[cfg(macos)]` or similar if they are in an environment that guarantees rearguard isdst.
- @robertbastian nobody seems to be willing to guarantee or document this though, it's such an old obscure flag that has been advised against for nearly decades
- @sffc The variant offsets table is an invention of ICU4X. ICU zoneinfo64, etc., have always used just a boolean. This is how zone calculations have worked for 20+ years and it's not clear that Casablanca is a big enough motivation for us to change the model. We have alternatives, like adding a new CLDR metazone.
- @robertbastian - and we're hitting limitations of this model, both with Morocco and with distinguishing Mountain Time from Mountain Time
- @sffc If we change the model... we should add data in CLDR for mapping between rearguard and the new model, and get buy-in from ICU.
- @sffc Also, rearguard and vanguard are already two models for this. Is the motivation really strong enough to add a third?

---

- @robertbastian - What do we think about using localized offset during ramadan until we get CLDR translations?
- @sffc - fine for now

PR #6715, opened 2025-06-27
Add datetime FFI test for #6688

sffc:
Waiting on https://github.com/rust-diplomat/diplomat/pull/898 before this can land

But when I do manage to run the test (manually exporting the missing functions from diplomat-runtime), the test passes. I can't reproduce the issue reported in #6688 via Rust code. @robertbastian


robertbastian:
Thanks for investigating

PR #6736, opened 2025-07-08
Fix `und-SA-u-ca-islamic`

### LINKED ISSUE #6559: DateTimeFormatter calendar fallback doesn't support und-SA-u-ca-islamic even though it should
`DateTimeFormatter::try_new(locale!("en-SA-u-ca-islamic").into(), ...)` should return an Umm al-Qura formatter but it doesn't.

It is a bug involving FormattableAnyCalendarKind::from_preferences.

robertbastian:
Fixes #6559

### DIFF
--- a/components/datetime/src/scaffold/calendar.rs
+++ b/components/datetime/src/scaffold/calendar.rs
@@ -336,6 +336,10 @@ fn test_calendar_fallback() {
         FormattableAnyCalendarKind::from_preferences(locale!("en-SA-u-ca-islamic").into()),
         FormattableAnyCalendarKind::HijriUmmAlQura
     );
+    assert_eq!(
+        FormattableAnyCalendarKind::from_preferences(locale!("en-IL-u-ca-islamic").into()),
+        FormattableAnyCalendarKind::Gregorian
+    );
 }
 
 /// A version of [`AnyCalendar`] for the calendars supported in the any-calendar formatter.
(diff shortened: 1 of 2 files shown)


zbraniecki (approved):

components/datetime/src/scaffold/calendar.rs
```diff
@@ -303,29 +303,41 @@ impl FormattableAnyCalendarKind {
...
         }
+
+        debug_assert!(false, "all locale-default calendars are supported");
+        // fall back to something non-Gregorian to make errors more obvious
+        FormattableAnyCalendarKind::Coptic
     }
 }
 
+#[test]
+fn asfd() {
```
sffc:
```suggestion
fn test_calendar_fallback() {
```

components/datetime/src/scaffold/calendar.rs
```diff
@@ -303,29 +303,41 @@ impl FormattableAnyCalendarKind {
...
+        FormattableAnyCalendarKind::Buddhist
+    );
+    assert_eq!(
+        FormattableAnyCalendarKind::from_preferences(locale!("en-TH").into()),
+        FormattableAnyCalendarKind::Buddhist
+    );
+    assert_eq!(
+        FormattableAnyCalendarKind::from_preferences(locale!("en-SA-u-ca-islamic").into()),
+        FormattableAnyCalendarKind::HijriUmmAlQura
+    );
```
sffc, line 338: Nit: Add a test for `Hijri(None)` in a different region, and make it fall back to e.g. `HijriTabularTypeIIThursday`

robertbastian: I don't think we reached consensus on a default Hijri calendar outside of Saudi-Arabia.

sffc (approved):

components/datetime/src/scaffold/calendar.rs
sffc: Ok, it's an easy thing to change now that the fallback algorithm is fixed

Issue #6628, opened 2025-05-22
Strict preference constructors


robertbastian:
Currently the `Locale` -> preferences constructors are infallible and drop unknown tags. We could add fallible constructors that fail on unknown tags.

PR #6737, opened 2025-07-08
Add Preferences::from_locale_strict

robertbastian:
#6628


zbraniecki (approved):

sffc (commented):
LGTM

components/locale_core/src/preferences/mod.rs
```diff
@@ -564,6 +543,49 @@ macro_rules! __define_preferences {
...
+                            Ok(Some(k)) => {
+                                $key = Some(k);
+                                continue;
+                            }
+                            Ok(None) => {}
+                            Err(_) => {
+                                is_err = true
+                            }
+                        }
+                    )*
```
sffc, line 572: Thought: Should it also fail if the key was not found?

robertbastian: No, because clients shouldn't need to know which keys each preferences object respects.

sffc:
Why is it `from_locale_strict` instead of `try_from_locale`?

robertbastian:
at this point because it's released

sffc:
https://github.com/unicode-org/icu4x/issues/7379

PR #6739, opened 2025-07-09
Prepare Dart package for publishing

mosuem:
Merge after https://github.com/dart-lang/native/pull/2417

This is an idea of how a `package:icu4x` could look like. Feel free to refactor at will to better suit your workflows - I would like to get something like this landed sometime soon to be able to iterate on it.

Matches https://github.com/dart-lang/i18n/pull/993.

The failure of the `fetch` and `check_hashes` workflows are expected as there are no published binaries yet - `skip-fetch` as a label in this repo would skip the fetching.


ffi/dart/test/icu_test.dart
```diff
@@ -9,31 +9,6 @@ void main() {
     expect(x.toString(), '1.4940300');
   });
 
-  test('LocaleFallbacker', () {
```
robertbastian: don't remove tests, these are failing because the binary is built without the relevant features. the icu4x publish should build with default_components and *maybe* experimental as well

mosuem: Just checking if the CI gets green.

ffi/dart/example/pubspec.yaml
robertbastian, line 1: this should be examples/dart

mosuem:
Regarding the binaries for upload - for `package:intl4x`, the binaries hosted on Github should only contain the symbols accessible through the intl4x API. The treeshaking is nice to have, but there is no reason to have users with treeshaking disabled download large binaries containing symbols they will never use.

We could upload two sets of binaries - `full` and `reduced`?

robertbastian:
How would the download hook know that only intl4x symbols are used? What if some project uses intl4x and other icu4x APIs directly? I think for now we should just host an everything binary and let treeshaking deal with the size.

mosuem:
> How would the download hook know that only intl4x symbols are used? What if some project uses intl4x and other icu4x APIs directly? I think for now we should just host an everything binary and let treeshaking deal with the size.

We could add a build option, so that we use a reduced set by default and can switch to the full set if required. But we can keep the full set for now - I am a bit worried about size implications though.

ffi/dart/pubspec.yaml
```diff
@@ -1,19 +1,52 @@
...
+      buildMode: checkout
+      checkoutPath: ../../
+
+dependency_overrides:
+  code_assets: 
+    path: ../../../native/pkgs/code_assets
+  hooks:
+    path: ../../../native/pkgs/hooks
+  native_toolchain_c: 
+    path: ../../../native/pkgs/native_toolchain_c
```
robertbastian: this doesn't work. if they're not published, can we pull them in as git dependencies?

.github/workflows/icu4x_artifacts.yml
robertbastian, line 1: I'd prefer if you put artifacts generation workflows in a separate PR. Artifacts might not actually be Dart-specific.

ffi/dart/lib/hook.dart
```diff
@@ -0,0 +1 @@
+export 'src/hook_helpers/link_helpers.dart' show SymbolKeeper;
```
robertbastian, line 1: why is this a public API?

.github/workflows/dart.yml
robertbastian, line 1: let's not merge any failing jobs, we can do this in a follow-up

ffi/dart/lib/hook.dart
mosuem: So that other packages can send their symbols to this package, see https://github.com/dart-lang/i18n/pull/993/files#diff-03d883e351f521ff45d1c403250f406302df7d2a545ccc7e2ef354877dea5e95R38

ffi/dart/pubspec.yaml
mosuem: We will not merge this PR until they are published.

.github/workflows/dart.yml
mosuem: It will not fail once all is published - otherwise I won't merge.

.github/workflows/icu4x_artifacts.yml
mosuem: Yeah feel free to refactor this however you like. This is just an example/template basically.

ffi/dart/lib/hook.dart
robertbastian: Is this something that could be upstreamed into the native_toolchain_c or hooks package?

mosuem: No, this is core to how linking works - the package which does the treeshaking must define the API for getting the  symbols to treeshake.

robertbastian (dismissed):

mosuem:
i don't have merge rights, so feel free to merge. Also if there is no icu4x_artifacts, the code for fetching the artifacts must be updated. Finally, a version `2.0.0-dev.0` is a bit unusual at least, normally experimental packages would have a `0.x` version number. The `2` suggests a stable API.

robertbastian:
> Finally, a version 2.0.0-dev.0 is a bit unusual at least, normally experimental packages would have a 0.x version number. The 2 suggests a stable API.

That's what the `-dev.0` is for. I want to keep the association with ICU4X releases.

robertbastian:
> if there is no icu4x_artifacts, the code for fetching the artifacts must be updated

I just don't want to add these in this PR because it needs more work.

robertbastian (approved):

PR #6742, opened 2025-07-10
Improve offsets data

robertbastian:
#6238

Based on a bug in `parse-zoneinfo` that I fixed


sffc (dismissed):
Seems reasonable and it makes the data cleaner

PR #6743, opened 2025-07-10
Enable experimental code in NPM

robertbastian:
We have bindings for it but they currently fail

### DIFF
--- a/ffi/capi/bindings/cpp/icu4x/DisplayNamesFallback.d.hpp
+++ b/ffi/capi/bindings/cpp/icu4x/DisplayNamesFallback.d.hpp
@@ -25,6 +25,8 @@ namespace capi {
 
 namespace icu4x {
 /**
+ * 🚧 This API is experimental and may experience breaking changes outside major releases.
+ *
  * See the [Rust documentation for `Fallback`](https://docs.rs/icu/2.0.0/icu/experimental/displaynames/enum.Fallback.html) for more information.
  */
 class DisplayNamesFallback {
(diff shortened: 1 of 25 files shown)


Manishearth (commented):
What does this mean in relation to our stable API guarantees on FFI?

We never formally decided anything for NPM. Should we?

(I know these APIs already exist and will just not work currently, but we should decide this if we're publishing something)

robertbastian:
- @Manishearth We should make sure that the experimental APIs are marked as experimental in our FFI docs
- @sffc The NPM package has other rough edges
- @robertbastian It is fine except for ABI bugs that we find and fix
- @sffc That, and we don't have a story around code/data slicing
- @Manishearth In Rust, semver is a signalling mechanism around when you should upgrade, and also readiness. It's not clear that readiness is as much of a signal on NPM.
- @robertbastian If you want a more optimized wasm file, you need to do your own build. I don't see a better way to do that.
- @sffc I would be onboard with documenting the APIs using TSDoc `@experimental`: https://tsdoc.org/pages/tags/experimental/


---

- @robertbastian A few directions we could go here. What's our policy?
- @hsivonen Are there clients?
- @robertbastian Yes, Dart needs display names.
- @hsivonen Does this mean committing to a design?
- @robertbastian I guess we need to figure out how to mark experimental over FFI. There isn't a great story. If we annotate the methods as being unstable, is that good enough?
- @sffc As a principle, I think we need to be more strict with FFI stability even than Rust stability. We don't have features to help us, we don't have semver to help us.

sffc:
See some more discussion on https://github.com/unicode-org/icu4x/issues/4689

sffc (approved):
OK, I still hope we can solve this in Diplomat itself but this is enough for me to approve the PR

PR #6744, opened 2025-07-10
Handle metazones that are not in `metazoneIds`

robertbastian:
Some metazones are not listed in the [`metazoneIds`](https://github.com/unicode-org/cldr/blob/1603ab764ed64c5d2de59380e8d84cbff3fcd485/common/supplemental/metaZones.xml#L1867-L2027). While this is a CLDR bug that should be fixed, we don't need a metazone to be listed there for it to exist.


provider/source/tests/data/cldr/cldr-core/supplemental/metaZones.json
```diff
@@ -86,12 +86,6 @@
               }
             ],
             "Bissau": [
-              {
-                "usesMetazone": {
-                  "_mzone": "Africa_FarWestern",
-                  "_to": "1975-01-01 01:00"
-                }
-              },
```
sffc: Issue: why did this file change?

provider/source/data/debug/timezone/names/generic/long/v1/en.json
```diff
@@ -21,23 +21,23 @@
...
     "63": "Greenland Time",
-    "64": "West Greenland Time",
-    "68": "Hawaii-Aleutian Time",
-    "78": "Irkutsk Time",
-    "85": "Korean Time",
-    "87": "Krasnoyarsk Time",
-    "91": "Lord Howe Time",
-    "100": "Mexican Pacific Time",
+    "64": "East Greenland Time",
+    "65": "West Greenland Time",
```
sffc: Question: why did the order change? I expected for the numbers to change, but for the entries to otherwise be in the same order as before.

provider/source/tests/data/cldr/cldr-core/supplemental/metaZones.json
robertbastian: I changed this to confirm that my change works but forgot to revert

provider/source/data/debug/timezone/names/generic/long/v1/en.json
robertbastian: The ids used to be sorted by short name, but now that we don't read the `metazoneIds` list anymore but instead implicitly construct the metazone set from the zones that are actually used, we only have the long names available. That's part of the point of this PR, as some metazones are missing short names.

The first commit changes this from short names to long names so the diff in the second commit is smaller.

sffc (approved):

PR #6751, opened 2025-07-15
Model Namibia as Central Africa Time

robertbastian:
Namibia did a biannual switch between Central Africa Time (summer) and West Africa Time (winter) for some decades. As TZDB models this using a rule, ICU/CLDR interpret it as a DST, which gives confusing display names: "West Africa Time" during the winter (correct), and "West Africa Summer Time" during the summer (uh what). "West Africa Summer Time" is not a name that I can find any references for outside of CLDR, because Namibia considered itself on Central Africa Time and used a _winter time_ (West Africa Time). TZDB uses "WAT" and "CAT" names.

With #6749 we can now specify an override rule to model this as base timezone changes, and, importantly, metazone changes. This way the display names can actually be "West Africa Time" and "Central Africa Time". And, because this was the only zone to use the West Africa Summer Time name, this removes it from data, saving 7.5kB for default data. The additional metazone transitions are only a 260 bytes and will go away completely in #6750. 

```
Africa/Windhoek  Sun Apr  3 00:00:00 2011 UT = Sun Apr  3 01:00:00 2011 WAT isdst=? gmtoff=3600
Africa/Windhoek  Sun Sep  4 01:00:00 2011 UT = Sun Sep  4 03:00:00 2011 CAT isdst=? gmtoff=7200
Africa/Windhoek  Sun Apr  1 00:00:00 2012 UT = Sun Apr  1 01:00:00 2012 WAT isdst=? gmtoff=3600
Africa/Windhoek  Sun Sep  2 01:00:00 2012 UT = Sun Sep  2 03:00:00 2012 CAT isdst=? gmtoff=7200
Africa/Windhoek  Sun Apr  7 00:00:00 2013 UT = Sun Apr  7 01:00:00 2013 WAT isdst=? gmtoff=3600
Africa/Windhoek  Sun Sep  1 01:00:00 2013 UT = Sun Sep  1 03:00:00 2013 CAT isdst=? gmtoff=7200
Africa/Windhoek  Sun Apr  6 00:00:00 2014 UT = Sun Apr  6 01:00:00 2014 WAT isdst=? gmtoff=3600
Africa/Windhoek  Sun Sep  7 01:00:00 2014 UT = Sun Sep  7 03:00:00 2014 CAT isdst=? gmtoff=7200
Africa/Windhoek  Sun Apr  5 00:00:00 2015 UT = Sun Apr  5 01:00:00 2015 WAT isdst=? gmtoff=3600
Africa/Windhoek  Sun Sep  6 01:00:00 2015 UT = Sun Sep  6 03:00:00 2015 CAT isdst=? gmtoff=7200
Africa/Windhoek  Sun Apr  3 00:00:00 2016 UT = Sun Apr  3 01:00:00 2016 WAT isdst=? gmtoff=3600
Africa/Windhoek  Sun Sep  4 01:00:00 2016 UT = Sun Sep  4 03:00:00 2016 CAT isdst=? gmtoff=7200
Africa/Windhoek  Sun Apr  2 00:00:00 2017 UT = Sun Apr  2 01:00:00 2017 WAT isdst=? gmtoff=3600
Africa/Windhoek  Sun Sep  3 01:00:00 2017 UT = Sun Sep  3 03:00:00 2017 CAT isdst=? gmtoff=7200
```
(I blanked out the `isdst` flag because it differs under rearguard so it's useless)

Based on #6755


sffc:
Do you have a citation that "Central Africa Time" is the preferred display name for the summer time in this time zone?

I'm not super excited about using the metazone mechanism for this, since it definitely is a DST rule. It looks like they aren't doing the time switch any more, but what if they start it again? The metazone table is not designed to handle routine transitions.

In the short term, I would rather it fall back to something like Specific Location ("Namibia Summer Time").

In the longer term, it seems like we keep being held back by the way CLDR stores these names. We want to use the name of one zone as one of the names of some other zone, but without changing the metazone. We should perhaps add an attribute to CLDR that allows metazone names to do horizontal inheritance from some other metazone, and then change the ICU4X data layout to store it efficiently.

robertbastian:
TZDB has a couple sources: https://github.com/eggert/tz/blob/be764d7bd3d6fd00fefdaf9027d9886b1e5cb741/africa#L1124-L1171

Wikipedia also: https://en.wikipedia.org/wiki/Time_in_Namibia#1994%E2%80%932017

sffc:
I feel like for both this and Morocco, what we want is data that is something like

| Metazone Name | Offsets | Generic | Standard | Daylight |
|---|---|---|---|---|
| Europe_Western | +0/+1 | Western European Time | Western European Winter Time | Western European Summer Time |
| Europe_Central | +1/+2 | Central European Time | Central European Winter Time | Central European Summer Time |
| Greenwich_Mean_Time | +0 | (no DST: use standard name) | Greenwich Mean Time | - |
| Africa_Western | +1 | (no DST: use standard name) | West Africa Time | - |
| Africa_Central | +2 | (no DST: use standard name) | Central Africa Time | - |
| United_Kingdom | +0/+1 | (generic location) | (inherit from Greenwich_Mean_Time) | British Summer Time |
| Morocco | +0/+1 | (generic location) | (inherit from Greenwich_Mean_Time) | (inherit from Europe_Central) |
| Namibia | +1/+2 | (generic location) | (inherit from Africa_Western) | (inherit from Africa_Central) |

When the specific name is inherited, it should use the one with the correct UTC offset. For example, Morocco Daylight inherits from Europe_Central, but it should use the Standard name in order for the offsets to match.

robertbastian:
The problem here is that metazones don't have fixed offsets. Some do, maybe we need to split these off, but many are just a country's zone and change offsets.

sffc:
I think it's fine; the name inherits based on the metazone offsets at the requested ZoneNameTimestamp.

robertbastian:
> based on the metazone offsets at the requested ZoneNameTimestamp.

Metazones don't have offsets. We could look at the golden zone inside the metazone, but then we need to start shipping that data as well.

PR #6755, opened 2025-07-17
Disambiguate inconsistent metazones

robertbastian:
This PR introduces new logic to discriminate DST differences within metazones for the generic non-location format. It extends the time zone periods data struct to keep track of which timezone is the "golden" zone for the metazone, and then checks whether the offsets used by the zone match the golden zone. There are two directions where they could disagree:
* The golden zone uses DST, the zone doesn't (e.g. `America/Phoenix` in `America_Mountain`, `Australia/Brisbane` in `Australia_Eastern`, etc.). In that case we use (\*drumm roll\*) generic partial location ("Mountain Time (Phoenix)", "Australian Eastern Time (Brisbane)"). I would actually prefer to use the standard name here ("Mountain Standard Time", "Australian Eastern Standard Time"), but those aren't currently loaded for the generic non-location format (they are for some zones, but only those that don't have a generic name, https://github.com/unicode-org/icu4x/pull/6115). It should be noted that this is not how the generic partial location format is defined in LDML: we only disambiguate if the offsets that are in use differ, LDML disambiguates *everything* that's not the golden zone for the user's locale (each region has it's own golden zone).
* The golden zone doesn't use DST, but the zone does (e.g. `Europe/London` in the `GMT` metazone). In that case, we fall back to the location format ("UK Time"), because, unlike in the first case, during DST the zone is on a time that the metazone never uses.

This also introduces a case for the specific non-location format:
* The golden zone doesn't use DST, but the daylight name is requested. Even if CLDR has a daylight name here (I think only Namibia right now), this is probably a CLDR invention, so don't use it and fall back to the specific location format ("Namibia Daylight Time")


components/datetime/src/format/time_zone.rs
```diff
@@ -156,7 +170,26 @@ impl FormatTimeZone for GenericNonLocationFormat {
             return Ok(Err(FormatTimeZoneError::Fallback));
         };
 
-        sink.write_str(name)?;
+        if offsets.daylight.is_none() && golden_offsets.daylight.is_some() {
+            // The MZ's golden uses daylight but we don't (e.g. Phoenix, Regina, Algiers,
+            // Brisbane).
+            // Disambiguate using the location.
+            // TODO: Use the standard name here
```
sffc:
```suggestion
            // TODO: Should we use the standard name here?
```

components/datetime/src/format/time_zone.rs
```diff
@@ -219,7 +252,18 @@ impl FormatTimeZone for SpecificNonLocationFormat {
...
+            return Ok(Err(FormatTimeZoneError::Fallback));
+        };
+
+        // Metazone names are defined for the offsets of the golden zone,
+        // so recalculate the variant just in case.
+        let variant = if offset == golden_offsets.standard {
+            TimeZoneVariant::Standard
+        } else if Some(offset) == golden_offsets.daylight {
+            TimeZoneVariant::Daylight
+        } else {
```
sffc: Question: What does this do differently than before?

components/datetime/tests/patterns/tests/time_zones.json
```diff
@@ -182,16 +221,16 @@
     "locale": "en",
     "datetime": "2013-08-15T12:00:00+01[Africa/Windhoek]",
     "expectations": {
-      "vvvv": "West Africa Time",
+      "vvvv": "Namibia Time",
       "zzzz": "West Africa Standard Time"
```
sffc, line 225: Question: should the Standard pattern for this metazone be just West Africa Time?

components/time/src/provider/mod.rs
```diff
@@ -290,6 +296,7 @@ struct TimeZonePeriodsSerde<'a> {
     >,
 
     pub offsets: ZeroVec<'a, VariantOffsets>,
+    pub goldens: ZeroVec<'a, u16>,
```
sffc: Question: Is this the only vec we currently have with metazones as the indices? Is there something else we can merge this with for better packing?

sffc (dismissed):
Praise: Nice work

components/datetime/src/format/time_zone.rs
robertbastian: In case a zone is not in the correct metazone (i.e. buggy CLDR), this only uses the metazone names that match the golden (i.e. metazone) offset. I don't know of any such cases, but I wouldn't be surprised if some exist.

components/datetime/tests/patterns/tests/time_zones.json
robertbastian: Yes it should. There are dozens of zones that don't observe DST but that have different generic and specific names, I don't know what the point of that is.

components/time/src/provider/mod.rs
robertbastian: yes it is

components/time/src/provider/mod.rs
```diff
@@ -69,42 +69,125 @@ impl AsULE for VariantOffsets {
     type ULE = [i8; 2];
 
     fn from_unaligned([std, dst]: Self::ULE) -> Self {
+        fn decode(encoded: i8) -> i32 {
```
sffc: Issue: It is confusing that now you have both `VariantOffsets` and `VariantOffsetsWithFlags` that both implement AsULE. Please delete `impl AsULE for VariantOffsets`. I don't think you are using it anywhere, but if you are, then just use `VariantOffsetsWithFlags`, I think.

components/time/src/provider/mod.rs
```diff
@@ -69,42 +69,125 @@ impl AsULE for VariantOffsets {
...
+/// two are used.
+#[derive(Debug, Clone, Copy, PartialEq, PartialOrd, Eq, Ord)]
+#[non_exhaustive]
+pub struct VariantOffsetsWithFlags {
+    /// The offsets
+    pub offsets: VariantOffsets,
+    /// See [`MetazoneInfo`]
+    pub uses_non_golden_variant: bool,
+    /// See [`MetazoneInfo`]
+    pub uses_custom_transitions: bool,
```
sffc: These are hard things to name. Suggestions:

- `has_extra_offsets` + `has_custom_transitions`
- `uses_non_golden_offsets` + `uses_non_golden_transitions`

I don't love the name `uses_non_golden_variant` because the term "variant" is one we are trying to phase out, but I won't object to it. I do think that we should be consistent between `non_golden` or `custom`.

Also I'm not sure if these are really two different flags. It seems like an enum:

```rust
enum MetazoneDivergenceKind {
    ExactMatch,
    ExtraOffsets,
    CustomTransitions,
}
```

I'd also be okay with naming these variants by what they do rather than what they are:

```rust
/// Defines what format this time zone should use in the generic zone style.
enum GenericNameType {
    /// Use the generic non-location name from the metazone, such as "Central European Time"
    NonLocation,
    /// Use the generic partial location name, such as "Central European Time (Algiers)"
    PartialLocation,
    /// Use the location name instead of the metazone name, such as "Ireland Time"
    Location,
}
```

I kind-of think that last model is cleanest, and then it allows datagen to figure out on its own what conditions to use to put each time zone in each bucket.

sffc (changes requested):

components/time/src/provider/mod.rs
robertbastian: VariantOffsets is on the VariantOffsetsCalculator API, and its AsULE implementation needs to stay stable because it's a public impl of a stable type, and because it's used by the legacy struct.

components/time/src/provider/mod.rs
robertbastian: I specifically don't want to use the last model, because while I'm using generic partial location for custom transitions now, I want to use the specific names in the future if there are no transitions.

components/time/src/provider/mod.rs
sffc: ok... please make sure this is covered in the legacy test, and add a comment clarifying that this custom ULE impl is not used in currently stable data structs, with a TODO to remove it in 3.0. Also consider marking the impl as deprecated.

components/time/src/provider/mod.rs
robertbastian: Also I much prefer "variants" over "offsets". We're phasing out the variant term on the public API for formattable time zones, but it's still everywhere in our data model as well as on VariantOffsetsCalculator.

sffc: ok, then I think use `MetazoneDivergenceKind`? And we essentially maintain a mapping from that enum to which behavior to use for generic names. It could even become a runtime option.

components/time/src/provider/mod.rs
robertbastian: Trait impls can't be deprecated and impl docs are not very discoverable, but I'll add a comment.

components/datetime/src/format/time_zone.rs
robertbastian: now checking this in datagen, currently this logs some errors which I'll fix in the next PR

components/time/src/provider/mod.rs
```diff
@@ -65,46 +65,90 @@ pub const MARKERS: &[DataMarkerInfo] = &[
...
+/// A [`VariantOffsets`] and a [`MetazoneMembershipKind`] packed into one byte.
+#[derive(Debug, Clone, Copy, PartialEq, PartialOrd, Eq, Ord)]
+pub struct VariantOffsetsWithMetazoneMembershipKind {
+    /// The offsets. Currently uses 3 bits.
+    pub offsets: VariantOffsets,
+    /// Metazone membership metadata. Currently uses 2 bits.
+    pub mzmsk: MetazoneMembershipKind,
+}
+
+impl From<VariantOffsets> for VariantOffsetsWithMetazoneMembershipKind {
```
sffc: Question: why do we need this impl? Seems like a footgun

robertbastian: ah this was needed when `VariantOffsetsWithFlags` was non-exhaustive, removed

PR #6761, opened 2025-07-21
Simplify `ZoneNameTimestamp` code

robertbastian:


components/time/src/zone/zone_name_timestamp.rs
```diff
@@ -187,13 +103,11 @@ impl ZoneNameTimestamp {
...
     pub fn from_zoned_date_time_iso(zoned_date_time: ZonedDateTime<Iso, UtcOffset>) -> Self {
-        let metadata = 0; // currently unused (reserved)
-        let ZonedDateTime { date, time, zone } = zoned_date_time;
-        let parts = ZoneNameTimestampParts::from_saturating_date_time_with_metadata(
-            DateTime { date, time },
-            metadata,
-        );
-        Self::from_parts(parts).adjust_zone(zone, UtcOffset::zero())
+        let qh = zoned_date_time.to_epoch_milliseconds_utc() / 1000 / 60 / 15;
+        let qh_clamped = qh.clamp(Self::far_in_past().0 as i64, Self::far_in_future().0 as i64);
```
sffc, line 107: Suggestion: this should be using `div_euclid`, but it technically doesn't matter since negative numbers go away with the `.clamp`. Still, it's good practice to always use `div_euclid` when dealing with time values.

components/time/src/zone/zone_name_timestamp.rs
```diff
@@ -213,66 +131,28 @@ impl ZoneNameTimestamp {
...
-            metadata,
-        }
-    }
-
-    fn from_parts(parts: ZoneNameTimestampParts) -> Self {
-        let metadata_shifted = (parts.metadata as u32) << 24;
-        debug_assert!(parts.quarter_hours_since_local_unix_epoch <= 0x00FFFFFF);
-        let qh_masked = parts.quarter_hours_since_local_unix_epoch & 0x00FFFFFF;
-        Self(metadata_shifted | qh_masked)
+        Self(0xFFFFFF)
```
sffc, line 149:
```suggestion
        // The top byte is reserved for future use, such as:
        // - A time zone UTC offset
        // - Extra bits for the epoch quarter-hours
        // - Bitmask to use the epoch quarter hour bits more efficiently
        Self(0xFFFFFF)
```

components/time/src/zone/zone_name_timestamp.rs
```diff
@@ -432,18 +312,4 @@ mod test {
...
-        let znt = ZoneNameTimestamp::from_unaligned(raw);
-        let roundtrip_znt =
-            ZoneNameTimestamp::from_zoned_date_time_iso(znt.to_zoned_date_time_iso());
-        let roundtrip_raw = roundtrip_znt.to_unaligned();
-
-        // [0..3] is the datetime. [3] is the metadata.
-        assert_eq!(raw.0[0..3], roundtrip_raw.0[0..3]);
-        assert_eq!(raw.0[3], 0x12);
-        assert_eq!(roundtrip_raw.0[3], 0);
-    }
```
sffc, line 448: Nit: This is still a valid test and it should be unaffected by your work

components/time/src/zone/zone_name_timestamp.rs
robertbastian: you only need to use Euclid when you decompose into multiple parts and you need the trailing parts to be positive. this is a rounding operation so division is the correct operation. it is also faster because it is a machine operation

components/time/src/zone/zone_name_timestamp.rs
robertbastian: the new struct only uses 24 bits which is why I'm removing all the unused metadata machinery

components/time/src/zone/zone_name_timestamp.rs
robertbastian: ditto

PR #6765, opened 2025-07-22
Encode DST offsets by lookup table

robertbastian:
#6238

I'm also reserving a value for a relative DST offset of zero, this might be useful if we want to add zones with permanent DST to metazones. For example, `America/Punta_Arenas` could be year-round "Chile Summer Time" if we set the offsets to (-03, -03).


robertbastian:
I'm reverting this in #6755 as the ULE implementation is used by the legacy struct as well. The test didn't catch this because it doesn't check a zone with DST

PR #6767, opened 2025-07-23
Fix some offsets and metazones

robertbastian:
Based on #6770.


provider/source/data/metaZonesPatched.json
```diff
@@ -127,6 +427,34 @@
...
+                  "_from": "1984-10-31 23:00",
+                  "_to": "2016-09-06 21:00"
+                }
+              },
+              {
+                "usesMetazone": {
+                  "_mzone": "Turkey",
+                  "_from": "2016-09-06 21:00"
+                }
+              }
```
sffc, line 457: Thought: We would save a lot of data if we mapped Turkey, Syria, Jordan, Saudi Arabia, and others all to the same metazone, maybe "Middle_East" or something. Maybe even throw in Moscow. :man_shrugging:

sffc (approved):

sffc (approved):

PR #6770, opened 2025-07-24
Move TZDB compatibility data to CLDR patch

robertbastian:
For now I'm forking the `metaZones.json` file in-repo, hoping I can get this into CLDR 48.

This adds the ability to add tzdb-abbreviation -> variant mappings to metazone periods. This can be used if the `isdst` flag is ambiguous (i.e. reaguard and main don't agree, datagen checks that there are overrides for these), or if we believe that the modeling as done in TZDB does not fit our metazone model (for example `Asia/Samarkand` has a winter where it stayed on DST, which TZDB models as a base offset change, but we can keep it in its metazone if we model it as a DST period).

This PR moves the current in-code overrides to the metazone data.

This approach, however, does not allow adding these overrides to periods where a zone is not in a metazone, so the offsets for Morocco will change from (+0, +1) (current code override) to (+1, +0). This is somewhat fine, because the variant doesn't have much meaning outside a metazone. I say somewhat, because in [CLDR-18074](https://unicode-org.atlassian.net/browse/CLDR-18074) I changed LDML to use the Specific Location format as a fallback from the Specific Non-Location format; I now consider this to have been a mistake and will try to revert it (I believe ICU4C has not implemented this anyway). 

#6767 uses this mechanism for more cleanups.

### DIFF
--- a/provider/source/src/time_zones/mod.rs
+++ b/provider/source/src/time_zones/mod.rs
@@ -236,7 +236,7 @@ impl SourceDataProvider {
                         offsets_and_metazones.entry(tz).or_default().push((
                             start,
                             os,
-                            curr_mz.and_then(|mz| mz.mzone.as_ref().map(|m| m.as_str())),
+                            curr_mz.and_then(|mz| mz.mzone.as_deref()),
                         ));
 
                         match (offsets.peek().as_ref(), mzs.peek().copied()) {


sffc:
It seems like there is a fair bit of novelty going on with the proposed JSON file changes; not just data fixes but new structure:

Screen Shot 2025-07-24 at 9 25 57 AM

I still don't think it is necessarily the job of this level of CLDR data to put a bandaid over the rearguard/vanguard thing. I think datagen, as a client of CLDR, should just use rearguard tzdb when generating its data, which we can do! So I don't see how this proposal benefits ICU4X.

If you want to move the code overrides to a data file, that seems fine, but it should be a small overrides file, not the whole big metaZones.json with less than 1% of the lines with a diff.

robertbastian:
> I still don't think it is necessarily the job of this level of CLDR data to put a bandaid over the rearguard/vanguard thing

It's not just a rearguard/vanguard thing, this lets CLDR decide which periods are considered DST. Currently TZDB does not necessarily always make sense, for example there's only half a year since the epoch where Morocco is on +1 and it's *not* marked DST, and for that half a year CLDR has to move it to a different metazone. See #6767, #6772, #6751.

This data can be used to make ICU agnostic to rearguard/vanguard, which is a big blocker for it using system data.

> datagen, as a client of CLDR, should just use rearguard tzdb when generating its data

Rearguard is a bandaid and TZDB could decide to drop it tomorrow because we've now had 6 years to migrate.

sffc (commented):
This PR is more than "Move TZDB compatibility data to CLDR patch" because it is adding -3600 as a DST offset.

I thought we had agreed to add -3600 as a new variant type but not "daylight".

robertbastian:
Yes that is true. The reason that I had to do this was because this model only allows variant overrides inside metazones, but Morocco is not currently in a metazone. If we create a `Morocco_Ramadan` metazone, we can add the `"+00": "ramadan"` override there.

The only thing that changes here is that the `VariantOffsetsCalculator` will return `standard: +01, daylight: +00` (previously `standard: +00, daylight: +01`). There is no change to datetime formatting since #6771.

sffc (approved):
Thank you!

PR #6772, opened 2025-07-24
Model `America/Punta_Arenas` as permanent DST in Chile metazone

robertbastian:
Based on #6770


components/time/src/zone/offset.rs
```diff
@@ -377,6 +377,12 @@ impl VariantOffsets {
             daylight: None,
         }
     }
+
+    /// Whether this offset set is considered permanent DST, i.e. standard and daylight
+    /// offsets are the same.
+    pub fn is_permanent_dst(self) -> bool {
```
sffc: Suggestions:

- `is_year_round_dst()`
- `is_year_round_daylight_time()`
- `uses_daylight_time_year_round()`
- `daylight_is_standard()`

Not really a fan of "permanent DST" because it is only "permanent" for as long as the metazone membership period, which isn't a external-facing concept.

Or, it's not clear you need this. You only use it in datagen. Add it if you think non-ICU4X users of VariantOffsets would find it useful.

provider/source/src/time_zones/mod.rs
```diff
@@ -291,13 +304,15 @@ impl SourceDataProvider {
...
                                     log::warn!("Offsets don't agree with metazone golden: {tz:?} - {golden:?}");
                                 }
 
                                 let kind = if os.daylight.is_some() && golden_os.daylight.is_none() {
                                     MetazoneMembershipKind::CustomVariants
-                                } else if os.daylight.is_none() && golden_os.daylight.is_some() {
+                                } else if os.daylight.is_none() && golden_os.daylight.is_some() || os.is_permanent_dst() {
                                     // TODO: this needs to look at actual transitions
                                     MetazoneMembershipKind::CustomTransitions
                                 } else {
```
sffc: Nit: I prefer keeping the two cases separate

```suggestion
                                } else if os.daylight.is_none() && golden_os.daylight.is_some() {
                                    // TODO: this needs to look at actual transitions
                                    MetazoneMembershipKind::CustomTransitions
                                } else if os.daylight.is_some() && os.daylight == os.standard {
                                    // ...
                                } else {
```

provider/source/src/time_zones/mod.rs
```diff
@@ -291,13 +304,15 @@ impl SourceDataProvider {
                                     .unwrap()
                                     .1;
 
-                                if golden_os.standard != os.standard {
+                                if (!os.is_permanent_dst() && golden_os.standard != os.standard)
+                                    || (os.is_permanent_dst() && golden_os.daylight != os.daylight)
```
sffc:
```suggestion
                                if golden_os.standard != os.standard || (os.standard == os.daylight && golden_os.daylight != os.daylight) {
```

provider/source/src/time_zones/mod.rs
robertbastian: no, this is a single case, the transitions differ. the current check is an approximation to "transitions differ", hence the TODO

provider/source/src/time_zones/mod.rs
robertbastian: that is not the same, `golden_os.standard` and `os.standard` don't match for permanent DST zones, but that's allowed as long as the daylights match

components/datetime/src/format/time_zone.rs
```diff
@@ -167,8 +167,8 @@ impl FormatTimeZone for GenericNonLocationFormat {
 
         if mz.kind == MetazoneMembershipKind::CustomTransitions {
             // Disambiguate using the location.
-            // TODO: Use the standard name here if zone only uses that
-            // (= has no transitions = `offsets.daylight.is_none()`).
+            // TODO: Use the specific name here if zone only uses that
+            // (= has no transitions = `offsets.daylight.is_none() || offsets.is_permanent_dst()`).
```
sffc: No more is permanent dst

robertbastian: thought I used this somewhere else

sffc (approved):

Issue #1941, opened 2022-05-25
Figure out what to do about Korean search collations


hsivonen:
Analyze the Korean search collations to decide if it makes sense to implement them as preprocessing of the character stream instead of different collation data. If the solution is to preprocess the character stream, consider optimizing the size of the jamo lookup table by storing only the part of CE32 that varies between the entries.

hsivonen:
@jungshik Does there exist a document that explains the semantics of searchjl? Specifically, is there documentation for the purpose of the this [prefix matching part](https://github.com/unicode-org/icu/blob/64b35481263ac4df37a28a9c549553ecc9710db2/icu4c/source/data/coll/ko.txt#L369-L379)?

PR #6778, opened 2025-07-28
Discuss supported options in CollationType docs

sffc:
It is not clear from `icu_collator` that search rules are not supported; they just silently fall back to the standard rules. I'm adding a link to the function in `icu_provider_export` where they can be added.

### DIFF
--- a/components/locale_core/src/preferences/extensions/unicode/keywords/collation.rs
+++ b/components/locale_core/src/preferences/extensions/unicode/keywords/collation.rs
@@ -35,15 +35,15 @@ enum_keyword!(
         ("dict" => Dict),
         /// The default Unicode collation element table order
         /// 
-        /// 🛈 `icu_collator` does not support the `ducet` ordering. See [`CollationType`].
+        /// 🛈 `icu::collator` does not support the `ducet` ordering. See [`CollationType`].
         ("ducet" => Ducet),
         /// Recommended ordering for emoji characters
         /// 
-        /// 🛈 `icu_collator` supports this ordering only in the root locale. See [`CollationType`].
+        /// 🛈 `icu::collator` supports this ordering only in the root locale. See [`CollationType`].
         ("emoji" => Emoji),
         /// European ordering rules
         /// 
-        /// 🛈 `icu_collator` supports this ordering only in the root locale. See [`CollationType`].
+        /// 🛈 `icu::collator` supports this ordering only in the root locale. See [`CollationType`].
         ("eor" => Eor),
         /// Phonebook style ordering (such as in German)
         ("phonebk" => Phonebk),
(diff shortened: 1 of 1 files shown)


components/locale_core/src/preferences/extensions/unicode/keywords/collation.rs
```diff
@@ -28,8 +28,16 @@ enum_keyword!(
         /// Pinyin ordering for Latin and for CJK characters (used in Chinese)
         ("pinyin" => Pinyin),
         /// Special collation type for string search
+        ///
+        /// Note: `icu_collator` does not include search collation rules by default.
```
robertbastian:
```suggestion
        /// Note: `icu::collator` does not include search collation rules by default.
```

components/locale_core/src/preferences/extensions/unicode/keywords/collation.rs
```diff
@@ -10,16 +10,40 @@ enum_keyword!(
     /// A Unicode Collation Identifier defines a type of collation (sort order).
     ///
     /// The valid values are listed in [LDML](https://unicode.org/reports/tr35/#UnicodeCollationIdentifier).
+    /// 
+    /// # Supported Values
+    /// 
+    /// Not all collation orderings are always supported by ICU4X:
```
robertbastian, line 16:
```suggestion
    /// Not all collation orderings are always supported:
```

components/locale_core/src/preferences/extensions/unicode/keywords/collation.rs
```diff
@@ -10,16 +10,40 @@ enum_keyword!(
...
+    /// For more information on how to add more orderings to a custom data build, see
+    /// [`ExportDriver::with_additional_collations`](https://docs.rs/icu_provider_export/latest/icu_provider_export/struct.ExportDriver.html#method.with_additional_collations).
     CollationType {
         /// A previous version of the ordering, for compatibility
         ("compat" => Compat),
         /// Dictionary style ordering (such as in Sinhala)
         ("dict" => Dict),
         /// The default Unicode collation element table order
+        /// 
+        /// 🛈 `icu::collator` does not support the `ducet` ordering. See [`CollationType`].
```
robertbastian, line 38: 🛈 is tofu for me

hsivonen: For me, too, on Apple platforms. The character seems appropriate for this purpose, though.

components/locale_core/src/preferences/extensions/unicode/keywords/collation.rs
```diff
@@ -28,8 +52,12 @@ enum_keyword!(
         /// Pinyin ordering for Latin and for CJK characters (used in Chinese)
         ("pinyin" => Pinyin),
         /// Special collation type for string search
+        ///
+        /// 🛈 `icu::collator` does not include search collation rules by default. See [`CollationType`].
         ("search" => Search),
         /// Special collation type for Korean initial consonant search
+        ///
+        /// 🛈 `icu::collator` does not include search collation rules by default. See [`CollationType`].
```
hsivonen, line 60: Also, even if built by datagen, the code doesn't support the data shape for this one. See #1941.

Please change this one to say something along the lines of "ICU4X does not support Korean search collations."

components/locale_core/src/preferences/extensions/unicode/keywords/collation.rs
robertbastian: I'm sure we can find an equivalent character that is more widely supported.

PR #6779, opened 2025-07-28
Remove Western Africa Time variant names

robertbastian:


components/datetime/tests/patterns/tests/time_zones.json
```diff
@@ -243,15 +243,15 @@
...
+      "zzzz": "West Africa Time"
     }
   },
   {
     "locale": "en",
     "datetime": "2013-09-15T12:00:00+02[Africa/Windhoek]",
     "expectations": {
       "vvvv": "Namibia Time",
-      "zzzz": "Namibia Daylight Time"
+      "zzzz": "GMT+02:00"
```
sffc, line 254: Please ensure there is a CLDR issue open to investigate the correct name for this historical specific zone.

provider/source/src/cldr_serde/time_zones/time_zone_names.rs
```diff
@@ -33,9 +33,31 @@ impl Metazone {
     }
 }
 
-#[derive(PartialEq, Debug, Clone, Deserialize)]
+#[derive(PartialEq, Debug, Clone)]
 pub(crate) struct Metazones(pub(crate) BTreeMap<String, Metazone>);
 
+// TODO(CLDR-18858): upstream
+impl<'de> Deserialize<'de> for Metazones {
```
sffc, line 40: Thought: Kinda strange place to put this; you have so much other customization logic in the time zone transform code. Why not there?

sffc (approved):
I think it's a better way to resolve this issue than your previous proposal, but it's definitely one we should prioritize upstreaming, since this is patching actual existing data, not just augmenting it.

components/datetime/tests/patterns/tests/time_zones.json
robertbastian: like CLDR-18858?

robertbastian: wait what do you mean "investigate the correct name"? i thought you were aligned on using localized offset? there is nowhere to specify any other name

provider/source/src/cldr_serde/time_zones/time_zone_names.rs
robertbastian: the transform code needs this in three different places (generic, standard, and specific data structs).

PR #6801, opened 2025-08-05
Temporarily commenting out unused units in the new CLDR data.

younies:
# Description

Temporarily commenting out unused units in the new CLDR data.

This is just until the main data is merged in [PR #6793](https://github.com/unicode-org/icu4x/pull/6793). Once that's done, we'll merge the new test data from common/testData/units/unitsTest.txt.


Manishearth (approved):
Why do these need to be commented?

robertbastian:
Yeah we don't need this, I'll just update the test file in the CLDR 48 PR.

younies:
@Manishearth & @robertbastian :  It is okay, but for me I prefer to comment them and then update the test cases in a different PR. but it is okay this way too.

robertbastian:
There should be a single PR that updates data to CLDR 48 and updates all test cases.

PR #6806, opened 2025-08-06
Fix nightly doc

robertbastian:


Manishearth (approved):

sffc (commented):
Why did it break?

robertbastian:
because nightly?

sffc:
Why did the crate `icu4x-datagen` stop resolving when it previously resolved just fine? Is it a bug we should report upstream?

Issue #6811, opened 2025-08-06
ICU Conformance: collation tests fail with locale root and strings containing U+FFFE


sven-oly:
[ICU4X collation 2.0.0](https://unicode-org.github.io/conformance/rust/icu77/collation/verifier_test_report.html) shows 5 tests failing (only!)
Note also that these tests pass in ICU4C, ICU4J, and NodeJS implementations of collation.

Each of these indicates locale 'root' and the test strings contain U+FFFE in combination with other characters. This seems to be an error in ICU4X collation. Note that earlier versions of ICU4X (1.3, 1.4, etc.) also have this set of failures.

The source for the collation tests is this file [testgen/icu77/collationtest.txt](https://github.com/unicode-org/conformance/blob/main/testgen/icu77/collationtest.txt), starting at line 360. This test file is used by ICU4C internal testing.

```
** test: U+FFFE on identical level
@ root
% strength=identical
* compare
# All of these control codes are completely-ignorable, so that
# their low code points are compared with the merge separator.
# The merge separator must compare less than any other character.
<1 \uFFFE\u0001\u0002\u0003
<i \u0001\uFFFE\u0002\u0003
<i \u0001\u0002\uFFFE\u0003
<i \u0001\u0002\u0003\uFFFE

* compare
# The merge separator must even compare less than U+0000.
<1 \uFFFE\u0000\u0000
<i \u0000\uFFFE\u0000
<i \u0000\u0000\uFFFE
```

That is processed by [collation.py](https://github.com/unicode-org/conformance/blob/main/testgen/generators/collation.py), producing a JSON file. Here are the JSON data for these 5 failures with labels "00160" to "00165" below. Note that "00163" passes, but that one ijust compares with s1 which is the empty string.

> 
> {
>  "test_type": "collation",
>  "tests": [
> ...
>  {
>    "compare_type": "<i",
>    "s1": "\ufffe\u0001\u0002\u0003",
>    "s2": "\u0001\ufffe\u0002\u0003",
>    "source_file": "collationtest.txt",
>    "line": 367,
>    "label": "00160",
>    "locale": "root",
>    "test_description": "U+FFFE on identical level",
>    "strength": "identical",
>    "hexhash": "8a6ec2612f6466afc149ca9b899ae40ab1a9dcf4"
>   },
>   {
>    "compare_type": "<i",
>    "s1": "\u0001\ufffe\u0002\u0003",
>    "s2": "\u0001\u0002\ufffe\u0003",
[...]

PR #6814, opened 2025-08-07
Make merge separator compare less than U+0000 on the identical strength

hsivonen:
See #6811.
    
This change does not fix the issue for sort keys.

Includes the test changeset from #6812.

### DIFF
--- a/components/collator/src/comparison.rs
+++ b/components/collator/src/comparison.rs
@@ -724,8 +724,8 @@ macro_rules! compare {
             }
             let ret = self.compare_impl(left_tail.$left_to_iter(), right_tail.$right_to_iter(), head.$left_to_iter().rev());
             if self.options.strength() == Strength::Identical && ret == Ordering::Equal {
-                return Decomposition::new(left_tail.$left_to_iter(), self.decompositions, self.tables).map(|c| if c != '\u{FFFE}' { c as i32 } else { -1i32 }).cmp(
-                    Decomposition::new(right_tail.$right_to_iter(), self.decompositions, self.tables).map(|c| if c != '\u{FFFE}' { c as i32 } else { -1i32 }),
+                return Decomposition::new(left_tail.$left_to_iter(), self.decompositions, self.tables).map(|c| if c != MERGE_SEPARATOR { c as i32 } else { -1i32 }).cmp(
+                    Decomposition::new(right_tail.$right_to_iter(), self.decompositions, self.tables).map(|c| if c != MERGE_SEPARATOR { c as i32 } else { -1i32 }),
                 );
             }
             ret


components/collator/src/comparison.rs
```diff
@@ -694,8 +694,8 @@ macro_rules! compare {
             }
             let ret = self.compare_impl(left_tail.$left_to_iter(), right_tail.$right_to_iter(), head.$left_to_iter().rev());
             if self.options.strength() == Strength::Identical && ret == Ordering::Equal {
-                return Decomposition::new(left_tail.$left_to_iter(), self.decompositions, self.tables).cmp(
-                    Decomposition::new(right_tail.$right_to_iter(), self.decompositions, self.tables),
+                return Decomposition::new(left_tail.$left_to_iter(), self.decompositions, self.tables).map(|c| if c != '\u{FFFE}' { c as i32 } else { -1i32 }).cmp(
+                    Decomposition::new(right_tail.$right_to_iter(), self.decompositions, self.tables).map(|c| if c != '\u{FFFE}' { c as i32 } else { -1i32 }),
```
hsivonen: The `if` condition is backwards due to the assumption that absent other information, LLVM treats the first branch as more likely and we _still_ after a decade [aren't allowed to say `unlikely` on non-nightly Rust](https://github.com/rust-lang/rust/issues/136873).

sffc (commented):
Is this the best place to put this logic? It seems un-elegant to put this directly into the main `compare` function macro, which to this point has just delegated to other functions and hasn't had any code-point-specific logic?

Are we sure that U+FFFE is the only code point with this special case? It is the one that the conformance test discovered, but I wonder if it was pointing at a bigger class of issues?

hsivonen:
> Is this the best place to put this logic?

Yes. It needs to go into a `.map()` where the identical level is implemented.

> It seems un-elegant to put this directly into the main `compare` function macro, which to this point has just delegated to other functions and hasn't had any code-point-specific logic?

It makes sense for the relevant iterators to be inside the macro, since the concrete types end up differing per macro instantiation. And, as noted above, the `.map()` needs to go where the iterators are. The `.map()` is tiny.

> Are we sure that U+FFFE is the only code point with this special case? It is the one that the conformance test discovered, but I wonder if it was pointing at a bigger class of issues?

It looks a lot like this isn't a bigger class of issues but a very special case.

In the case of ICU4C sort key generation, only U+FFFE is special-cased: https://searchfox.org/mozilla-central/rev/f6a806c38c459e0e0d797d264ca0e8ad46005105/intl/icu/source/i18n/bocsu.cpp#132

sffc:
I do see U+FFFE specifically called out in the CLDR collator spec, so maybe this really is just a single isolated special case:

https://www.unicode.org/reports/tr35/dev/tr35-collation.html#CLDR_Collation_Algorithm

sffc:
Why is this a 4 line change but https://github.com/unicode-org/icu4x/pull/6823 a much bigger change? 

How come there aren't tests failing right now given an inconsistency in behavior between sort keys and compare?

hsivonen:
> Why is this a 4 line change but #6823 a much bigger change?

Signed 32-bit integers offer a way to easily switch from `char` to something that expands the value space below U+0000, so the fix here is simple.

In the sort key case, the requirement is that `0u8` is reserved for C string compat, `1u8` is reserved for level separator, and `2u8` is the merge separator, so there's a need to put everything else at `3u8` or above. Instead of implementing a new scheme to shift UTF-8 like that, the exact solution from ICU4C was implemented. That solution _also_ compresses the length of the resulting byte sequence, which is where the complexity comes from.

> How come there aren't tests failing right now given an inconsistency in behavior between sort keys and compare?

Because we in general lack test coverage to show consistency between sort keys and compare. It would make sense to fuzz them.

sffc (approved):
I approve but I'd like @echeran to also take a look.

hsivonen:
Thanks. I'm assuming that @echeran is doing a post-landing review. Landing now to avoid merge conflicts showing up and to avoid keeping the tree in a state where regular comparison and sort keys don't agree.

PR #6816, opened 2025-08-07
Jump forward throwing away work instead of panic in release builds

hsivonen:
In the context of #6788, @kornelski suggested on Mastodon getting rid of panics in the hot loop.

This changeset replaces some panics with jumps forward when debug assertions are disabled when jumping forward simply throws away fast-path work. (In the UTF-8 cases, there are still panic sites left, but in those cases state visible outside the fast path has already been modified, so there's no obvious way to jump forward in a manner that clearly couldn't have more severe effects than a panic.)

Unfortunately, at least taken alone, this changeset doesn't make things faster with the fast trie type (it has interesting effects with the small trie type, but anyone who cares about UTF-16 perf should be using the fast trie type; making that the default is out of scope here).

But since this doesn't make things worse, either, and we generally want to avoid generating panic code, I'm opening the PR.

Somewhat distressingly, replacing panics with error returns in the `IsNormalizedSink`s made perf worse! I don't know if this can be generalized for jump forward instead of panic vs. jump to early return instead of panic.

### DIFF
--- a/components/normalizer/src/lib.rs
+++ b/components/normalizer/src/lib.rs
@@ -1808,6 +1808,9 @@ impl<'data> DecomposingNormalizerBorrowed<'data> {
                             // Fast-track succeeded!
                             continue 'fastest;
                         }
+                        // This deliberately isn't panic-free, since the code pattern
+                        // that was OK for the composing counterpart regressed
+                        // English and French performance if done here, too.
                         decomposition.delegate = pending_slice[pending_slice.len() - code_unit_iter.as_slice().len() - 1..].chars();
                         break 'fastest;
                     }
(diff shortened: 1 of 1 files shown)


components/normalizer/src/lib.rs
```diff
@@ -2529,97 +2546,109 @@ impl<'data> ComposingNormalizerBorrowed<'data> {
...
-                    let upcoming = unsafe { char::from_u32_unchecked(upcoming32) };
-                    let upcoming_with_trie_value = CharacterAndTrieValue::new(upcoming, trie_value);
-                    // We need to fall off the fast path.
-                    composition.decomposition.pending = Some(upcoming_with_trie_value);
-                    #[expect(clippy::indexing_slicing)]
-                    let mut consumed_so_far = pending_slice[..pending_slice.len() - code_unit_iter.as_slice().len() - upcoming.len_utf16()].chars();
-                    // `unwrap` OK, because we've previously managed to read the previous character
-                    #[expect(clippy::unwrap_used)]
-                    {
+                        // SAFETY: upcoming32 can no longer be a surrogate.
```
Manishearth, line 2617: nit: please mark all the other comments above around surrogateness with SAFETY, and justify it at every loop exit

hsivonen: Isn't this out of scope for this changeset, since this in pre-existing code that merely gets re-indended?

Manishearth: I didn't know it was just being reindented; it wasn't immediately obvious. I guess that's fine.

In general I would like normalizer code to have far more safety comments than it does, it makes it much harder to review PRs like this.

components/normalizer/src/lib.rs
```diff
@@ -1808,6 +1808,9 @@ impl<'data> DecomposingNormalizerBorrowed<'data> {
                             // Fast-track succeeded!
                             continue 'fastest;
                         }
+                        // This deliberately isn't panic-free, since the code pattern
+                        // that was OK for the composing counterpart regressed
+                        // English and French performance if done here, too.
```
Manishearth, line 1813: Can `.get(...).unwrap_or_default()` or something similar work? We really should avoid panicky code.

hsivonen: Not, AFAICT, in a way that I could reason about not to have worse outcomes than panic if the code ever runs. But this comment doesn't add any panicky code. It just explains why this particular bit (among many) was not changed here.

hsivonen:
Thanks. I added https://github.com/unicode-org/icu4x/issues/6835 about the requested safety comments.

PR #6818, opened 2025-08-07
Update to Rust 1.89

robertbastian:


components/datetime/src/provider/pattern/runtime/pattern.rs
```diff
@@ -123,7 +123,7 @@ impl Pattern<'_> {
         }
     }
 
-    pub(crate) fn as_borrowed(&self) -> PatternBorrowed {
+    pub(crate) fn as_borrowed(&self) -> PatternBorrowed<'_> {
```
sffc, line 126: This is a big flip-flop in terms of clippy. In the past, I had to remove `'_` lifetimes, and now it wants us to add them back. But, I like the new behavior because I think being explicit is better.

robertbastian: it's more nuanced now: https://blog.rust-lang.org/2025/08/07/Rust-1.89.0/#:~:text=Self%2Devident%20it%20has%20a%20lifetime

Issue #6820, opened 2025-08-07
Deprecate time zone variant APIs


robertbastian:
#6754

sffc:
Notes from ICU4X WG:

- @robertbastian I'd like to move the enum to be defined in the provider module and re-exported as deprecated, if that's possible.
- @sffc Seems fine, but we should still have a definition of what these variants mean.
- @robertbastian - if it's in the provider module, it's just a data thing, so the defitions are just "this is the CLDR display names concept"
- @sffc That's fine; we should still say it in the docs. Are we sure that this will be true, like with the Ramadan time?
- @robertbastian Yes, we shouldn't add a sundown variant unless CLDR adds it.

robertbastian:
Outstanding question: do we deprecate `VariantOffsetsCalculator`?

sffc:
> Outstanding question: do we deprecate `VariantOffsetsCalculator`?

Yes

PR #6822, opened 2025-08-07
Move `TimeZoneVariant` to provider module

robertbastian:
#6820


components/time/src/provider/mod.rs
```diff
@@ -65,6 +64,21 @@ pub const MARKERS: &[DataMarkerInfo] = &[
...
+#[cfg_attr(feature = "datagen", derive(serde::Serialize, databake::Bake))]
+#[cfg_attr(feature = "datagen", databake(path = icu_time::provider))]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize))]
+#[non_exhaustive]
+pub enum TimeZoneVariant {
+    /// The variant corresponding to `"standard"` in CLDR.
+    Standard = 0,
+    /// The variant corresponding to `"daylight"` in CLDR.
+    Daylight = 1,
+}
```
sffc: I think we should retain the notice that was there previously

```suggestion
pub enum TimeZoneVariant {
    /// The variant corresponding to `"standard"` in CLDR.
    ///
    /// The semantics vary from time zone to time zone. The time zone display
    /// name of this variant may or may not be called "Standard Time".
    Standard = 0,
    /// The variant corresponding to `"daylight"` in CLDR.
    ///
    /// The semantics vary from time zone to time zone. The time zone display
    /// name of this variant may or may not be called "Daylight Time".
    Daylight = 1,
}
```

sffc (dismissed):
I'm happy with this TimeZoneVariant definition if it is in the provider module and its sole purpose is to represent the CLDR standard/daylight strings according to how CLDR defines them. ICU4X should avoid remapping the CLDR names to different buckets.

components/time/src/provider/mod.rs
robertbastian: I think it's too much detail but if you insist

sffc (changes requested):
Issue: Please remove the dart bin files

robertbastian:
whoops

sffc (approved):

Manishearth (approved):

PR #6823, opened 2025-08-08
Port BOCSU algorithm from ICU4C for identical level sort keys

rs-sac:
Here is the requested code ported from ICU4C to make U+FFFE sort below U+0000 in sort keys at the identical level.

I include some tests derived from the bug report that did not pass before but now do.

Because we no longer write directly to the sink during normalization, the SinkAdapter can be deleted.

Related: #6811

### DIFF
--- a/components/collator/src/comparison.rs
+++ b/components/collator/src/comparison.rs
@@ -1821,7 +1821,7 @@ impl CollatorBorrowed<'_> {
         S: CollationKeySink + ?Sized,
         S::State: Default,
     {
-        self.write_sort_key_impl(s.chars(), sink, |nfd, sink| nfd.normalize_to(s, sink))
+        self.write_sort_key_impl(s.chars(), sink)
     }
 
     /// Given potentially invalid UTF-8, write the sort key bytes up to the collator's strength.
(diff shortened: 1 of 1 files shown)


components/collator/src/comparison.rs
```diff
@@ -1835,9 +1865,11 @@ impl CollatorBorrowed<'_> {
             let nfd =
                 DecomposingNormalizerBorrowed::new_with_data(self.decompositions, self.tables);
             sink.write_byte(&mut state, LEVEL_SEPARATOR_BYTE)?;
-            let mut adapter = SinkAdapter::new(sink, &mut state);
+
+            let mut tmp = Vec::new();
```
hsivonen: Instead of creating a temporary buffer here, please use `normalize_iter` on the normalizer to get an iterator that can be passed to `write_identical_level` without an intermediate buffer.

Then the sinks can be simplified to remove the parts that were meant for receiving output from `normalize`.

rs-sac: Sorry, I missed that.

hsivonen (approved):
LGTM. Thank you!

PR #4514, opened 2024-01-05
Diplomat bump: Dart native

robertbastian:

PR #6826, opened 2025-08-08
Add CI to build and publish compiled binaries

robertbastian:
#4689


.github/workflows/artifacts-build.yml
```diff
@@ -586,6 +586,160 @@ jobs:
...
+            include_data: false
+    runs-on: ${{ matrix.os }}
+    env:
+      FEATURES: ${{ matrix.include_data == true && 'default_components,experimental,buffer_provider,compiled_data' || 'default_components,experimental,buffer_provider' }}
+      FILENAME_SUFFIX: ${{ matrix.compiletype == 'static' && matrix.include_data == true && '-with_data' || '' }}
+
+    steps:
+    - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9
+      with:
+        submodules: true
```
sffc: Question: submodules?

.github/workflows/artifacts-build.yml
```diff
@@ -586,6 +586,160 @@ jobs:
...
+        cp target/aarch64-pc-windows-msvc/release/icu4x-datagen.exe bin/icu4x-datagen-aarch64-pc-windows-msvc.exe
+        cp target/x86_64-pc-windows-msvc/release/icu4x-datagen.exe bin/icu4x-datagen-x86_64-pc-windows-msvc-datagen.exe
+    
+    - name: Upload artifacts
+      uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02
+      with:
+        name: ffi-${{matrix.os}}-${{ matrix.compiletype }}${{ matrix.compiletype == 'static' && matrix.include_data == true && '-with_data' || '' }}-libs
+        path: bin
+        
+  release-ffi-libs:
```
sffc, line 721: Question: Have you tested this action, since it only runs on tags?

.github/workflows/artifacts-build.yml
```diff
@@ -586,6 +586,160 @@ jobs:
...
+        echo "aarch64-unknown-linux-gnu.linker=\"aarch64-linux-gnu-gcc\"" >> .cargo/config.toml
+        echo "armv7-unknown-linux-gnueabihf.linker=\"arm-linux-gnueabihf-gcc\"" >> .cargo/config.toml
+        echo "riscv64gc-unknown-linux-gnu.linker=\"riscv64-linux-gnu-gcc\"" >> .cargo/config.toml
+        
+    - name: Build Linux
+      if: matrix.os == 'ubuntu-latest'
+      run: |
+        (cd ffi/dart && dart pub get)
+        
+        dart ffi/dart/lib/src/hook_helpers/build_libs.dart --working_directory . --file bin/icu4x-2-${{ matrix.compiletype }}${{ env.FILENAME_SUFFIX }}-armv7-linux-androideabi --target armv7-linux-androideabi --compile_type ${{ matrix.compiletype }} --cargo_features ${{ env.FEATURES }}
```
sffc, line 641: Question: Why are you using a Dart script to build these, instead of a cargo make script or something?

ffi/dart/tool/regenerate_hashes.dart
```diff
@@ -1,30 +1,48 @@
...
+    'aarch64-unknown-linux-gnu',
+    'i686-unknown-linux-gnu',
+    'riscv32gc-unknown-linux-gnu',
+    'riscv64gc-unknown-linux-gnu',
+    'x86_64-unknown-linux-gnu',
+    'aarch64-apple-darwin',
+    'x86_64-apple-darwin',
+    'aarch64-pc-windows-msvc',
+    'i686-pc-windows-msvc',
+    'x86_64-pc-windows-msvc',
```
sffc, line 33: Observation: You have this list, or things derived from this list, in multiple places. Seems fragile?

.github/workflows/artifacts-build.yml
```diff
@@ -586,6 +586,160 @@ jobs:
...
+        echo "[target]" >> .cargo/config.toml
+        echo "aarch64-linux-android.linker=\"$ANDROID_NDK_LATEST_HOME/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android30-clang\"" >> .cargo/config.toml
+        echo "armv7-linux-androideabi.linker=\"$ANDROID_NDK_LATEST_HOME/toolchains/llvm/prebuilt/linux-x86_64/bin/armv7a-linux-androideabi30-clang\"" >> .cargo/config.toml
+        echo "i686-linux-android.linker=\"$ANDROID_NDK_LATEST_HOME/toolchains/llvm/prebuilt/linux-x86_64/bin/i686-linux-android30-clang\"" >> .cargo/config.toml
+        echo "x86_64-linux-android.linker=\"$ANDROID_NDK_LATEST_HOME/toolchains/llvm/prebuilt/linux-x86_64/bin/x86_64-linux-android30-clang\"" >> .cargo/config.toml
+        echo "riscv64-linux-android.linker=\"$ANDROID_NDK_LATEST_HOME/toolchains/llvm/prebuilt/linux-x86_64/bin/riscv64-linux-android35-clang\"" >> .cargo/config.toml
+        echo "aarch64-unknown-linux-gnu.linker=\"aarch64-linux-gnu-gcc\"" >> .cargo/config.toml
+        echo "armv7-unknown-linux-gnueabihf.linker=\"arm-linux-gnueabihf-gcc\"" >> .cargo/config.toml
+        echo "riscv64gc-unknown-linux-gnu.linker=\"riscv64-linux-gnu-gcc\"" >> .cargo/config.toml
```
sffc, line 634: Question: Why can't you use lld for everything? Why do you need to install and point to custom linkers for each platform

.github/workflows/artifacts-build.yml
robertbastian: I've run in up to the release step, see the commit history

.github/workflows/artifacts-build.yml
robertbastian: because we already have a dart build script. I plan to migrate this in the future (but we'll always need a Dart build script anyway)

.github/workflows/artifacts-build.yml
robertbastian: this is existing code that currently lives in dart-lang/i18n, but previously already lived in ICU4X where you already reviewed it: https://github.com/unicode-org/icu4x/pull/4514/files#diff-d1bb081e6d04ced14e89691280b752ae6124cee2b13d0af35840b7bf3db32a07

ffi/dart/tool/regenerate_hashes.dart
robertbastian: the whole hash code needs to be cleaned up but I can only really work on it once we have published artifacts

sffc (approved):

sffc:
- @robertbastian Can I ship the package on pub.dev as 2.0.0-dev0?
- @sffc Yes, I think that is the correct number to use, and I think it's better to have a package out there, even if we decide to tweak it later with more features / slicing / etc.
- @Manishearth LGTM. A bit worried about the stability stuff.
- @sffc Let's confirm with @moseum. We're meeting later today.
- @robertbastian I think @moseum is okay with this; he preferred 0.2 but is okay with 2.0.0-dev0.
- @sffc I think releasing a prerelease artifact doesn't constitute a Technical Decision and therefore we can approve it here in the Working Group. But, FYI @zbraniecki.

PR #6843, opened 2025-08-13
Add `calendrical_calculations::gregorian`

robertbastian:
The book uses the term "ISO calendar" for the ISO week calendar, which is another (unimplemented) calendar. 

We should follow its naming in this crate.


Cargo.toml
```diff
@@ -187,7 +187,7 @@ icu_experimental_data = { version = "~0.3.0", path = "provider/data/experimental
 
 # Utils
 bies = { version = "0.2.2", path = "utils/bies", default-features = false }
-calendrical_calculations = { version = "0.2.0", path = "utils/calendrical_calculations", default-features = false }
+calendrical_calculations = { version = "0.3.0-dev", path = "utils/calendrical_calculations", default-features = false }
```
Manishearth: issue: we can't do this, RataDie is exported.


we need to do this in a way that allows icu_calendar to still use the old versions.

Manishearth (changes requested):
I'm not really sold on this. This is one of the few modules that will get called outside of implementing that module's calendar, calling it iso (which is more accurate, and more broadly applicable) is good. The book naming is useful for verifying code correctness but I don't think we need to match it exactly. We have a fair amount of important stuff that doesn't come from the book already, like keviyah.

Furthermore, helpers like the week thing from my draft PR or the Neri/Schneider code could probably go in here.

Cargo.toml
robertbastian: eww

robertbastian:
> calling it iso (which is more accurate, and more broadly applicable) is good

But it's not? ISO is a specialized Gregorian calendar and interchange format, but it's first and foremost a Gregorian calendar, and all the calculations here are general Gregorian calculations, like the calculations in the `julian` module are Julian calculations.

I was actually looking for Gregorian logic in this crate and couldn't find anything. It took me a while (and checking the book) to figure out that everything that is called gregorian* in the book is called iso* in our crate (while the book has iso* functions that mean something completely different).

Manishearth:
> I was actually looking for Gregorian logic in this crate and couldn't find anything

I'd say the same for looking for iso logic. I'm happy with a reexport if that helps, and ensuring that the lisp method names are mentioned in the docs.

robertbastian:
Gregorian is by far the more common name for this calendar, and if the book uses it, we should use it at least in this crate.

I honestly don't understand why we have an `Iso` calendar in `icu_calendar` at all. It's just the Gregorian calendar, it's not formattable, what's the point? The only difference is the eras, but you can always get ISO behaviour through `extended_year`.

Manishearth:
We have it as a primitive, basically. And the problem with "gregorian" is that often that includes the switchover, which you really don't want to have to think about when working with an ISO datetime primitive.

Gregorian is the more used one but when writing datetime _code_ "ISO" is thrown around a lot. The name is useful too.

I'm in favor of making this more discoverable for gregorian users, I'm against making it less discoverable for ISO users or making a breaking change. I think there's plenty of stuff we can do that satisfies both.

robertbastian:
> And the problem with "gregorian" is that often that includes the switchover, which you really don't want to have to think about when working with an ISO datetime primitive.

The only APIs that expose the year 0 difference between these calendars are `year_info`, which is basically only for formatting (something that ISO doesn't support) and `new_from_codes` where they agree on the extended year but not on the eras. `Date::try_new_iso` and `Date::try_new_gregorian` are identical APIs. Anyway, that's `icu_calendar`, but the book just doesn't use ISO calendar terminology outside of the ISO week calendar.

> I'm against making it less discoverable for ISO users or making a breaking change

So what if we want to implement the ISO week calendar from the book? That should own the `calendrical_calculations::iso` namespace, including functions like `fixed-from-iso` etc.

Manishearth:
> The only APIs that expose the year 0 difference between these calendars

I meant gregorian-julian switchover. Just "Gregorian" is ambiguous. When implementing a Gregorian calendar, that's something you _should_ check for and it's fine for that to be something you have to. When doing other calendrical stuff you should not have to worry about your main calendrical primitive being ambiguous.



> So what if we want to implement the ISO week calendar from the book? That should own the `calendrical_calculations::iso` namespace, including functions like `fixed-from-iso` etc.

I'd call it `iso_week`; I'd be against using the `fixed-from-iso` naming anyway.

This crate is more than a wrapper around the Reingold calculations. It's _mostly_ that, but it's more, too.

utils/calendrical_calculations/src/lib.rs
```diff
@@ -36,27 +36,24 @@
...
-pub mod iso;
-/// The Julian calendar
 pub mod julian;
-/// The persian calendar
 pub mod persian;
-/// Representation of Rata Die (R.D.) dates, which are
-/// represented as the number of days since ISO date 0001-01-01.
 pub mod rata_die;
+
+#[deprecated(since = "0.2.1", note = "use `gregorian`")]
```
Manishearth: I guess my problem with this is that I think _most_ of the current users of this API should be calling it ISO, not gregorian.

Only calendar/src/cal/gregorian.rs should call it gregorian.

Definitely the zoneinfo64 code shouldn't be calling it "gregorian"; time zone code I have seen is very consistent about using "ISO".

Manishearth: I'm okay with the reexport, I'm not okay with this being considered second class.

robertbastian: > Definitely the zoneinfo64 code shouldn't be calling it "gregorian"; time zone code I have seen is very consistent about using "ISO".

zoneinfo64 should definitely call Gregorian, that's the calendar. I don't know where we got the concept of ISO as a calendar from (temporal?), but it's not universal (ICU4C/J doesn't have it).

Manishearth: Temporal, yes, but I've seen the notion of "ISO date" when reading standards a fair amount.

robertbastian: `icu::calendar` now implements the ISO calendar through `AbstractGregorian`

robertbastian:
Ok, so

* it is the Gregorian calendar
* the book calls it the Gregorian calendar
* the book calls *a different calendar* ISO
* the main user of this calendar is `abstract_gregorian.rs`
* all calendars in `calendrical_calculations` (including the very similar Julian calendar) are proleptic, so we don't have to call it ISO for it to be proleptic
    * in `calendrical_calculations`, all years are extended years, and all calendars have a year 0
* it's weird for the `easter` method to be in the `iso` module
* time zone code I've seen [uses "Gregorian"](https://data.iana.org/time-zones/tzdb-2022b/theory.html#:~:text=The%20tz%20database%20models%20time%20using%20the%20proleptic%20Gregorian%20calendar)

Manishearth:
I think that's more convincing but I'm still reluctant to do a deprecation just for that.

I don't think ISO is _wrong_ here, I am slowly getting convinced that Gregorian is _better_, but very slightly so.

robertbastian:
I would like to get this changed to not have this ISO terminology proliferate.

Manishearth:
I don't see that as a problem, really. ISO terminology is good, in my book.

robertbastian:
I think it's problematic: 
image

Manishearth:
I'd like to hear more opinions on this. To me the ISO week thing is obscure.

robertbastian:
Well that's what you get when you ask Gemini what an "iso year" is.

sffc:
The book calls it `fixed-from-gregorian` and `gregorian-from-fixed`, and the `calendrical_calculations` crate aims to implement Reingold (it is a separate crate directly as a result of the Reingold license), so it seems to make sense for the function names to reflect the names from Reingold.

I'm open to keeping the `iso` names around as aliases, but it's probably better to promote a single canonical answer. What I don't think we should do is to break `calendrical_calculations` in a way that makes `icu_calendar` need a hard dependency update in a way that violates our internal dependencies policy.

Manishearth:
The calendrical_calculations crate does more than just Reingold (the keviyah stuff is non-Reingold). But point taken on correspondence.

I don't really wish to spend time pushing for my relatively weakly held position, so if both of you think this should be renamed and deprecated, that's fine by me.

PR #6846, opened 2025-08-13
Support `$PUB_CACHE` in Dart checkout mode

robertbastian:


sffc (approved):
Rubber stamp

ffi/dart/hook/build.dart
```diff
@@ -173,20 +174,35 @@ final class LocalMode extends BuildMode {
...
     print('Running in `checkout` mode');
     if (checkoutPath == null) {
       throw ArgumentError(
         'Specify the ICU4X checkout folder with the `checkoutPath` key in your '
         'pubspec build options.',
       );
     }
-    if (!File.fromUri(checkoutPath!.resolve('Cargo.lock')).existsSync()) {
+    if (checkoutPath.path.contains('\$PUB_CACHE')) {
+      final envVars = Platform.environment;
```
mosuem, line 188: The environment variables are filtered for build hooks IIRC, so that will probably not work.

robertbastian: I tried it, it works

mosuem:
I don't understand - how would the ICU4X repo be in the pub cache? Normally the cache would only have `package:icu4x`, which does not have any rust code in it, just bindings.

ffi/dart/hook/build.dart
mosuem: That's surprising to me - https://github.com/dart-lang/native/blob/e0cd7d4555491d96a784ad362c569be1b53fa2f7/pkgs/hooks_runner/lib/src/build_runner/build_runner.dart#L545 does not list the pub cache env var. 

@dcharkes any idea why this is not filitered?

robertbastian:
It would be if you use a git dependency. But if this is not supposed to work it's not a good idea.

ffi/dart/hook/build.dart
dcharkes: This should not be accessible indeed. I have no idea why it is not filtered.

PR #6861, opened 2025-08-14
Fix `RataDie::until`

robertbastian:
`until` is actually `since`

#6386

### DIFF
--- a/utils/calendrical_calculations/src/rata_die.rs
+++ b/utils/calendrical_calculations/src/rata_die.rs
@@ -70,8 +70,13 @@ impl RataDie {
     }
 
     /// Calculate the number of days between two `RataDie` in a const-friendly way
-    pub const fn until(self, rhs: Self) -> i64 {
+    pub const fn since(self, rhs: Self) -> i64 {
         self.0 - rhs.0
+    }
+
+    /// Calculate the number of days between two `RataDie` in a const-friendly way
+    pub const fn until(self, rhs: Self) -> i64 {
+        rhs.0 - self.0
     }
 
     /// Adds a number of days to this `RataDie` in a const-friendly way
@@ -134,7 +139,7 @@ impl SubAssign<i64> for RataDie {
 impl Sub for RataDie {
     type Output = i64;
     fn sub(self, rhs: Self) -> Self::Output {
-        self.until(rhs)
+        self.since(rhs)
     }
 }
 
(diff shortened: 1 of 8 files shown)


robertbastian:
I renamed `until` to `since`, and then added a new `until`.

utils/calendrical_calculations/src/rata_die.rs
```diff
@@ -70,10 +70,15 @@ impl RataDie {
     }
 
     /// Calculate the number of days between two `RataDie` in a const-friendly way
-    pub const fn until(self, rhs: Self) -> i64 {
+    pub const fn since(self, rhs: Self) -> i64 {
```
sffc, line 79: Please add a docs test. I think it may have been less likely that we landed this issue if we'd had a docs test.

utils/calendrical_calculations/src/rata_die.rs
```diff
@@ -70,10 +70,15 @@ impl RataDie {
...
         self.0 - rhs.0
     }
 
+    /// Calculate the number of days between two `RataDie` in a const-friendly way
+    pub const fn until(self, rhs: Self) -> i64 {
```
sffc, line 90: Suggestion: since `until` is tainted, please mark it `#[deprecated]` to inform people that the sign changed, and we can un-deprecate in a future release.

sffc (changes requested):
^

utils/calendrical_calculations/src/rata_die.rs
robertbastian: The idea here is to consider it a bug fix

robertbastian:
> we should be careful not to break people

I don't think this will break people, it would rather unbreak them because they would expect a certain behaviour from `until`.

There are no dependents on `calendrical_calculations` or `icu_calendar@^2` on crates.io.

sffc (dismissed):
Please release calendrical calculations from the 2.0 release branch with this fix to reduce the amount of time people have to encounter this bug and build in wrong assumptions

Manishearth (approved):
In favor of releasing, from the 2.0 branch

sffc:
> > we should be careful not to break people
> 
> I don't think this will break people, it would rather unbreak them because they would expect a certain behaviour from `until`.

`since` and `until` are the two most confusing functions in Temporal. I still get their sign mixed up. If I had written software using one, I would experiment to see what sign it gives me, and then continue on with the rest of my program. I would not assume that the function had a bug.

sffc (approved):

PR #6886, opened 2025-08-20
Let prop map lookups reach the trie code with char range info and inlining

hsivonen:
We shouldn't be inhibiting inlining here and we shouldn't be throwing away `char` range information.


robertbastian (approved):
makes sense

PR #6896, opened 2025-08-25
Align rearguard metazone variant overrides with CLDR

robertbastian:
Follow-up #6893


sffc (commented):
I'm confused, I thought rearguard didn't have negative DST

robertbastian:
We only need rearguard semantics while zone is in a metazone, and in fact with the current CLDR overrides we can only have overrides while a zone is in a metazone. Casablanca has not been in a metazone since TZDB started modeling it as negative DST.

sffc (changes requested):
OK, so CLDR is agnostic to the guardedness of Casablanca. So then ICU4X needs to make a choice one way or the other. Why not stick with Rearguard like we've always done?

robertbastian:
Not sure what you mean by "CLDR is agnostic to the guardedness of Casablanca".

> Why not stick with Rearguard like we've always done?

Because "we've always done it this way" is not a convincing argument and I'm trying to get a way from rearguard.

sffc (changes requested):
I feel like this is the 4th time we've had this conversation and it's not a good use of my time to rehash everything. But basically can we please just keep the current (rearguard) model and only adopt changes as CLDR adopts them.

sffc:
> Not sure what you mean by "CLDR is agnostic to the guardedness of Casablanca".

CLDR does not define the offsets for Casablanca post-2018 and therefore it is not the decision-maker. We are the decision-maker in ICU4X.

> Because "we've always done it this way" is not a convincing argument and I'm trying to get a way from rearguard.

It's a very convincing argument. We are trying to get to CLDRguard, and until CLDRguard defines this behavior, we shouldn't proactively adopt something else.

sffc:
Could we make UTS 35 say "the DST semantics are those of TZDB Rearguard except where there are overrides"?

robertbastian:
> Could we make UTS 35 say "the DST semantics are those of TZDB Rearguard except where there are overrides"?

No. We don't have access to rearguard during datagen, we had hardcoded this before but that is not a long term solution.

Outside a metazone it *does not matter*.

sffc:
If it doesn't matter at all for anything (I guess with no Specific Location there isn't any case where it is used), then can we remove this data from TimezonePeriodsV1?

robertbastian:
We could, however there is some value knowing that Casblanca can be +1 or +0, without judging which one is standard time.

We also wouldn't be improving data size, as the offsets need to be there for metazone periods and making this variably-typed is only going to increase size.

robertbastian:
> it is harmful to change this time zone to vanguard

Not vanguard, main. The default TZDB variant that many other systems are also using.

PR #6899, opened 2025-08-26
Add Christian holidays to `Gregorian` and `Julian`

robertbastian:
The Hijri calendars are basically exclusively religious, so we have precedent for something like this.


components/calendar/src/cal/julian.rs
```diff
@@ -208,6 +208,11 @@ impl Julian {
     pub fn new() -> Self {
         Self
     }
+
+    /// Returns the date of (Orthodox) Easter in the given year.
+    pub fn easter(year: i32) -> Date<Self> {
+        Date::from_rata_die(calendrical_calculations::julian::easter(year), Self)
+    }
```
sffc, line 215: Thought: Should this go in the Coptic calendar, too?

robertbastian: Those two calendars are close enough to the Julian calendar that they use the dates as calculated by the Julian calendar: https://en.wikipedia.org/wiki/Coptic_calendar#Easter. So they should be constructed like `Julian::easter(y).to_calendar(Coptic/Ethiopian)`.

In fact, the only use case I see for `Julian::easter` is `Julian::easter(y).to_calendar(c)`, as no country actually uses the Julian calendar, and it's not in `AnyCalendar`.

PR #6915, opened 2025-09-03
Advertise that zerotrie uses safe Rust

sffc:


utils/zerotrie/src/lib.rs
```diff
@@ -10,6 +10,13 @@
 //!
 //! There are multiple variants of [`ZeroTrie`] optimized for different use cases.
 //!
+//! # Safe Rust
+//!
+//! All runtime lookup code in this crate is 100% safe Rust.
+//!
+//! A small amount of unsafe Rust is used in the builder code as well as for implementing
+//! unsafe traits such as [`VarULE`](zerovec::ule::VarULE).
```
Manishearth: nit: something like "unsafe traits for zero-copy frameworks"

robertbastian (changes requested):
There are transmutes outside the builder code and ULE implementations.

Please add a `#[cfg_attr(not(feature = "zerovec"), deny(unsafe_code))]` to keep this claim valid.

utils/zerotrie/src/lib.rs
```diff
@@ -23,6 +23,15 @@
 //!
 //! There are multiple variants of [`ZeroTrie`] optimized for different use cases.
 //!
+//! # Safe Rust
+//!
+//! All runtime lookup code in this crate is 100% safe Rust.
+//!
+//! A small amount of unsafe Rust is used in these situations:
+//!
+//! - Casting references after checking for invariants
```
Manishearth:
```suggestion
//! - Casting references after checking for invariants in the builder code
```

To make it clear that this isn't "runtime lookup".

sffc: But as @robertbastian pointed out, there are transmutes outside the builder code, mainly in the zerotrie bytes constructors.

We could get rid of most or all of these if we were okay depending on a crate like ref-cast or if we would land the icu4x-internal solution to ref-cast that I proposed a long time ago.

Manishearth: ref-cast is still an internally-unsafe crate. I'm a bit confused: we make the claim that runtime lookup code is 100% safe, but you're saying that the bytes constructors are not.

Are those not runtime lookup as well? They feel like "runtime" deps.


I'm trying to reconcile the claim in the first line with the claims in the bullet points. They should be consistent, and the bullet points should be clear about where the unsafe lies.

"We don't use unsafe except for the few cases we do, where we first check invariants" isn't a very interesting claim, that is the default state that all unsafe crates should be in.

sffc: The transmutes are not in the runtime lookup code path and you can use ZeroTrie without touching those code paths.

I'm trying to draw a contrast to other trie-like data structures in the Rust ecosystem that use unsafe code deep in the lookup code.

sffc: In other words: you can have a `ZeroTrie<&[u8]>` and you don't ever touch the transmutes. You only touch them if you have a `&ZeroTrie<[u8]>`

Manishearth: Got it. Can you just say "Casting references after checking for invariants in the builder code and when deserializing from bytes"?

Maybe even include the type thing you just said. I think it helps shore up the claim.

utils/zerotrie/src/lib.rs
```diff
@@ -47,6 +56,9 @@
 //! [`LiteMap`]: litemap::LiteMap
 //! [`BTreeMap`]: alloc::collections::BTreeMap
 
+// To back up the claim in the docs:
+#![cfg_attr(not(feature = "zerovec"), deny(unsafe_code))]
```
robertbastian: this gets disabled with `--all-features`, which is what we usually run. I'd prefer if you actually put `allow`s where needed.

utils/zerotrie/src/lib.rs
robertbastian: I don't think invariants have anything to do this with. The unsafe code is `&A -> &B` where `#[repr(transparent)] struct B(A)`. This is always a valid transmute, but there is no safe way to write this. 

```suggestion
//! - Constructing unsized transparent newtypes (i.e. https://github.com/rust-lang/rust/issues/18806)
```

utils/zerotrie/src/builder/bytestr.rs
```diff
@@ -13,26 +13,30 @@ use alloc::boxed::Box;
 pub(crate) struct ByteStr([u8]);
 
 impl ByteStr {
+    #[allow(unsafe_code)] // approved in crate-level safety docs
```
robertbastian:
```suggestion
    #[allow(unsafe_code)] // transparent newtype casts are documented
```

utils/zerotrie/src/lib.rs
Manishearth: Invariants are relevant: transmute has some invariants, and we do uphold them here.

But yes, "I used unsafe after checking invariants" is somewhat of a vacuous statement since you should always be doing that

utils/zerotrie/src/lib.rs
Manishearth: We should use `forbid` here instead to back up the claim.

Manishearth: Except we can't because of the from bytes API. nvm

utils/zerotrie/src/lib.rs
robertbastian: ah I was reading this like there were additional domain-specific invariants, which there shouldn't be

utils/zerotrie/src/lib.rs
```diff
@@ -23,6 +23,15 @@
 //!
 //! There are multiple variants of [`ZeroTrie`] optimized for different use cases.
 //!
+//! # Safe Rust
+//!
+//! All runtime lookup code in this crate is 100% safe Rust.
+//!
+//! A small amount of unsafe Rust is used in these situations:
+//!
+//! - Constructing unsized transparent newtypes (i.e. <https://github.com/rust-lang/rust/issues/18806>)
```
Manishearth: issue: we should explain which API surface pulls this code in. That way people can go "oh i'm not using that code anyway"

utils/zerotrie/src/lib.rs
```diff
@@ -47,6 +56,9 @@
 //! [`LiteMap`]: litemap::LiteMap
 //! [`BTreeMap`]: alloc::collections::BTreeMap
 
+// To back up the claim in the docs:
+#![deny(unsafe_code)]
```
Manishearth, line 61: should we apply forbid to  most of the submodules?

utils/zerotrie/src/lib.rs
sffc: I don't see the value in this. Supporting a list seems like it would get stale. The audience of these comments are not clients who want to avoid unsafe code at all costs. The audience are developers who are weighing the tradeoffs between different data structures and trie crates on crates.io.

utils/zerotrie/src/lib.rs
sffc: Doesn't `#!` turn it on for all submodules?

Manishearth: forbid, not deny. Forbid = "cannot `allow`"

sffc: ok. I don't see the advantage to that. CI won't let code land if it isn't annotated.

utils/zerotrie/src/lib.rs
Manishearth: Why would it get stale? Do we plan to expand the amount of unsafe in this crate? If so, I'm not convinced of the usefulness of this claim at all. To be clear, I'm not asking for an exhaustive list of APIs, just a description of the types of APIs. In this case, I think it's just "building from slices of tuples".

People who are weighing tradeoffs still care about where the tradeoffs apply. "This crate uses more unsafe, but the unsafe is in a part of the code that will not be security sensitive for me" is great to know.

The specific operation behind the unsafe is usually not that convincing at the margin. It's sometimes useful to understand motivation (e.g. I'll always be wary of crates that say "we use unchecked indexing for performance" because 99% of the time they've benchmarked nothing), but it's not a positive signal, only the lack of a negative one.


------

Also, I looked closer at the usage of unsafe in this crate: it is theoretically unsound. `(&ByteStr, usize)` is not safe to transmute from `(&str, usize)`, at least not according to my current understanding of unsafe rules. In practice I don't expect this to be a real problem anytime soon.

`str` guarantees its own layout, which is sufficient for _direct_ transmutation. However, within a `repr(Rust)` type, all bets are off: Rust is e.g. allowed to treat `(i8, usize)` and `(u8, usize)` as having different layouts. People sometimes test with `-Zrandomize-layout`. I _believe_ this can't happen when the difference is just two `repr(transparent)` types[^1], but it can happen for two types with the same layout.

It is worth asking Rust for clarity on this. Rust has historically been very reluctant to make commitments about what layout optimizations it is allowed to do, but this might be something where they're fine doing so. Current rustc doesn't do too much here but it could choose to in the future. Happy to engage upstream on this if desired.

 [^1]: I'm not actually sure of this either: I think Rust is allowed to give `struct Foo(u8)` and `struct Bar(u8)` different layouts if it really wants;

sffc: I'm not sure what is actionable from this comment. I want to either land this PR or close it.

I created this PR when I was in a RustConf talk where someone was advertising their superfast trie impl, and I looked at it and the whole thing is super unsafe, like zerovec-level unsafe. I wanted to put in a quick note about zerotrie being quite safe in comparison.

Manishearth: Here's the actionable thing: say we use a little amount of unsafe for transparent newtype casting in ZeroTrie builder code when building from slices of tuples. It's a small caveat and it makes this much clearer to me.

utils/zerotrie/src/lib.rs
Manishearth: This is mostly paired with my other comment; it proves the claim about "no unsafe except in builder code" for anyone who bothers to look.

sffc:
Thanks for the approval.

Lots of commits for a 32-line change. I might not have started this if I had known it was going to take this long.

Manishearth (approved):

PR #6930, opened 2025-09-09
Implement `u` and `g` fields

robertbastian:


components/datetime/src/scaffold/get_field.rs
```diff
@@ -205,6 +219,13 @@ impl<C: Calendar, A: AsCalendar<Calendar = C>, Z> GetField<DayOfYear> for ZonedD
     }
 }
 
+impl<C: Calendar, A: AsCalendar<Calendar = C>, Z> GetField<RataDie> for ZonedDateTime<A, Z> {
+    #[inline]
+    fn get_field(&self) -> RataDie {
+        self.date.to_rata_die()
```
sffc, line 225: Question/Issue: is `to_rata_die` efficient in general? The fields are evaluated eagerly, and this is wasteful in the 99% case where the pattern doesn't contain this field. There's a long-standing open issue to read the fields lazily but it hasn't been a priority because all of them are currently fast to compute.

components/datetime/src/scaffold/dynamic_impls.rs
```diff
@@ -37,6 +37,7 @@ impl DateInputMarkers for DateFieldSet {
     type MonthInput = datetime_marker_helper!(@input/month, yes);
     type DayOfMonthInput = datetime_marker_helper!(@input/day_of_month, yes);
     type DayOfYearInput = datetime_marker_helper!(@input/day_of_year, yes);
+    type RataDieInput = datetime_marker_helper!(@input/rata_die, yes);
```
sffc, line 40: Thought/Suggestion: Since none of the stock patterns have this field, you could disable it here, and enable it only in power user APIs… hmm, maybe a new marker type? Not sure where to put it.

robertbastian: does it hurt here, if it's never set?

components/datetime/src/scaffold/get_field.rs
robertbastian: It's very optimised in Gregorian, so also in Buddhist, Indian, Japanese, Roc. It's also cheap in Chinese, Dangi, Hijri because we have precomputed year info. Julian, Ethiopian, and Coptic also have implementations that look cheap, Persian is the only one that might be a bit less efficient.

components/datetime/src/scaffold/dynamic_impls.rs
sffc: actually I guess `DateFieldSet` is the only place you currently have it turned on, which seems like the right place to put it (on the dynamic field set), so that the static field sets stay fast.

components/datetime/src/format/input.rs
```diff
@@ -126,6 +129,7 @@ impl DateTimeInputUnchecked {
             day_of_month: GetField::<D::DayOfMonthInput>::get_field(input).into_option(),
             iso_weekday: GetField::<D::DayOfWeekInput>::get_field(input).into_option(),
             day_of_year: GetField::<D::DayOfYearInput>::get_field(input).into_option(),
+            rata_die: GetField::<D::RataDieInput>::get_field(input).into_option(),
```
sffc, line 132: NB: `GetField::<D::RataDieInput>` resolves to `GetField<()>` when using static field sets. It resolves to `GetField<RataDie>` when using dynamic field sets, however.

PR #6952, opened 2025-09-17
Implement Ethopian in terms of Coptic

robertbastian:
These are the same calendar, just using different eras.

This is the same situation as Gregorian, Iso, Buddhist, Japanese, and Roc being the same calendar using different eras.

We might want to make these calendars generic in the future, and have something like `Coptic<CopticEra>`, `Coptic<EthiopianEraStyle>`, `Gregorian<JapaneseEras>`, `Gregorian<Iso>`, etc.


components/calendar/src/cal/gregorian.rs
```diff
@@ -40,13 +40,9 @@ use tinystr::tinystr;
...
 pub struct Gregorian;
 
-#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, PartialOrd, Ord)]
-/// The inner date type used for representing [`Date`]s of [`Gregorian`]. See [`Date`] and [`Gregorian`] for more details.
-pub struct GregorianDateInner(pub(crate) IsoDateInner);
-
 impl crate::cal::scaffold::UnstableSealed for Gregorian {}
 impl Calendar for Gregorian {
-    type DateInner = GregorianDateInner;
+    type DateInner = IsoDateInner;
```
sffc: Issue: I think we have the `GregorianDateInner` type because it's public API and we don't want to lock into a specific representation

components/calendar/src/cal/ethiopian.rs
```diff
@@ -64,57 +67,9 @@ pub enum EthiopianEraStyle {
...
-        } else {
-            365
-        }
-    }
-}
-
 impl crate::cal::scaffold::UnstableSealed for Ethiopian {}
 impl Calendar for Ethiopian {
-    type DateInner = EthiopianDateInner;
+    type DateInner = CopticDateInner;
```
sffc: Issue: This should stay as `EthiopianDateInner` but that type could change to be a newtype over `CopticDateInner`

components/calendar/src/cal/ethiopian.rs
```diff
@@ -294,8 +236,8 @@ impl Date<Ethiopian> {
         if era_style == EthiopianEraStyle::AmeteAlem {
             year -= INCARNATION_OFFSET;
         }
```
sffc: Issue: I was also looking at this code last week, and I think this operation seems wrong. I think it should be

```suggestion
        if era_style == EthiopianEraStyle::AmeteMihret {
            year += INCARNATION_OFFSET;
        }
```

because it seems like the year was stored in Amete Alem for both era styles. In your case, you still need this change if you are trying to store the year in Anno Martyrum.

components/calendar/src/cal/roc.rs
```diff
@@ -50,13 +50,9 @@ const ROC_ERA_OFFSET: i32 = 1911;
...
 pub struct Roc;
 
-/// The inner date type used for representing [`Date`]s of [`Roc`]. See [`Date`] and [`Roc`] for more info.
-#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, PartialOrd, Ord)]
-pub struct RocDateInner(IsoDateInner);
-
 impl crate::cal::scaffold::UnstableSealed for Roc {}
 impl Calendar for Roc {
-    type DateInner = RocDateInner;
+    type DateInner = IsoDateInner;
```
sffc: Issue: Keep the newtype

components/calendar/src/cal/gregorian.rs
robertbastian: * it's unstable public API
* Buddhist already uses `IsoDateInner` without wrapping

components/calendar/src/cal/ethiopian.rs
robertbastian: why

components/calendar/src/cal/ethiopian.rs
robertbastian: this seems unrelated to this PR

components/calendar/src/cal/gregorian.rs
sffc: Associated types are not unstable; you can get DateInners from various places and you might have saved them in a variable. I know I do that. I imagine temporal_rs or some other client who integrates deeply with icu_calendar might do that.

If Buddhist doesn't have a newtype, that's a bug

robertbastian: associated types are obtained only from the `Calendar` trait, which is unstable

sffc: It's Sealed, not unstable. We can't just go and change the signatures of functions on the trait. It is sealed so we can add new functions.

components/calendar/src/cal/ethiopian.rs
sffc: I guess it's fine if the tests still pass

components/calendar/src/cal/gregorian.rs
robertbastian: it's `UnstableSealed`

robertbastian: image

sffc: `UnstableSealed` is the same as `Sealed` except it means people can still implement it if they are okay with the trait changing. It's about implementers, not users

robertbastian: fact is we document this as being allowed to break semver

sffc: `icu::calendar::Calendar` is a trait that is sealed, and it is also a trait that people can construct and call methods from. It is implemented by a dozen public types in the crate. This is not controversial.

components/calendar/src/cal/ethiopian.rs
robertbastian: this code path doesn't seem to be covered by tests. however I'm not touching the alem vs mihret logic here, so it should be fixed and tested in a different PR

sffc:
Plumbing `Ethiopian` to wrap `Coptic` is a fine change if you wanted to make it. My issue was only with changing the public associated type `DateInner`.

sffc:
Technically I think adding a type alias would also suffice from a semver perspective, but I still strongly prefer the newtypes. (Not just here but also ideally in Chinese and Hijri)

components/calendar/src/cal/gregorian.rs
Manishearth: The intent of that documentation was that this trait is unstable to implement. The trait implementors [are not going to have the same banner](https://docs.rs/icu_calendar/latest/icu_calendar/cal/struct.Dangi.html#trait-implementations).

We should maybe fix the banner to be clearer about calling the trait, but at the moment a person looking at the docs for an individual calendar will not know that the underlying types are stable.

So this is not unstable public API, and I don't think that's a matter that we should continue to debate.

Now, _that said_, I am somewhat okay with minor breaking stuff around this provided we do not remove any names from the crate (we add type aliases, like we did in Dangi). We probably should have started with type aliases from the get-go, or started with individual newtypes, to give us freedom to change things.

So if we retain a deprecated GregorianDateInner type alias, that's great.

Manishearth:
> Technically I think adding a type alias would also suffice from a semver perspective

It's _technically_ not fully good from a semver perspective but the semver risk is very slight: it only matters for people doing `impl MyTrait for Type1` and `impl MyTrait for Type2` because now you have clashing impls.

But as with the Chinese/Hijri stuff, I'm okay with this level of breakage. We should not _remove_ names from the crate; so we should introduce deprecated aliases.

components/calendar/src/cal/ethiopian.rs
```diff
@@ -64,53 +67,9 @@ pub enum EthiopianEraStyle {
 #[derive(Copy, Clone, Debug, Hash, Default, Eq, PartialEq, PartialOrd, Ord)]
 pub struct Ethiopian(pub(crate) bool);
 
-/// The inner date type used for representing [`Date`]s of [`Ethiopian`]. See [`Date`] and [`Ethiopian`] for more details.
-///
-/// The year is stored as Amete Alem year
+#[allow(missing_docs)] // not actually public
```
sffc, line 65: Nit: Let's not add `#[missing_docs]`. I would like this type to be public but have no methods on it.

It might be a bug that this type isn't exported, since it can be accessed by `Coptic::DateInner`.

sffc (changes requested):
Also, let's wait to merge this until I merge Date::from_fields. I've already needed to rebase it on top of Hijri and Chinese, which was fine because those PRs blocked other work, but this PR strikes me as one that could wait a couple of days and be harmless, and it's small enough that it can be rebased on top of Date::from_fields.

components/calendar/src/cal/ethiopian.rs
robertbastian: I'm pretty sure it's intended that these types are only nameable as associated types. If they don't have a public API, why would they need docs.

Manishearth: I think a very short line of docs is good here. We don't need more.

Yes, these are mostly assoc types, but people might stumble upon them via search. We should probably improve the docs for that case (the current docs aren't great!), but we shouldn't remove them.

components/calendar/src/cal/ethiopian.rs
Manishearth: This is something we should fix in the from_fields PR, really, or separately.

Manishearth (approved):
r+ on approach, would also prefer from_fields to land first.

components/calendar/src/cal/ethiopian.rs
robertbastian: > the current docs aren't great!

the current docs are not public, because this is not a public type

Manishearth: ah, huh. That's fine then

Manishearth: Surprised the lint fires.

sffc:
One other thing is missing from the PR: _why?_

For Hijri, there are a lot of different sightings, and the Hijri trait offers convenient functionality utilizing the YearInfo, like arithmetic, ISO conversion, etc., which would be hard to replicate in userland. Same with Chinese.

But I don't see the same motivation applying to ISO-based calendars. You don't _gain_ any functionality by plugging in your own epoch. Do you?

robertbastian:
This deduplicates code that the compiler will not necessarily know to deduplicated (because the diffs are at the very bottom, in `calendrical_calculations`).

Storing the Ethiopian year as a Coptic year is also way clearer wrt which era is used internally.

components/calendar/src/cal/ethiopian.rs
sffc: Don't `#[doc(hidden)]` a non-exported type, then?

components/calendar/src/cal/ethiopian.rs
```diff
@@ -68,70 +70,28 @@ pub enum EthiopianEraStyle {
...
-            366
-        } else {
-            365
-        }
-    }
-}
+pub struct EthiopianDateInner(CopticDateInner);
 
 impl DateFieldsResolver for Ethiopian {
     type YearInfo = i32;
```
sffc, line 71: Nit: This code is confusing enough that I'd like you to please add a comment saying what year this YearInfo is. It used to be the variant-specific ethiopic year, but now it is the coptic year.

components/calendar/src/cal/ethiopian.rs
```diff
@@ -68,70 +70,28 @@ pub enum EthiopianEraStyle {
 #[derive(Copy, Clone, Debug, Hash, Default, Eq, PartialEq, PartialOrd, Ord)]
 pub struct Ethiopian(pub(crate) bool);
```
sffc: Nit: Document what this bool indicates

components/calendar/src/cal/ethiopian.rs
```diff
@@ -68,70 +70,28 @@ pub enum EthiopianEraStyle {
...
+            (EthiopianEraStyle::AmeteAlem, "aa") => Ok(era_year - COPTIC_OFFSET),
             (_, _) => Err(DateError::UnknownEra),
         }
     }
 
     #[inline]
     fn year_info_from_extended(&self, extended_year: i32) -> Self::YearInfo {
-        extended_year
+        extended_year + if !self.0 { INCARNATION_OFFSET } else { 0 } - COPTIC_OFFSET
     }
```
sffc, line 90: Suggestion:

```rust
    fn year_info_from_extended(&self, extended_year: i32) -> Self::YearInfo {
        let amete_alem_year = if self.0 { extended_year } else { extended_year + INCARNATION_OFFSET };
        let coptic_year = amete_alem_year - COPTIC_OFFSET;
        coptic_year
    }
```

components/calendar/src/cal/ethiopian.rs
robertbastian: I'm not doing it for fun. The compiler complains

components/calendar/src/cal/ethiopian.rs
robertbastian: well you didn't add a comment that it is the variant-specific ethiopic year either...

sffc: Previously it was a few lines away from the CalendarArithmetic impl but now it's in a different file.

sffc:
```
thread 'format::datetime::tests::extended_year' panicked at components/datetime/src/format/datetime.rs:688:9:
assertion `left == right` failed: 
  left: "10"
 right: "-5490"
```

I assume this is a bug fix but please verify

robertbastian:
Yes this is the thing you noted in https://github.com/unicode-org/icu4x/pull/6952#discussion_r2356087656

PR #6970, opened 2025-09-22
Change range for Chinese data to 1912-2125

robertbastian:
We should put the cutoff to "historical Chinese" at 1912, as that's when today's algorithm started being used (and it's also more than 100 years ago, a value that was agreed on in #5778).

In the other direction, add hardcoded data until 2125 (100 years in the future).

### DIFF
--- a/components/calendar/src/cal/chinese.rs
+++ b/components/calendar/src/cal/chinese.rs
@@ -140,14 +140,14 @@ fn fallback_approximation<CB: ChineseBased>(related_iso: i32) -> LunarChineseYea
 ///
 /// Accurate calculation according to [GB/T 33661-2017] is computationally
 /// [expensive](https://ytliu0.github.io/ChineseCalendar/computation.html#modern),
-/// so this type includes precomputed data for the years 1901-2100.
+/// so this type includes precomputed data for the years 1912-2125.
 ///
-/// If performance is required beyond 2100, clients can implement their own
+/// If performance is required beyond 2125, clients can implement their own
 /// [`Rules`] type using data computed by [`LunarChineseYearData::gb_t_33661_2017`],
 /// or from sources like the excellent compilation by
 /// [Yuk Tung Liu](https://ytliu0.github.io/ChineseCalendar/table.html).
 ///
-/// Dates before 1901 are mainly of historical interest, and were calculated
+/// Dates before 1912 are mainly of historical interest, and were calculated
 /// using different [methods](https://ytliu.epizy.com/Shixian/Shixian_summary.html)
 /// than [GB/T 33661-2017]. These algorithms and data are not included for binary
 /// size reasons. If accuracy is required in this range clients can implement their
--- a/components/calendar/src/cal/chinese/chinese_data.rs
+++ b/components/calendar/src/cal/chinese/chinese_data.rs
@@ -7,7 +7,7 @@
 
 use crate::provider::chinese_based::PackedChineseBasedYearInfo;
 
-pub const STARTING_YEAR: i32 = 1901;
+pub const STARTING_YEAR: i32 = 1912;
 
 #[rustfmt::skip]
 pub const DATA: &[PackedChineseBasedYearInfo] = {
@@ -15,17 +15,6 @@ pub const DATA: &[PackedChineseBasedYearInfo] = {
     let l = true; // long
     let s = false; // short
     &[
-        PackedChineseBasedYearInfo::new(1901, [s, l, s, s, l, s, l, s, l, l, l, s, s], None, iso(1901, 2, 19)),
-        PackedChineseBasedYearInfo::new(1902, [l, s, l, s, s, l, s, l, s, l, l, l, s], None, iso(1902, 2, 8)),
-        PackedChineseBasedYearInfo::new(1903, [s, l, s, l, s, s, l, s, s, l, l, s, l], Some(6), iso(1903, 1, 29)),
-        PackedChineseBasedYearInfo::new(1904, [l, l, s, l, s, s, l, s, s, l, l, s, s], None, iso(1904, 2, 16)),
-        PackedChineseBasedYearInfo::new(1905, [l, l, s, l, l, s, s, l, s, l, s, l, s], None, iso(1905, 2, 4)),
-        PackedChineseBasedYearInfo::new(1906, [s, l, l, s, l, s, l, s, l, s, l, s, l], Some(5), iso(1906, 1, 25)),
-        PackedChineseBasedYearInfo::new(1907, [s, l, s, l, s, l, l, s, l, s, l, s, s], None, iso(1907, 2, 13)),
-        PackedChineseBasedYearInfo::new(1908, [l, s, s, l, l, s, l, s, l, l, s, l, s], None, iso(1908, 2, 2)),
-        PackedChineseBasedYearInfo::new(1909, [s, l, s, s, l, s, l, s, l, l, l, s, l], Some(3), iso(1909, 1, 22)),
-        PackedChineseBasedYearInfo::new(1910, [s, l, s, s, l, s, l, s, l, l, l, s, s], None, iso(1910, 2, 10)),
-        PackedChineseBasedYearInfo::new(1911, [l, s, l, s, s, l, s, s, l, l, s, l, l], Some(7), iso(1911, 1, 30)),
         PackedChineseBasedYearInfo::new(1912, [l, s, l, s, s, l, s, s, l, l, s, l, s], None, iso(1912, 2, 18)),
         PackedChineseBasedYearInfo::new(1913, [l, l, s, l, s, s, l, s, s, l, s, l, s], None, iso(1913, 2, 6)),
         PackedChineseBasedYearInfo::new(1914, [l, l, s, l, s, l, s, l, s, s, l, s, l], Some(6), iso(1914, 1, 26)),
@@ -214,6 +203,32 @@ pub const DATA: &[PackedChineseBasedYearInfo] = {
         PackedChineseBasedYearInfo::new(2097, [l, s, l, s, s, s, l, s, l, l, s, l, s], None, iso(2097, 2, 12)),
         PackedChineseBasedYearInfo::new(2098, [l, l, s, l, s, s, s, l, s, l, s, l, s], None, iso(2098, 2, 1)),
         PackedChineseBasedYearInfo::new(2099, [l, l, s, l, l, s, s, l, s, s, l, s, l], Some(3), iso(2099, 1, 21)),
+        PackedChineseBasedYearInfo::new(2100, [l, l, s, l, s, l, s, l, s, s, l, s, s], None, iso(2100, 2, 9)),
+        PackedChineseBasedYearInfo::new(2101, [l, l, s, l, l, s, l, s, l, s, s, l, s], Some(8), iso(2101, 1, 29)),
+        PackedChineseBasedYearInfo::new(2102, [l, s, l, l, s, l, s, l, l, s, l, s, s], None, iso(2102, 2, 17)),
+        PackedChineseBasedYearInfo::new(2103, [s, l, s, l, s, l, s, l, l, l, s, l, s], None, iso(2103, 2, 7)),
+        PackedChineseBasedYearInfo::new(2104, [s, l, s, s, l, s, l, s, l, l, s, l, l], Some(6), iso(2104, 1, 28)),
+        PackedChineseBasedYearInfo::new(2105, [s, l, s, s, l, s, s, l, l, s, l, l, s], None, iso(2105, 2, 15)),
+        PackedChineseBasedYearInfo::new(2106, [l, s, l, s, s, l, s, s, l, s, l, l, s], None, iso(2106, 2, 4)),
+        PackedChineseBasedYearInfo::new(2107, [l, l, s, l, s, s, l, s, s, l, s, l, l], Some(5), iso(2107, 1, 24)),
+        PackedChineseBasedYearInfo::new(2108, [l, s, l, s, l, s, l, s, s, l, s, l, s], None, iso(2108, 2, 12)),
+        PackedChineseBasedYearInfo::new(2109, [l, s, l, l, s, l, s, l, s, s, l, s, l], Some(10), iso(2109, 1, 31)),
+        PackedChineseBasedYearInfo::new(2110, [s, l, l, s, l, l, s, l, s, s, l, s, s], None, iso(2110, 2, 19)),
+        PackedChineseBasedYearInfo::new(2111, [l, s, l, s, l, l, s, l, s, l, l, s, s], None, iso(2111, 2, 8)),
+        PackedChineseBasedYearInfo::new(2112, [s, l, s, l, s, l, s, l, l, s, l, l, s], Some(7), iso(2112, 1, 29)),
+        PackedChineseBasedYearInfo::new(2113, [l, s, s, l, s, s, l, l, s, l, l, l, s], None, iso(2113, 2, 16)),
+        PackedChineseBasedYearInfo::new(2114, [s, l, s, s, l, s, s, l, s, l, l, l, s], None, iso(2114, 2, 6)),
+        PackedChineseBasedYearInfo::new(2115, [s, l, l, s, s, l, s, s, l, s, l, l, l], Some(5), iso(2115, 1, 26)),
+        PackedChineseBasedYearInfo::new(2116, [s, l, s, l, s, l, s, s, l, s, l, l, s], None, iso(2116, 2, 14)),
+        PackedChineseBasedYearInfo::new(2117, [s, l, l, s, l, s, l, s, s, l, s, l, s], None, iso(2117, 2, 2)),
+        PackedChineseBasedYearInfo::new(2118, [s, l, l, s, l, l, s, l, s, s, l, s, l], Some(4), iso(2118, 1, 22)),
+        PackedChineseBasedYearInfo::new(2119, [s, l, l, s, l, s, l, s, l, s, l, s, s], None, iso(2119, 2, 10)),
+        PackedChineseBasedYearInfo::new(2120, [l, s, l, s, l, s, l, l, s, l, s, l, s], Some(8), iso(2120, 1, 30)),
+        PackedChineseBasedYearInfo::new(2121, [l, s, l, s, l, s, l, s, l, l, s, l, s], None, iso(2121, 2, 17)),
+        PackedChineseBasedYearInfo::new(2122, [s, l, s, l, s, s, l, s, l, l, l, s, s], None, iso(2122, 2, 7)),
+        PackedChineseBasedYearInfo::new(2123, [l, l, s, s, l, s, s, l, s, l, l, l, s], Some(6), iso(2123, 1, 27)),
+        PackedChineseBasedYearInfo::new(2124, [l, s, l, s, l, s, s, l, s, l, l, s, s], None, iso(2124, 2, 15)),
+        PackedChineseBasedYearInfo::new(2125, [l, l, s, l, s, l, s, s, l, s, l, s, s], None, iso(2125, 2, 3)),
     ]
 };
 
@@ -226,10 +241,10 @@ fn test_against_hong_kong_observatory_data() {
         Date,
     };
 
-    let mut related_iso = 1900;
+    let mut related_iso = 1911;
     let mut lunar_month = MonthCode::new_normal(11).unwrap();
 
-    for year in 1901..=2100 {
+    for year in 1912..=2100 {
         println!("Validating year {year}...");
 
         for line in ureq::get(&format!(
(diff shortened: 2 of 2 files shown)


sffc:
> We should put the cutoff to "historical Chinese" at 1912, as that's when today's algorithm started being used (and it's also more than 100 years ago, a value that was agreed on in #5778).

Further down in that thread we revise the "minimum of 100 years" to "what the HKO ships"

robertbastian:
Further down in that thread you say you plan to ship HKO data; that does not make it an agreed-upon plan. There is no relevance to the year 1900 other than that it starts a Gregorian century, we have correct data for 1899 as well.

> when anyone currently alive has a birthday

If you want that to be our criterion for including data, please back it up with some evidence that this results in a cutoff at 1900.

sffc:
The "person alive" criterion would imply 1908 IIRC which we round to 1900 because it signals that it is a human picked arbitrary cutoff

robertbastian:
I can find no evidence that anyone born in China before 1912 is alive today: https://en.wikipedia.org/wiki/List_of_the_oldest_people_by_country

sffc:
Also the PMO tables explicitly mark when pre-1912 diverges, and they impact some minor solar terms but nothing that causes days or months to shift, which I think I posted in one of the issues

sffc:
Also doesn't this unfix the 1906 issue you fixed last week? https://github.com/unicode-org/icu4x/issues/6454

Basically our policy as I understand it is that we aim to match ground truth for historical dates going back arbitrarily far, with an implicit statute of limitations when either data gets too big or we can't independently verify the data correctness.

robertbastian:
> Also doesn't this unfix the 1906 issue you fixed last week? https://github.com/unicode-org/icu4x/issues/6454

For all we know that same issue exists in 1899. I still haven't seen an argument why 1900 is a more epochal year for the Chinese calendar than 1912.

robertbastian:
> with an implicit statute of limitations when either data gets too big or we can't independently verify the data correctness.

"Data getting too big" is an arbitrary cutoff. We can independently verify data before 1900 as well, the PMO published this book for example: https://www.las.ac.cn/front/book/detail?id=34e0e2b422dd15f2231febddb7d764b8. We can do this arbitrarily far back, e.g. using the sources listed in https://ytliu0.github.io/ChineseCalendar/rules.html.

The fact remains that anything before 1912 does not follow GB/T 33661-2017/Calendrical Calculations, and everything after does. This is the most natural place to draw the line between "historical data" and "current data".

Manishearth:
My position is similar to Shane's, 1900 is a good cutoff and we have published data for it from a trusted source. The argument that we don't actually have people with birthdays before 1912 is compelling, though.

robertbastian:
We have published data from a trusted source since at least 1645

Manishearth:
Not the argument I was making.

1900 is a decent cutoff, _and_ we have published data from a trusted source, not _because_ we have published data from a trusted source.

Why I think it is a decent cutoff is roughly along the lines of what Shane said about artificial boundaries and birthdays.

robertbastian:
> Why I think it is a decent cutoff is roughly along the lines of what Shane said about artificial boundaries and birthdays.

Please elaborate? Shane's birthday argument does not hold water.

1912 is an epochal year for China, with the end of the Qing dynasty and the new Calendar rules. 1900 is completely arbitrary because China did not even use the Gregorian calendar then.

Manishearth:
The birthday argument is about getting a minimum, I like adding some padding where possible, and here it is quite possible.

1912 being epochal is somewhat compelling, though.

robertbastian:
> The "person alive" criterion would imply 1908

Still waiting for a source for this.

sffc:
> > The "person alive" criterion would imply 1908
> 
> Still waiting for a source for this.

Last I checked the oldest person alive was born in 1908 but now it's 1909; still rounds to 1900.

robertbastian:
Still no source

Manishearth:
Ethel Caterham
https://en.m.wikipedia.org/wiki/List_of_the_verified_oldest_people

Manishearth:
Amusingly the oldest verified living _Chinese_ person according to that list was born in 1913.

That's probably not a coincidence, this list is of verified people (there are a lot of _claims_, many of them later revealed to be fake) and I suspect Chinese recordskeeping changed with the ROC.

Though the Qing were pretty good at keeping records....

robertbastian:
> Ethel Caterham

Well she's not Chinese is she

Manishearth:
Right, and that's because this cutoff is a pan-calendar concept (in @sffc's version of it, and mostly mine). This is also why 1900 is a "round number" here: it's a pan-calendar concept, and this is a round number in ISO.

The idea is to have a cutoff where we guarantee pan-calendar ground truth accuracy as much as possible, which I think is generally a good idea because then our entire API surface becomes valid for those dates. It is not _challenging_ to do this; we have the data and we have the algorithms, the only "problem" is that it's a bit weird to be doing something different for a short span of 12 years. Overall I think I think that weirdness is a small cost to pay, and it'll be moot if we switch to a different approximation pre-1900 anyway.


Individual calendars can, on top of the 1900 cutoff, make guarantees of ground truth accuracy for wider ranges.

sffc:
The year number 1900 just got codified into the Temporal specification.

https://github.com/tc39/proposal-temporal/pull/3152

robertbastian:
No. That spec change is about how to find a reference year, i.e. to give 1900-1972 precedence over 1972-2035. That does not matter for this issue, because no Chinese reference year is in the range 1900-1912 (and even if one was, there are already reference years before the modern range anyway).

sffc:
The basis for the Temporal change is: "historical dates might change as more is discovered about how the calendar was used". This is exactly the case with 1906. The ground truth differs from the approximation we use. If we set 1912 as the start date, then dates between 1900 and 1912 are ones we consider subject to change, but the Temporal spec says that we should do that only prior to 1900.

sffc:
WG notes:

- @robertbastian Hardcoded chinese data is 1900-2100. (a) want to extend 100y so 2125
- @robertbastian This is a performance thing after 1912, we match the spec
- @robertbastian Before 1912 we diverge from ground truth in the algorithm, so hardcoding is actually about correctness
- @robertbastian It's weird for us to say "we use the right algorithm 1912 onwards, then have hardcoded data, then we are proleptic"
- @hsivonen How bad would it be to say this is correct from 1912 onwards and is proleptic using the same algorithm before that?
- @Manishearth We plan to be proleptic with a Pinqi approximation that we haven't completely figured out. I like having ground-truth invariants that, for everyone, from 1900 we match ground truth. It is easy for us to do this. Calendars can extend that range.
- @Manishearth I view these things as having a "ground truth validity", "data source", and "algorithm" as different things. I think it's fine for us to say "The chinese calendar follows Chinese ground truth from 1900 onwards, which is verified against the PMO data, using the GB/T algorithm from 1912 onwards and a different one before that. Outside of this range, we fall back to the GB/T algorithm, but this may change."
- @sffc Note that when the 1900 cutoff was something we initially thought of there was a Japanese or Chinese person alive. But either way I want to round down.
- @robertbastian 1912 is when the Qing dynasty fell, it's much more recognizeable
- @sffc 1900 is a globally recognizeable year. These are the dates we guarantee globally. This applies to Chinese as using the PMO data from 1900.
- @sffc The whole point of these calendars is to define ground truth, even if it's to say "lunar ground truth" it is ground truth.
- @hsivonen what is our UAQ start date?
- @robertbastian 1882.
- @hsivonen and Japanese?
- @robertbastian 1868, Meiji
- @sffc The thing I care about is a single global year range where all calendars have a well defined ground truth, and 1900 is an excellent choice for that year range. Maybe there are other choices, which we can debate.
- @manishearth one cop out is to define ±100y from 2025 as the global range, 1925 is the mandated minimum, and then individual calendars opt in to more range (Chinese would opt in to 1912)
- @sffc ±100y doesn't cover live-person birthdays. Also, once we choose a start year, we should stick to it.
- @hsivonen Without taking a position for a global range, observing that the most constraining calendar is UAQ, which depends on an authority.
- @sffc before I knew this was controversial we landed the referenceYear change to Temporal that codified 1900. That year is derived from the birthday heuristic.
- @manishearth I don't think that algorithm is super relevant because MonthDay doesn't interoperate between calendars *at all*.
- @sffc (1) I believe there should be a global range. (2) I think aligning it with the birthday heuristic, which is the same heuristic used for the MonthDay reference year algorithm, is a sensible choice.
- @manishearth generally find the MonthDay link to be not a big difference here since it doesn't interoperate
- @sffc it's linked because the referenceYear algorithm should return a year-month-day that isn't subject to implementation-defined approximation
- @nekevss I'm sympathetic to reference year not dictating the decisions for other things
- @nekevss Generally don't think that starting dates should move around
- @manishearth i think we should be okay with expanding starting dates
- @nekevss yes!
- @hsivonen The birthday criterion is useful for informing what to attempt. I think rounding to 1900 is less strongly justified. A bunch of these calendars are "easily proleptic", for the rest it's somewhat relevant as to what the adoption start / scheme is. E.g. for Japanese I'm sympathetic to the JDK reasoning ("they start with Jan 1 6 meiji", which is when the current calendar scheme was adopted) for Meiji rather than Temporal, but they both end up with the same conclusion.
- @manishearth yeah, that makes some sene. I do think in the Chinese case the calendar changed within the same shape and that's different from changing calendar system


No conclusion yet.

sffc:
@Manishearth made a point yesterday afternoon that changed my perspective a bit.

My hardline has been interop between Temporal implementations. This is what has been driving the Intl Era Month Code proposal, the strict rules for Reference Year, etc.

What @Manishearth pointed out is that _dates of birth_ (PlainDates as opposed to MonthDays) are themselves not defined by the calendar but are rather defined as a specific Rata Die. Therefore, serializing a date of birth to an IXDTF string and parsing it in another implementation is guaranteed to work, regardless of which set of Chinese approximations an engine happens to be using.

Therefore, if an engine has the wrong mapping between ISO and a calendar for someone's birthday, it is not an _interop_ bug; it is an engine-specific user bug.

To be clear, an engine-specific user bug is still one that I care a lot about. But, it's one where I'm more amenable to considering calendar-specific solutions instead of global solutions. In other words, I still prefer the ICU4X Chinese calendar to match PMO ground truth through 1900, but it's more because "well the data is there and it costs only a few extra bytes and then we fully align with PMO and can move on" and not "we need to enforce 1900 due to interop concerns".

---

The place where interop concerns still come up is if any MonthDay reference dates land in this range. We happen to be lucky and they don't. However, if they did, then this would be a hard objection from me at least until we updated the Temporal spec to use different reference years.

robertbastian:
Can you suggest phrasing for the docs of [`China`](https://unicode-org.github.io/icu4x/rustdoc/icu/calendar/cal/chinese/struct.China.html) for the status quo (which you're arguing for)? Note that the current docs are not accurate for the range 1901-1912

sffc:
Sure. Something like,

> ICU4X implements calendar rules according to GB/T 33661-2017 from 1912 through 2100. Outside that range, ICU4X attempts to match ground truth where there is unambiguous, high-quality data, and falls back to an arbitrary proleptic approximation for distant dates.

sffc:
WG notes:

- @sffc What is the core issue?
- @robertbastian Because GB/T only works from 1912 onwards, and 1900-1912 comes from written data. The only reason we have 1900-1912 is because @sffc had Temporal spec associated with it.
- @robertbastian Maybe this is blocked on deciding the requirements for the simple approximation?
- @sffc This is not because of Temporal any more. It's because our default assumption is that we want to match ground truth. Unfortunately we need to pick a cutoff date. The official data source starts in 1900, the PMO, and then there's the living person heuristic. Those are two specific reasons to use 1900 specifically. And the Temporal spec 1900 is one additional weak signal. But we could also consider 1800, 1823, 1645, 2000 BCE. If we go before 1900 as a universal year cutoff, though, we need to consider the impact on Japanese / Buddhist using lunar calendars, etc.
- @hsivonen I don't find the argument about Japanese lunar (...) compelling. When we discuss the calendars we have we should not let that discussion lead to "well then we have to add these historical calendars". We should take that off the table
- @hsivonen as for what to do before 1912... 1912 is a natural cutoff for the current algorithm. 1645 is where the available data starts. Also a natural starting point. Taking 12y of data that's not the current algorithm, but also doesn't go as far back ... seems somewhat arbitrary.
- @robertbastian 3b per year. So ~1kb
- (discussion about Easter)
- @hsivonen the general way you do festivals is you subscribe to a Festivals calendar, you don't need to compute them locally. But you do want to be able to locally compute the birthdays of your contacts. The oldest living person is relevant only in the context of that cultural sphere.
- @robertbastian: Current table of oldest known people: 

image

- @robertbastian Genreally disagree with "validity range for all calendars". Nobody has asked for this; buddhist is already incorrect before 1940, so that's the earliest that all calendars are correct
- @hsivonen the saka calendar was taken into use in 1958
- @robertbastian Qing and GBT can be considered different calendars, with a change in 1912. We don't model the Julian/Gregorian changes, why should we model this change? If we started in 1912, this calendar would be the GBT calendar
- @hsivonen ...
- @sffc GB/T can't be proleptic because it's impossible to calculate.
- @sffc A point of disagreement is whether we use a universal date or a calendar-specific date. If we use a universal date, we can use the person-living rule, or we can use our own rule.
- @robertbastian We could add 3 generations before 1900, too. Are there any other reasons to use 1900 than the person-living rule?
- @sffc Yes, it's where the PDF file starts.
- @robertbastian That seems arbitrary; I can order the book.
- @hsivonen By the way, I think the genealogy use case is scope creep that we shouldn't accept for the project.
- @sffc I'd like to summarize the points/disagreements.

Key points that have been brought up in discussion:

1. Propleptic GBT is expensive to calculate so it would have to be hardcoded - at that point we can hardcode actual data
2. The first year that used the GBT algorithm was 1912. This is similar to 1582 in the Julian/Gregorian calendar
3. Official data (from the Purple Mountain Observatory) is available from 1900 onwards. The PMO also publishes books that start before 1900. Data based on research is available from 722 BCE.
4. The oldest person currently living (with a documented birthday) was born in 1909. The person-living rule, which is an arbitrary but grounded tool that could be used for deciding a universal range of dates that calendars ought to support, would point to the nearest round decade, 1900.
5. The person-living rule was used by CLDR to decide that the modern Japanese calendar doesn't need eras before Meiji. It's not clear whether the person-living rule is used by other software.
7. Within the cultural sphere of this calendar, the oldest living (documented) person is not born before 1912.
8. The Temporal specification currently encodes the year 1900 as the first year for MonthDay reference years.
9. The Thai Buddhist calendar that we support is only valid from ~1940 and is incorrectly extended before that. The śaka calendar didn't exist before ~1950.

sffc:
A mild breakthrough from the CLDR Design WG meeting:

- @robertbastian believes that a "calendar" is an algorithm, one that can be projected forward and backward.
    - The north star is to export a set of algorithms.
    - "Chinese" means GB/T. To support dates before 1912, if there is demand, we could add "Qing" to encode the algorithm used during that period. @robertbastian also believes that there is not sufficient demand for that.
- @sffc believes that a "calendar" is a ground truth of how a certain region or group of people identify dates.
    - The north star is to export a set of ground truth rules.
    - "Chinese" means the calendar published by the Purple Mountain Observatory. Dates before 1912 should correspond to whatever the PMO has published.

Example implication for Julian/Gregorian:

- @robertbastian believes that we should export a Julian calendar and a Gregorian calendar, and _never_ export a combined Julian/Gregorian calendar with switchover. The user should figure out how they want to combine those two calendars.
- @sffc believes that it's in scope for ICU4X to export a Julian/Gregorian calendar with configurable switchover date. In BCP-47, the different calendars could be identified as `juligreg-1582` and `juligreg-1752`, for example.

CC @Manishearth

robertbastian:
The issue I see with Shane's view is that we won't be able to model the history of even the 20th century for every region without adding completely new calendar algorithms. The Soviet Union for example used very interesting calendars between 1918 and 1940, but because today they use the Gregorian calendar, we are not discussing adding these calendars. The region-based approach is a lot of work to do correctly, which is why I much prefer the algorithmic approach.

> @robertbastian believes that we should never export a combined Julian/Gregorian calendar with switchover. The user should figure out how they want to combine those two calendars.

I don't believe that. I'm very open to adding something like a `CalendarTransition` type that delegates between two calendars. But this type would have a very limited API; no from-fields, no arithmetic, no formatting, basically just ISO and RD conversions.

Manishearth:
This is interesting, and helps me understand where everyone is coming from.

My position is, essentially, _neither_.

I think this is a matter of _individual_ calendar _definition_ and _identity_. This is similar to my stance on https://github.com/unicode-org/icu4x/issues/6963#issuecomment-3340677483 . Ground truth is a potential parameter to a definition: some types of definitions need it, others don't, calendars do not inherently need to be defined as ground truth.

The "Gregorian calendar" is well-defined "as an algorithm". The "Gregorian-julian switchover calendar" is also mostly well-defined as an algorithm except it has a parameter, so to get a well-defined algorithm you need something like "Gregorian-Julian switchover, 17 Feb 1753" (Sweden, assuming in this notation that the date provided is the day the switchover happened in the Julian calendar).

Note that what each _code_ means is a matter of matching a calendar code up to some such expanded definition. So the word "Solar Buddhist calendar" is itself not well-defined, but it is possible to match it to a well-defined algorithm with ease.


The chinese calendar, _can_, too, be defined as an algorithm. However, such definitions are inherently less useful. We could very well expose ChineseGBTProleptic, or ChineseQingProleptic (using the math from that french book), or whatever.

But a _useful_ Chinese calendar, which is ideally our default `Chinese` calendar, would be something that matches ground truth. Here, instead of the calender being well-defined as an algorithm, it is defined based on a combination of some future- and past- extensible math approximating what we expect the calendar to look like, as well as ground truth. Ground truth becomes a parameter to the abstract calendar definition here.

What we're essentially debating here, is precisely which definition to pick (or, which parameters to pick). PMO dates post 1912 are a slam dunk, before 1912 using the Qing dates still keeps the calendar _useful_. We could very well pick the PMO data as far back as it goes if we really wanted, there's just a usefulness-codesize tradeoff here, and usefulness already tapers off the further back you go.

In the extreme abstract when talking about calendar identity I think about "would people on the ground think they are switching calendars" as the core idea.

The Gregorian-Julian or Buddhist switchover can be rather easily framed as switching calendars entirely, whereas in China at 1912 nobody would have considered the calendar to have changed: they just would have maybe noticed that some older almanacs are out of date; but forward-looking almanacs have always been somewhat unreliable.

(This isn't to say that a calendar can't have the _explicit_ identity as a switchover calendar; I'm just saying that I don't consider a Qing-GBT-PMO Chinese calendar to be a switchover calendar)

I think this essential difference is what we sometimes get bogged down on when trying to talk about underlying algorithm changes in the abstract in a pan-calendar way: underlying algorithm changes can happen in many different ways!


--------------------

> The issue I see with Shane's view is that we won't be able to model the history of even the 20th century for every region without adding completely new calendar algorithms. The Soviet Union for example used very interesting calendars between 1918 and 1940, but because today they use the Gregorian calendar, we are not discussing adding these calendars. The region-based approach is a lot of work to do correctly, which is why I much prefer the algorithmic approach.


I don't think the region-based approach mandates that we do this. The Soviet switchover can be rather easily framed as switching calendar systems entirely, with the rule of thumb above.

If there was a convincing reason to add support for a Soviet switchover calendar, I don't think we _couldn't_ add it. We'd probably call have to call it something like "Soviet-Russia calendar (1900-2000)" or something where it is defined as matching ground truth in that region for those years, and proleptically Gregorian.

(I don't think we ought to support this, but .... we can)

> I don't believe that. I'm very open to adding something like a `CalendarTransition` type that delegates between two calendars. But this type would have a very limited API; no from-fields, no arithmetic, no formatting, basically just ISO and RD conversions.

I think our crate should be able to support a Gregorian-Julian switchover calendar as a full fledged calendar. I think both our crate model and Temporal's model are _able_ to, though there are some ambiguities about how exactly to handle `daysInMonth` and first-day-of-month stuff. Ambiguities where a decision needs to be made, not ones which have any inherent contradiction, so that's fine.

sffc:
A visualization of my mental model for how far forward and backward to match ground truth in the Chinese calendar:

Chinese Calendar Value Deltas (1)

I consider the "δ Effort" at 1911 to be small, since we already ship tables. I consider the two "δ Trust" discontinuities at 1900 and 2100 to be the biggest changes.

There could also be another "δ Trust" somewhere in the mid-1800s where multiple provinces had competing authorities with different methods.

I'm happy to order a book for PMO data before 1900, in which case that triangle would change from a "δ Trust" to a "δ Effort" of roughly the same magnitude.

sffc:
I can buy into @Manishearth's mental model. I think a productive next step would be for @Manishearth to make a pull request with clearly documented definitions about which calendars are proleptic algorithms and which calendars try to match ground truth. Then we can debate individual points. I think agreeing on these definitions would resolve not only this issue but several others.

sffc:
Question, do we consider this a 2.1 blocking concern? My impression was that we did. But, if we are okay shipping 2.1 with what's currently on main (PMO data starting in 1900), then we don't need to block 2.1 on the final decision here.

Manishearth:
It's stretch, IMO. We have never committed to any specific calendar details here, and we don't _need_ to start now. It is nice to try.

Manishearth:
>  I think a productive next step would be for @Manishearth to make a pull request with clearly documented definitions about which calendars are proleptic algorithms and which calendars try to match ground truth

I'm still going to try this, but a worry I have is that the type of precise definition I have in mind is not actually appropriate for documentation.

PR #6971, opened 2025-09-22
Add MixedMeasureUnit struct for handling CLDR mixed units

younies:
# Description

This commit introduces:
* The `MixedMeasureUnit` struct, which allows for the representation of mixed measurement units, such as "foot-and-inch" and "kilometer-and-meter". 
* The implementation includes a method to create a `MixedMeasureUnit` from a string representation, ensuring that only valid combinations of single units are accepted. 
* Additionally, tests have been added to verify the functionality and error handling for invalid mixed units.

### DIFF
--- a/components/experimental/src/measure/mixed_measureunit.rs
+++ b/components/experimental/src/measure/mixed_measureunit.rs
@@ -1,0 +1,132 @@
+// This file is part of ICU4X. For terms of use, please see the file
+// called LICENSE at the top level of the ICU4X source tree
+// (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
+use crate::measure::{measureunit::MeasureUnit, parser::InvalidUnitError};
+
+use super::single_unit_vec::SingleUnitVec;
+
+/// The [`MixedMeasureUnit`] struct represents a CLDR mixed unit,
+/// which is a combination of one or more single units used together to express a measurement.
+///
+/// # Examples
+/// - `foot-and-inch`
+/// - `kilometer-and-meter`
+///
+/// Note: Compound units such as `meter-per-second` or units with a constant denominator are not supported in mixed units.
+#[derive(Debug, Eq, Clone, PartialEq)]
+pub struct MixedMeasureUnit {
+    /// Contains the single mixed units.
+    pub(crate) mixed_units: SingleUnitVec,
+}
+
+impl MixedMeasureUnit {
+    /// Returns a slice of the mixed units contained within this mixed unit.
+    pub fn try_from_str(mixed_units_str: &str) -> Result<MixedMeasureUnit, InvalidUnitError> {
+        let mixed_units_strs = mixed_units_str.split("-and-");
+        let mut mixed_units = SingleUnitVec::Zero;
+        for unit in mixed_units_strs {
+            let unit = MeasureUnit::try_from_str(unit)?;
+            let internal_single_units = unit.single_units();
+            if internal_single_units.len() > 1 {
+                return Err(InvalidUnitError);
+            }
+            if unit.constant_denominator != 0 {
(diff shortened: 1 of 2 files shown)


components/experimental/src/measure/mixed_measureunit.rs
```diff
@@ -0,0 +1,134 @@
...
+
+use crate::measure::{measureunit::MeasureUnit, parser::InvalidUnitError};
+
+use super::single_unit_vec::SingleUnitVec;
+
+/// The [`MixedMeasureUnit`] struct represents a CLDR mixed unit,
+/// which is a combination of one or more single units used together to express a measurement.
+///
+/// # Examples
+/// - `meter` - a special case of a mixed unit that contains only one single unit.
```
robertbastian: nit: don't list the special case as the first example

components/experimental/src/measure/mixed_measureunit.rs
```diff
@@ -0,0 +1,134 @@
...
+/// - `foot-and-inch`
+/// - `kilometer-and-meter`
+///
+/// Note: Compound units such as `meter-per-second` or units with a constant denominator are not supported in mixed units.
```
robertbastian: "units with a constant denominator" such as?

components/experimental/src/measure/mixed_measureunit.rs
```diff
@@ -0,0 +1,134 @@
...
+    /// Contains the single mixed units.
+    pub(crate) mixed_units: SingleUnitVec,
+}
+
+impl MixedMeasureUnit {
+    /// Returns a slice of the mixed units contained within this mixed unit.
+    pub fn try_from_str(mixed_units_str: &str) -> Result<MixedMeasureUnit, InvalidUnitError> {
+        // '-and-' is the separator for the mixed units and it is allowed to appear in the start or end of the string.
+        let mixed_units_strs = mixed_units_str.split("-and-");
+        let mut mixed_units = SingleUnitVec::Zero;
```
robertbastian, line 29: nit: `::Zero` should be `::Empty`?

components/experimental/src/measure/mixed_measureunit.rs
```diff
@@ -0,0 +1,134 @@
...
+mod tests {
+    use crate::measure::provider::{
+        si_prefix::{Base, SiPrefix},
+        single_unit::SingleUnit,
+    };
+
+    use super::*;
+
+    #[test]
+    fn test_mixed_measure_unit_from_str() {
```
robertbastian, line 59: issue: what does `kilometer-and-meter` test that `foot-and-inch` doesn't? what does `meter` test that `foot-and-inch` doesn't?

the reason why I think excessive testing like this is an issue is that these tests don't test the whole construct-format cycle, which is the actual API. instead, they test internal representation, which might need to be changed in the future, and then you need to update every single one of these test cases.

I'd prefer if you added the whole construct-format functionality in the same PR and tested that instead of the internal representation.

components/experimental/src/measure/mixed_measureunit.rs
younies: https://github.com/unicode-org/icu4x/pull/6973

components/experimental/src/measure/mixed_measureunit.rs
younies: done

components/experimental/src/measure/mixed_measureunit.rs
younies: done

components/experimental/src/measure/mixed_measureunit.rs
robertbastian: liter-per-100-kilometers falls under compound units, no? are there any units that have a purely constant denominator that this would apply to?

PR #6978, opened 2025-09-23
Remove unused dependencies

robertbastian:
This runs the `unused_crate_dependencies` lint in the `depcheck` job.

Unfortunately it's hard to globally enable this lint (i.e. in our lint boilerplate in each crate, or in `clippy-all`), as it triggers a lot of false positives for dev targets (examples/benches/tests) if they don't use all dependencies and all dev dependencies (https://github.com/rust-lang/rust/issues/95513). It also triggers with `-Zbuild-std` (https://github.com/rust-lang/rust/issues/122105), so the FFI jobs would fail.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1187,7 +1187,6 @@ name = "icu_benchmark_memory"
 version = "0.0.0"
 dependencies = [
  "clap",
- "libc",
  "serde_json",
 ]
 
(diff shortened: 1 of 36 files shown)


robertbastian:
The semver issue is not breaking as there is no functionality behind the `write16` feature.

Manishearth (approved):

PR #6991, opened 2025-09-26
Use `serde_core` where possible

robertbastian:


Manishearth (approved):
Great, a lot of utils decoupled from syn.

robertbastian:
can't fix GN because the scripts don't work on Mac. @sffc

Manishearth (dismissed):

sffc (dismissed):
Fantastic!

There's probably some crate somewhere that exports `macro_rules` based Serde impls, which may allow us to fix up even some component crates.

sffc:
We're pretty close in `icu_locale_core`, since we already have mostly manual impls. I see one derive, which I think is a pass-through derive that would be trivial to also implement manually.

PR #6993, opened 2025-09-27
Remove GN example and CI test

sffc:
I made this when we were prototyping Chrome/V8 integrations. These experiments have landed, so ICU4X compatibility with GN is now enforced in an actual real-world project using GN, so we don't need to test it here. Plus, the test is brittle (https://github.com/unicode-org/icu4x/pull/6991#issuecomment-3339239447).


Manishearth (approved):
yay

note: I have found people asking about gn integration in the past so the example is nice to point to but I guess we can just point people to chrome and say "gnaw works with icu4x". Note that chrome doesn't use gnaw they have a weird custom thing

PR #6999, opened 2025-09-29
Make `Rules` traits unstable

### LINKED ISSUE #6962: Should the Rules traits be marked UnstableSealed?
I'm unconvinced that these traits definitely have the shape that we will always want. We already found one gap, #6959.

The point of these traits is to make it _easier_ to implement these calendars in userland. _Crucially_, clients can _already do this_ by implementing the UnstableSealed trait Calendar.

We're not confident about the shape of Calendar, so why should we be any more confident about the shape of Rules?

If we mark these traits as UnstableSealed, it still allows users to implement them in userland, it's just that they commit to the unstable contracts that we already require in other contexts like `try_new_unstable`.

@robertbastian @Manishearth

robertbastian:
Fixes #6962

### DIFF
--- a/components/calendar/src/cal/chinese.rs
+++ b/components/calendar/src/cal/chinese.rs
@@ -79,12 +79,19 @@ pub struct LunarChinese<X>(pub X);
 ///
 /// For example, the [`China`] type implements the rules that are used in
 /// China, and the [`Dangi`] type implements the rules used in Korea.
-pub trait Rules: Clone + core::fmt::Debug {
+///
+/// <div class="stab unstable">
+/// 🚫 This trait is sealed; it should not be implemented by user code. If an API requests an item that implements this
+/// trait, please consider using a type from the implementors listed below.
+///
+/// It is still possible to implement this trait in userland (since `UnstableSealed` is public),
+/// do not do so unless you are prepared for things to occasionally break.
+/// </div>
+pub trait Rules: Clone + core::fmt::Debug + crate::cal::scaffold::UnstableSealed {
     /// Returns data about the given year.
     fn year_data(&self, related_iso: i32) -> LunarChineseYearData;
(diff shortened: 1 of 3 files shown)


sffc (commented):
Issue: The methods should also have `unstable` in the name or else be doc hidden.

robertbastian:
I'm following the precedent set by the calendar trait, which doesn't do either of that

robertbastian:
- @Manishearth Are they publicly callable?
- @sffc Better question are they stably callable?
- @robertbastian The way I understood our discussion last week was that we might need to add new methods, not change existing methods. If they are hidden, they can't be easily implemented.
- @sffc We've made mistakes, intentionally or unintentionally, of making UnstableTrait functions be stable, and then we can't change the signatures, which has constrained us. For example, some `icu_datetime` scaffolding traits have been problematic.
- @robertbastian Ideally we should have an unstable feature.
- @robertbastian I tried to make a PR but it didn't land.
- @Manishearth Are we on the same page that these should not be publicly callable?
- @robertbastian I'm not
- @sffc To me we need to graduate these to make them publicly callable
- @Manishearth Then we should mark the trait methods as doc hidden. `unstable` prefixes is a mandatory breakage.
- @robertbastian But then it is hard to implement. so we should just make it fully sealed
- @Manishearth You can still have a bound on the trait
- @sffc Any new trait or function should have an incubation period. I want people to be able to implement this in 2.1 and give us feedback, like for the Iran Hijri.
- @robertbastian I'm not sure we'll get feedback like that
- @Manishearth I think more important is what it means for the rest of icu4x, not for users. As long as users can implement this it's fine.
- @robertbastian Should we just have an experimental feature?
- @sffc Yes that also works (or unstable feature). As long as it is semver compatible.

Issue #6787, opened 2025-07-31
New property values for Unicode 17


sffc:
We may want to add things from https://unicode-org.atlassian.net/browse/ICU-23038 to ICU4X.

@robertbastian said: We don't have blocks or JoiningGroup, but we do need to add the LineBreak and Script values.

PR #7025, opened 2025-10-02
Update ICU data to 78

### LINKED ISSUE #7034: `icuexportdata` name changes in ICU 78
This was apparently discussed in https://docs.google.com/document/d/1CtPdH8E-UoEYP1IXGsPiyNATdyDpJeeCsgv6yFK2PrA/edit but I don't think anyone reached out to us.

Previous:

`https://github.com/unicode-org/icu/releases/download/{tag}/icuexportdata_{tag.replace('/', '_')}.zip`

Now:

`https://github.com/unicode-org/icu/releases/download/{tag}/icu4x-exportdata-{tag.replace("release-", "")}.zip`

(this is the pattern for regular releases, it's unclear to me how it will look for ad-hoc releases (like `icu4x/2025-05-21/77.x`). The document has some proposals but no decision)

@echeran

robertbastian:
#6787

Fixes #7034

### DIFF
--- a/tools/make/download-repo-sources/src/main.rs
+++ b/tools/make/download-repo-sources/src/main.rs
@@ -181,7 +181,9 @@ fn main() -> eyre::Result<()> {
         cached(&format!(
             "https://github.com/unicode-org/icu/releases/download/{}/icu4x-icuexportdata-{}.zip",
             SourceDataProvider::TESTED_ICUEXPORT_TAG,
-            SourceDataProvider::TESTED_ICUEXPORT_TAG.replace("release-", "").replace("icu4x-", "")
+            SourceDataProvider::TESTED_ICUEXPORT_TAG
+                .replace("release-", "")
+                .replace("icu4x-", "")
         ))
         .with_context(|| "Failed to download ICU ZIP".to_owned())?,
         expand_paths(ICUEXPORTDATA_GLOB, true),


components/collator/src/provider.rs
```diff
@@ -586,7 +586,7 @@ impl CollationSpecialPrimariesValidated<'static> {
           // Last primaries
           1286,
           3072,
-          3488,
```
robertbastian, line 589: @hsivonen is this change allowed?

hsivonen: I thought the point of this hard-coded data was that it's frozen to the state of CLDR that existed at ICU4X 2.0 such that we can fill in the values when datagen hasn't generated them, but now with newer CLDR and datagen, we should get these from data.

That is, I think this code change shouldn't be made, but extra care is needed for testing that values from new data take precedence.

robertbastian: some constructor fails when I don't change this, need to investigate then

robertbastian: > but now with newer CLDR and datagen, we should get these from data.

has any of this been implemented? no code path seems to try to get this from data, they all use the hardcoded value

Manishearth (dismissed):

components/collator/src/provider.rs
robertbastian: ah it  has

provider/source/src/lib.rs
```diff
@@ -234,16 +234,21 @@ impl SourceDataProvider {
...
+            format!(
+                "https://github.com/unicode-org/icu/releases/download/{tag}/icu4x-exportdata-{}.zip",
+                tag.replace("release-", "")
+            )
+        } else {
+            format!(
+                "https://github.com/unicode-org/icu/releases/download/{tag}/icuexportdata_{}.zip",
+                tag.replace('/', "-")
+            )
+        };
```
sffc: Suggestion:

- Keep the `/` to `-` substitution, in addition to the "release-" substitution
- Special-case "release-77.1" and any other releases that this datagen version is compatible with

components/collator/src/provider.rs
```diff
@@ -633,7 +609,10 @@ impl CollationSpecialPrimariesValidated<'_> {
         // into Compiler Explorer shows that the panic
         // is optimized away.
```
hsivonen, line 623: Is the above comment still true after this change?

components/collator/src/provider.rs
```diff
@@ -633,7 +609,10 @@ impl CollationSpecialPrimariesValidated<'_> {
...
         #[expect(clippy::indexing_slicing)]
-        let field = self.compressible_bytes[usize::from(b >> 4)];
+        let field = self
+            .last_primaries
+            .get(MaxVariable::Currency as usize + usize::from(b >> 4))
+            .unwrap_or_else(|| CollationSpecialPrimaries::HARDCODED_FALLBACK[usize::from(b >> 4)]);
```
hsivonen: To keep the above comment true, shouldn't the compressibility bits be referenced as a fixed-length array in the collator constructor instead of doing slice access on every call to this method?

hsivonen (changes requested):
AFAICT, this pessimizes the compressibility check that before this change was formulated to validate the length upon construction so that bound checking was elided at access time. Perhaps that's not super-important, but it still seems bad to break that property if we could avoid breaking that property.

components/collator/src/provider.rs
robertbastian: done

provider/source/src/lib.rs
sffc: Based on the agreement with the ICU TC, this change is no longer necessary.

PR #7027, opened 2025-10-02
More Hijri changes

robertbastian:
#6963

### DIFF
--- a/components/calendar/src/cal/hijri.rs
+++ b/components/calendar/src/cal/hijri.rs
@@ -33,6 +33,8 @@ mod ummalqura_data;
 /// There are many variants of this calendar, using different lunar observations or calculations
 /// (see [`Rules`]).
 ///
+/// This type supports only variants where months are either 29 or 30 days.
+///
 /// # Era codes
 ///
 /// This calendar uses two era codes: `ah`, and `bh`, corresponding to the Anno Hegirae and Before Hijrah eras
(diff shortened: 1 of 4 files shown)


components/calendar/src/cal/hijri.rs
```diff
@@ -260,6 +313,36 @@ impl Rules for TabularAlgorithm {
...
     }
 
+    fn ecma_reference_year(&self, month_code: types::MonthCode, day: u8) -> Result<i32, DateError> {
+        let Some((ordinal_month, false)) = month_code.parsed() else {
+            return Err(DateError::UnknownMonthCode(month_code));
+        };
+
+        Ok(match (ordinal_month, day) {
+            (1, _) => 1392,
+            (2, 30) => 1389,
```
Manishearth, line 323: issue: This is wrong, tabular calendars have "short" even months and "long" odd months (except for month 12, depending on the type of year).

So `(even, 30)` should never return a different number than `(even, 29)` unless it happens to be on the 1972 border

I think your compute code is broken.

Manishearth: https://github.com/boa-dev/temporal/pull/490/changes#diff-c80069b1554213a8876734963381635b585a009ea2c5e76de1fa5134a4d1d43bR195 has working values, though they're written slightly differently

robertbastian: this PR is only hardcoding the values that this code was already producing.

the code for this was added in https://github.com/unicode-org/icu4x/pull/6910/changes#diff-1faca0b96f658eda57e86d07a1603ab917b7411013c5f333051f053fd25500ca. that PR also added the `reference_year.rs` test, so that is broken as well?

robertbastian: (so it's not my compute code, it's @sffc's)

Manishearth: Got it. Either way, it's being fixed in https://github.com/unicode-org/icu4x/pull/7641

PR #7028, opened 2025-10-02
Allow different Hijri calendars to use different names

robertbastian:
#7005

### DIFF
--- a/components/datetime/src/scaffold/calendar.rs
+++ b/components/datetime/src/scaffold/calendar.rs
@@ -91,10 +91,65 @@ impl CldrCalendar for Indian {
     type SkeletaV1 = DatetimePatternsDateIndianV1;
 }
 
-impl<R: hijri::Rules> CldrCalendar for Hijri<R> {
+/// [`hijri::Rules`]-specific formatting options.
+/// 
+/// See [`CldrCalendar`].
+/// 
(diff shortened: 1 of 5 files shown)


sffc (approved):

components/datetime/src/scaffold/calendar.rs
```diff
@@ -91,12 +91,67 @@ impl CldrCalendar for Indian {
...
+/// }
+/// 
+/// impl FormattableHijriRules for MyRules {
+///     type YearNamesV1 = <hijri::UmmAlQura as FormattableHijriRules>::YearNamesV1;
+///     type MonthNamesV1 = <hijri::UmmAlQura as FormattableHijriRules>::MonthNamesV1;
+///     type SkeletaV1 = <hijri::UmmAlQura as FormattableHijriRules>::SkeletaV1;
+/// }
+/// ```
+// TODO: default associated types would be nice (https://github.com/rust-lang/rust/issues/29661)
+pub trait FormattableHijriRules: hijri::Rules {
```
sffc: UnstableSealed?

robertbastian: well actually `hijri::Rules` is `UnstableSealed`, do we need to use one per crate?

PR #7029, opened 2025-10-02
Clarify that the Hijri rule sets are ICU4X Hijri

sffc:
#6963

### DIFF
--- a/components/calendar/src/cal/hijri.rs
+++ b/components/calendar/src/cal/hijri.rs
@@ -95,7 +95,7 @@ pub trait Rules: Clone + Debug + crate::cal::scaffold::UnstableSealed {
     }
 }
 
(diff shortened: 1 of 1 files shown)


components/calendar/src/cal/hijri.rs
```diff
@@ -95,7 +95,7 @@ pub trait Rules: Clone + Debug + crate::cal::scaffold::UnstableSealed {
     }
 }
 
-/// An astronomical simulation for a particular location.
+/// [Hijri] rules based on an astronomical simulation for a particular location.
```
```suggestion
/// [`Hijri`] rules based on an astronomical simulation for a particular location.
```

```suggestion
/// [`Hijri`] [`Rules`] based on an astronomical simulation for a particular location.
```

PR #7036, opened 2025-10-06
Document proleptic behavior of the Japanese calendars in more detail

hsivonen:


components/calendar/src/cal/japanese.rs
```diff
@@ -17,7 +17,7 @@ use tinystr::tinystr;
 /// The months and days are identical to that of the Gregorian calendar, however the years are counted
 /// differently using the Japanese era system.
 ///
-/// This calendar only contains eras after Meiji, for all historical eras, check out [`JapaneseExtended`].
+/// The [Japanese calendar] was taken into use starting from 6 Meiji (1873 ISO/Gregorian), so years before 6 Meiji are proleptic. This implementation uses the Meiji era proleptically back to and including 1868-10-23 ISO and uses the Gregorian eras for 1868-10-22 ISO and earlier dates. For a variant that uses approximations of historical Japanese eras proleptically, check out [`JapaneseExtended`].
```
robertbastian: I would rephrase this. Japan has always used the "Japanese" calendar. this could say that Japan started using the Gregorian calendar, which this implements, then

robertbastian: nit: not a big fan of "yyyy-MM-dd *ISO*". it's already the ISO format, I don't think it needs to be called out like that

robertbastian:
```suggestion
/// The [Japanese calendar] was taken into use starting from 6 Meiji (1873 CE), so years before 6 Meiji are proleptic. This implementation uses the Meiji era proleptically back to and including 1868-10-23 ISO and uses the Gregorian eras for 1868-10-22 ISO and earlier dates. For a variant that uses approximations of historical Japanese eras proleptically, check out [`JapaneseExtended`].
```

robertbastian: "from 6 Meiji" - please give the exact date here instead of below

robertbastian: "was taken into use" is a bit awkward phrasing

hsivonen: I believe the exact date to be January 1st, 6 Meiji, based on OpenJDK docs, but Wikipedia doesn't state the day, so I'm shy to state the exact day explicitly, even though "6 Meiji" kinda suggests "from the start of 6 Meiji".

hsivonen: Changed to "adopted". (I used "starting" instead of "in", since logically the decision has been taken in an earlier year than the decision taking effect.)

hsivonen: Changed to CE.

hsivonen: Omitted "ISO".

hsivonen: Rephrased.

hsivonen: Furthermore, since the implementation is proleptic to the start of the Meiji era anyway, the exact actual date of use isn't that important for the docs.

hsivonen: I guess https://en.wikipedia.org/wiki/Adoption_of_the_Gregorian_calendar#Countries_that_used_lunisolar_calendars confirms January 1st.

hsivonen: Adjusted the text.

components/calendar/src/cal/japanese.rs
```diff
@@ -17,7 +17,7 @@ use tinystr::tinystr;
 /// The months and days are identical to that of the Gregorian calendar, however the years are counted
 /// differently using the Japanese era system.
 ///
-/// This calendar only contains eras after Meiji, for all historical eras, check out [`JapaneseExtended`].
+/// The calendar scheme using months and days identical to the Gregorian calendar was adopted starting with January 1st, 6 Meiji (1873-01-01), so years before 6 Meiji are proleptic. This implementation uses the Meiji era proleptically back to and including 1868-10-23 and uses the Gregorian eras for 1868-10-22 and earlier dates. For a variant that uses approximations of historical Japanese eras proleptically, check out [`JapaneseExtended`].
```
robertbastian: I'm not a fan of the description as a "calendar scheme using months and days identical to the Gregorian calendar". It is very much a Gregorian calendar, just with Japanese era names, and that's how Wikipedia for example describes it as well.

hsivonen: Tweaked wording.

PR #7044, opened 2025-10-07
Ignore tidy failure in beta, nightly

robertbastian:
https://github.com/rust-lang/rust/issues/147069


sffc:
I hope it's temporary; they gave the ticket P2 Medium priority which doesn't make it sound like it's likely to get fixed in time for 1.91 stable

PR #7053, opened 2025-10-09
Code to optimize the choice of new moon

sffc:
Trying to pick the best value for the new moon. Output of this test:

```
OptimizationResult:
    Solver:        BrentOpt
    param (best):  0.5961012372892227
    cost (best):   244.805785003407
    iters (best):  2
    iters (total): 6
    termination:   Solver converged
    time:          621.786µs
```

which equates to about 14:18 UTC on Jan 6, 2000, as compared to 18:14 on that day as we currently have.

CC @robertbastian

### DIFF
--- a/components/calendar/src/cal/chinese/simple.rs
+++ b/components/calendar/src/cal/chinese/simple.rs
@@ -34,6 +34,9 @@ const MEAN_GREGORIAN_SOLAR_TERM_LENGTH: Milliseconds =
 /// [Astronomical Almanac (1992)]: https://archive.org/details/131123ExplanatorySupplementAstronomicalAlmanac/page/n302/mode/1up
 const MEAN_SYNODIC_MONTH_LENGTH: Milliseconds = day_fraction_to_ms!(295305888531 / 10000000000i64);
 
+#[cfg(test)]
+const MEAN_SYNODIC_MONTH_LENGTH_F64: f64 = 29.5305888531;
+
 /// Number of milliseconds in a day.
 const MILLISECONDS_IN_EPHEMERIS_DAY: i64 = 24 * 60 * 60 * 1000;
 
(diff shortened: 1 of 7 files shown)


components/calendar/src/cal/chinese/simple.rs
```diff
@@ -160,3 +163,60 @@ impl super::LunarChineseYearData {
...
+        center_moment: Moment,
+        new_moon_number: i32,
+    }
+
+    impl CostFunction for NewMoons {
+        type Param = f64;
+        type Output = f64;
+        fn cost(&self, param: &f64) -> Result<f64, Error> {
+            let candidate = self.center_moment + *param;
+            Ok(self.new_moon_moments.iter().map(|(i, x)| (candidate - *x) - self.mean_synodic_length * (self.new_moon_number - i) as f64).map(|v| v * v).sum::<f64>())
```
robertbastian, line 188: this uses mean absolute error for the new moon dates. but the date of the new moon is not the only thing our algorithm approximates.

the cost should be the count of months where the simple approximation doesn't match the GB/T calculations for >2100 (and potentially the ground truth for <1900, weighted appropriately)

components/calendar/src/cal/chinese/simple.rs
```diff
@@ -160,3 +163,60 @@ impl super::LunarChineseYearData {
...
+        }
+    }
+
+    #[test]
+    fn calculate_new_moon_and_mean_synodic_month() {
+        let jan1900 = fixed_from_gregorian(1900, 1, 1);
+        let jan2000 = fixed_from_gregorian(2000, 1, 1);
+        let jan2100 = fixed_from_gregorian(2100, 1, 1);
```
robertbastian, line 196: this is the one range where the approximation isn't used, so why optimise over that?

components/calendar/src/cal/chinese/simple.rs
```diff
@@ -160,3 +163,60 @@ impl super::LunarChineseYearData {
...
+        let new_moon_0 = Astronomical::num_of_new_moon_at_or_after(Moment::try_from_rata_die(jan1900).unwrap());
+        let new_moon_mid = Astronomical::num_of_new_moon_at_or_after(Moment::try_from_rata_die(jan2000).unwrap());
+        let new_moon_n = Astronomical::num_of_new_moon_at_or_after(Moment::try_from_rata_die(jan2100).unwrap());
+        let new_moon_moments: Vec<(i32, Moment)> = (new_moon_0..=new_moon_n).map(|i| (i, Astronomical::nth_new_moon(i))).collect();
+
+        let cost_fn = NewMoons {
+            new_moon_moments,
+            mean_synodic_length: MEAN_SYNODIC_MONTH_LENGTH_F64,
+            center_moment: Moment::try_from_rata_die(fixed_from_gregorian(2000, 1, 6)).unwrap(),
```
robertbastian, line 205: issue: our `LocalMoment` is milliseconds-i64. optimising a `Moment`, which is day-f64, is not equivalent

components/calendar/src/cal/chinese/simple.rs
```diff
@@ -34,6 +34,9 @@ const MEAN_GREGORIAN_SOLAR_TERM_LENGTH: Milliseconds =
 /// [Astronomical Almanac (1992)]: https://archive.org/details/131123ExplanatorySupplementAstronomicalAlmanac/page/n302/mode/1up
 const MEAN_SYNODIC_MONTH_LENGTH: Milliseconds = day_fraction_to_ms!(295305888531 / 10000000000i64);
 
+#[cfg(test)]
+const MEAN_SYNODIC_MONTH_LENGTH_F64: f64 = 29.5305888531;
```
robertbastian, line 38: issue: this is not an exact version of `MEAN_SYNODIC_MONTH_LENGTH`

robertbastian:
> which equates to about 14:18 UTC on Jan 6, 2000, as compared to 18:14 on that day as we currently have.

18:14 *which actually happened*

if you want to do this, we should optimise the _choice_ of actual new moon, instead of inventing one

PR #7061, opened 2025-10-10
Fix month constraining in reference year

sffc:
#6790


sffc:
I'm going to merge this, but I would like https://github.com/tc39/proposal-temporal/issues/3163 to be resolved.

PR #7066, opened 2025-10-10
Update MSRV to 1.83

robertbastian:
In policy and needed for #7065


utils/zerotrie/src/builder/litemap.rs
```diff
@@ -46,7 +46,7 @@ where
     }
 }
 
-// TODO(MSRV 1.83): Make this more infallible by calculating the required length,
+// TODO(sffc): Make this more infallible by calculating the required length,
```
robertbastian, line 49: @sffc I don't know what this means

sffc:
This is in policy for ICU4X component crates. But, after releasing new versions of everything in 2.0, I think we should by default keep utility crates on the lower MSRV until the individual utils crates need the bump.

utils/zerotrie/src/builder/litemap.rs
sffc: https://github.com/unicode-org/icu4x/pull/7085

sffc:
Proposal:

- Set MSRV 1.82 in the Cargo.toml of all utils crates
- Allow individual utils crates to opt for a different MSRV when they need a new feature
- Utils crates can adopt the N-6 MSRV without needing any justification if they are undergoing a major release
- This policy has no impact on ICU component crates

sffc:
WG Discussion on the above policy (https://github.com/unicode-org/icu4x/pull/7066#issuecomment-3409265315):

- @Manishearth Seems fine. Let's update the policy to say that utils crates don't get automatically bumped, but can be bumped to the ICU4X MSRV without approval if they need the version for any reason.
- @zbranicki In large organizations, Rust versions don't get bumped often, so I'm happy with policies to strengthen low MSRV requirements.

Conclusion: yep, + policy edit

LGTM: @manishearth @sffc @zbraniecki

PR #7073, opened 2025-10-13
Remove `AnyDateInner::kind`

robertbastian:
This is only used in an unreachable panic.

Uncontroversial part of #7071


sffc:
Lgtm

PR #7109, opened 2025-10-16
Use PotentialUtf8 for era code in DateFields

sffc:
A study for #7104


sffc:
Backed up at https://github.com/sffc/omnicu/tree/archive/potential-date-fields

PR #7110, opened 2025-10-16
Add test for month code error types

sffc:
#7010


sffc:
The last commit in this PR can be resurrected and written to work with the Month type.

Manishearth:
Yeah that seems like a good idea

sffc:
Tests migrated to #7723

PR #7130, opened 2025-10-20
Consolidate constructor code in `ArithmeticDate`

robertbastian:

### DIFF
--- a/components/calendar/src/calendar_arithmetic.rs
+++ b/components/calendar/src/calendar_arithmetic.rs
@@ -191,6 +191,7 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
         day: u8,
         calendar: &C,
     ) -> Result<Self, DateError> {
+        let year = range_check(year, "year", VALID_YEAR_RANGE)?;
         let year = if let Some(era) = era {
             calendar.year_info_from_era(era.as_bytes(), year)?
         } else {
@@ -198,7 +199,13 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
         };
         let month = calendar
             .ordinal_month_from_code(&year, month_code, Default::default())
-            .map_err(|_| DateError::UnknownMonthCode(month_code))?;
+            .map_err(|e| match e {
+                MonthCodeError::InvalidMonthCode
+                | MonthCodeError::UnknownMonthCodeForCalendar
+                | MonthCodeError::UnknownMonthCodeForYear => {
+                    DateError::UnknownMonthCode(month_code)
+                }
+            })?;
 
         let day = range_check(day, "day", 1..=C::days_in_provided_month(year, month))?;
 


components/calendar/src/calendar_arithmetic.rs
```diff
@@ -169,13 +174,135 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
         ArithmeticDate { year, month, day }
     }
 
-    /// Construct a new arithmetic date from a year, month ordinal, and day, bounds checking
-    /// the month and day according to the overflow parameter.
-    pub(crate) fn try_from_builder(
-        builder: ArithmeticDateBuilder<C::YearInfo>,
+    pub(crate) const fn cast<C2: DateFieldsResolver<YearInfo = C::YearInfo>>(
```
sffc, line 177: Nit: This doesn't seem like a safe cast because we use `i32` as the YearInfo in most calendars, but the i32 has different epochs.

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -169,13 +174,135 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
...
+        // to return structural type errors (`NotEnoughFields`) before checking for range errors.
+        // This isn't behavior we *must* have, but it is not much additional work to maintain
+        // so we make an attempt.
+        let year = match (fields.era, fields.era_year) {
+            (None, None) => match fields.extended_year {
+                Some(extended_year) => calendar.year_info_from_extended(range_check(
+                    extended_year,
+                    "year",
+                    VALID_YEAR_RANGE,
+                )?),
```
sffc, line 258: Suggestion (optional): @Manishearth had factored this code into a separate function since it gets called in multiple places here, I think in order to make the range check be more consistent. I prefer to restore the helper function.

(process comment: this is a style concern where reasonable people could draw different conclusions as to whether it's better to inline the function or not so I won't block on it)

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -169,13 +174,135 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
...
+        calendar: &C,
+    ) -> Result<Self, DateError> {
+        let year = if let Some(era) = era {
+            calendar.year_info_from_era(era.as_bytes(), year)?
+        } else {
+            calendar.year_info_from_extended(year)
+        };
+        let month = calendar
+            .ordinal_month_from_code(&year, month_code, Default::default())
+            .map_err(|_| DateError::UnknownMonthCode(month_code))?;
```
sffc: Nit: Match on the error, in case the error variants for the internal function are refactored in the future. (ok for follow up)

components/calendar/src/calendar_arithmetic.rs
robertbastian: the cast here is for things like `Ethiopian`: it parses into a YMD tuple using a Coptic year, but the type is `ArithmeticDate<Ethiopian>`, so we need to cast to store that as a `ArithmeticDate<Coptic>`. 

we could avoid this is we made `ArithmeticDate` *not* be generic in a `CalendarArithmetic`, but instead generic in a `YearInfo`. Then the constructors would each be generic, and we'd store `ArithmeticDate<i32>` or `ArithmeticDate<HijriYearData>`.

components/calendar/src/calendar_arithmetic.rs
robertbastian: I inlined this because the range check for the era year is already inlined, so now all three checks are here

components/calendar/src/calendar_arithmetic.rs
sffc: I thought we handled the Ethiopian case by having an EthiopianDateInner wrap a CopticDateInner?

robertbastian: yes, and we need an `ArithmeticDate<Coptic>` to put in the `CopticDateInner`

sffc (approved):

PR #7131, opened 2025-10-20
Improve `AnyCalendar` docs

### LINKED ISSUE #7097: Would like to see docs tests illustrating how you can't compare DateAny across calendars
Would like to see docs tests illustrating how you can't compare DateAny across calendars

_Originally posted by @sffc in https://github.com/unicode-org/icu4x/pull/7090#pullrequestreview-3336853833_

robertbastian:
Fixes #7097


sffc (commented):
This is sufficient, thanks

Issue #7111, opened 2025-10-16
Bikeshed name LunarChinese


sffc:
I think this was mentioned in comment threads or maybe on Slack but it never had an official issue.

There are pros and cons to the name `LunarChinese`. Let's get a list of alternatives and run a ballot.

sffc:
My understanding of the current situation:

It was asserted that the name "lunar chinese" reflects the type of calendar structure, and there is not another well-known name for that calendar structure. The original calendar structure was invented in China, not in Korea.

The main pushback has been that the word "chinese" favors China over other regions that also use the calendar system.

I will use a Large Language Model for a task that it is good at: naming things.

<details>
(start of conversation)

## Shane

Tell me about the traditional calendar used in China, the traditional calendar used in Korea, and how they relate to each other.

## LLM

The traditional calendars used in China and Korea, while having distinct cultural names and some unique observances, are fundamentally the same type of calendar system and are deeply related.

Here is a breakdown of the traditional calendars and their relationship:

### The Traditional Calendars

Both the traditional Chinese and Korean calendars are **lunisolar calendars**. This means they are based on a combination of:

* **Lunar Cycles:** The months are aligned with the phases of the moon. Typically, a month starts on the day of the astronomical new moon.
* **Solar Year:** The calendar is periodically synchronized with the solar year (the Earth's orbit around the sun) to keep the seasons in line. This synchronization is achieved by adding an extra month (a "leap month" or intercalary month) about seven times every 19 years.

**Key Features Shared by Both:**

* **Lunisolar Nature:** Both calendars use the lunisolar system.
* **Solar Terms:** Both systems use 24 solar terms (known as *Jiéqì* in Chinese or *Jeolgi* in Korean) to mark changes in season, weather, and agricultural activities throughout the year.
* **Sexagenary Cycle:** Both utilize the 60-year cycle (combining the 10 Heavenly Stems and 12 Earthly Branches, which are associated with the 12 Zodiac animals) for naming years and other time units.
* **Purpose:** While the **Gregorian calendar** is now used for official and daily purposes in both countries, the traditional lunisolar calendar remains vital for dating traditional holidays, festivals (like Lunar New Year and Mid-Autumn Festival), and for calculating auspicious dates for events like weddings.

### Relationship Between the Calendars

The traditional Korean calendar is historically **derived from and heavily influenced by the traditional Chinese calendar**.

[...]

PR #7143, opened 2025-10-22
Deprecate convenience constructors of lunisolar calendars

robertbastian:
The fact that they use ordinal months is a footgun. They should be replaced by constructors taking `month_code: &str` in the future, but that's not ready yet, so `try_new_from_codes` can be used in the meantime.

Related #7111

Related #7095

### DIFF
--- a/components/calendar/benches/date.rs
+++ b/components/calendar/benches/date.rs
@@ -152,7 +152,10 @@ fn date_benches(c: &mut Criterion) {
         "calendar/chinese_cached",
         &fxs,
         icu::calendar::cal::LunarChinese::new_china(),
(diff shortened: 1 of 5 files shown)


Manishearth (approved):
I have a very very slight preference for doing this we have the replacement ready. I'm comfortable doing it early for chinese (https://github.com/unicode-org/icu4x/issues/7111) because we're churning that area anyway.

But I think the ordinal month thing is just a big footgun which makes this API hard to use. I find myself using the Hijri convenience constructors a lot, but never Hebrew, since I don't want to hit the footgun.

robertbastian:
can you be more specific?

sffc:
#7156

PR #7146, opened 2025-10-22
Align and clean up Hijri and Chinese code

robertbastian:
These calendars are very similar code-wise, using a `PackedXYearData` for month lengths. Hijri has better implementations for RD, so using those for Chinese as well. Also removing indirection by removing trivial methods on `XYearData` that just delegate to `PackedXYearData`.


sffc (commented):
Defer to @Manishearth on the full review. Seems good but make sure that the numbers are handled correctly.

Manishearth (approved):
I liked having this split up into smaller subroutines like md_from_rd as a way of being self-documenting, but this code is also just simpler, so still an improvement.

PR #7162, opened 2025-10-23
Partially back out #6755

robertbastian:
The current approach has both false positives and false negatives:

* `Australia/Brisbane` formats as "Eastern Australian Time" instead of "Eastern Australian Time (Brisbane)", because it has used DST in the past (and the current logic doesn't detect when DST is abolished). I don't consider this blocking, as it's the same behaviour as before #6755.
* `Asia/Pyongyang` formats as "Korea Time (North Korea)" instead of "Korea Time", because the golden zone `Asia/Seoul` has used DST in the past. This I consider launch blocking because it doesn't make sense.

I don't have time to implement the correct algorithm for 2.1, it's quite involved.


Manishearth (dismissed):

sffc:
I don't consider excessively specific formats like "Korea Time (North Korea)" to be blocking

PR #7186, opened 2025-10-28
Omit the trailing NO_CE when iterating over the case level buffer

### LINKED ISSUE #7145: Collator::write_sort_key_to assertion failure with CaseLevel::ON
When using `Collator` with `case_level` option set to `CaseLevel::ON`, then the `write_sort_key_to` function will fail on the debug assertion on line https://github.com/unicode-org/icu4x/blob/eee596432397461a55d4c19844b67d927d45bfaa/components/collator/src/comparison.rs#L2303

Comparing with ICU4C implementation, I suspect this is because the icu4x code also includes the trailing `NO_CE`. 

The icu4c has this: https://github.com/unicode-org/icu/blob/892d4498b9b70f36c42c0b6c87244b9dca89ca62/icu4c/source/i18n/collationkeys.cpp#L640C46-L640C76 

But it is not present in the icu4x implementation.

hsivonen:
ICU4C does this, but and the inclusion of the NO_CE in iteration was an oversight in porting.

Fixes #7145

### DIFF
--- a/components/collator/src/comparison.rs
+++ b/components/collator/src/comparison.rs
@@ -2306,15 +2306,20 @@ impl CollatorBorrowed<'_> {
 
             // Write pairs of nibbles as bytes, except separator bytes as themselves.
             let mut b = 0;
-            for c in &cases.buf {
-                debug_assert_eq!(*c & 0xf, 0);
-                debug_assert_ne!(*c, 0);
-                if b == 0 {
-                    b = *c;
-                } else {
-                    sink.write_byte(state, b | (*c >> 4))?;
-                    b = 0;
+            if let Some((last, head)) = cases.buf.split_last() {
+                debug_assert_eq!(*last, 1); // The trailing NO_CE
+                for c in head {
+                    debug_assert_eq!(*c & 0xf, 0);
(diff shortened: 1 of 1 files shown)


hsivonen:
CC @rs-sac

hsivonen:
See the issue for a link to the corresponding ICU4C code.

components/collator/src/comparison.rs
```diff
@@ -2306,15 +2306,20 @@ impl CollatorBorrowed<'_> {
...
-            for c in &cases.buf {
-                debug_assert_eq!(*c & 0xf, 0);
-                debug_assert_ne!(*c, 0);
-                if b == 0 {
-                    b = *c;
-                } else {
-                    sink.write_byte(state, b | (*c >> 4))?;
-                    b = 0;
+            if let Some((last, head)) = cases.buf.split_last() {
+                debug_assert_eq!(*last, 1); // The trailing NO_CE
```
robertbastian, line 2310: observation: the C++ code just ignores the `NO_CE`, we `debug_assert!` on it which I guess is better

PR #7224, opened 2025-11-06
Add `Julian` to `AnyCalendar`

robertbastian:
I want be able to write code that is generic over all calendars. We include more pointless calendars than Julian, and the code size for Julian is neglible (compared to e.g. simulated Hijri). Julian is the only calendar that is not in `AnyCalendar`.

## Changelog

`icu_calendar`: Add `Julian` to `AnyCalendar`


Manishearth (changes requested):
Not in favor, this was a deliberate omission. Our bar for AnyCalendar is higher than our bar for "is in the crate"/

If this type were to be made today the RGSA and Japanext entries would have been removed too.

sffc:
WG notes:

- @robertbastian Julian is the only calendar in icu_calendar not in AnyCalendar. There are calendars in there with more code size. We can't write code generically over all calendars.
- @Manishearth That's intentional; the bar for adding to AnyCalendar is high.
- @nekevss If something is added to AnyCalendar, does it get reflected in V8?
- @robertbastian There are already calendars in AnyCalendar that aren't in V8
- @nekevss We could expose the calendars in temporal_rs even if not in the engines
- @sffc I thought JapaneseExtended wasn't in AnyCalendar. And HijriSimulated
- @robertbastian No they're all there
- @sffc I thought we removed them
- @robertbastian No
- @manishearth the history here is that RGSA was never really supported, and japanext
- @robertbastian I have a need for Julian in AnyCalendar in order to write code that tests all calendars. What is the cost exactly?
- @hsivonen islamic-rgsa isn't in AnyCalendar; Simulated is. They both shouldn't be exposed to the Web. The two calendars in AnyCalendar that aren't Web calendars are JapaneseExtended and HijriSimulated, and I agree they shouldn't be there.
- @hsivonen About cost, there's a technical thing and an accidental Web exposure. If it's the case that referring to AnyCalendar means LTO can't remove the code for calendars in AnyCalendar, then an implementation of Temporal ships that code. (But would that imply calling it TemporalCalendar?)
- @hsivonen Why not Julian, which is well known? First, it's not in CLDR, which is a signal. We know Julian is exceptionally well-known for historical purposes. We know that there's a large population that uses it for religious purposes, but it isn't used for civil purposes.
- @hsivonen If we add it to AnyCalendar, there's a risk it gets exposed to the Web, because people who write glue code shouldn't need to be experts on what to expose and what not to expose. That's how we've gotten things like Japanese with extended eras being exposed. @nekevss is asking that question already even today. So we shouldn't put things here without use cases.
- @Manishearth The decision should be about user needs, not cost. Writing tests isn't a user need. About code size cost, it's not a lot, but we shouldn't add things that we want to remove. I'm also considering patching out these calendars in V8. As an actual client need, needing more patches is bad. About writing tests, we can write something that takes generics, or something that takes a trait object? There are ways to do it.
- @sffc For testing, can we make Calendar trait objects? (Rob: no.) Or make a macro?
- @sffc Julian seems motivated because it allows making dates before Gregorian switchover.
- @robertbastian I don't think all AnyCalendars are civil. Coptic, HijriTabular, etc.
- @hsivonen Coptic is used in Egypt for farming, I think? HijriTabular has a lot of motivation based on extended conversation.
- @robertbastian Temporal_rs could expose its own enum with the Temporal-specific use case.
- @robertbastian ICU implements the Julian transition.
- @Manishearth I would rather have this be an async discussion.
- @nekevss I was mentioning V8 earlier because it has a glue layer. But I think that question is answered.
- @Manishearth It's something that should be carefully considered. It's an informal public contract
- @sffc We have FormattableAnyCalendar in icu_datetime, which is the one that doesn't have JapaneseExtended
- @sffc I think Julian is motivated for AnyCalendar. But not sure that AnyCalendar is the long-term solution for testing.
- @hsivonen We shouldn't use ICU4X testing as a motivational use case.
- @hsivonen For doing Julian/Gregorian switchover calculation, why would you want AnyCalendar in-between?
- @robertbastian Because of why we have AnyCalendar. I think it's not reasonable for it to just exclude Julian.
- @robertbastian AnyCalendar currently is not the set of all calendars, it is not the set of all calendars for the Web, it is not the set of all calendars for formatting.
- @robertbastian 
- @nekevss I don't think you need to restrict AnyCalendar to what is available on the Web
- @sffc I do think we should be more clear on the definition of AnyCalendar. I agree with @robertbastian that it seems ill-defined right now.

Next steps:

- @robertbastian to file an issue to discuss AnyCalendar's definition
- List out all of the use cases for having a multi-calendar enum, and what the solutions are for those cases

sffc:
I support doing this; is it ready for a technical review?

robertbastian:
yes

hsivonen:
Why? Adding this to `AnyCalendar` wouldn't be making `AnyCalendar` match Temporal.

robertbastian:
It's part of the discussion in https://github.com/unicode-org/icu4x/issues/7320. We have not decided that `AnyCalendar` should be the Temporal set.

sffc (approved):
Would like some more tests, maybe an example showing how to use this to manually implement a Julian/Gregorian switchover.

Before we merge this though we should get alignment from the TC on the goals of AnyCalendar.

Manishearth (approved):
Architecturally sound.

Goals-wise, as Shane said we should get alignment on AC. The codesize cost is less, but on the other hand our active clients (Firefox via Henri, to some extent me, though codesize mitigates most of my worries) would prefer AC be Temporal-focused.

The way I see it, the "more useless" calendars we include are included because they have been uplifted by BCP-47/CLDR/Temporal. Julian has no such dynamic in its favor.

That said, Julian is a useful calendar to have, and does not really need many caveats to be used meaningfully (unlike the other calendars under debate). The switchover thing is confusing, but it's pretty unambiguous when you are using a "julian" calendar that there is no switchover (it's actually more ambiguous for Gregorian!).

hsivonen:
> That said, Julian is a useful calendar to have, and does not really need many caveats to be used meaningfully (unlike the other calendars under debate).

There does appear to be yet another caveat for the "Historical Accuracy" docs that applies to whether the calendar can be used directly for day-accurate mapping of historical dates Feb 24 through Feb 28, inclusive, in a leap year in region / historical period combinations that had a particularly [impractical interpretation](https://en.wikipedia.org/wiki/Bissextus) of what it means for Feb 24 to be the leap day.

(To be clear, I think we should not add support for non-month-final leap days to the ICU4X calendar model.)

hsivonen:
> pretty unambiguous when you are using a "julian" calendar

Except when saying "Julian" to mean ["Revised Julian"](https://en.wikipedia.org/wiki/Revised_Julian_calendar) and actually using Gregorian calendar software until 2800, since Revised Julian does not get software implementations. 

(To be clear, I think we should not provide a Revised Julian implementation.)

sffc (approved):
I was going to leave a comment asking you to add a docs test, but I added one myself.

robertbastian:
I'm not sure we should have such an extended example on just a single `AnyCalendar` variant. We don't have docs tests for any of the `AnyCalendar` variants, so I don't understand why you see this as a requirement for this PR.

sffc (approved):

Issue #14, opened 2020-03-24
String encodings (UTF-8/UTF-16)


sffc:
We want to make OmnICU able to be ported to other environments where strings are not necessarily UTF-8.  For example, if called via FFI from Dart or WebAssembly, we may want to process strings that are UTF-16.  How should we deal with strings in OmnICU that makes it easy to target UTF-16 environments while not hurting performance in UTF-8 environments?

macchiati:
That's a tricky topic, because of the indexing and performance
considerations. For (logically) immutable strings, perhaps the most general
scheme is something like Swift pre-version-5, and have a String object that
provides "views" of the object as either UTF-8, UTF-16, or UTF-32 (code
points). Internally it could use something like
https://swift.org/blog/utf8-string/#breadcrumbs for performance. But this
just scratches the surface of the topic.

Mark

zbraniecki:
@Manishearth @hsivonen

sffc:
I was thinking one option here could be that we have a file like,

```rust
// typedefs.rs

// Rust UTF-8 environment
#[cfg(...)]
pub use std::string::String;

// UTF-16 environment
#[cfg(...)]
pub use omnicu::types::U16String as String;
```

And then, all public OmnICU APIs should import the String symbol from *typedefs.rs*, and the definition varies by configuration option.  As long as *omnicu::types::U16String* is duck-type-compatible with std::string::String, then we should be OK, and we don't cause any performance regressions in Rust environments.

Manishearth:
IMO it would be better to use a trait than use cfgs

hsivonen:
I expect in Gecko whether we want to perform a given operation on UTF-8, on UTF-16, or, depending on caller, both, will vary over time. In the case of `encoding_rs`, Gecko is pretty firmly in the "both" situation, but for specific operations to be provided by OmnICU, it could easily be one or the other for a given operation.

I also expect there will be some operations where the Gecko caller wants UTF-16 for the time being but performing a conversion at the boundary and running the operation on UTF-8 inside OmnICU would be acceptable. When writing our conversions between UTF-16 and UTF-8, it's been a goal of mine to make it so fast that people wouldn't shy away from using UTF-8 for new code just because the new code needs to be called by old code that has UTF-16 at hand.

That is to say:

* I think having having one big UTF-16 vs. UTF-8 build config flag isn't a good match for Gecko.
[...]

PR #7232, opened 2025-11-09
Fix line breaking for Khmer text (issue #7218)

### LINKED ISSUE #7218: Double Line Break Opportunities Around Spaces in Khmer Text
## Summary

The `icu_segmenter` crate (version 2.1.1) incorrectly places line break opportunities **both before and after** space characters in Khmer text, creating consecutive invalid break points that isolate spaces as standalone segments.

## Environment

- **Crate**: `icu_segmenter` v2.1.1
- **Features**: `compiled_data`
- **Rust version**: (rustc 1.91.0)
- **Locale**: Khmer (`km`)

## Expected Behavior

Line break opportunities should appear **after** space characters only, following UAX #14. Spaces should remain attached to the preceding word or be consumed as part of the break opportunity, not isolated as standalone segments.

For the text `"អស់ នឹង មាន"`, expected breaks would be:

```
Position 10: after space → "អស់ " | "នឹង"
Position 20: after space → "អស់ នឹង " | "មាន"
```

## Actual Behavior

The segmenter creates **two consecutive break points** around each space character:

```rust
Position 9:  before space → "អស់" | " នឹង"
Position 10: after space  → "អស់ " | "នឹង"
Position 19: before space → "អស់ នឹង" | " មាន"
Position 20: after space  → "អស់ នឹង " | "មាន"
```

This creates segments where a single space character is isolated between two break opportunities.

### Code

```rust
use icu_segmenter::LineSegmenter;
use icu_segmenter::options::LineBreakOptions;
use icu_locale_core::{locale, LanguageIdentifier};

fn main() {
    let khmer = "អស់ នឹង មាន ចន្លោះ។";
    let km_locale: LanguageIdentifier = locale!("km").into();

    let mut options = LineBreakOptions::default();
    options.content_locale = Some(&km_locale);

    let seg = LineSegmenter::new_auto(options);
    let breaks: Vec<usize> = seg.segment_str(khmer).collect();

    println!("Breaks: {:?}", breaks);

    // Show the problem: consecutive positions around spaces
    for window in breaks.windows(2) {
        let segment = &khmer[window[0]..window[1]];
        println!("[{}->{}] '{}'", window[0], window[1], segment);
    }
}
```

### Cargo.toml

```toml
[dependencies]
icu_segmenter = { version = "2.1.1", features = ["compiled_data"] }
icu_locale_core = "2.1.1"
```

### Output

```
Breaks: [0, 9, 10, 19, 20, 29, 30, 48, 52, 61, 70, 71, 80]

[0->9] 'អស់'
[9->10] ' '        ⚠️ ISOLATED SPACE
[10->19] 'នឹង'
[19->20] ' '       ⚠️ ISOLATED SPACE
[20->29] 'មាន'
[29->30] ' '       ⚠️ ISOLATED SPACE
[30->48] 'ចន្លោះ'
[48->52] '។ '
[52->61] 'អស់'
[61->70] 'នឹង'
[70->71] ' '       ⚠️ ISOLATED SPACE
[71->80] 'មានចន្'
```

## Verification Against ICU4C

The official ICU segmentation demo does **NOT** show this behavior:
https://icu4c-demos.unicode.org/icu-bin/icusegments#2/km_KH

Testing the same Khmer text on the ICU4C demo shows correct single break opportunities after spaces, not the double breaks observed in the Rust implementation.

## Scope

This bug affects:

1. **All segmenter modes**: LSTM, Dictionary, and Auto all show identical incorrect behavior
2. **With and without locale**: The bug appears regardless of whether Khmer locale is specified
3. **Real-world impact**: This affects applications like Typst and any other Rust software using `icu_segmenter` for Khmer text

## Additional Observations

- Tested with all four segmentation methods:
  
  - `new_lstm()` - Shows bug
  - `new_dictionary()` - Shows bug
  - `new_auto()` - Shows bug
  - No locale (default) - Shows bug
- The double-break pattern occurs consistently at every space in the text
  
- The behavior differs from ICU4C, suggesting an issue specific to the Rust port
  

## Potential Causes

Since ICU4C works correctly but ICU4X doesn't, the issue likely lies in:

1. Break position calculation logic specific to the Rust implementation
2. Post-processing of break iterator results
3. Khmer-specific compiled data generation or loading
4. Incorrect handling of the Unicode Line_Break property for spaces in Khmer context

## Workaround

Users can filter consecutive breaks as a temporary solution:

```rust
fn filter_double_breaks(breaks: &[usize], text: &str) -> Vec<usize> {
    let mut filtered = Vec::new();
    let mut i = 0;

    while i < breaks.len() {
        let current = breaks[i];

        if i + 1 < breaks.len() {
            let next = breaks[i + 1];

            // Check if segment between current and next is a single space
            if next == current + 1 && text.as_bytes().get(current) == Some(&b' ') {
                filtered.push(next); // Keep break after space, skip before
                i += 2;
                continue;
            }
        }

        filtered.push(current);
        i += 1;
    }

    filtered
}
```

## Related

- This may be related to how spaces are classified in the Line_Break property for Southeast Asian scripts
- Similar patterns might exist for Thai, Lao, or Myanmar if they use similar space handling

## Request

Please investigate and fix the double-break issue for Khmer (and potentially other Southeast Asian languages). This bug causes incorrect line breaking behavior that affects text layout and rendering quality.

kenton-r:
## Problem
Spaces in Khmer text create isolated segments with breaks before AND after each space:
- Input: `"អស់ នឹង មាន"`
- Output: `['អស់', ' ', 'នឹង', ' ', 'មាន']` ❌ (isolated spaces)

## Solution Overview
Fix two issues that prevented proper space handling:
1. **language.rs**: Spaces were split into separate chunks → complex segmenter never saw them
2. **line.rs**: Complex segmenter wasn't triggered for SA×SPACE×SA sequences

### `components/segmenter/src/complex/language.rs`

**What changed**: Don't split text on whitespace characters

**Lines ~62 & ~105**: Modified both UTF-8 and UTF-16 iterators to skip whitespace when checking for language changes

**Effect**: Khmer phrases with spaces stay together as one chunk: `"អស់ នឹង"` instead of `"អស់"`, `" "`, `"នឹង"`

---

### `components/segmenter/src/line.rs`

**What changed**: Handle SA×SPACE×SA (complex script + space + complex script) sequences

**4 changes**:

1. **~Line 1070**: Add `peek_past_spaces_for_sa()` helper
   - Looks ahead past consecutive spaces to check if SA continues

2. **~Line 880**: Extend complex breaking trigger  
   - Changed from: only trigger for `SA × SA`
   - Changed to: trigger for `SA × SA` OR `SA × SPACE × SA`

3. **~Line 908**: Suppress UAX#14 breaks
   - Don't break at `SA × SP` if SA continues after space(s)

4. **~Lines 1165 & 1198**: Include spaces in text collection
   - Complex segmenter sees full phrases with spaces: `"អស់ នឹង"`

**Effect**: Complex segmenter (LSTM/dictionary) handles the entire SA×SPACE×SA sequence intelligently

---

## Result
- Before: `[0, 9, 10, 19, 20, ...]` (double breaks)
- After: `[0, 9, 19, 29, ...]` (single breaks)
- Spaces properly included with words: `['អស់', ' នឹង', ' មាន']` ✅

## Impact
Fixes line breaking for Khmer. Also possibly Thai, Lao, and Myanmar scripts. 
Matches ICU4C behavior.

---

## Changelog

`icu_segmenter`:
* Fix line breaking for complex scripts (Khmer, Thai, Lao, Myanmar) to properly handle spaces within SA segments, matching ICU4C behavior (fixes #7218)


sffc:
Thanks for the contribution!

Please add tests for this behavior. Also, please assert that _word break_ continues to break around spaces, and only _line break_ gets the new behavior.

sffc (commented):
Please add tests.

makotokato:
UTF-16 and UTF-8 result are different on current fix.  Also, I need more works to write tests completely.

sffc:
Thanks for adding tests.

@makotokato @aethanyc PTAL.

makotokato:
(I will look this at this weekend)

components/segmenter/src/complex/mod.rs
robertbastian, line 1: issue: this PR should not required changes to the `complex` module, which is also used by the word segmenter, as well as by an in-progress full rewrite of the line segmenter

components/segmenter/src/line.rs
```diff
@@ -127,6 +127,95 @@ const ZW: u8 = 53;
...
+///   marks the elided continuation of the preceding word/phrase.
+/// * **U+0EC6** `ໆ` LAO KO LA — Lao repetition mark, analogous to
+///   THAI MAIYAMOK.
+/// * **U+0EAF** `ຯ` LAO ELLIPSIS — Lao abbreviation marker, analogous
+///   to THAI PAIYANNOI.
+/// * **U+17D7** `ៗ` KHMER SIGN LEK TOO — Khmer repetition mark.
+///
+/// This is a typographic correction layered on top of the dict/LSTM
+/// output, not a UAX #14 rule. It is applied only inside the line-
+/// break pipeline; the word segmenter (which uses the word-boundary
```
robertbastian: this seems like a workaround for a shortcoming of the LSTM/dictionary models. should the models be updated to do this correctly? can the code be moved into the `complex` module?

why are you not applying this to the word segmenter? wouldn't it suffer from the same issue?

robertbastian (commented):
I'm in the process of rewriting the line segmenter, and my implementation doesn't have this issue. I don't know if it's worth investing more time into this PR

sffc:
It seems like there are multiple problems being fixed here. If there are issues _inside_ the complex breaker, should we improve the dictionary or the training data for the LSTM?

kenton-r:
Fundamentally the word segmenter is supposed to mark break points around all words, spaces, and punctuation. This is the correct behavior for a word segmenter.

However this is not the default behavior needed for a line segmenter. There are to be breaks only after spaces. Some punctuation marks have breaks after and others before. These are defined with UAX#14. 

There is no need to have UAX#14 awareness within a word segmenter, but it is needed within a line segmenter. Otherwise you will need to be filtering out all of the inappropriate breaks when you call to a word segmenter to get line segmentation.

My thought was that having a dedicated line segmenter route that was UAX#14 aware and did not include the confusion caused by a wrod segmenter was the better way to handle this.

kenton-r:
For the extra Thai, Lao, Khmer characters that are not handled correctly by the LSTM/dictionary models. It would be better to add that to them than just a workaround.

makotokato:
Basically, hard coding rules isn't good fix.

Proposal to fix this is,
1. `Y::line_handle_complex_language(self, left_codepoint)` (https://github.com/unicode-org/icu4x/blob/main/components/segmenter/src/line.rs#L850) fills `self.result_cache`. So if `result` is returned, you should check `result_cache` size.  If size is 1, push `AL` to `lb9_left` then `continue 'a`
2. When using `self.result_cache` at first (https://github.com/unicode-org/icu4x/blob/main/components/segmenter/src/line.rs#L811) in `next()`, if `result_cache` is last one, push `AL` to `lb9_left` then `continue`.

[kenton-r](https://github.com/kenton-r), could you try my proposal?

kenton-r:
Thanks for your efforts in rewriting the line segmenter. I want to make sure whichever solution is finalized on the break points for spaces and punctuation for Southeast Asian languages are handled correctly. 

### UAX#14 Table Driven
Thank you for the idea of using the table driven approach. I tried your suggestion (replace `lb_class_forbids_break_before(class)` with `matches!(data.get_break_state_from_table(AL, right_prop), BreakState::Keep)` inside the existing filter closure, using LB1's `SA→AL` resolution). It looks interesting on paper, but it fails three regression tests in this PR for reasons that are worth recording:

1. LB28 over-suppresses at script transitions. `(AL | HL) × (AL | HL)` was written for within-script letter sequences (`a+b` inside one English word). When LB1 resolution makes a Thai→Latin boundary look like `AL × AL` to the table, LB28's blanket no-break forbids the dict's meaningful script-transition break. Test `complex_script_letter_keeps_break_before` (`ภาษาไทยabc` → expect a break before `abc`) fails under the table-driven version. Same problem for `AL × HL` and `AL × NU` via LB23.

2. LB19a under-suppresses opening quotes. Unicode 15.1 split the old `× QU` rule into context-dependent LB19a/LB19b variants, where `× QU_PI` now requires a specific East-Asian left context that AL does not satisfy. The table therefore allows line breaks immediately before `"`, `"`, `«`, etc. when the preceding (resolved) class is AL. Typographically this is wrong — opening punctuation attaches to what follows, a line shouldn't end with ". Tests `complex_script_punctuation_no_break_before` (auto) and `complex_script_punctuation_no_break_before_lstm` (LSTM) fail under the table-driven version.

So the takeaway is that the hand-list in this PR isn't a duplicate of UAX #14 — it's a typographic-attachment classifier for the SA→next-character seam, which is a position UAX #14 doesn't address directly. The table is the right source of truth for AL-on-the-left in general; it's the wrong source for AL-by-LB1-resolution-on-the-left specifically. 

### On architecture
The two paths in complex/mod.rs: I think that keeping the word-segmentation entry point `complex_language_segment_*` and the line-break entry point `complex_language_line_breaks_*` separate is better than threading a flag through a single function. The word path has no business touching UAX #14 — WordSegmenter consumes raw dict/LSTM word boundaries, including ones the line-break path needs to suppress (e.g., before MAIYAMOK ๆ). Folding them together would either pull UAX #14 awareness into the complex module (it currently imports zero LB data) or push a "line-break mode" flag down through it, both of which I think your rewrite will want to avoid for the same reasons. The seam is the closure parameter `forbids_break_before`, which keeps each side independent.

### The SA-mark override list
We could reframe it to `dict_boundary_before_iteration_mark_override`. This is a separate concern from the rule layer. The five characters (THAI MAIYAMOK ๆ, THAI PAIYANNOI ฯ, LAO KO LA ໆ, LAO ELLIPSIS ຯ, KHMER LEK TOO ៗ) have class SA → LB1 → AL, so `AL × AL` in the table says "break allowed" — the spurious boundary doesn't come from any UAX #14 rule, it comes entirely from the dict/LSTM treating these marks as separate words. Since the marks semantically attach to the preceding syllable, the boundary is wrong by definition of the character regardless of context. The right long-term fix is in the dictionary data or LSTM training corpus upstream of icu4x-data.

For now having this temporary small override is the smallest-surface patch the consumer (line segmenter) can apply at the boundary, and the function name and docstring can be changed to make clear that it can be deleted entirely once the dict data is corrected.

### On your rewrite
If the new line segmenter integrates dict/LSTM output with the spec's state machine directly, the expression of these two layers changes — the typographic-attachment classifier moves inside the state machine instead of being a closure on the outside, and the SA-mark override moves to wherever dict post-processing happens — but the underlying spec gaps don't go away. `AL × AL` (LB28) will still over-suppress script transitions and `AL × QU_PI` (LB19a) will still allow line breaks before opening quotes, because those are properties of the spec rules themselves, not of where the lookup happens. So whichever path you take, expect to need equivalent adjustments. Happy to defer to your rewrite if it's close to landing; if it's further out, I'd suggest taking this PR as a stopgap so 2.2 users hit a correct line breaker for Thai/Khmer/Lao/Burmese in the meantime — the analysis in this thread should save the rewrite from re-discovering the LB28/LB19a edge cases regardless.

If you're keeping the two-path shape in the rewrite, this PR's structure should transfer directly; if you're folding them, I'd appreciate hearing the reasoning.

PR #7234, opened 2025-11-11
Add Month constants

robertbastian:
Hebrew already uses these in tests. This way callers can just write `Month::ADAR_I` without having to figure out that it's `Month::leap(5)` (not `Month::leap(6)` or `Month::new(6)`).

The names are CLDR root display names.

### DIFF
--- a/components/calendar/src/cal/ethiopian.rs
+++ b/components/calendar/src/cal/ethiopian.rs
@@ -267,7 +267,7 @@ impl Month {
     pub const GENBOT: Month = Month::new(9);
     pub const SENE: Month = Month::new(10);
     pub const HAMLE: Month = Month::new(11);
-    pub const NÄHASE: Month = Month::new(12);
+    pub const NEHASSE: Month = Month::new(12);
     pub const PAGUMEN: Month = Month::new(13);
 }
 


components/calendar/src/cal/ethiopian.rs
```diff
@@ -254,6 +254,23 @@ impl Calendar for Ethiopian {
...
+    pub const HEDAR: Month = Month::new(3);
+    pub const TAHSAS: Month = Month::new(4);
+    pub const TER: Month = Month::new(5);
+    pub const YEKATIT: Month = Month::new(6);
+    pub const MEGABIT: Month = Month::new(7);
+    pub const MIAZIA: Month = Month::new(8);
+    pub const GENBOT: Month = Month::new(9);
+    pub const SENE: Month = Month::new(10);
+    pub const HAMLE: Month = Month::new(11);
+    pub const NÄHASE: Month = Month::new(12);
```
sffc: Observation: you have a `Ä` here, but you dropped the `Ḥ` from Heshvan down below.

Probably we should transliterate these into ASCII, but I've long been amenable to non-ASCII identifiers for specific concepts like these. If converting to ASCII, avoid the saka/shaka issue by transliterating and not just dropping the diacritics.

components/calendar/src/cal/coptic.rs
```diff
@@ -233,6 +234,23 @@ impl Calendar for Coptic {
     }
 }
 
+#[allow(missing_docs)]
```
sffc, line 237: Issue: They should at least be documented with which calendars they represent

sffc:
I think I would prefer a solution with 7 calendar-specific month enums, which can have roundtrip conversions to Month.

```rust
#[derive(Debug, Copy, Clone, PartialEq, Eq)]
pub enum EthiopianMonth {
    Meskerem,
    Tekemt,
    // ...
}

impl EthiopianMonth {
    pub fn to_month(self) -> Month {
        match self {
            Self::Meskerem => Month::new(1),
            Self::Tekemt => Month::new(2),
            // ...
        }
    }

    pub fn try_from_month(month: Month) -> Option<Self> {
        match month.to_tuple() {
            (1, false) => Some(Self::Meskerem),
            (2, false) => Some(Self::Tekemt),
            // ...
            _ => None
        }
    }
}
```

And the calendar-specific constructors can take the calendar-specific month enum directly.

components/calendar/src/cal/ethiopian.rs
robertbastian: whoops, should be "Nehasse" according to CLDR

PR #7238, opened 2025-11-11
Update prebuilt postcard files in date-picker-data.md

sffc:
They were last generated in 2023.

Autosubmit = true


robertbastian:
also needs to go in the release branches

PR #7239, opened 2025-11-12
Fix JavaScript errors in date-picker-data.md

sffc:
It works now:

image


sffc:
I pushed this and #7238 to release/2.1.

The data files might work on 2.0, but I'm not sure, and it's no longer the latest release, so I didn't bother pushing to release/2.0

PR #7246, opened 2025-11-16
Tests for RAdaBoost and CNN segmenters

opnuub:
@sffc 

The program test ```rust_matches_python_probs()``` loads ```python_test_output.txt``` that contains the flattened probability matrix (for cnn) or adaboost score (for radaboost) using ```python_test_output()```. 

I've also added ```#[cfg(test)]```for ```as_flat_slice(&self)``` in ```helper.rs``` to load the output cnn probability matrix.

### DIFF
--- a/components/segmenter/tests/cnn/main.rs
+++ b/components/segmenter/tests/cnn/main.rs
@@ -562,9 +562,9 @@ fn rust_matches_python_probs() -> Result<(), Box<dyn std::error::Error>> {
 
     assert_eq!(probs.len(), python.len());
 
-    let tol = 1e-6_f32;
+    let tol = 1e-5;
     for (i, (&got, &expected)) in probs.iter().zip(python.iter()).enumerate() {
-        let diff = (got - expected).abs();
+        let diff = (got - expected).abs() / expected.abs();
         assert!(
             diff <= tol,
             "mismatch at index {i}: got={got:.8e}, expected={expected:.8e}, diff={diff:.3e}"


components/segmenter/tests/cnn/main.rs
```diff
@@ -536,3 +546,29 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+        .map_err(|_| "validation/conversion failed".to_string())?;
+    let segmenter = CnnSegmenter::new(&cnndata);
+
+    let thai = "ปัญหาความแตกต่างที่เกิดขึ้นระหว่างความเป็นธรรมในทางสังคมกับความเป็นธรรมทางกฎหมาย".to_string();
+    let out = segmenter.segment_str(&thai);
+    let probs = out.probs.as_flat_slice();
+
+    assert_eq!(probs.len(), python.len());
+
+    let tol = 1e-6_f32;
```
sffc: Please either check the exact floating point value (floating point arithmetic works the same in Python and Rust), or check that the floats are within 1e-6% of each other, not the absolute value 1e-6. For example, `1.00875393e-11` and `4.57440706e-07` are within 1e-6 of each other, but they are very different values. For example, `abs(actual - expected) / expected < 1e-6`

opnuub: Hi Shane, I've just changed to relative difference of 1e-5 but not more (e.g. 1e-6). After some testing, I've concluded that the reasons can be due to
1. different matmul implementations (numpy vs what I wrote in rust),
2. different exp implementations (numpy vs rust ```exp()```)

sffc: 1e-5 percentage is fairly large in float space, more than I would have expected, but this is fine for now

sffc (approved):

PR #7259, opened 2025-11-26
fixed the reexports to shows in icu::datetime same as icu_datetime

### LINKED ISSUE #7247: Fix re-export docs (may depend on rustdoc fixes)
Previous issues: https://github.com/unicode-org/icu4x/issues/6069, https://github.com/unicode-org/icu4x/pull/6104, https://github.com/unicode-org/icu4x/issues/5915

We previously agreed to add docs to our re-exported input and preference types, the ones in `icu_component::input` and `icu_component::preferences`, that indicate that they are re-exports.

However, the resulting docs don't look the same in `icu_component` and `icu::component`. Here's an example:

Image

https://docs.rs/icu_datetime/latest/icu_datetime/preferences/index.html

Image

https://docs.rs/icu/latest/icu/datetime/preferences/index.html

Our re-export disclaimer is in `icu_datetime::preferences` but not `icu::datetime::preferences`.

SATVIKsynopsis:
Fixes #7247
Fixed the re-export disclaimer issue for `icu::datetime::preferences`

The re-export disclaimer was appearing correctly in `icu_datetime::preferences`, but not in `icu::datetime::preferences`.
This happened because `#[doc(inline)]` on the datetime re-export caused the full documentation from the original types to be inlined, which replaced the intended disclaimer.

To fix this, the datetime re-export in `components/icu/src/lib.rs` is now explicitly defined, and the preferences submodule is overridden with the correct re-export disclaimers.

This update ensures that both `icu_datetime::preferences` and `icu::datetime::preferences` display consistent and accurate documentation.

image


SATVIKsynopsis:
Thank you for the feedback. I understand the maintenance concern manually keeping the re-exports in sync would be error prone and goes against keeping the `icu` crate as simple re-exports.

I see that this will likely need to be resolved at the rustdoc level. do you have any suggestions for alternative approaches, or should wait for rustdoc to provide better support for this?

sffc:
I'd encourage you to consider contributing to rustdoc.

SATVIKsynopsis:
Thank you for the suggestion. I will look into contributing to rustdoc to address this issue.

I will close this PR since the consensus is to wait for a proper rustdoc solution rather than work around it in the codebase. I appreciate the guidance.

PR #7269, opened 2025-12-04
Fix script values

robertbastian:
It's missing Unicode 17 (and 16?) values, and some discriminants are actually wrong. What happened there is that ICU changed them between 78 beta and RC and we didn't notice.

### DIFF
--- a/ffi/dart/.gitattributes
+++ b/ffi/dart/.gitattributes
@@ -1,0 +1,1 @@
+lib/src/** linguist-generated=true
--- a/ffi/mvn/.gitattributes
+++ b/ffi/mvn/.gitattributes
@@ -1,0 +1,1 @@
+src/main/kotlin/** linguist-generated=true
--- a/ffi/npm/.gitattributes
+++ b/ffi/npm/.gitattributes
@@ -1,0 +1,1 @@
+lib/** linguist-generated=true


components/properties/src/props.rs
```diff
@@ -891,6 +897,14 @@ impl Script {
 }
 }
 
+impl Script {
+    // Doesn't actually exist!
+    #[doc(hidden)]
+    #[allow(non_upper_case_globals)]
+    #[deprecated]
+    pub const Chisoi: Script = Self(254);
```
robertbastian, line 905: Chisoi was removed from Unicode 17 late in the cycle. It will probably be in Unicode 18, so it might have a value like 211. Question is: do we gamble on it becoming 211, or do we assign 254 and then later change it again? Or do we actually break semver and remove it?

sffc: I vote to break semver and release this as 2.1.1 asap before people start needing it.

sffc: I guess the doc hidden is fine.

sffc (approved):
Thanks for catching this. Please file a follow-up issue to add a test for this.

components/properties/src/props.rs
robertbastian: with doc hidden we still need to decide on a value

robertbastian:
> add a test for this.

How is that meant to work? I would have to teach an LLM to read the Unicode standard or ICU4C docs.

There has been a process failure at some point, because Unicode removed this very late in the cycle, something that basically never happens, and nobody here was aware of that.

sffc:
> How is that meant to work? I would have to teach an LLM to read the Unicode standard or ICU4C docs.

Maybe we can stick uscript.h into icuexportdata.zip and write a super simple parser to extract the ICU4C values from it. Or, icuexportdata.zip can just dump them into a toml file.

> There has been a process failure at some point, because Unicode removed this very late in the cycle, something that basically never happens, and nobody here was aware of that.

Markus sent an email on 2025-07-28 announcing the removal of Chisoi from ICU4C, which I forwarded to icu4x-core. If our process is to implement stuff based on Markus's email, then we didn't succeed at doing that.

robertbastian:
> Maybe we can stick uscript.h into icuexportdata.zip and write a super simple parser to extract the ICU4C values from it. Or, icuexportdata.zip can just dump them into a toml file.

And then? How do you test identifier names? Note our names don't match ICU4C's names exactly.

> Markus sent an email on 2025-07-28 announcing the removal of Chisoi from ICU4C, which I forwarded to icu4x-core.

This email announced the *addition* of Chisoi. The removal happened later and I don't see an email about it.

robertbastian:
Discussion 2025-12-04: 

- @robertbastian I noticed this when I was looking at script enum values and they were off by one.
- @Manishearth It sounds like a bug in the release.
- @Manishearth Using script value 254 in the mean time seems fine.
- @sffc I think we should make a patch release.
- @sffc Why don't we just delete Chisoi? I don't think we should ship an enum value that isn't in the ICU4C enum.
- @Manishearth We could set it to unknown.
- @sffc I'm okay setting it to unknown.
- @robertbastian Other values that are unassigned don't return unknown; they return None. Also this doesn't work with FFI because every variant needs a unique discriminant
- @Manishearth We have an unchecked constructor from an arbitrary u16.
- @Manishearth I think we should set it to a very large number
- @sffc I think the ~1 month since the 2.1 release is short enough in human time that we should just delete it.
- @Manishearth I don't want to semver break
- @robertbastian I don't want to semver break
- @sffc OK fine. But do please ship 2.1.1 regardless to fix the others.

sffc:
Discussion about process:

- @robertbastian We need a better process for this. We have a UTC liaison.
- @sffc I'm the liaison, and I did forward this to the icu4x-core list.
- @Manishearth We shouldn't rely on the liaison.
- @sffc I agree, liaisons are human, and humans are bad at tracking changes.
- @robertbastian ICU is in a better position because Markus is chair of UTC and he knows all of what's going on.
- @robertbastian We should be briefed on the changes to Unicode every release. Not just property changes, but also normalizer, etc.
- @Manishearth I attend UTC but not all of it.
- @sffc I don't think it's a good use of our time to actively engage with UTC; it is a big time sink, and requires active effort. I think we should focus on conformance testing.
- @robertbastian We should give UTC feedback more actively during the development.
- @sffc UTC is more closely involved with ICU4C and ICU4J; I trust that they publish implementable algorithms. I'd rather be a downstream client with automated testing.
- @robertbastian We need a better solution for keeping up-to-date with Unicode in all our components. For Segmenter we rely on Makoto who sometimes updates it.
- @sffc I want segmenter to be automated.
- @Manishearth I think we can automate detecting when there is a bug, but not automating the implementation.
- @robertbastian I don't like that we lag behind.
- @sffc It's a discussion we should have with the TCG. ICU4X lags behind Unicode. The only way to fix that is to either invest a lot more time to engage with UTC, or convince UTC to implement their algorithms in Rust code.
- @robertbastian - If we accept this status quo I think we need to do a better job at documenting which parts of ICU4X conform to which Unicode versions, and where people can complain
- @sffc I strongly agree with that. You should be able to open a component and tell which Unicode version it implements.
- @Manishearth I think Segmenter is the only actual problem here.
- @sffc I don't think it's only Segmenter, but Segmenter happens to change more regularly than the other components. We have a problem with Properties.

PR #7270, opened 2025-12-04
Add conversions for properties/locale scripts

robertbastian:
#4705


sffc:
I want these conversions to exist.

I'm not sure about putting them in a From impl (which is not super discoverable) and furthermore making the From impl gated on a non-obvious feature.

sffc (commented):
I verified that the feature notice is visible on https://unicode-org.github.io/icu4x/rustdoc/icu/locale/subtags/struct.Script.html#impl-From%3CScript%3E-for-Script so this is lgtm

PR #7272, opened 2025-12-04
Add conversions for `unicode_bidi::BidiClass`

robertbastian:
This is already done internally in our `BidiDataSource` impl.

### DIFF
--- a/components/properties/src/bidi.rs
+++ b/components/properties/src/bidi.rs
@@ -68,6 +68,9 @@ pub enum BidiPairedBracketType {
     None,
 }
 
+#[cfg(feature = "unicode_bidi")]
+use crate::props::BidiClass;
+
 /// Implements [`unicode_bidi::BidiDataSource`] on [`CodePointMapDataBorrowed<BidiClass>`](crate::CodePointMapDataBorrowed).
 ///
 /// ✨ *Enabled with the `unicode_bidi` Cargo feature.*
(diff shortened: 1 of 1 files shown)


components/properties/src/bidi.rs
```diff
@@ -151,3 +158,34 @@ impl unicode_bidi::data_source::BidiDataSource
...
+            unicode_bidi::BidiClass::RLE => BidiClass::RightToLeftEmbedding,
+            unicode_bidi::BidiClass::RLO => BidiClass::RightToLeftOverride,
+            unicode_bidi::BidiClass::PDF => BidiClass::PopDirectionalFormat,
+            unicode_bidi::BidiClass::NSM => BidiClass::NonspacingMark,
+            unicode_bidi::BidiClass::BN => BidiClass::BoundaryNeutral,
+            unicode_bidi::BidiClass::FSI => BidiClass::FirstStrongIsolate,
+            unicode_bidi::BidiClass::LRI => BidiClass::LeftToRightIsolate,
+            unicode_bidi::BidiClass::RLI => BidiClass::RightToLeftIsolate,
+            unicode_bidi::BidiClass::PDI => BidiClass::PopDirectionalIsolate,
+        }
```
robertbastian, line 189: observation: `unicode_bidi::BidiClass` is exhaustive

sffc (approved):
It makes sense for the `From` impl to be gated on the feature that adds the crate dependency that includes the type that the `From` is referencing.

PR #7281, opened 2025-12-05
Validate properties constants

robertbastian:
More typos!

Based on #7280


components/properties/src/props.rs
```diff
@@ -72,6 +74,25 @@ macro_rules! create_const_array {
...
+        #[test]
+        fn $consts_test() {
+            $(
+                assert_eq!(
+                    crate::names::PropertyNamesLong::<$enum_ty>::new().get($enum_ty::$i).unwrap()
+                        // Rust identifiers use camel case
+                        .replace('_', "")
+                        // We use Ethiopian
+                        .replace("Ethiopic", "Ethiopian")
+                        // Nastaliq is missing a long name?
```
sffc, line 87: Should we open issues to discuss these discrepancies?

robertbastian: maybe, it's unclear to me when there should be a long name:
https://github.com/unicode-org/icu/blob/e68834de862b61aae50b031a19b73782850b749d/icu4c/source/test/cintltst/cucdapi.c#L439

PR #7285, opened 2025-12-05
reduced the size of arithmetic calendar

### LINKED ISSUE #7201: Reduce binary code size of Date::try_from_fields
Date::try_from_fields for AnyCalendar uses approximately 10 kB of binary code size according to [Chromium data](https://chrome-supersize.firebaseapp.com/viewer.html?load_url=https%3A%2F%2Fstorage.googleapis.com%2Fchromium-binary-size-trybot-results%2Fandroid-binary-size%2F2025%2F10%2F29%2F2478540%2FTrichrome32.ssargs.sizediff#focus=648).

There may be low hanging fruit to reduce this number.

The main culprit is likely the polymorphic function [`ArithmeticDate<C>::from_fields`](https://github.com/unicode-org/icu4x/blob/ebbbccd3912e108595cca4fc50c6365a9cdbbceb/components/calendar/src/calendar_arithmetic.rs#L202), which is 150 lines of code copied for each calendar. The challenge is that the function calls several trait functions, such as `year_info_from_extended` and `ordinal_month_from_code`. The code size can likely be reduced by making an inner non-generic function that returns an intermediate partially constructed ArithmeticDate and then calling the trait functions at the end, instead of interspersed through the middle of the function.

Any work on this issue should be supplemented with before and after measurements of code size. The easiest way to measure would be to add an example using `Date<AnyCalendar::try_from_fields` and measure the binary size difference of the example using the appropriate compilation options.

SATVIKsynopsis:
Fixes #7201 

This pull request refactors the `from_fields()` method in `calendar_arithmetic.rs` as described in the issue. 

The shared parsing logic for day, month, and year fields has been moved into a new helper function `parse_fields_shared()` along with a `SharedFields` struct. 

This keeps the logic in one place and reduces the overall code size.

### DIFF
--- a/components/calendar/src/calendar_arithmetic.rs
+++ b/components/calendar/src/calendar_arithmetic.rs
@@ -2,6 +2,8 @@
 // called LICENSE at the top level of the ICU4X source tree
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
+extern crate alloc;
+
 use calendrical_calculations::rata_die::RataDie;
 
 use crate::duration::{DateDuration, DateDurationUnit};
@@ -11,12 +13,14 @@ use crate::error::{
 };
 use crate::options::{DateAddOptions, DateDifferenceOptions};
 use crate::options::{DateFromFieldsOptions, MissingFieldsStrategy, Overflow};
-use crate::types::{DateFields, Month};
+use crate::types::{DateFields, ValidMonthCode};
 use crate::{types, Calendar, DateError, RangeError};
 use core::cmp::Ordering;
 use core::fmt::Debug;
 use core::hash::{Hash, Hasher};
 use core::ops::RangeInclusive;
+use alloc::vec::Vec;
+
 
 /// This is checked by constructors. Internally we don't care about this invariant.
 pub const VALID_YEAR_RANGE: RangeInclusive<i32> = -1_000_000..=1_000_000;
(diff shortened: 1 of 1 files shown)


SATVIKsynopsis:
Hi. I see that several CI checks are failing after this refactor.

Since this change moves shared parsing logic into `parse_fields_shared()`, some behavior differences may appear. Please let me know which parts of the original behavior must remain exactly the same, and I’ll update the code accordingly.

Happy to adjust anything needed.

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -25,14 +29,22 @@ pub const VALID_YEAR_RANGE: RangeInclusive<i32> = -1_000_000..=1_000_000;
 /// date types. Because this range slightly exceeds the [`VALID_YEAR_RANGE`], only
 /// the valid year range is checked in constructors. Only the `Date::from_rata_die`
 /// constructor actually uses this, but for clamping instead of erroring.
-// This range is the tightest possible range that includes all valid years for all
-// calendars, this is asserted in [`test_validity_ranges`].
+///
+/// The longest average year length is 365.25 (Julian, Coptic). We round up to 366 to make
+/// it round and capture year offsets from the Gregorian calendar, which underlies RD.
```
robertbastian, line 34: you merged wrong, this PR recovers some pretty old code

sffc:
Thanks for your interest in ICU4X!

Please also see my comment on the issue:

> Any work on this issue should be supplemented with before and after measurements of code size

We don't accept optimizations unless they are backed by real numbers. 

I'm going to close this PR because it also has the merge issue and CI failures. Please open a new PR once you've fixed it. Thank you!

PR #7286, opened 2025-12-05
Relax some bounds

robertbastian:
The derive puts a bound on `C`, which we don't need.

### DIFF
--- a/components/time/src/types.rs
+++ b/components/time/src/types.rs
@@ -2,7 +2,7 @@
 // called LICENSE at the top level of the ICU4X source tree
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
-use icu_calendar::{types::RataDie, AsCalendar, Calendar, Date, Iso, RangeError};
+use icu_calendar::{types::RataDie, AsCalendar, Date, Iso, RangeError};
 
 use crate::zone::UtcOffset;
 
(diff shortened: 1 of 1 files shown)


components/time/src/types.rs
```diff
@@ -193,10 +193,34 @@ pub struct DateTime<A: AsCalendar> {
...
+impl<A: AsCalendar + Clone> Clone for DateTime<A> {
+    fn clone(&self) -> Self {
+        Self {
+            date: self.date.clone(),
+            time: self.time,
+        }
+    }
+}
+
+impl<A> Copy for DateTime<A> where A: AsCalendar + Copy {}
```
sffc: It looks like Copy and Clone have the same bounds as the derive; please use the derive unless the bounds can be simplified (like for PartialEq)

robertbastian: actually they should delegate to `Date`'s bounds

sffc: Yes they should, but if Date's bounds are the same as the ones generated by the derive, then let's please use the derive

robertbastian: they're specifically not, the derive puts bounds on all generics

sffc: Your first commit had

```
impl<A: AsCalendar + Clone> Clone for DateTime<A
```

Which should be identical to what the derive generates

sffc: According to https://unicode-org.github.io/icu4x/rustdoc/icu/calendar/struct.Date.html#impl-Clone-for-Date%3CA%3E the bound for Date is

```
impl<A> Clone for Date<A>
where
    A: AsCalendar + Clone,
```

The bound you currently have is

```
impl<A: AsCalendar> Clone for DateTime<A>
where
    Date<A>: Clone,
```

which by induction is equivalent to

```
impl<A: AsCalendar> Clone for DateTime<A>
where
    A: AsCalendar + Clone,
```

which is exactly what `#[derive(Clone)]` generates.

My comment is that I think we should do a manual impl when the desired bound is not what the standard derive generates, and we should use the standard derive when the desired bound is what the standard derive generates

robertbastian: It's only equivalent as long as the `Date` bound doesn't change. It's brittle to make this assumption.

sffc: I agree with you, but this is a complaint to make to rust lang team. There's probably already an issue about this.  In the meantime, we're not going to do this treatment to all derives.

robertbastian: So you're blocking this PR because?

sffc: I'd prefer a separate PR for PartialEq/Eq and Clone/Copy but I won't keep blocking this

PR #7287, opened 2025-12-05
Deprecate `Date::new_from_iso`/`Date::to_iso`

robertbastian:
We should stop treating the ISO calendar as special. Conversions are available through `to_calendar()` like for all other calendars, and there shouldn't be two ways of doing things.

This also cleans up old code in `AnyCalendar` that converts through ISO, which skips the optimisations in `.to_calendar()`.

### DIFF
--- a/components/calendar/src/date.rs
+++ b/components/calendar/src/date.rs
@@ -616,7 +616,11 @@ impl<A> Copy for Date<A> where A: AsCalendar + Copy {}
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::types::Weekday;
+    use crate::{
+        cal::{Buddhist, Hebrew},
+        types::Weekday,
+        Gregorian,
+    };
 
     #[test]
     fn test_ord() {
@@ -664,5 +668,16 @@ mod tests {
             Date::try_new_iso(2020, 1, 21).unwrap().weekday(),
             Weekday::Tuesday,
         );
+    }
+
+    #[test]
+    fn test_to_calendar() {
+        let date = Date::try_new_gregorian(2025, 12, 9).unwrap();
+        // These conversions use the AbstractGregorian fast path
+        let date2 = date.to_calendar(Buddhist).to_calendar(Gregorian);
+        // These conversions go through RataDie
+        let date3 = date.to_calendar(Hebrew).to_calendar(Gregorian);
+        assert_eq!(date, date2);
+        assert_eq!(date2, date3);
     }
 }


components/datetime/src/scaffold/dynamic_impls.rs
```diff
@@ -38,7 +38,7 @@ impl DateInputMarkers for DateFieldSet {
     type DayOfMonthInput = datetime_marker_helper!(@input/day_of_month, yes);
     type DayOfYearInput = datetime_marker_helper!(@input/day_of_year, yes);
     type RataDieInput = datetime_marker_helper!(@input/rata_die, yes);
-    type DayOfWeekInput = datetime_marker_helper!(@input/day_of_week, yes);
+    type WeekdayInput = datetime_marker_helper!(@input/weekday, yes);
```
sffc: Depending on how we feel about our policy, this might not be semver compatible because clients can stably refer to an associated type of a scaffold trait. It requires the word "scaffold" but not "unstable" in source

sffc: Should be discussed in #7288

components/calendar/src/cal/roc.rs
```diff
@@ -127,50 +127,50 @@ mod test {
...
         );
-        assert_eq!(roc_from_rd.day_of_month().0, case.expected_day,
-            "Failed day_of_month check from RD: {case:?}\nISO: {iso_from_rd:?}\nROC: {roc_from_rd:?}");
 
-        let iso_from_case = Date::try_new_iso(case.iso_year, case.iso_month, case.iso_day)
-            .expect("Failed to initialize ISO date for {case:?}");
-        let roc_from_case = Date::new_from_iso(iso_from_case, Roc);
-        assert_eq!(iso_from_rd, iso_from_case,
-            "ISO from RD not equal to ISO generated from manually-input ymd\nCase: {case:?}\nRD: {iso_from_rd:?}\nManual: {iso_from_case:?}");
+        let roc_from_case = Date::try_new_roc(
```
sffc, line 168: In this test, you removed the `new_from_iso` and now only create from the ROC dates and the RD.

Are we still satisfactorily covering the `has_cheap_iso_conversion` code path?

```rust
        let inner = if c1.has_cheap_iso_conversion() && c2.has_cheap_iso_conversion() {
            c2.from_iso(c1.to_iso(self.inner()))
        } else {
            // `from_rata_die` precondition is satified by `to_rata_die`
            c2.from_rata_die(c1.to_rata_die(self.inner()))
        };
```

robertbastian: there are a couple of tests that exercise this code, it's shared between all `AbstractGregorian` calendars, so we don't necessarily need to test it on Roc

it might make sense to add a few test specifically for this code in `date.rs`.

components/calendar/src/date.rs
```diff
@@ -674,4 +669,15 @@ mod tests {
             Weekday::Tuesday,
         );
     }
+
+    #[test]
+    fn test_to_calendar() {
+        let date = Date::try_new_gregorian(2025, 12, 9).unwrap();
+        // These conversions use the AbstractGregorian fast path
+        let date2 = date.to_calendar(Buddhist).to_calendar(Gregorian);
```
sffc, line 677: Suggestion: It seems better to test this for all calendars, not just Buddhist

Manishearth (approved):
Generally not a fan of churning the APIs here more but I guess it's fine.

PR #7288, opened 2025-12-05
Replace `Date::day_of_week` by `Date::weekday`

robertbastian:
#7228

Also replace other usages of "day of week" that are not the numeric day of the week (pretty much all of them).

### DIFF
--- a/components/datetime/src/scaffold/fieldset_traits.rs
+++ b/components/datetime/src/scaffold/fieldset_traits.rs
@@ -38,7 +38,7 @@ pub trait DateInputMarkers: UnstableSealed {
     /// Marker for resolving the day-of-year input field.
     type RataDieInput: IntoOption<RataDie>;
     /// Marker for resolving the weekday input field.
-    type WeekdayInput: IntoOption<Weekday>;
+    type DayOfWeekInput: IntoOption<Weekday>;
 }
 
 /// A trait associating types for date formatting in a specific calendar
@@ -193,7 +193,7 @@ pub trait AllInputMarkers<R: DateTimeMarkers>:
     GetField<<R::D as DateInputMarkers>::YearInput>
     + GetField<<R::D as DateInputMarkers>::MonthInput>
     + GetField<<R::D as DateInputMarkers>::DayOfMonthInput>
-    + GetField<<R::D as DateInputMarkers>::WeekdayInput>
+    + GetField<<R::D as DateInputMarkers>::DayOfWeekInput>
     + GetField<<R::D as DateInputMarkers>::DayOfYearInput>
     + GetField<<R::D as DateInputMarkers>::RataDieInput>
     + GetField<<R::T as TimeMarkers>::HourInput>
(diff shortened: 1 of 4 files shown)


components/calendar/src/date.rs
```diff
@@ -243,16 +243,69 @@ impl<A: AsCalendar> Date<A> {
...
+    /// The day-of-month of this date.
     #[inline]
-    pub fn months_in_year(&self) -> u8 {
-        self.calendar.as_calendar().months_in_year(self.inner())
+    pub fn day_of_month(&self) -> types::DayOfMonth {
+        self.calendar.as_calendar().day_of_month(&self.inner)
     }
 
-    /// The number of days in the year of this date.
+    /// The day-of-month of this date.
```
sffc: Docs for wrong function

components/datetime/src/scaffold/fieldset_traits.rs
```diff
@@ -37,8 +37,8 @@ pub trait DateInputMarkers: UnstableSealed {
     type DayOfYearInput: IntoOption<DayOfYear>;
     /// Marker for resolving the day-of-year input field.
     type RataDieInput: IntoOption<RataDie>;
-    /// Marker for resolving the day-of-week input field.
-    type DayOfWeekInput: IntoOption<Weekday>;
+    /// Marker for resolving the weekday input field.
+    type WeekdayInput: IntoOption<Weekday>;
```
sffc: Depending on how we feel about our policy, this might not be semver compatible because clients can stably refer to an associated type of a scaffold trait. It requires the word "scaffold" but not "unstable" in source

robertbastian: > /// 🚧 This trait is considered unstable; it may change at any time, in breaking or non-breaking ways,

components/calendar/src/date.rs
robertbastian: good catch, this is preexisting

components/datetime/src/scaffold/fieldset_traits.rs
sffc: https://github.com/unicode-org/icu4x/issues/6659#issuecomment-3416501430

robertbastian: I don't see how a comment discussing the unstable feature is relevant here

robertbastian: If anything, that comment says that this trait will live behind the unstable feature, so everything about it is unstable (as the comment currently says).

sffc: The trait as it currently lives on main is UnstableSealed in a stable module\* with stable methods. This implies that the trait is not implementable, and therefore we can add to it, but it is stably usable in bounds, and methods are stably callable.

\* Unless we wish to categorize the module named "scaffold" to imply unstable.

robertbastian: I don't see why they wouldn't be unstable. They are used for bounds on unstable constructors. When this was designed, I was under the impression that all of this was unstable, as it is documented as unstable

sffc: I had intended them to be stably usable in bounds, unstably implementable, and unstably callable. But the draft policy for how to enforce that came after 2.0, which means we need to decide what to do here.

robertbastian: Then the docs are wrong

PR #7297, opened 2025-12-09
Add example for measuring Date::try_from_fields code size

SATVIKsynopsis:
This PR adds a standalone example using release-opt-size and the benchmark macros so we can measure code size separately from the main library changes.

Only the example and its dependencies are included here.

### DIFF
--- a/components/icu/examples/date_try_from_fields.rs
+++ b/components/icu/examples/date_try_from_fields.rs
@@ -6,8 +6,8 @@
 icu_benchmark_macros::instrument!();
 use icu_benchmark_macros::println;
 
-use icu_calendar::types::DateFields;
-use icu_calendar::{AnyCalendar, AnyCalendarKind, Date};
+use icu::calendar::types::DateFields;
+use icu::calendar::{AnyCalendar, AnyCalendarKind, Date};
 
 const CALENDAR_KINDS: &[AnyCalendarKind] = &[
     AnyCalendarKind::Buddhist,


components/calendar/examples/date_try_from_fields.rs
```diff
@@ -0,0 +1,29 @@
...
+use icu_calendar::error::DateFromFieldsError;
+use icu_calendar::types::DateFields;
+use icu_calendar::{AnyCalendar, AnyCalendarKind, Date};
+
+#[no_mangle]
+pub extern "C" fn main() {
+    real_main().unwrap();
+}
+
+fn real_main() -> Result<(), DateFromFieldsError> {
```
robertbastian:
```suggestion
#![no_main] // https://github.com/unicode-org/icu4x/issues/395
icu_benchmark_macros::instrument!();
use icu_benchmark_macros::println;

use icu_calendar::error::DateFromFieldsError;
use icu_calendar::types::DateFields;
use icu_calendar::{AnyCalendar, AnyCalendarKind, Date};

fn main() {
```

components/calendar/examples/date_try_from_fields.rs
```diff
@@ -0,0 +1,29 @@
...
+    let cal = AnyCalendar::new(AnyCalendarKind::Iso);
+
+    let mut fields = DateFields::default();
+    fields.extended_year = Some(2025);
+    fields.month_code = Some(b"M07");
+    fields.day = Some(8);
+
+    let date = Date::try_from_fields(fields, Default::default(), cal)?;
```
robertbastian:
```suggestion
    let date = Date::try_from_fields(fields, Default::default(), cal).unwrap();
```

components/calendar/Cargo.toml
```diff
@@ -28,6 +28,7 @@ ixdtf = { workspace = true, optional = true }
 tinystr = { workspace = true, features = ["zerovec"] }
 zerovec = { workspace = true, features = ["derive"] }
 
+writeable = { workspace = true, optional = true }
```
robertbastian: should be a dev-dependency

sffc:
@robertbastian Would you rather this live here in `icu_calendar` or should we put it in `icu`?

components/calendar/examples/date_try_from_fields.rs
```diff
@@ -0,0 +1,29 @@
...
+use icu_calendar::types::DateFields;
+use icu_calendar::{AnyCalendar, AnyCalendarKind, Date};
+
+#[no_mangle]
+pub extern "C" fn main() {
+    real_main().unwrap();
+}
+
+fn real_main() -> Result<(), DateFromFieldsError> {
+    let cal = AnyCalendar::new(AnyCalendarKind::Iso);
```
sffc: Suggestion: loop over multiple values of AnyCalendarKind, to make it less likely that the compiler will try to inline this one special case.

robertbastian:
I think `icu` is better

SATVIKsynopsis:
integrated the fixes and changes suggested below:

switched to the size-optimized icu_benchmark_macros::instrument! structure and added the CALENDAR_KINDS loop to prevent compiler over-optimization. error handling is simplified with .unwrap().

moved the writeable dependency to [dev-dependencies] in cargo.toml.

SATVIKsynopsis:
Huge number of CI tests are failing which i think are due to introduction of new macros in the example file. How to fix these issues?

robertbastian:
Have you taken a look at any of the failures? Every single one of them says

```
Caused by:
  dev-dependencies are not allowed to be optional: `writeable`
```

SATVIKsynopsis:
CI fixes applied and pushed. All checks should pass now. let me know if any other refinements are needed.

components/calendar/examples/date_try_from_fields.rs
```diff
@@ -0,0 +1,33 @@
...
+fn main() {
+    for &kind in CALENDAR_KINDS {
+        let cal = AnyCalendar::new(kind);
+
+        let mut fields = DateFields::default();
+        fields.extended_year = Some(2025);
+        fields.month_code = Some(b"M07");
+        fields.day = Some(8);
+
+        let date = Date::try_from_fields(fields, Default::default(), cal.clone()).unwrap();
```
sffc: Don't clone the calendar, and it's fine to pass the Result into the println fn

```suggestion
        let date = Date::try_from_fields(fields, Default::default(), Ref(&cal));
```

components/icu/examples/date_try_from_fields.rs
```diff
@@ -0,0 +1,33 @@
...
+        let cal = AnyCalendar::new(kind);
+
+        let mut fields = DateFields::default();
+        fields.extended_year = Some(2025);
+        fields.month_code = Some(b"M07");
+        fields.day = Some(8);
+
+        let date = Date::try_from_fields(fields, Default::default(), cal.clone()).unwrap();
+
+        println!("Constructed date for {:?} = {:?}", kind, date);
```
sffc: Question @robertbastian, will this println work with the macro? How do you test it?

components/calendar/Cargo.toml
```diff
@@ -36,6 +36,8 @@ icu_locale = { workspace = true, optional = true }
```
sffc, line 36: As @robertbastian suggested, please add this example to the `icu` crate instead of the `icu_calendar` crate, next to all the other examples. Then you won't need to change Cargo.toml

SATVIKsynopsis:
I have implemented the requested changes:

The date_try_from_fields.rs example has been moved from icu_calendar/examples/ to icu/examples/.

The implementation uses Ref(&cal) instead of cloning, as suggested.

components/calendar/Cargo.toml
```diff
@@ -36,6 +36,8 @@ icu_locale = { workspace = true, optional = true }
 
 [dev-dependencies]
 icu_provider = { path = "../../provider/core", features = ["logging"] }
+writeable = { workspace = true} 
+icu_benchmark_macros = { path = "../../tools/benchmark/macros" }
 icu = { path = "../../components/icu", default-features = false }
```
sffc: Please undo the changes in this file; you don't need them anymore

SATVIKsynopsis: Should i also remove cargo files from this pr history?

sffc: No it will get squashed when merged

components/icu/examples/date_try_from_fields.rs
robertbastian: no, because `kind` and `date` do not implement `Writeable.`

test with `cargo make wasm-wasm-examples`.

components/icu/examples/date_try_from_fields.rs
```diff
@@ -0,0 +1,41 @@
...
+
+        let date = Date::try_from_fields(fields, Default::default(), Ref(&cal));
+
+        let kind_str = format!("{:?}", kind);
+        let date_str = match date {
+            Ok(date) => format!("{:?}", date),
+            Err(err) => format!("Error: {:?}", err),
+        };
+
+        println!("Constructed date for {:?} = {:?}", kind_str, date_str);
```
robertbastian: using `format!` defeats the purpose of `icu_benchmark_macros` and will pull in all of `std::fmt`. Use the `println!` macro with a type that implements `Writeable`, such as `u8`.

```suggestion
        println!("The day is {}", date.unwrap().day_of_month().0);
```

sffc: Please let's not use `.unwrap()` though. How about

```rust
match date {
    Ok(date) => println!("The day is {}", date.day_of_month().0),
    Err(_e) => println!("Not a valid date")
}
```

robertbastian: Why? We use `unwrap` and `expect` in basically all our examples: https://github.com/search?q=repo%3Aunicode-org%2Ficu4x+path%3Acomponents%2Ficu%2Fexamples+unwrap+OR+expect&type=code

Binary size benchmarks are built with `panic-immediate-abort`

sffc: Yep ok

sffc (approved):
This is good now although you can put back the unwrap if you like

robertbastian (approved):

PR #7299, opened 2025-12-10
Build CodePointTries in datagen

robertbastian:
Currently we use pre-built CPTs exported from ICU4C. ICU4C also exports the raw data, so we should build the tries ourselves. This is a step towards reading the UCD directly #4602.

However, the `icu_codepointtrie_builder` WASM backend seems to not be able to handle this. Datagen times out after 6 hours 🤷


robertbastian:
There must be a memory leak in `wasmi`. It generates the properties data correctly and then later grinds to a halt.

sffc:
Make sure our usage of wasmi isn't leaking memory.

robertbastian:
We're not using unsafe code, so it can't be our fault.

sffc:
I mean, without studying in depth how wasmi works, I'm not sure the best practices for cleaning up or reusing wasm modules, memories, etc.

robertbastian:
Done better in #7904

Issue #7303, opened 2025-12-11
Handle out-of-range values in Dense ZeroTrie


sffc:
Currently we panic when trying to build a dense ZeroTrie with values that don't fit in the dense matrix. What we should do instead:

1. Select the row offset that gets the greatest number of values into the dense matrix.
2. Put all out-of-range values into the primary trie, as we do with sparse rows.

Docs for background: https://unicode-org.github.io/icu4x/rustdoc/zerotrie/dense/struct.ZeroAsciiDenseSparse2dTrieOwned.html

Cdog1526:
Hey, could I try to handle this?

sffc:
I believe this is fixed by #7305; do you agree @Cdog1526 ?

PR #7305, opened 2025-12-11
Dense zero trie Out of Range handling #7303

Cdog1526:
Working on https://github.com/unicode-org/icu4x/issues/7303.

When creating a dense sparse zerotrie, if the incoming values are too spread out for a dense matrix:
- Select the row offset that gets the greatest number of values into the dense matrix.
- Put all out-of-range values into the primary trie, as we do with sparse rows.

Also, added a test case for this. I am not certain that this test case is sufficient, this is my first contribution here.


utils/zerotrie/tests/dense_test.rs
```diff
@@ -137,3 +137,39 @@ fn test_short_subtags() {
...
 }
+
+#[test]
+fn test_dense_sparse_window_selection() {
+    //Make initial Btree (not using enumerate)
+    let row_width = usize::from(u16::MAX); // Densetype max
+    let far_low = 0;
+    let cluster_start = 50;
+    let cluster_vals = [cluster_start, cluster_start + 2, cluster_start + 3];
+    let far_high = cluster_start + row_width + 100;
```
sffc, line 147: Suggestion: also add

```
cluster_start + row_width - 3
cluster_start + row_width - 2
cluster_start + row_width - 1
cluster_start + row_width
cluster_start + row_width + 1
```

Maybe insert them not in order so that the sorting matters more. And maybe also insert duplicate values, which are allowed.

utils/zerotrie/tests/dense_test.rs
```diff
@@ -137,3 +137,39 @@ fn test_short_subtags() {
...
+    assert_eq!(trie.get("p", "a"), Some(cluster_vals.get(0)));
+    assert_eq!(trie.get("p", "b"), Some(cluster_vals.get(1)));
+    assert_eq!(trie.get("p", "c"), Some(cluster_vals.get(2)));
+    assert_eq!(trie.get("p", "low"), Some(far_low));
+    assert_eq!(trie.get("p", "high"), Some(far_high));
+
+    let dense_size = check_encoding(trie);
+    let simple_size = make_simple_ascii_trie(&data).byte_len();
+
+    println!("Dense size: {}, Simple size: {}", dense_size, simple_size);
```
sffc: Please add assertions on the two sizes. We assert the sizes because they should be stable, and because you can see what they are from reading the code. They should only change if we change the data structure or builder algorithm.

utils/zerotrie/tests/dense_test.rs
```diff
@@ -137,3 +137,39 @@ fn test_short_subtags() {
...
+    // Build the 2d trie.
+    let dense =
+        ZeroAsciiDenseSparse2dTrieOwned::try_from_btree_map_str(&data, b'/').unwrap();
+    let trie = dense.as_borrowed();
+
+    assert_eq!(trie.get("p", "a"), Some(cluster_vals.get(0)));
+    assert_eq!(trie.get("p", "b"), Some(cluster_vals.get(1)));
+    assert_eq!(trie.get("p", "c"), Some(cluster_vals.get(2)));
+    assert_eq!(trie.get("p", "low"), Some(far_low));
+    assert_eq!(trie.get("p", "high"), Some(far_high));
```
sffc: Please use `check_data` which does these checks and more.

sffc (commented):
Nice work! Some suggestions

utils/zerotrie/tests/dense_test.rs
Cdog1526: Added!

utils/zerotrie/tests/dense_test.rs
Cdog1526: Added!

Cdog1526:
@sffc I see that two tests are failing right now, but I'm not really sure how to recreate that on my side - is there something I can run or look at or is this related to me? And is there anything else I should do?

sffc:
The branch might be stale. I updated it to main

Cdog1526:
Ah awesome, thanks! Let me know if you need anything else to review.

utils/zerotrie/src/builder/dense.rs
```diff
@@ -28,6 +28,31 @@ pub(crate) struct DenseSparse2dAsciiWithFixedDelimiterBuilder<'a> {
...
+        let mut sorted_vals: Vec<usize> = values.values().cloned().collect();
+        let row_width = usize::from(DenseType::MAX);
+        sorted_vals.sort_unstable();
+        let mut bot = 0;
+        let mut best = 0;
+        let mut best_index = 0;
+        for top in 0..sorted_vals.len() {
+            while bot < top {
+                let top_val = sorted_vals.get(top).copied().unwrap_or(0);
+                let bot_val = sorted_vals.get(bot).copied().unwrap_or(0);
```
sffc: Issue: Here and everywhere else you changed to use `.unwrap_or(0)`:

It is easy to see that both `top` and `bot` are `< sorted_vales.len()`, so typically our style is to use the index operator and add a suppression with comment like

```
#[allow(clippy::indexing_slicing)] // bot < top and top < sorted_vals.len()
```

We do the `.unwrap_or` style in GIGO situations, where the input _could_ be invalid, but in this case it is an easily provable code constraint.

utils/zerotrie/tests/dense_test.rs
```diff
@@ -137,3 +137,42 @@ fn test_short_subtags() {
...
+    //Make initial Btree (not using enumerate)
+    let row_width = usize::from(u16::MAX); // Densetype max
+    let far_low = 0;
+    let cluster_start = 50;
+    let cluster_vals = [cluster_start, cluster_start + 2, cluster_start + 3];
+    let far_high = cluster_start + row_width + 100;
+
+    let mut inner = BTreeMap::new();
+    inner.insert("low", far_low);
+    inner.insert("a", cluster_vals.first().copied().unwrap_or(0));
```
sffc: Nit: Please inline the `cluster_vals` values to make it more consistent with the rest of the map insertions.

sffc (approved):

sffc (approved):

PR #7310, opened 2025-12-12
Export an Encodeable instead of a VarULE and use Index32

sffc:
Breaking change in `icu_provider` extracted from #7309.

The trait unfortunately wasn't implemented with EncodeAsVarULE in mind, so this is a breaking change in icu_provider in a trait that isn't used by anything except hello_world (which lives in that crate) and is only implemented by macros that are versioned along with the crate. Do we care? Please speak up.

### DIFF
--- a/provider/core/src/hello_world.rs
+++ b/provider/core/src/hello_world.rs
@@ -49,21 +49,12 @@ impl<'a> ZeroFrom<'a, str> for HelloWorld<'a> {
     }
 }
 
-crate::data_struct!(HelloWorld<'data>, manual_varule);
-
-impl icu_provider::ule::MaybeAsVarULE for HelloWorld<'_> {
-    type EncodedStruct = str;
-}
-
-impl icu_provider::ule::MaybeEncodeAsVarULE for HelloWorld<'_> {
-    type EncodeableStruct<'a>
-        = &'a str
-    where
-        Self: 'a;
-    fn maybe_as_encodeable<'a>(&'a self) -> Option<Self::EncodeableStruct<'a>> {
-        Some(&*self.message)
-    }
-}
+crate::data_struct!(
+    HelloWorld<'data>,
+    varule: str,
+    #[cfg(feature = "export")]
+    encode_as_varule: |v: &HelloWorld<'_>| &*v.message
+);
 
 data_marker!(
     /// Marker type for [`HelloWorld`].
--- a/provider/core/src/varule_traits.rs
+++ b/provider/core/src/varule_traits.rs
@@ -70,8 +70,22 @@ macro_rules! data_struct {
     };
     (
         $ty:ty,
-        manual_varule
+        varule: $varule:ty,
+        $(#[$attr:meta])*
+        encode_as_varule: $encode_as_varule:expr
     ) => {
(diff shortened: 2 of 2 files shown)


sffc:
cargo-semver-check suggested that I add a default for the associated type, but when I do that, I get an error saying that associated type defaults are not stable. https://github.com/rust-lang/rust/issues/29661

sffc:
Here's the semver check report

```
    Checking icu_provider v2.1.1 -> v2.1.1 (no change; assume patch)
     Checked [   0.027s] 184 checks: 180 pass, 4 fail, 0 warn, 16 skip

--- failure trait_associated_type_added: non-sealed public trait added associated type without default value ---

Description:
A non-sealed trait has gained an associated type without a default value, which breaks downstream implementations of the trait
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#trait-new-item-no-default
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.45.0/src/lints/trait_associated_type_added.ron

Failed in:
  trait associated type icu_provider::ule::MaybeEncodeAsVarULE::EncodeableStruct in file /home/runner/work/icu4x/icu4x/provider/core/src/varule_traits.rs:36

--- failure trait_method_added: pub trait method added ---

Description:
A non-sealed public trait added a new method without a default implementation, which breaks downstream implementations of the trait
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#trait-new-item-no-default
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.45.0/src/lints/trait_method_added.ron

Failed in:
  trait method icu_provider::ule::MaybeEncodeAsVarULE::maybe_as_encodeable in file /home/runner/work/icu4x/icu4x/provider/core/src/varule_traits.rs:41

--- failure trait_method_missing: pub trait method removed or renamed ---

Description:
A trait method is no longer callable, and may have been renamed or removed entirely.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#major-any-change-to-trait-item-signatures
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.45.0/src/lints/trait_method_missing.ron

Failed in:
  method maybe_encode_as_varule of trait MaybeEncodeAsVarULE, previously in file /home/runner/work/icu4x/icu4x/semver-checks/target/semver-checks/git-origin_main/d6f790cc393030a5d1382de3f18bbc857d401ae6/provider/core/src/varule_traits.rs:35

--- failure trait_no_longer_dyn_compatible: trait no longer dyn compatible ---

Description:
Trait is no longer dyn compatible, which breaks `dyn Trait` usage.
        ref: https://doc.rust-lang.org/stable/reference/items/traits.html#object-safety
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.45.0/src/lints/trait_no_longer_dyn_compatible.ron

Failed in:
  trait MaybeEncodeAsVarULE in file /home/runner/work/icu4x/icu4x/provider/core/src/varule_traits.rs:34

     Summary semver requires new major version: 4 major and 0 minor checks failed
```

If I switch the function name back to what it was before and add a default associated type (not possible in stable Rust), I think 3/4 of those errors go away.

But I'm going to merge this for now since @Manishearth approved it.

PR #7312, opened 2025-12-12
Add `make_any_calendar` macro

robertbastian:
This macro generates an any calendar with the given list of variants. This PR just uses it internally, but the idea is to expose it so that users can define domain-specific any calendars.

### DIFF
--- a/components/calendar/src/any_calendar.rs
+++ b/components/calendar/src/any_calendar.rs
@@ -783,12 +783,9 @@ impl fmt::Display for AnyCalendarKind {
     }
 }
 
-/// TODO
 pub trait AnyCalendarable: Calendar + Sized {
-    /// TODO
     type Identity: PartialEq + Eq + Copy;
 
-    /// TODO
     fn identity(&self) -> Self::Identity;
 }
 


components/calendar/src/any_calendar.rs
```diff
@@ -813,30 +716,36 @@ pub enum AnyCalendarKind {
     Roc,
 }
 
+impl CalendarPreferences {
+    /// Selects the [`CalendarAlgorithm`] appropriate for the given [`CalendarPreferences`].
+    pub fn resolved_algorithm(self) -> CalendarAlgorithm {
```
sffc, line 714: Observation: new API

components/calendar/src/any_calendar.rs
```diff
@@ -881,6 +790,101 @@ impl fmt::Display for AnyCalendarKind {
     }
 }
 
+/// TODO
```
sffc: Observation: New API

Issue: Please add docs since this is landing directly into a stable crate, or open a follow-up issue in the 2.2 milestone.

components/calendar/src/any_calendar.rs
```diff
@@ -28,211 +25,402 @@ define_preferences!(
...
+            }
+        }
+
+        $(#[$any_date_meta])*
+        #[doc = concat!("The inner date type for [`", stringify!($any_calendar_ident), "`]")]
+        #[derive(Clone, PartialEq, Eq, Debug, Copy)]
+        pub enum $any_date_ident {
+            $(
+                #[doc = concat!("A date for a [`", stringify!($ty), "`] calendar")]
+                $variant(<$ty as $crate::Calendar>::DateInner, <$ty as AnyCalendarable>::Identity),
```
sffc, line 70: Observation: you're adding the second field Identity to every variant, but before it was only in HijriTabular variant. Is this a semver issue? The semver check doesn't seem to notice it

robertbastian: the date variants are not public

sffc: oh I see, AnyDateInner is not exported. Cool.

components/calendar/src/any_calendar.rs
```diff
@@ -28,211 +25,402 @@ define_preferences!(
...
+    /// assert!(any_gregorian_date <= any_gregorian_date);
+    /// assert!(any_japanese_date <= any_japanese_date);
+    /// assert!(!(any_gregorian_date <= any_japanese_date) && !(any_japanese_date <= any_gregorian_date));
+    /// ```
+    #[non_exhaustive]
+    AnyCalendar,
+
+    #[non_exhaustive]
+    AnyDateInner,
+    #[cfg(feature = "unstable")],
```
Manishearth, line 394: nit: I don't think this should be an input to the macro.

components/calendar/src/any_calendar.rs
Manishearth: Does this trait need to be public yet? We can treat it as internal for now.

The way I see it, this change opens up the floor for making more AC-like types, OR for making a public macro, both of which can be discussed independently of this PR if this PR has no public API changes.

robertbastian: It's not public yet, but will be in https://github.com/unicode-org/icu4x/pull/7314

components/calendar/src/any_calendar.rs
robertbastian: how would you make this work cross-crate? require the `unstable` feature to use the macro?

components/calendar/src/any_calendar.rs
robertbastian: Yes. I don't want to generate an `AnyCalendarKind` for every `AnyCalendar`; I think the type is generally not needed. The resolution that DTF needs can be done entirely in the `CaledarAlgorithm` space, so it should live on `CalendarPreferences`.

components/calendar/src/any_calendar.rs
Manishearth: I'm not yet convinced I want this to work cross crate. I would like to land this change without having to think about that broader issue, or having public API changes to consider.

components/calendar/src/any_calendar.rs
Manishearth: I would like to fully decouple the publicness of this macro from adding the macro.

One very valid route is that we do not make this macro public but we make a handful of AnyCalendar-like types that contain useful collections for users.

Another very valid route is that we do make this macro public but we tweak it to not need this type of trait. Or we rename the trait. I don't want to have to think about that here, I would prefer landing this so we can iterate on top of it.

sffc:
I would appreciate if you would manually list out the new exported APIs in this PR (really, it should be a process requirement). I am approving based on my understanding:

1. `CalendarPreferences::resolved_algorithm`
    - Is this the best name for the function?
2. Some new From and TryFrom impls between concrete calendars, AnyCalendar, and references thereof. Some of these already existed but not all of them.

There are other things marked `pub` but which are not being exported.

Manishearth:
I guess AnyCalendarable is not exported because it's in a private module. I'd prefer those things not be marked `pub`, but that's fine I guess.

robertbastian:
The new APIs are

* `CalendarPreferences::resolved_algorithm`
* `impl PartialEq<$ty> for AnyCalendar`
* `impl From<$ty> for AnyCalendar`
* `impl TryFrom<AnyCalendar> for $ty`
* `impl<'a> TryFrom<&'a AnyCalendar> for &'a $ty`

The last three are currently available as `IntoAnyCalendar::to_any(self)`, `IntoAnyCalendar::from_any(AnyCalendar) -> Result<Self, AnyCalendar>`, and `IntoAnyCalendarfrom_any_ref(&AnyCalendar) -> Option<&Self>`. The `PartialEq` impl is useful for avoiding calendrical calculations when converting dates, such as in `Date::convert_any`. It's also used internally by the `AnyCalendar` implementation to check calendar equality for parameterized calendars.

components/calendar/src/any_calendar.rs
```diff
@@ -881,6 +783,98 @@ impl fmt::Display for AnyCalendarKind {
...
+impl AnyCalendarable for KoreanTraditional {
+    type Identity = ();
+
+    fn identity(&self) -> Self::Identity {}
+}
+impl AnyCalendarable for Ethiopian {
+    type Identity = ();
+
+    fn identity(&self) -> Self::Identity {}
+}
```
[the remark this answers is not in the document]
sffc: The `impl PartialEq for AnyCalendar` seems to be new in this PR. I think I agree with Gemini that we should not equate the two Ethiopian variants as equal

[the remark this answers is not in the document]
robertbastian: We currently allow mixing dates between both Ethiopian variants. I don't know why we do that, and I doubt it's fully consistent, but that's why I implemented it like this.

PR #7318, opened 2025-12-15
Clean up `FormattableAnyCalendar`

robertbastian:
Trying to reduce usage of `AnyCalendarKind` and `FormattableAnyCalendarKind`.

### DIFF
--- a/components/datetime/src/neo.rs
+++ b/components/datetime/src/neo.rs
@@ -548,8 +548,7 @@ where
     {
         let selection = DateTimeZonePatternSelectionData::try_new_with_skeleton(
             &FormattableAnyCalendarNamesLoader::<<FSet::D as DateDataMarkers>::Skel, _>::new(
-                provider_p,
-                calendar.kind(),
+                provider_p, &calendar,
             ),
             &<FSet::T as TimeMarkers>::TimeSkeletonPatternsV1::bind(provider_p),
             &FSet::GluePatternV1::bind(provider_p),
@@ -567,12 +566,10 @@ where
         };
         let result = names.load_for_pattern(
             &FormattableAnyCalendarNamesLoader::<<FSet::D as DateDataMarkers>::Year, _>::new(
-                provider,
-                calendar.kind(),
+                provider, &calendar,
             ),
             &FormattableAnyCalendarNamesLoader::<<FSet::D as DateDataMarkers>::Month, _>::new(
-                provider,
-                calendar.kind(),
+                provider, &calendar,
             ),
             &<FSet::D as DateDataMarkers>::WeekdayNamesV1::bind(provider),
             &<FSet::T as TimeMarkers>::DayPeriodNamesV1::bind(provider),
--- a/components/datetime/src/scaffold/calendar.rs
+++ b/components/datetime/src/scaffold/calendar.rs
@@ -299,7 +299,7 @@ impl IntoFormattableAnyCalendar for Roc {}
 
 // keep in sync with IntoFormattableAnyCalendar
 #[derive(Debug, Clone, Copy, PartialEq)]
-pub(crate) enum FormattableAnyCalendarKind {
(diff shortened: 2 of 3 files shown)


sffc:
Can you give an overview? What problem does this solve?

/gemini review

components/datetime/src/neo.rs
```diff
@@ -425,7 +425,7 @@ size_test!(
 pub struct DateTimeFormatter<FSet: DateTimeNamesMarker> {
     pub(crate) selection: DateTimeZonePatternSelectionData,
     pub(crate) names: RawDateTimeNames<FSet>,
-    pub(crate) calendar: UntaggedFormattableAnyCalendar,
+    pub(crate) calendar: FormattableAnyCalendar,
```
sffc, line 428: Observation: I think I had introduced `UntaggedFormattableAnyCalendar` because it reduced the stack size of `DateTimeFormatter`. But, the tests that enforce stack size are still passing, so maybe there's a new niche where the extra field can fit now.

components/datetime/src/scaffold/calendar.rs
```diff
@@ -347,235 +348,243 @@ impl FormattableAnyCalendarKind {
...
+            Hijri(Some(HijriCalendarAlgorithm::Umalqura)) => {
+                AnyCalendar::HijriUmmAlQura(cal::Hijri::new_umm_al_qura())
+            }
             Japanese => AnyCalendar::Japanese(cal::Japanese::new()),
             Persian => AnyCalendar::Persian(cal::Persian),
             Roc => AnyCalendar::Roc(cal::Roc),
+            Iso8601 | Hijri(_) => {
+                // unsupported
+                prefs.calendar_algorithm = None;
+                return Self::new(prefs);
```
sffc: Observation/Issue: This logic that used to live in 1 place now lives in 3 places

components/datetime/src/neo.rs
robertbastian: FormattableAnyCalendar is now just untagged

components/datetime/src/scaffold/calendar.rs
robertbastian: fixed

components/datetime/src/scaffold/calendar.rs
```diff
@@ -424,107 +415,68 @@ impl FormattableAnyCalendar {
...
             Persian => AnyCalendar::Persian(cal::Persian),
             Roc => AnyCalendar::Roc(cal::Roc),
+            Iso8601 | Hijri(_) => {
+                // unsupported
+                prefs.calendar_algorithm = None;
+                return Self::try_new_unstable(provider, prefs);
+            }
+            _ => {
+                // unknown
+                AnyCalendar::Gregorian(Gregorian)
```
sffc, line 479: Nit: Debug-assert and probably use the Coptic fallback

robertbastian:
> Commits 1 and 2 are fine. I'm not convinced commit 3 "reduce FormattableAnyCalendarKind usage" is an improvement.

If you think commits 1 and 2 are fine, and there's not anything inherently wrong with commit 3, you could just approve this PR to avoid us both investing more time in it.

> I think it's useful to separate out in code where you need the actual calendar (for conversion, etc) and where you just need the kind (for data loading, etc).

As a general principle it's not good style to pass around derived data instead of letting the receiver of that data derive it themselves. Computing data as late as possible makes refactoring easier, and avoids computing data that will not even be used.

sffc:
I have trouble reviewing with PRs like this because it's not clear what the goal is. In my opinion, commit 1 is slightly positive, commit 2 is neutral, and commit 3 is slightly negative, and none of the commits impact the user, and none of the commits block other PRs. We should approve PRs when they improve the project, not just because someone did the work

sffc (approved):
Overall this is more positive than negative so I'll approve it

PR #7322, opened 2025-12-16
Remove `JapaneseExtended`

robertbastian:
Replace by a reexport of `Japanese`.

This is in preparation for [CLDR-11400](https://unicode-org.atlassian.net/browse/CLDR-11400)

## Changelog

icu_calendar: Replace `JapaneseExtended` with reexport of `Japanese`, since CLDR no longer includes pre-Meiji eras.


Manishearth (approved):
Changing type identity (two distinct types are now the same type) is technically a stability hazard, but a minor one at that.

I _think_ there are no issues with removing the data markers? New-data-old-code is broken but only at the key availability level.

We should also deprecate the AnyCalendarKind and AnyCalendar enum variants.

sffc:
There's an issue for this: #6491

sffc:
Do you have measurements on the binary size impact?

robertbastian:
I tried to keep this PR minimal, so I don't think the codesize differences here are significant. The only code that is actually removed is the binary search over eras, and some constructors (none that have heavy monomorphisations).

Take a look at the fingerprints files for data size impact.

#7323, which requires this, has more data size improvements, code size improvements, and memory usage improvements.

hsivonen:
> I don't currently see a path where I would approve this unless

Does https://github.com/unicode-org/icu4x/issues/7422#issuecomment-3747381709 mean you can now approve this?

robertbastian:
> The CLDR TC actually removes the historical era data (not just an announcement that they're considering it)

This is happening

provider/source/src/datetime/semantic_skeletons.rs
```diff
@@ -578,13 +578,425 @@ impl_datetime_skeleton_datagen!(DatetimePatternsDateGregorianV1, DatagenCalendar
...
-);
-impl_datetime_skeleton_datagen!(
-    DatetimePatternsDateJapanextV1,
-    DatagenCalendar::JapaneseExtended
-);
+impl_datetime_skeleton_datagen!(DatetimePatternsDateJapaneseV1, DatagenCalendar::Japanese);
 impl_datetime_skeleton_datagen!(DatetimePatternsDatePersianV1, DatagenCalendar::Persian);
 impl_datetime_skeleton_datagen!(DatetimePatternsDateRocV1, DatagenCalendar::Roc);
+
+#[test]
```
sffc: I love tests but where did these tests come from?

sffc (commented):
This appears to be a clean removal without any side-effects like making AnyCalendar be Copy which I would want to be discussed separately.

I would still ideally like this to be synced with a CLDR pull request that deletes the data, but I'll drop the blocking review

provider/source/src/datetime/semantic_skeletons.rs
robertbastian: merge

robertbastian:
> I would still ideally like this to be synced with a CLDR pull request that deletes the data, but I'll drop the blocking review

How do you envision this working? We wait for CLDR 49 integration? Or we just wait for https://github.com/unicode-org/cldr/pull/5173 to be merged, which is in the next few hours?

PR #7328, opened 2025-12-18
Add full support for `U` datetime symbol

robertbastian:
Per LDML: 

> If the calendar does not provide cyclic year name data, or if the year value to be formatted is out of the range of years for which cyclic name data is provided, then numeric formatting is used (behaves like 'y').

### DIFF
--- a/components/datetime/src/pattern/formatter.rs
+++ b/components/datetime/src/pattern/formatter.rs
@@ -257,8 +257,10 @@ impl TryWriteable for FormattedDateTimePattern<'_> {
 #[cfg(test)]
 #[cfg(feature = "compiled_data")]
 mod tests {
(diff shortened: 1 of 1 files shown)


sffc (changes requested):
Meh, it's in LDML, but seems like a footgun. Returning an error if `U` is used in a non-cyclic calendar seems better, to inform clients that they should be using `y`. Let's please discuss with CLDR first.

robertbastian:
We can also return an error and format `y`. You already do that in other error cases.

sffc:
> We can also return an error and format `y`. You already do that in other error cases.

I'd approve if you did this

robertbastian:
Please approve

sffc (approved):

PR #7330, opened 2025-12-18
Add paragraph about crescent moon visibility to Hijri docs

sffc:
Split from #7325

### DIFF
--- a/components/calendar/src/cal/hijri.rs
+++ b/components/calendar/src/cal/hijri.rs
@@ -70,8 +70,8 @@ mod ummalqura_data;
 ///
 /// 1. [`UmmAlQura`], used in Saudi Arabia, is based on official predictions of crescent
 ///    timings published by the KACST.
-/// 2. [`TabularAlgorithm`], used by certain denominations such as Dawoodi Bohra, is
-///    based on a proleptic approximation of the length of a lunar year.
+/// 2. [`TabularAlgorithm`] is based on a proleptic approximation of the length of a lunar year.
+///    See the docs for information on the branches of Islam using it.
 ///
 /// # Calendar drift
 ///
@@ -278,6 +278,10 @@ impl Rules for UmmAlQura {
 /// See [`TabularAlgorithmEpoch`] and [`TabularAlgorithmLeapYears`] for customization.
 ///
 /// The most common version of these rules uses [`TabularAlgorithmEpoch::Friday`] and [`TabularAlgorithmLeapYears::TypeII`].
+///
+/// Tabular Islamic rules are used in denominations such as Dawoodi Bohra and other branches of Ismailism.
+/// Be sure to select the correct leap year and epoch parameters according to your use case. If ICU4X does
+/// not implement the parameters you need, please file an issue.
 ///
 /// When constructed with [`TabularAlgorithmLeapYears::TypeII`], and either [`TabularAlgorithmEpoch::Friday`] or [`TabularAlgorithmEpoch::Thursday`],
 /// this corresponds to the `"islamic-civil"` and `"islamic-tbla"` [CLDR calendars](https://unicode.org/reports/tr35/#UnicodeCalendarIdentifier) respectively.


components/calendar/src/cal/hijri.rs
```diff
@@ -57,6 +57,18 @@ mod ummalqura_data;
 ///
 /// There are either 6 or 7 30-day months, so the length of the year is 354 or 355 days.
 ///
+/// # Crescent moon visibility
+///
+/// According to Islam, months begin when an observer first sees the crescent moon.
```
robertbastian: when exactly? the same day? there's something about sunset, no?

components/calendar/src/cal/hijri.rs
```diff
@@ -57,6 +57,18 @@ mod ummalqura_data;
...
+///
+/// For centuries, astronomers have been developing criteria for predicting crescent moon
+/// visibility. However, most regions that use the Hijri calendar rely on observations that
+/// are impacted by atmospheric phenomena, meaning such predictions are only an approximation
+/// of ground truth.
```
robertbastian: this calendar is inherently observer-based. everyone can look at the sky and make their own "ground truth", so it's not really ground truth

components/calendar/src/cal/hijri.rs
```diff
@@ -57,6 +57,18 @@ mod ummalqura_data;
...
+///
+/// The primary exception is Saudi Arabia, where the KACST publishes official predictions of
```
robertbastian: primary exception to what? if it's to "most regions", it should be in that paragraph

components/calendar/src/cal/hijri.rs
Manishearth: A new moon can basically only occur after sunrise, you cannot see a new moon after sunset.

robertbastian: this is about crescent moons, not new moons, right?

I'm mainly asking if the day you see the crescent is day 0 or 1. that should be documented

Manishearth: A crescent moon occurs immediately after the moment of the new moon. The moment of the new moon can only occur during the day.

Whether that crescent is distinct enough is a different question.

Islam doesn't have "something about sunset" IIRC, it's Judaism which does that.

But yes, day 1 of the month is when the crescent moon is visible. Fair to want that documented.

components/calendar/src/cal/hijri.rs
sffc: Primary exception to the previous paragraph: "observations that are impacted by atmospheric phenomena"

components/calendar/src/cal/hijri.rs
sffc: My goal in this section is for readers to better understand the nature of the different Hijri Rules using language that is clear and accurate but doesn't add new invariants.

components/calendar/src/cal/hijri.rs
sffc: The term "ground truth" is used elsewhere in these docs. But, I rewrote this and my rewrite doesn't use the term.

components/calendar/src/cal/hijri.rs
sffc: I rewrote this to not frame it as an exception to a rule.

components/calendar/src/cal/hijri.rs
robertbastian: if you're trying to be accurate then say "on the day an observer first sees the crescent moon"

components/calendar/src/cal/hijri.rs
```diff
@@ -59,15 +59,21 @@ mod ummalqura_data;
...
+/// Currently, most groups that use the Hijri calendar rely on human observations of the
+/// crescent moon, which are impacted by atmospheric phenomena. As a result, it is not
+/// possible to predict the month start dates ahead of time.
 ///
-/// The primary exception is Saudi Arabia, where the KACST publishes official predictions of
-/// crescent timings for multiple centuries. See [`UmmAlQura`].
+/// However, some groups use criteria that are not impacted by these phenomena and are
+/// therefore suitable for future prediction.
+///
+/// ICU4X ships two of these:
```
robertbastian: `icu` or `icu_calendar` users might not know what "ICU4X" is. just say "this crate"

sffc: done

components/calendar/src/cal/hijri.rs
sffc: added "on the day"

components/calendar/src/cal/hijri.rs
```diff
@@ -57,6 +57,22 @@ mod ummalqura_data;
...
+/// Currently, most groups that use the Hijri calendar rely on human observations of the
+/// crescent moon, which are impacted by atmospheric phenomena. As a result, it is not
+/// possible to predict the month start dates ahead of time.
+///
+/// However, some groups use criteria that are not impacted by these phenomena and are
+/// therefore suitable for future prediction. This crate ships two:
+///
+/// 1. [`UmmAlQura`], used in Saudi Arabia, is based on official predictions of crescent
+///    timings published by the KACST.
+/// 2. [`TabularAlgorithm`], used by certain denominations such as Dawoodi Bohra, is
```
robertbastian: nit: you consistenly use "groups" earlier, but here you suddenly use "denomination". I haven't seen the term "denomination" associated with Islam much, according to [Wikipedia](https://en.wikipedia.org/wiki/Religious_denomination#Islam) "sects" or "schools" might be better terms. or just continue using "groups"

robertbastian (approved):

components/calendar/src/cal/hijri.rs
sffc: I wanted a word that covers both regions and sects.

https://en.wikipedia.org/wiki/Dawoodi_Bohra calls it a "denomination".

robertbastian: I don't see the need to cover regions, as this is not official in any region, so it's not a geographic thing.

components/calendar/src/cal/hijri.rs
```diff
@@ -57,6 +57,22 @@ mod ummalqura_data;
 ///
 /// There are either 6 or 7 30-day months, so the length of the year is 354 or 355 days.
 ///
+/// # Crescent moon visibility
+///
+/// According to Islam, months begin on the day when the crescent moon is visible in the sky.
```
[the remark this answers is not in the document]
sffc: /gemini Good point. I tweaked the language. It now reads: "According to Islam, the first day of a month begins when the crescent moon is visible at sunset." Is this better?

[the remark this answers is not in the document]
sffc: /gemini I think your phrasing is potentially less clear if you consider days to begin at sunset rather than midnight. How about: "According to Islam, the first day of a month begins after the crescent moon is visible at sunset."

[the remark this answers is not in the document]
sffc: I talked with you Gemini in another window on the following phrasing: "According to Islam, a new month begins the evening when the crescent moon is visible at sunset." And you said: 'That statement is excellent. It is concise, accurate, and avoids the "day/day after" ambiguity by using the word "evening."' So I will go with that version. Thanks.

components/calendar/src/cal/hijri.rs
robertbastian: > But yes, day 1 of the month is when the crescent moon is visible

this seems to disagree with what Gemini is arguing in the other thread. I was also under the assumption that it was the next day (i.e. you spot the crescent, and then celebrate Eid the next day). [this](https://www.moonsighting.org.uk/moon/sighting-reports/report-1447-07.html#:~:text=However%2C%20on%2021%20December%202025%2C%20the%20moon%20may%20be%20visible%20in%20Morocco%2C%20but%20%5Bmay%5D%20not%20be%20in%20the%20UK.%20Hence%2C%20Rajab%20is%20likely%20to%20start%20on%2022%20%26%2023%20December%202025%20in%20Morocco%20and%20the%20UK) says something similar

Manishearth: Oh, yes, sorry, you're right, I had my wires crossed. The islamic "day" does traditionally also begin after sunset.

sffc:
I consider this mergeable. Since we just had two weeks of holidays, I will wait for a few more days for feedback.

components/calendar/src/cal/hijri.rs
hsivonen: Wikipedia appears to suggest a different leap year cycle for Dawoodi Bohra. Are we sure that they use this leap year cycle and epoch?

Manishearth: Yeah, they might be using TypeIV.

hsivonen: Let's not claim Dawoodi Bohra using this then. AFAICT, the justification for `islamic-tbla` is Microsoft legacy and the justification for `islamic-civil` is IBM legacy. (I think one of them, not sure which one, matches historical official status in the Ottoman Empire, but that's not why either is here.)

Manishearth: Little frustrating that we have BCP47 codes for a bunch of legacy unused calendars but we do not for actually in-use calendars.

I remember when I researched this I confirmed that tbla was using the same epoch as the Bohra/Fatimid calendar, but I probably missed the leap year pattern difference.

Manishearth (approved):

components/calendar/src/cal/hijri.rs
```diff
@@ -263,6 +279,10 @@ impl Rules for UmmAlQura {
 ///
 /// The most common version of these rules uses [`TabularAlgorithmEpoch::Friday`] and [`TabularAlgorithmLeapYears::TypeII`].
 ///
+/// Tabular Islamic rules are used in denominations such as Dawoodi Bohra and other branches of Ismailism.
+/// Be sure to select the correct leap year and epoch parameters according to your use case. If ICU4X does
```
robertbastian, line 283:
```suggestion
/// Be sure to select the correct leap year and epoch parameters according to your use case. If this type does
```

sffc: https://github.com/unicode-org/icu4x/pull/7392

PR #7334, opened 2025-12-19
Untangle Japanese calendars

robertbastian:
Hopefully uncontroversial parts of #7322 and #7323 

This splits the implementation of `Japanese` and `JapaneseExtended` so that the complexity of `JapaneseExtended` doesn't contribute to `Japanese`'s code size.

The datagen part correctly implements the `inheritEras` logic, instead of hardcoding it for Japanese, as CLDR is considering changes to the model.


sffc (approved):
Seems fine; would be nice to share a bit of code where the logic is the same, like the first 20 lines of the GregorianYears function impls.

Manishearth (approved):
fine by me.

Issue #4413, opened 2023-12-06
Support u-rg in region priority fallback


sffc:
There is a comment in the fallback code:

```
// NOTE: -u-rg is not yet supported; when it is, this test should be updated
```

I think `-u-rg` should be supported in region priority fallback.

Should probably consult with the CLDR design group.

sffc:
We really should support `-u-rg`; it's not a low priority feature. It directly impacts i18n correctness.

One way to support both this and `-u-sd` would be for `LocaleFallbacker` to take all of these things as parameters and store them in internal slots, similar to how it currently deals with variants. It can still emit `LanguageIdentifier`s.

@robertbastian

sffc:
So for `-u-rg` I think we can still emit `LanguageIdentifer` but it's not clear to me how we'd do that for `-u-sd` unless we changed the data model.

Ideally we should support something like

```rust
pub struct DataLocale {
    pub language: Language,
    pub script: Script,
    pub region: Region,
    pub sd_only: SubdivisionOnly,
    pub variant: Variant
}

pub struct SubdivisionOnly(TinyStr6)
```

and then you'd get strings like

- `en-US`
- `en-US-ca`

PR #7337, opened 2025-12-22
Fix locale region override

robertbastian:
Simpler version of #7335

https://github.com/unicode-org/icu4x/issues/4413

### DIFF
--- a/components/locale_core/src/preferences/extensions/unicode/keywords/regional_subdivision.rs
+++ b/components/locale_core/src/preferences/extensions/unicode/keywords/regional_subdivision.rs
@@ -24,21 +24,7 @@ struct_keyword!(
             .ok_or(PreferencesParseError::InvalidKeywordValue)
     },
     |input: RegionalSubdivision| {
-        let mut raw = [0; 8];
-        raw[0] = input.0.region.into_raw()[0];
-        raw[1] = input.0.region.into_raw()[1];
-        raw[2] = input.0.region.into_raw()[2];
-        let len = input.0.region.as_str().len();
-        debug_assert!((2..=3).contains(&len));
-        #[allow(clippy::indexing_slicing)] // safe
-        {
-            raw[len] = input.0.suffix.into_raw()[0];
-            raw[len + 1] = input.0.suffix.into_raw()[1];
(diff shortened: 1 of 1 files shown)


components/datetime/src/pattern/names.rs
```diff
@@ -371,7 +371,7 @@ where
 size_test!(
     FixedCalendarDateTimeNames<icu_calendar::Gregorian>,
     typed_date_time_names_size,
-    328
+    336
```
sffc, line 374: Observation: Unfortunate that it increases stack size by 8 bytes

components/locale_core/src/extensions/unicode/subdivision.rs
```diff
@@ -133,7 +133,11 @@ impl SubdivisionId {
 
     /// Convert to [`Subtag`]
     pub fn into_subtag(self) -> Subtag {
-        let result = self.region.to_tinystr().concat(self.suffix.to_tinystr());
+        let result = self
+            .region
+            .to_tinystr()
+            .to_ascii_lowercase()
```
sffc, line 139: Question: is this a bugfix?

components/locale_core/src/preferences/locale.rs
```diff
@@ -65,16 +66,13 @@ impl From<&crate::Locale> for LocalePreferences {
             .unicode
             .keywords
             .get(&crate::extensions::unicode::key!("sd"))
-            .and_then(|v| v.as_single_subtag().copied());
+            .and_then(|v| SubdivisionId::try_from_str(v.as_single_subtag()?.as_str()).ok());
```
sffc, line 69: Suggestion: Add comments about invalid subtags being silently ignored?

Suggestion/Follow-up: Check that the subdivision region matches the subtag region?

components/locale_core/src/extensions/unicode/subdivision.rs
robertbastian: yes, it wasn't lowercasing the region as required and your test failed

components/locale_core/src/preferences/locale.rs
robertbastian: it's existing behaviour, we need to do more work here anyway

PR #7340, opened 2025-12-23
Exploratory: Policy on Bugs in Stable Crates

sffc:
This doc is marked "draft"; I would prefer to land it and iterate from there.

Background: see #7320

PR #7342, opened 2025-12-23
Change `AstronomicalSimulation`

robertbastian:
## Changelog

icu_calendar: Use UmmAlQura under the hood for AstronomicalSimulation.
  - Clients wishing to use the Reingold simulation are encouraged to write their own implementation of `hijri::Rules` using code from `calendrical_calculations`.

### DIFF
--- a/components/calendar/src/cal/hijri.rs
+++ b/components/calendar/src/cal/hijri.rs
@@ -1667,9 +1667,6 @@ mod test {
     }
 
     #[test]
-    fn test_regression_5069_obs() {}
-
-    #[test]
     fn test_regression_6197() {
         let calendar = Hijri::new_umm_al_qura();
 


components/calendar/src/any_calendar.rs
```diff
@@ -670,8 +670,6 @@ pub enum AnyCalendarKind {
     /// This corresponds to the `"hebrew"` [CLDR calendar](https://unicode.org/reports/tr35/#UnicodeCalendarIdentifier).
     Hebrew,
     /// The kind of an [`HijriSimulated`], Mecca calendar
-    ///
-    /// This corresponds to the `"islamic-rgsa"` [CLDR calendar](https://unicode.org/reports/tr35/#UnicodeCalendarIdentifier).
```
sffc, line 763: Please say something like "there is currently no CLDR calendar that corresponds to this"

components/calendar/src/any_calendar.rs
```diff
@@ -764,7 +762,7 @@ impl TryFrom<CalendarAlgorithm> for AnyCalendarKind {
             Hijri(Some(HijriCalendarAlgorithm::Civil)) => {
                 Ok(AnyCalendarKind::HijriTabularTypeIIFriday)
             }
-            Hijri(Some(HijriCalendarAlgorithm::Rgsa)) => Ok(AnyCalendarKind::HijriSimulatedMecca),
+            Hijri(Some(HijriCalendarAlgorithm::Rgsa)) => Err(()),
```
sffc, line 834: We don't return `Err` elsewhere in this fn; what is the impact?

components/calendar/src/cal/hijri.rs
```diff
@@ -127,6 +127,45 @@ pub trait Rules: Clone + Debug + crate::cal::scaffold::UnstableSealed {
...
+/// These simulations are unofficial and are known to not necessarily match sightings
+/// on the ground. Unless you know otherwise for sure, instead of this variant, use
+/// [`UmmAlQura`], which uses the results of KACST's Mecca-based calculations.
+///
+/// As floating point arithmetic degenerates for far-away dates, this falls back to
+/// the tabular calendar at some point.
+///
+/// The precise behavior of this calendar may change in the future if:
+/// - We decide to tweak the precise astronomical simulation used
+/// - We decide to expand or reduce the range where we are using the astronomical simulation.
```
```suggestion
/// [`Hijri`] [`Rules`] based on astronomical simulations published by the KACST.
///
/// This is identical to [`UmmAlQura`].
///
/// The precise behavior of this calendar may change in the future if:
/// - We decide to tweak the precise astronomical simulation used
/// - We decide to expand or reduce the range where we are using the astronomical simulation.
```

components/calendar/src/cal/hijri.rs
```diff
@@ -141,36 +180,30 @@ pub trait Rules: Clone + Debug + crate::cal::scaffold::UnstableSealed {
...
 /// [^1]: See [calendrical_calculations::islamic::observational_islamic_from_fixed]
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
-pub struct AstronomicalSimulation {
-    pub(crate) location: SimulatedLocation,
-}
-
-#[derive(Clone, Debug, Copy, PartialEq, Eq)]
-pub(crate) enum SimulatedLocation {
+#[non_exhaustive]
+pub enum ReingoldSimulation {
```
sffc: Issue: I prefer the old model where the fields are private because we might want fields to tweak behavior besides just the location, perhaps the crescent moon criterion, or how far back to run calculations or something

sffc (commented):
I think this direction is consistent with #7340, since it takes advantage of the bullet point "We decide to tweak the precise astronomical simulation used" and keeps the type `AstronomicalSimulation` as representing an astronomical simulation, just a different one than we had been shipping before.

components/calendar/src/any_calendar.rs
robertbastian: yes we do, for `-u-ca-islamic`

components/calendar/src/cal/hijri.rs
robertbastian: we can still change things because this is unstable. I prefer exposing one and not two unstable types

Manishearth: At the very least we should then have some scary internal comment "TODO(before stabilization): decide whether fields are private or not.". I fear accidentally stabilizing it.

and also perhaps add this to a tracking issue somewhere.

components/calendar/src/cal/hijri.rs
```diff
@@ -178,12 +178,15 @@ impl Rules for AstronomicalSimulation {
 /// [^1]: See [calendrical_calculations::islamic::observational_islamic_from_fixed]
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 #[non_exhaustive]
+#[cfg(feature = "unstable")] // TODO(before stabilisation): design API
```
Manishearth:
```suggestion
#[cfg(feature = "unstable")] // TODO(before stabilisation): should this be an enum? a struct with private fields?
```

I wanted to call out specifically that this is an important thing to consider, and that this aspect is in flux.

components/calendar/src/cal/hijri.rs
```diff
@@ -127,6 +127,40 @@ pub trait Rules: Clone + Debug + crate::cal::scaffold::UnstableSealed {
...
+/// [`Hijri`] [`Rules`] based on astronomical simulations published by the KACST.
+///
+/// This is identical to [`UmmAlQura`].
+///
+/// The precise behavior of this calendar may change in the future if:
+/// - We decide to tweak the precise astronomical simulation used
+/// - We decide to expand or reduce the range where we are using the astronomical simulation.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[non_exhaustive]
+pub struct AstronomicalSimulation;
```
hsivonen, line 154: Why isn't this deprecated if this duplicates Umm al-Qura behavior and exists for semver?

components/calendar/src/any_calendar.rs
hsivonen: Thank you for removing this mapping.

components/calendar/src/cal/hijri.rs
```diff
@@ -141,36 +175,33 @@ pub trait Rules: Clone + Debug + crate::cal::scaffold::UnstableSealed {
...
 /// [^1]: See [calendrical_calculations::islamic::observational_islamic_from_fixed]
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
-pub struct AstronomicalSimulation {
-    pub(crate) location: SimulatedLocation,
-}
-
-#[derive(Clone, Debug, Copy, PartialEq, Eq)]
-pub(crate) enum SimulatedLocation {
+#[non_exhaustive]
+#[cfg(feature = "unstable")] // TODO(before stabilisation): should this be an enum? a struct with private fields?
```
hsivonen: I'm posting an accepting review, because putting this behind a flag is better than the status quo.

components/calendar/src/cal/hijri.rs
robertbastian: I'm trying to do this in as small chunks as possible to make *some* progress

components/calendar/src/cal/hijri.rs
```diff
@@ -127,6 +127,40 @@ pub trait Rules: Clone + Debug + crate::cal::scaffold::UnstableSealed {
     }
 }
 
+/// [`Hijri`] [`Rules`] based on astronomical simulations published by the KACST.
+///
+/// This is identical to [`UmmAlQura`].
```
sffc: I know I suggested this language previously, but I want to change it again

```suggestion
/// [`Hijri`] [`Rules`] based on astronomical simulations.
///
/// Currently, this uses simulation results published by the KACST.
/// It is therefore currently identical to [`UmmAlQura`].
```

components/calendar/src/any_calendar.rs
```diff
@@ -670,8 +670,6 @@ pub enum AnyCalendarKind {
     /// This corresponds to the `"hebrew"` [CLDR calendar](https://unicode.org/reports/tr35/#UnicodeCalendarIdentifier).
     Hebrew,
     /// The kind of an [`HijriSimulated`], Mecca calendar
```
sffc:
```suggestion
    /// The kind of a [`Hijri`] calendar with [`AstronomicalSimulation`] rules
    ///
    /// ICU4X does not map any CLDR calendar algorithm to this calendar kind.
```

components/calendar/src/any_calendar.rs
sffc: I think this creates a new code path. `-u-ca-islamic` doesn't make it into this impl; it is remapped in `resolved_algorithm()` above. This new case means that `-u-ca-islamic-rgsa` maps unconditionally to Gregorian due to the `unwrap_or(Self::Gregorian)` in `AnyCalendarKind::new`. It should instead map to the locale default.

sffc:
Consensus: As agreed in #7422, this can be handled by migration notes; we don't need to retain the ReingoldSimulation type.

components/calendar/src/any_calendar.rs
robertbastian: please stop using the term ICU4X in docs

components/calendar/src/any_calendar.rs
robertbastian: should it go to the locale default, or should it behave like `-u-ca-islamic` which goes to the locale default Hijri?

sffc: You've now specifically documented, on `resolved_algorithm`,

```
    /// This is guaranteed to return a [`CalendarAlgorithm`] that successfully converts into an
    /// [`AnyCalendarKind`] (i.e. not `-u-ca-islamic` or `-u-ca-islamic-rgsa`).
```

I feel like `islamic-rgsa` should have the same behavior as `islamic-zzz`.

components/calendar/tests/reingold.rs
```diff
@@ -0,0 +1,72 @@
...
+/// The simulations are pre-computed for Gregorian years 1900 to 2140, falling back to
+/// a tabular approximation outside that range.
+///
+/// The precise behavior of this calendar may change in the future if:
+/// - We decide to tweak the precise astronomical simulation used
+/// - We decide to expand or reduce the range where we are using the astronomical simulation.
+///
+/// [^1]: See [calendrical_calculations::islamic::observational_islamic_from_fixed]
+#[derive(Debug, Clone, Copy)]
+pub struct ReingoldSimulation;
```
sffc, line 19: Praise: good to retain this code even if outside of src. Is tests the right place, or should it be in examples?

robertbastian: I don't want to advertise this

components/calendar/src/any_calendar.rs
robertbastian: done

components/calendar/src/any_calendar.rs
```diff
@@ -769,23 +769,23 @@ pub enum AnyCalendarKind {
     ///
     /// This corresponds to the `"hebrew"` [CLDR calendar](https://unicode.org/reports/tr35/#UnicodeCalendarIdentifier).
     Hebrew,
-    /// The kind of an [`HijriSimulated`], Mecca calendar
+    /// The kind of a [`HijriSimulated`] calendar
     ///
-    /// This corresponds to the `"islamic-rgsa"` [CLDR calendar](https://unicode.org/reports/tr35/#UnicodeCalendarIdentifier).
+    /// This does not correspond to a CLDR calendar.
```
sffc, line 763: Issue: Well, actually it does correspond to a CLDR calendar, just not one that ICU4X chooses to implement. The previous language is still correct.

components/calendar/src/any_calendar.rs
```diff
@@ -813,28 +813,37 @@ pub enum AnyCalendarKind {
...
-    /// use icu::calendar::preferences::{CalendarAlgorithm, CalendarPreferences};
+    /// use icu::calendar::preferences::{CalendarAlgorithm, CalendarPreferences, HijriCalendarAlgorithm};
     /// use icu::locale::locale;
     ///
     /// assert_eq!(CalendarPreferences::from(&locale!("und")).resolved_algorithm(), CalendarAlgorithm::Gregory);
     /// assert_eq!(CalendarPreferences::from(&locale!("und-US-u-ca-hebrew")).resolved_algorithm(), CalendarAlgorithm::Hebrew);
     /// assert_eq!(CalendarPreferences::from(&locale!("und-AF")).resolved_algorithm(), CalendarAlgorithm::Persian);
     /// assert_eq!(CalendarPreferences::from(&locale!("und-US-u-rg-thxxxx")).resolved_algorithm(), CalendarAlgorithm::Buddhist);
+    /// assert_eq!(
+    ///     CalendarPreferences::from(&locale!("und-US-u-ca-islamic")).resolved_algorithm(),
```
sffc: Nit: We should have a test for what happens with `islamic-zzz`, `islamic-rgsa`, and `buddhist-zzz`.

components/calendar/src/any_calendar.rs
```diff
@@ -813,28 +813,37 @@ pub enum AnyCalendarKind {
...
     /// ```
     pub fn resolved_algorithm(self) -> CalendarAlgorithm {
         let region = self.locale_preferences.region();
         let region = region.as_ref().map(|r| r.as_str());
         // This is tested to be consistent with CLDR in icu_provider_source::calendar::test_calendar_resolution
         match self.calendar_algorithm {
-            Some(CalendarAlgorithm::Hijri(None)) => match region {
-                Some("AE" | "BH" | "KW" | "QA" | "SA") => {
-                    CalendarAlgorithm::Hijri(Some(HijriCalendarAlgorithm::Umalqura))
+            Some(CalendarAlgorithm::Hijri(None | Some(HijriCalendarAlgorithm::Rgsa))) => {
```
hsivonen: Is it infeasible to make `rgsa` fall back to the locale default calendar at this point, consistent with TG2?

(Out of scope for this PR, but making the behavior for unqualified `islamic` differ from 402 creates an unnecessary complication and source of bugs for 402 implementations.)

robertbastian: It's not. However, `-u-ca-islamic-rgeg` falls back to the default Hijri calendar...

components/calendar/src/any_calendar.rs
robertbastian: the `AnyCalendarKind`s correspond to ICU4X calendars, `CalendarAlgorithm`s correspond to CLDR calendars. The ICU4X calendar `Hijri<AstronomicalSimulation>` does not correspond to a `CalendarAlgorithm` (it's not UAQ because it might still change), and therefore doesn't correspond to a CLDR calendar.

components/calendar/src/any_calendar.rs
hsivonen: I guess that, too, is then in principle out of scope for this PR. Filed #7447.

components/calendar/src/any_calendar.rs
sffc: "CLDR calendar" literally links to UTS 35, not ICU4X CalendarAlgorithm.

sffc (dismissed):

components/calendar/src/any_calendar.rs
robertbastian: `CalendarAlgorithm` is the list of UTS 35 calendar preferences, they're the same thing.

"CLDR calendar" linking to UTS 35 does not invalidate my argument:

> The ICU4X calendar `Hijri<AstronomicalSimulation>` does not correspond to a `CalendarAlgorithm` (it's not UAQ because it might still change), and therefore doesn't correspond to a CLDR calendar.

components/calendar/src/cal/hijri.rs
```diff
@@ -1912,17 +1667,7 @@ mod test {
...
-
-        let dt = Date::try_new_hijri_with_calendar(1390, 1, 30, cal).unwrap();
-
-        assert_eq!(Date::from_rata_die(dt.to_rata_die(), cal), dt);
-
-        let dt = Date::try_new_iso(2000, 5, 5).unwrap().to_calendar(cal);
-
-        assert!(dt.day_of_month().0 > 0);
-    }
+    fn test_regression_5069_obs() {}
```
sffc: If this test is obsolete, please delete it

components/calendar/src/any_calendar.rs
sffc: ok

components/calendar/src/cal/hijri.rs
robertbastian: yes this moved into the new example

sffc (approved):
it implements the working group consensus.

Issue #3214, opened 2023-03-21
Add configuration to use system BLAS in ML calculations


sffc:
We currently perform our ML calculations on the CPU. We have managed to make them about as fast as possible on the CPU by using SIMD / vectorization. However, it may improve performance further if we could use the GPU. I think the easiest and best way to do this is to invoke the system BLAS.

A concern I have is that these LSTM calculations, although slow, are still smaller faster than most ML model inferences, as far as I believe. I'm not convinced that delegating to the GPU to left-multiply / dot-product some NxN matrices (where N < 100) is going to give us the wins we'd be hoping to see.

Also, we probably don't want to invoke discrete GPU because those use more battery power. It's probably fine if we use the embedded GPU.

This seems to be a good article on the topic: https://www.erikpartridge.com/2019-03/rust-ml-simd-blas-lapack

CC @FrankYFTang

PR #7344, opened 2025-12-24
GEMV Optimisation of CNN

opnuub:
GEMV Optimisation instead of BLAS [(#3214](https://github.com/unicode-org/icu4x/issues/3214)) due to nature of Conv1D. Achieved similar speedups. Due to the monotonic nature of softmax, also removed softmax at inference and thus updated ```python_test_output.txt``` 

@sffc

### DIFF
--- a/components/segmenter/tests/cnn/main.rs
+++ b/components/segmenter/tests/cnn/main.rs
@@ -185,7 +185,7 @@ impl RawCnnData {
 
         let cnn_data_float32 = CnnDataFloat32::try_from_parts(
             model,
-            self.dic.iter().map(|(k, &v)| (k.clone(), v)).collect(),
+            self.dic.iter().map(|(k, &v)| (*k, v)).collect(),
             ndarray_to_cnn_matrix2(embedding)?,
             ndarray_to_cnn_matrix3(cnn_w1)?,
             ndarray_to_cnn_matrix1(cnn_b1)?,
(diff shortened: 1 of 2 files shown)


components/segmenter/tests/cnn/main.rs
```diff
@@ -137,7 +144,7 @@ impl RawCnnMatrix {
 #[derive(Deserialize, Debug)]
 pub struct RawCnnData {
     model: String,
-    dic: HashMap<String, u16>,
+    dic: HashMap<char, u16>,
```
sffc, line 147: Good to use `char` when possible over `String`

components/segmenter/tests/cnn/python_test_output.txt
```diff
@@ -1,320 +1,320 @@
-9.81858909e-01
```
sffc, line 1: Why did these numbers change? Is there a corresponding Python change?

sffc (approved):
Nice progress!

components/segmenter/tests/cnn/python_test_output.txt
opnuub: Yes, I removed the softmax activation function in Python as well and printed the output. This is a small optimisation that works because softmax is monotonic and we are doing top-1 prediction (argmax).

PR #7347, opened 2025-12-24
Add docs tests for SubdivisionId and preferences types

sffc:
Split from #7335

### DIFF
--- a/components/locale_core/src/extensions/unicode/subdivision.rs
+++ b/components/locale_core/src/extensions/unicode/subdivision.rs
@@ -66,12 +66,13 @@ impl_tinystr_subtag!(
 ///     subtags::region,
 /// };
 ///
-/// let ss = subdivision_suffix!("eng");
+/// // "zzzz" means "unknown subdivision"
+/// let ss = subdivision_suffix!("zzzz");
 /// let region = region!("gb");
 ///
 /// let si = SubdivisionId::new(region, ss);
 ///
-/// assert_eq!(si.to_string(), "gbeng");
+/// assert_eq!(si.to_string(), "gbzzzz");
 /// ```
 #[derive(Debug, PartialEq, Eq, Clone, Hash, PartialOrd, Ord, Copy)]
 #[non_exhaustive]


components/locale_core/src/extensions/unicode/subdivision.rs
```diff
@@ -131,6 +166,35 @@ impl SubdivisionId {
...
+    /// assert_writeable_eq!(subdivision.suffix, "eng");
+    ///
+    /// // Value is empty (special value "true")
+    /// // See also: https://unicode-org.atlassian.net/browse/CLDR-19163
+    /// assert!(matches!(
+    ///     SubdivisionId::try_from_value(&value!("true")),
+    ///     Err(ParseError::InvalidExtension),
+    /// ));
+    /// ```
+    pub fn try_from_value(value: &Value) -> Result<Self, ParseError> {
```
robertbastian: it's weird to have `try_from_value` but `into_subtag`

given that `try_from_subtag` is already public, and we don't have `try_from_value` on any other preference, and this type represents a single subtag, this should be `try_from_subtag`. but that function is very trivial, so I'm not sure this is worth adding.

sffc: It is into_subtag because all SubdivisionIds are single subtags. It's about getting the raw value.

It is try_from_value because most of the time when you want one of these you start from a Value because that's the type coming from Locale. It's about interop.

I'm open to adding to_value to complete the quartet if your concern is consistency.

sffc: We have impl TryFrom &Value on the preference enums. Example:

https://unicode-org.github.io/icu4x/rustdoc/icu_locale_core/preferences/extensions/unicode/keywords/enum.CalendarAlgorithm.html

I prefer named functions but would change this to a TryFrom impl if it helps it get merged.

robertbastian: And we have `TryFrom<Value>` for `RegionOverride` and `RegionalSubdivision`, which are the preferences. This type is not a preference, it's what's stored in the preference.

robertbastian: And https://github.com/unicode-org/icu4x/pull/7341 uses those code paths

robertbastian (changes requested):
.

components/locale_core/src/extensions/unicode/subdivision.rs
robertbastian: I've split the relevant part of #7341 into #7360

sffc: ok, and apparently `RegionalSubdivision` has a Deref impl to `SubdivisionId` which means we effectively already have a `TryFrom` it's just super indirect.

robertbastian: it's not indirect, it's direct on the outer types which you should be using instead of the inner type

sffc: Well in the tests here I care about SubdivisionId and not whether it is originating from a region preference or a subdivision preference.

robertbastian: that's a test setup problem which doesn't justify adding public APIs. just construct through `RegionalSubdivision` or `RegionOverride`

sffc:
> f897a83

Split to #7362

components/locale_core/src/extensions/unicode/subdivision.rs
```diff
@@ -61,17 +61,64 @@ impl_tinystr_subtag!(
...
 ///
-/// let ss = subdivision_suffix!("zzzz");
-/// let region = region!("gb");
+/// // American English with British user preferences
+/// let locale = locale!("en-US-u-rg-gbzzzz");
 ///
-/// let si = SubdivisionId::new(region, ss);
+/// let normal_region = locale.id.region.unwrap();
+/// let rg_extension_value = locale.extensions.unicode.keywords.get(&key!("rg")).unwrap();
+/// let subdivision = RegionOverride::try_from(rg_extension_value.clone()).unwrap();
```
robertbastian: keep the names aligned - `rg` is `RegionOverride`, `sd` is `RegionalSubdivision`

```suggestion
/// let rg_extension_value = locale.extensions.unicode.keywords.get(&key!("rg")).unwrap();
/// let region_override = RegionOverride::try_from(rg_extension_value.clone()).unwrap();
```

robertbastian:
```suggestion
/// let sd_extension_value = locale.extensions.unicode.keywords.get(&key!("sd")).unwrap();
/// let subdivision = RegionalSubdivision::try_from(rg_extension_value.clone()).unwrap();
```

components/locale_core/src/extensions/unicode/subdivision.rs
```diff
@@ -61,17 +61,64 @@ impl_tinystr_subtag!(
 /// # Examples
 ///
 /// ```
-/// use icu::locale::{
-///     extensions::unicode::{subdivision_suffix, SubdivisionId},
-///     subtags::region,
-/// };
+/// use icu::locale::locale;
```
robertbastian: I'd prefer if this example lived on both `RegionOverride` and `RegionalSubdivision`, instead of on this inner type

components/locale_core/src/extensions/unicode/subdivision.rs
```diff
@@ -61,17 +61,64 @@ impl_tinystr_subtag!(
...
+///
+/// When the value can't be parsed:
+///
+/// ```
+/// use icu::locale::extensions::unicode::SubdivisionId;
+/// use icu::locale::ParseError;
+///
+/// // Value is too short
+/// assert!(matches!(
+///     SubdivisionId::try_from_str("zz"),
```
robertbastian: move the `try_from_str` error cases to `try_from_str`

robertbastian (approved):
thanks!

PR #7355, opened 2025-12-29
implemented ISO 8601 date duration parsing

### LINKED ISSUE #7030: Add ISO-8601 style Duration parsing
It would be convenient for users and for writing unit tests if we could parse ISO-8601 duration strings into DateDuration.

Examples:

- `P1D` => `{ days: 1 }`
- `P2M1D` => `{ months: 2, days: 1 }`
- `P77Y` => `{ years: 77 }`

SATVIKsynopsis:
Fixes #7030

This change adds support for parsing ISO 8601 date only duration strings into `DateDuration`. The implementation validates the overall structure, rejects time components and duplicate units, and correctly handles negative values and numeric overflow.

### DIFF
--- a/components/calendar/src/duration.rs
+++ b/components/calendar/src/duration.rs
@@ -265,31 +265,26 @@ impl DateDuration {
         Self::try_from_utf8(s.as_bytes())
     }
 
-    /// Parses an ISO 8601 date-only duration from UTF-8 encoded bytes.
-    ///
-    /// This function operates directly on byte slices in order to support
-    /// unvalidated UTF-8 inputs.
-    ///
-    /// Supported units are years (`Y`), months (`M`), weeks (`W`), and days (`D`).
-    /// Time based components are rejected.
+    /// See [`Self::try_from_str`].
     pub fn try_from_utf8(code_units: &[u8]) -> Result<Self, DateDurationParseError> {
-        let mut index = 0;
-        let length = code_units.len();
+        let mut s = code_units;
 
         let mut is_negative = false;
-        if index < length && code_units[index] == b'-' {
-            is_negative = true;
-            index += 1;
-        } else if index < length && code_units[index] == b'+' {
-            return Err(DateDurationParseError::PlusNotAllowed);
+        match s {
+            [b'-', rest @ ..] => {
+                is_negative = true;
+                s = rest;
+            }
+            [b'+', ..] => return Err(DateDurationParseError::PlusNotAllowed),
+            _ => {}
         }
 
-        if index >= length || code_units[index] != b'P' {
-            return Err(DateDurationParseError::InvalidStructure);
+        match s {
+            [b'P', rest @ ..] => s = rest,
+            _ => return Err(DateDurationParseError::InvalidStructure),
         }
-        index += 1;
 
-        if index >= length {
+        if s.is_empty() {
             return Err(DateDurationParseError::InvalidStructure);
         }
 
@@ -303,66 +298,65 @@ impl DateDuration {
         let mut seen_weeks = false;
         let mut seen_days = false;
 
-        while index < length {
-            if code_units[index] == b'T' {
+        while !s.is_empty() {
+            if matches!(s, [b'T', ..]) {
                 return Err(DateDurationParseError::TimeNotSupported);
             }
+
             let mut value: u64 = 0;
             let mut has_digits = false;
 
-            while index < length && code_units[index].is_ascii_digit() {
-                has_digits = true;
+            while let [b @ b'0'..=b'9', rest @ ..] = s {
                 value = value
                     .checked_mul(10)
-                    .and_then(|v| v.checked_add((code_units[index] - b'0') as u64))
+                    .and_then(|v| v.checked_add((b - b'0') as u64))
                     .ok_or(DateDurationParseError::NumberOverflow)?;
-                index += 1;
+                s = rest;
+                has_digits = true;
             }
 
             if !has_digits {
                 return Err(DateDurationParseError::MissingValue);
             }
 
-            if index >= length {
-                return Err(DateDurationParseError::InvalidStructure);
-            }
-
-            match code_units[index] {
-                b'Y' => {
+            match s {
+                [b'Y', rest @ ..] => {
                     if seen_years {
                         return Err(DateDurationParseError::DuplicateUnit);
                     }
                     years =
                         u32::try_from(value).map_err(|_| DateDurationParseError::NumberOverflow)?;
                     seen_years = true;
+                    s = rest;
                 }
-                b'M' => {
+                [b'M', rest @ ..] => {
                     if seen_months {
                         return Err(DateDurationParseError::DuplicateUnit);
                     }
                     months =
                         u32::try_from(value).map_err(|_| DateDurationParseError::NumberOverflow)?;
                     seen_months = true;
+                    s = rest;
                 }
-                b'W' => {
+                [b'W', rest @ ..] => {
                     if seen_weeks {
                         return Err(DateDurationParseError::DuplicateUnit);
                     }
                     weeks =
                         u32::try_from(value).map_err(|_| DateDurationParseError::NumberOverflow)?;
                     seen_weeks = true;
+                    s = rest;
                 }
-                b'D' => {
+                [b'D', rest @ ..] => {
                     if seen_days {
                         return Err(DateDurationParseError::DuplicateUnit);
                     }
                     days = value;
                     seen_days = true;
+                    s = rest;
                 }
                 _ => return Err(DateDurationParseError::InvalidStructure),
             }
-
-            index += 1;
         }
 
         Ok(Self {


sffc (commented):
Hello again! Please add unit tests and docs tests. Cover as many edge cases as you can think of. I can't review this further until it is properly tested.

SATVIKsynopsis:
I have added unit tests covering valid iso 8601 date only durations as well as common invalid cases and also added as many edge cases as i can think of.

components/calendar/src/duration.rs
```diff
@@ -303,3 +419,107 @@ impl DateDuration {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn parse_single_unit_durations() {
+        let d = DateDuration::try_from_str("P1D").unwrap();
+        assert_eq!(d.days, 1);
+        assert!(!d.is_negative);
```
sffc: Here and elsewhere: please compare equality of the entire struct. Like this:

```suggestion
        assert_eq!(d, DateDuration {
            days: 1,
            ..Default::default()
        });
```

components/calendar/src/duration.rs
```diff
@@ -156,6 +156,15 @@ pub enum DateDurationUnit {
     /// Duration in days
     Days,
 }
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum DateDurationParseError {
```
sffc, line 163: Please document how each of these enum variants are reachable. We try to do this for all new error enums. See DateFromFieldsError for an example:

https://unicode-org.github.io/icu4x/rustdoc/icu/calendar/error/enum.DateFromFieldsError.html

Every variant has a docs test showing how that variant is reachable.

I think you might also need to export this enum. The docs tests will not compile until you export it.

If you add these docs tests, you don't need the separate unit tests for errors that you have down below. You can convert those into docs tests.

SATVIKsynopsis:
i have updated the assertions to compare the full struct and added docs tests for each `DateDurationParseError` variant documenting how each error can occur.

components/calendar/src/duration.rs
```diff
@@ -157,6 +157,93 @@ pub enum DateDurationUnit {
...
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use icu::calendar::types::DateDuration;
+    ///
+    /// assert!(DateDuration::try_from_str("P").is_err());
+    /// assert!(DateDuration::try_from_str("P1").is_err());
+    /// assert!(DateDuration::try_from_str("PX1D").is_err());
+    /// ```
```
sffc: Please assert that it is this particular error variant.

Something more like

```suggestion
    /// assert_eq!(DateDuration::try_from_str("P"), Err(DateDurationParseError::InvalidStructure));
    /// assert_eq!(DateDuration::try_from_str("P1"), Err(DateDurationParseError::InvalidStructure));
    /// assert_eq!(DateDuration::try_from_str("PX1D"), Err(DateDurationParseError::InvalidStructure));
    /// ```
```

SATVIKsynopsis:
i have updated the doc tests to assert the exact `DateDurationParseError` variant returned instead of just checking the `is_err()`

SATVIKsynopsis:
All CI checks pass now.

SATVIKsynopsis:
thanks for pointing that out and very happy to answer this.

ixdtf is a general iso 8601 parser and supports a wider set of duration forms which also include time components.

`DateDuration` is intentionally a date only type like Y/M/W/D with calendar specific behavior and error handling. using ixdtf here would still require extra validation and mapping to enforce those constraints.

for that reason, this parser is intentionally made instead of reusing the ixdtf parser.

components/calendar/src/types.rs
```diff
@@ -15,6 +15,9 @@ use zerovec::ule::AsULE;
 pub use crate::duration::{DateDuration, DateDurationUnit};
```
sffc: Nit:

```suggestion
pub use crate::duration::{DateDuration, DateDurationUnit, DateDurationParseError};
```

And remove it from below.

components/calendar/src/duration.rs
```diff
@@ -16,6 +16,8 @@
 /// # Example
 ///
 /// ```rust
+/// # #[cfg(feature="unstable")]
+/// # {
```
sffc: Issue: Why do you need this cfg? It wasn't there before and the test worked just fine. We run our tests in CI with `--all-features`.

components/calendar/src/duration.rs
```diff
@@ -157,6 +160,100 @@ pub enum DateDurationUnit {
...
+pub enum DateDurationParseError {
+    /// The input does not follow the expected ISO 8601 date only duration structure.
+    ///
+    /// This error occurs when the duration string is incomplete,
+    /// or contains unexpected characters.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # #[cfg(feature="unstable")]
```
sffc: Here and elsewhere: we run our tests with `--all-features` so you shouldn't put in these feature gates in docs tests. They are prone to getting stale.

sffc:
> `ixdtf` has an ISO 8601 duration parser available, and it is currently an optional dependency on `icu_calendar`.
> 
> Is there an intended benefit to rewriting specifically the date duration parsing vs. using the already available parser?

I see this PR primarily as a convenience for users of DateDuration, not intended to be used by Temporal. We should evaluate whether we want to switch to ixdtf duration parsing before graduating DateDuration.

SATVIKsynopsis:
i have resolved the nitpick and removed the unnecessary unstable feature gates from the doctests.

SATVIKsynopsis:
Hi. Just checking if anything else is required from my side. Thanks

components/calendar/src/duration.rs
```diff
@@ -157,6 +157,94 @@ pub enum DateDurationUnit {
     Days,
 }
 
+/// Errors that can occur when parsing an ISO 8601 date only duration string.
```
robertbastian:
```suggestion
/// Errors that can occur when parsing an ISO 8601 date-only duration string.
```

components/calendar/src/duration.rs
```diff
@@ -157,6 +157,94 @@ pub enum DateDurationUnit {
...
+    DuplicateUnit,
+
+    /// A numeric value exceeded or was more than the supported range.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use icu::calendar::types::{DateDuration, DateDurationParseError};
+    ///
+    /// assert_eq!(DateDuration::try_from_str("P9999999999999555555588888888"), Err(DateDurationParseError::NumberOverflow));
```
robertbastian: nit: this is a 94-bit number, lets use a smaller one to demonstrate this

components/calendar/src/duration.rs
```diff
@@ -157,6 +157,94 @@ pub enum DateDurationUnit {
...
+    /// Only negative durations using a leading `-` are supported.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use icu::calendar::types::{DateDuration, DateDurationParseError};
+    ///
+    /// assert_eq!(DateDuration::try_from_str("+P1D"), Err(DateDurationParseError::PositiveNotAllowed));
+    /// ```
+    PositiveNotAllowed,
```
robertbastian: the character is the plus sign. positive durations are very much allowed

```suggestion
    PlusNotAllowed,
```

components/calendar/src/duration.rs
```diff
@@ -166,6 +254,113 @@ impl DateDuration {
...
+        chars.next();
+
+        while let Some(mut ch) = chars.next() {
+            let mut value: i64 = 0;
+            let mut has_digits = false;
+
+            while ch.is_ascii_digit() {
+                has_digits = true;
+
+                let digit = ch as i64 - '0' as i64;
```
robertbastian:
```suggestion
            while let Some(digit) = ch.to_digit(10) {
                has_digits = true;
```

components/calendar/src/duration.rs
```diff
@@ -166,6 +254,113 @@ impl DateDuration {
...
+                        return Err(DateDurationParseError::DuplicateUnit);
+                    }
+                    seen_days = true;
+                    days = value;
+                }
+                _ => return Err(DateDurationParseError::InvalidStructure),
+            }
+        }
+
+        let years = u32::try_from(years).map_err(|_| DateDurationParseError::NumberOverflow)?;
```
robertbastian: suggestion: put these casts in the match statement

components/calendar/src/duration.rs
```diff
@@ -166,6 +254,113 @@ impl DateDuration {
...
+        let mut seen_years = false;
+        let mut seen_months = false;
+        let mut seen_weeks = false;
+        let mut seen_days = false;
+
+        let mut chars = s.chars();
+        chars.next();
+
+        while let Some(mut ch) = chars.next() {
+            let mut value: i64 = 0;
```
robertbastian: issue: this value can never be negative, so don't use a signed number. this will make the casts later clearer

```suggestion
            let mut value: u64 = 0;
```

components/calendar/src/duration.rs
```diff
@@ -166,6 +254,113 @@ impl DateDuration {
             ..Default::default()
         }
     }
+}
+
+impl DateDuration {
+    /// Parses an ISO 8601 date only duration string into a [`DateDuration`].
+    /// This function supports year, month, week and day components and rejects time based durations and returns an error for invalid structures or duplicate units.
+    pub fn try_from_str(s: &str) -> Result<Self, DateDurationParseError> {
```
robertbastian, line 264: we usually write our parsing functions in terms of UTF-8 bytes, so that we can accept unvalidated strings over FFI


```suggestion
    pub fn try_from_str(s: &str) -> Result<Self, DateDurationParseError> {
        Self::try_from_utf8(s.as_bytes())
    }

    pub fn try_from_utf8(code_units: &[u8]) -> Result<Self, DateDurationParseError> {
```

SATVIKsynopsis:
I moved the core parsing logic into `try_from_utf8` so it works directly on utf-8 bytes, as suggested. `try_from_str` is now just a thin wrapper calling it. Also made the nit changes as required.

components/calendar/src/duration.rs
```diff
@@ -257,101 +257,113 @@ impl DateDuration {
...
+
+    /// Parses an ISO 8601 date-only duration from UTF-8 encoded bytes.
+    ///
+    /// This function operates directly on byte slices in order to support
+    /// unvalidated UTF-8 inputs.
+    ///
+    /// Supported units are years (`Y`), months (`M`), weeks (`W`), and days (`D`).
+    /// Time based components are rejected.
+    pub fn try_from_utf8(code_units: &[u8]) -> Result<Self, DateDurationParseError> {
+        let mut index = 0;
```
robertbastian: suggestion: try not using indexing. you should be able to do the same thing you did before, where you keep track of the remaining `&[u8]` slice. with `&[u8]` you can actually pattern-match (which you can't with `&str`):

```rust
if let &[b'x', rest @ ..] = s {
    ...
    s = rest;
}

File: components/calendar/src/duration.rs [hunk 15]
```diff
@@ -257,101 +257,113 @@ impl DateDuration {
 }
 
 impl DateDuration {
-    /// Parses an ISO 8601 date only duration string into a [`DateDuration`].
-    /// This function supports year, month, week and day components and rejects time based durations and returns an error for invalid structures or duplicate units.
+    /// Parses an ISO 8601 date-only duration string into a [`DateDuration`].
```
robertbastian, line 260: nit: put the main docs on `try_from_str`, and just put "See [`Self::try_from_str`]" on `try_from_utf8`

SATVIKsynopsis:
I updated `try_from_utf8` to use slice based pattern matching instead of indexing so as to avoid panics and clippy warnings. i also moved the main documentation to `try_from_str` from `try_from_utf8`  as suggested.

SATVIKsynopsis:
Please let me know if anything else needs updating.

SATVIKsynopsis:
Ah got it. re requested the review now. Thanks

sffc (approved):
Let's get this landed. Thank you!

SATVIKsynopsis:
Thank you for the review and guidance. Really appreciate it.

PR #7358, opened 2025-12-29
Add missing convenience API for Basic_Emoji and EmojiSet

### LINKED ISSUE #7356: Add EmojiSet::for_char()
Spawned off of https://github.com/unicode-org/icu4x/issues/6049

We should have `EmojiSet::for_char`, similar to [`EnumeratedProperty::for_char()`](https://unicode-org.github.io/icu4x/rustdoc/icu_properties/props/trait.EnumeratedProperty.html#method.for_char)

Manishearth:
Fixes #7356

Adds:

 - `EmojiSet::for_char()`, `::for_str()`, which matches method naming on `BinaryProperty` and `EnumeratedProperty`
 - FFI methods `EmojiSetData::basic_emoji_for_char()`/`...for_str()`, which matches method naming on `CodePointSetData`.

These are convenience APIs based on client requests (see #6049, #6048), and do not expose new functionality.

Note: While there was previous consensus on adding such helpers APIs, we never _explicitly_ talked about emoji properties; we only talked about binary/enum props. A valid resolution here is to close this PR and the linked issue as wontfix.

### DIFF
--- a/ffi/capi/src/properties_unisets.rs
+++ b/ffi/capi/src/properties_unisets.rs
@@ -79,11 +79,7 @@ pub mod ffi {
         #[diplomat::rust_link(icu::properties::props::EmojiSet::for_str, FnInTrait)]
         #[cfg(feature = "compiled_data")]
         pub fn basic_emoji_for_str(s: &DiplomatStr) -> bool {
-            use icu_properties::props::EmojiSet;
-            let Ok(s) = core::str::from_utf8(s) else {
-                return false;
-            };
-            BasicEmoji::for_str(s)
+            icu_properties::EmojiSetData::new::<BasicEmoji>().contains_utf8(s)
         }
     }
 }


ffi/capi/src/properties_unisets.rs
```diff
@@ -65,5 +65,23 @@ pub mod ffi {
...
+
+        /// Get the `Basic_Emoji` value for a given character, using compiled data
+        #[diplomat::rust_link(icu::properties::props::EmojiSet::for_str, FnInTrait)]
+        #[cfg(feature = "compiled_data")]
+        pub fn basic_emoji_for_str(s: &DiplomatStr) -> bool {
+            use icu_properties::props::EmojiSet;
+            let Ok(s) = core::str::from_utf8(s) else {
+                return false;
+            };
+            BasicEmoji::for_str(s)
```
robertbastian: observation: this does a full UTF-8 validation, which is unnecessary for the lookup. I see that the non-convenience API *also* does this, we should improve that by adding `contains_utf8` to the `EmojiSet` type.

it might make sense to define the FFI convenience API in terms of the FFI API, not in terms of the Rust convenience API. This way, we only have one code location where we do the UTF-8 validation, and won't forget to fix this if we fix the other location

Manishearth: > it might make sense to define the FFI convenience API in terms of the FFI API

That forces an allocation, one of the reasons that FFI convenience APIs are necessary (arguably, FFI convenience APIs are more necessary than Rust ones).

But we can define the convenience API in terms of the regular Rust API instead of the convenience API, to solve your concern.

robertbastian: yes sorry that's what I meant

robertbastian: wait it's not because that requires a validated string. anyway with my pr it's not a problem anymore

PR #7361, opened 2025-12-30
Add `From<&Value>` for struct preferences

robertbastian:
This is a breaking change for callers of `struct_keyword!`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 - Components
+    - `icu_locale_core`
+        - The `struct_keyword!` macro was changed to operate on references for conversions (unicode-org#7361)
     - `icu_properties`
         - Fix some property constants (unicode-org#7269, unicode-org#7281, unicode-org#7284)
         - Add conversions for `unicode_bidi::BidiClass` (unicode-org#7272)


components/locale_core/src/preferences/extensions/unicode/macros/struct_keyword.rs
```diff
@@ -40,13 +40,29 @@ macro_rules! __struct_keyword {
...
             }
         }
 
         impl From<$name> for $crate::extensions::unicode::Value {
             fn from(input: $name) -> $crate::extensions::unicode::Value {
+                (&input).into()
+            }
+        }
+
+        impl From<&$name> for $crate::extensions::unicode::Value {
```
sffc, line 64: Not sure we need the other direction because these are always `Copy` right? Or not?

robertbastian: https://docs.rs/icu/latest/icu/locale/preferences/extensions/unicode/keywords/struct.DictionaryBreakScriptExclusions.html

Manishearth (approved):
This just adds an impl, yes? In what way is this a breaking change: is it because the client may already have an impl, or something else?

robertbastian:
> In what way is this a breaking change

The closure expected by the macro takes a reference now.

Manishearth:
Ah! I was looking from a public API perspective.

Yeah that makes sense. We should document this in the changelog.

But I _think_ this breaks agreed upon versioning that older versions of icu_component should be able to depend on newer versions of icu_provider, yes?

Can we make the macro optionally take a reference so it still supports older code?

robertbastian:
All our call sites of `struct_keyword!` are in-crate, but it is public.

Manishearth:
Ah! That's fine then.

zbraniecki:
What's the rationale for this change?

robertbastian:
removing the clone during the locale-preference conversion

sffc:
@zbraniecki, you can see the impact of the change in this PR in components/locale_core/src/preferences/locale.rs

We implement `impl From<&Locale> for LocalePreferences`. Do you think that function should have a different signature, perhaps one that reads already-extracted Values?

I do think this change is positive, though I'm not super happy that it breaks semver. It shouldn't break any ICU4X crates, though, since we only use this macro within icu_locale_core.

PR #7363, opened 2025-12-30
Add `CodePointInversionListAndStringList::contains_utf8`

robertbastian:
For #7358

### DIFF
--- a/components/collections/src/codepointinvliststringlist/mod.rs
+++ b/components/collections/src/codepointinvliststringlist/mod.rs
@@ -173,6 +173,20 @@ impl<'data> CodePointInversionListAndStringList<'data> {
         self.str_list.binary_search(s).is_ok()
     }
 
+    /// See [`Self::contains_str`]
+    pub fn contains_utf8(&self, s: &[u8]) -> bool {
+        use utf8_iter::Utf8CharsEx;
+        let mut chars = s.chars();
+        if let Some(first_char) = chars.next() {
+            if chars.next().is_none() {
+                return self.contains(first_char);
+            }
+        }
+        self.str_list
+            .binary_search_by(|t| t.as_bytes().cmp(s))
+            .is_ok()
+    }
+
     ///
     /// # Examples
     /// ```
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1346,6 +1346,7 @@ dependencies = [
  "serde",
  "serde_json",
  "toml",
+ "utf8_iter",
  "yoke",
  "zerofrom",
  "zerovec",
--- a/components/collections/Cargo.toml
+++ b/components/collections/Cargo.toml
@@ -25,6 +25,7 @@ yoke = { workspace = true, features = ["derive"] }
 zerofrom = { workspace = true, features = ["derive"] }
 zerovec = { workspace = true, features = ["derive", "yoke"] }
 potential_utf = { workspace = true, features = ["zerovec"] }
+utf8_iter = { workspace = true }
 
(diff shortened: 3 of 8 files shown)


components/collections/src/codepointinvliststringlist/mod.rs
```diff
@@ -173,6 +173,20 @@ impl<'data> CodePointInversionListAndStringList<'data> {
         self.str_list.binary_search(s).is_ok()
     }
 
+    /// See [`Self::contains_str`]
```
Manishearth, line 176: nit: docs should say this is like contains_str but does not require validation, and ideally say that invalid UTF8 will always return false.

Manishearth (approved):
Naming matches what we do elsewhere.

cc @sffc, new API

components/collections/src/codepointinvliststringlist/mod.rs
sffc: The names of the functions on this type were carefully designed with regard to input types and return types since Rust doesn't have function overloading. Would be good for @echeran to review.

robertbastian: this follows our naming convention as decided in https://github.com/unicode-org/icu4x/issues/4931

sffc: It is also consistent with the icu_collections-specific conventions decided in https://github.com/unicode-org/icu4x/issues/2413 ?

robertbastian: I don't think #2413 discussed strings anywhere

sffc: It flows naturally for `contains`, but what if we extend this to `get` (for example, on CodePointTrie)

We have:

- `get(char) -> T`
- `get32(u32) -> T`
- `get32_u32(u32) -> u32`

We don't currently have `get_u32(char) -> u32` but that's what the logical signature would be.

So, what's the signature of `get_str` or `get_utf8`?

I'm okay if we say it will be

- `get_str(str) -> T`
- `get_utf8([u8]) -> T`
- `get_str_u32(str) -> u32`
- `get_utf8_u32([u8]) -> u32`

But, a potential alternative design would be something like

- `getstr(str) -> T`
- `getutf8([u8]) -> T`
- `getstr_u32(str) -> u32`
- `getutf8_u32([u8]) -> u32`

Even if we were to pick that second set of function names, it's not immediately clear if that design would translate over to the function names of the `contains` functions.

---

All this said, we already have `contains_str`, so adding `contains_utf8` is not a novel naming pattern in that particular type.

robertbastian: `CodePointInversionListAndStringList` does not have `get` functions because it's a set, not a map. There is no map collection that is indexed by strings.

sffc (approved):
Approving because we already have `contains_str`, so `contains_utf8` is parallel.

hsivonen:
This created a circular dependency with #6925. :frowning_face:

hsivonen:
This created rather odd semantics that if the unvalidated UTF-8 argument contains a single encoding error, it matches U+FFFD, but if the argument contains multiple encoding errors, it won't match strings of U+FFFD.

Perhaps it would make sense to validate the argument for UTF-8, return false if the validation fails, and delegate to `contains_str` if the validation succeeds.

robertbastian:
I don't quite follow, can you be more specific? Which inputs produce unexpected results?

Issue #5193, opened 2024-07-08
Add security policy


robertbastian:
Tracking issue for:
- [ ] https://github.com/unicode-org/icu4x/security/code-scanning/92

Manishearth:
This needs us to determine a set of people who can receive security bugs and set up bug reporting either through GitHub or elsewhere.

PR #7365, opened 2025-12-31
Add SECURITY.md

Manishearth:
Fixes https://github.com/unicode-org/icu4x/issues/5193

This is modeled after https://github.com/unicode-org/icu, except we can use GitHub for reporting instead of. "Maintainers" on GitHub can see these issues. We could also choose to include icu4x-core as an alternate reporting route.

I arbitrarily picked a sensible policy for past versions, we can iterate on this.

Note that the "security vulnerability report form" looks different for people who are maintainers vs people who are not, so ignore the fact that that link says "Open a draft security advisory", non-maintainers should see something like https://github.com/mozilla/fxa/security/advisories/new.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,7 +2,7 @@
 
 ## Reporting a Vulnerability
 
-If you have discovered a security vulnerability in this project, please report it privately. **Do not disclose it as a public issue.**
+If you have discovered a security vulnerability in ICU4X crates or any utility crates maintained in this repository, please report it privately. **Do not disclose it as a public issue.**
 This gives us time to work with you to fix the issue before public exposure, reducing the chance that the exploit will be used before a patch is released.
 
 Please submit the bug report using the [security vulnerability report form](https://github.com/unicode-org/icu4x/security/advisories/new) on GitHub.
@@ -11,8 +11,9 @@ Please provide the following information in your report:
 
  * A description of the vulnerability and its impact
  * How to reproduce the issue
+ * The versions you know are impacted (this list need not be exhaustive)
 
-This project is maintained by volunteers on a reasonable-effort basis. As such, we ask that you give us 90 days to work on a fix before public exposure.
+We ask that you give us 90 days to work on a fix before public exposure.
 
 
 ## Supported Versions


Manishearth:
Note that https://github.com/orgs/unicode-org/teams/icu4x-sc is everyone who has "maintainer" access to security issues. If we want a smaller group, we should make a separate mailing list or something.

SECURITY.md
```diff
@@ -0,0 +1,24 @@
+# Security Policy
+
+## Reporting a Vulnerability
```
sffc, line 3: Suggestion: Should we call out that this covers the utils projects, too?

SECURITY.md
```diff
@@ -0,0 +1,24 @@
...
+
+Please provide the following information in your report:
+
+ * A description of the vulnerability and its impact
+ * How to reproduce the issue
+
+This project is maintained by volunteers on a reasonable-effort basis. As such, we ask that you give us 90 days to work on a fix before public exposure.
+
+
+## Supported Versions
```
sffc, line 19: Suggestion: I'm not sure this section adds a lot of value. I think I'd rather just add a bullet point "impacted versions" in the list of information to include in the report. If you feel the need, say something like "We will review reports impacting any version and will prioritize those impacting recently published versions."

SECURITY.md
```diff
@@ -0,0 +1,24 @@
...
+This gives us time to work with you to fix the issue before public exposure, reducing the chance that the exploit will be used before a patch is released.
+
+Please submit the bug report using the [security vulnerability report form](https://github.com/unicode-org/icu4x/security/advisories/new) on GitHub.
+
+Please provide the following information in your report:
+
+ * A description of the vulnerability and its impact
+ * How to reproduce the issue
+
+This project is maintained by volunteers on a reasonable-effort basis. As such, we ask that you give us 90 days to work on a fix before public exposure.
```
robertbastian: is it though? it's maintained by the Unicode Consortium, which is backed by very resourceful tech companies. if we want to fix something quickly we can.

SECURITY.md
robertbastian: I like this section

SECURITY.md
Manishearth: I copied this from the ICU4C reporting instructions. I'll tweak.

SECURITY.md
Manishearth: It's pretty standard to include _some_ idea of which versions are "security patch supported" in security policies; and GitHub's default template encourages this.

Manishearth: I think this section accurately sets expectations: we want reports for all versions, we may fix them for some, we have a near-guarantee for the most recent version.

sffc (approved):

Issue #6549, opened 2025-05-07
Consider adding TimeZone constructor from IANA


sffc:
Currently if you want to create a TimeZone from an IANA ID, you need to know to create an `IanaParser`.

It would be more discoverable, especially on FFI, if TimeZone itself had a function with the following signature

```rust
impl TimeZone {
    pub fn try_from_iana_str(iana_str: &str, parser: IanaParserBorrowed) -> Self { ... }
}
```

If we add this constructor, we should make it the default constructor for the WASM Demo.

robertbastian:
SG, however it's not fallible so it should just be `from_iana_str`. We'd also need `from_iana_utf8`.

Cdog1526:
https://github.com/unicode-org/icu4x/pull/7368
Made a mr for this! 

However, I'm running into an issue while testing. This seems to be related to a file that I have not touched at all, but as a result I can't test my work. I'm running the following command for testing, is there something incorrect that I'm doing or is this just an existing issue?
 cargo test --package icu_time --lib -- zone::test_from_iana_utf8 --exact --nocapture --all-features

   Compiling icu v2.1.1 (C:\Users\bucke\OneDrive\Documents\Projects\icu4x\components\icu)
error[E0599]: no function or associated item named `try_new_with_buffer_provider` found for struct `offset::VariantOffsetsCalculator` in the current scope
   --> components\time\src\zone\offset.rs:400:39
    |
194 | pub struct VariantOffsetsCalculator {
    | ----------------------------------- function or associated item `try_new_with_buffer_provider` not found for this struct
...
400 |     let c = VariantOffsetsCalculator::try_new_with_buffer_provider(
    |                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function or associated item not found in `offset::VariantOffsetsCalculator`

sffc:
I think ` --all-features` is in the wrong place. It should be after `cargo test` and before `--`

PR #7368, opened 2026-01-02
Added constructors for timezone from iana string and utf

Cdog1526:
Added constructors for TimeZone from iana str and iana utf8 as per #6549


sffc:
> I'm not convinced of the design. We might want to have `from_iana_str` be `#[cfg(feature = "compiled_data")]` and not require an `IanaParser`.

I would support that design

robertbastian:
Please add the API behind the `compiled_data` feature, and also add it to FFI (`ffi/capi/src/timezone.rs`).

PR #7371, opened 2026-01-05
Fix normalize_utf8 example in C API documentation

Kayd-06:
### Summary:

The `normalize_utf8` example in the C API documentation was incorrect and duplicated
the `normalize` example.

This PR updates the example to correctly demonstrate `normalize_utf8` usage.

### Changes:

Fixed the `normalize_utf8` example in C API documentation.

### DIFF
--- a/ffi/capi/src/locale_core.rs
+++ b/ffi/capi/src/locale_core.rs
@@ -150,7 +150,31 @@ pub mod ffi {
         }
 
         /// Normalizes a locale string.
+        ///
+        /// # Examples
+        /// ```c
+        /// DiplomatWrite write;
+        /// char buffer[32];
+        /// diplomat_write_init(&write, buffer, sizeof(buffer));
+        ///
+        /// icu4x_locale_normalize("en-us", &write);
+        ///
+        /// // buffer now contains "en-US"
+        /// ```
         #[diplomat::rust_link(icu::locale::Locale::normalize, FnInStruct)]
+
+        /// Normalizes a UTF-8 locale string.
+        ///
+        /// # Examples
+        /// ```c
+        /// DiplomatWrite write;
+        /// char buffer[32];
+        /// diplomat_write_init(&write, buffer, sizeof(buffer));
+        ///
+        /// icu4x_locale_normalize_utf8("en-us", &write);
+        ///
+        /// // buffer now contains "en-US"
+        /// ```
         #[diplomat::rust_link(icu::locale::Locale::normalize_utf8, FnInStruct, hidden)]
         pub fn normalize(
             s: &DiplomatStr,


ffi/capi/src/locale_core.rs
```diff
@@ -150,7 +150,31 @@ pub mod ffi {
         }
 
         /// Normalizes a locale string.
+        ///
+        /// # Examples
+        /// ```c
+        /// DiplomatWrite write;
```
robertbastian, line 156: issue: these docs are used for all FFI-languages (C, CPP, Dart, JS/TS, Kotlin), so it's not appropriate to add C examples here

Kayd-06: Thanks for pointing this out, that makes sense.

I missed that this documentation is shared across all FFI languages.
I’ll remove the C-specific example from here.

Could you please suggest the appropriate place for C-only examples,
or should this example live elsewhere (e.g. C API specific docs)?

Manishearth: You filed this PR because "the normalize_utf8 C API documentation is incorrect".

Which documentation are you talking about, there? Should this not attempt to fix that?

Our C examples are in tutorials/c. But I'm not sure we need an example for this particular function.

Kayd-06: Thanks for the clarification.

The issue I was trying to address is that the existing example for
'normalize_utf8' was effectively duplicating the 'normalize' example,
which made it confusing from a C API perspective.

Given that these docs are shared across all FFI languages, I agree that
adding a C-specific example here is not appropriate.

Based on your comment, the best fix for this PR would be to simply remove
the example from this location and avoid duplicating behavior already
covered elsewhere.

I’m happy to update the PR accordingly, unless you’d prefer a different
approach.

Manishearth: Again, which example are you talking about? There is currently no example in C API documentation here.

Manishearth: Again, which example are you talking about? There is currently no example in C API documentation here.

Kayd-06: Thanks for the clarification, that’s on me.

You’re right, this is not C API documentation specifically.
What I meant to refer to was the example block in the shared FFI docs
for `normalize_utf8` in `ffi/capi/src/locale_core.rs`.

My concern is that the current example content is confusing because it
duplicates the behaviour shown for 'normalize', without clearly
demonstrating what is specific to 'normalize_utf8'.

Given that these docs are shared across all FFI languages, I agree that
adding language-specific examples here is not appropriate.

The simplest fix would be to remove this example block entirely to avoid
confusion, unless you think there is value in keeping a language-agnostic
example here.

Happy to update the PR accordingly.

PR #7376, opened 2026-01-06
Respect `-u-rg` in calendar resolution

robertbastian:
#4413 

Should respect `-u-rg` per https://github.com/unicode-org/cldr/blob/main/common/supplemental/rgScope.xml

### DIFF
--- a/components/calendar/src/any_calendar.rs
+++ b/components/calendar/src/any_calendar.rs
@@ -711,6 +711,18 @@ pub enum AnyCalendarKind {
 
 impl CalendarPreferences {
     /// Selects the [`CalendarAlgorithm`] appropriate for the given [`CalendarPreferences`].
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use icu::calendar::preferences::{CalendarAlgorithm, CalendarPreferences};
+    /// use icu::locale::locale;
+    ///
+    /// assert_eq!(CalendarPreferences::from(&locale!("und")).resolved_algorithm(), CalendarAlgorithm::Gregory);
+    /// assert_eq!(CalendarPreferences::from(&locale!("und-US-u-ca-hebrew")).resolved_algorithm(), CalendarAlgorithm::Hebrew);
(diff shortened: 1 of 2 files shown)


Manishearth (approved):
Note: this changes the behavior of `LocalePreferences::region()`. LocalePreferences' API is mostly for internal use and in the context of preferences, so I think this makes sense. Robert noted that this PR could be implemented in other ways too if requested. I think this is fine for now and we can move things around if we want.

Manishearth:
An additional thing that was noted was that _typically_ LocalePreferences goes throguh `to_data_locale_region_priority`/`to_data_locale_language_priority`, calendar resolution does not. These codepaths work off of raw values (as they should, since they require a different treatment of region settings).

We can in the future add `region_tag` vs `region_override` or something

sffc (commented):
Seems good, would be good to get a third opinion from @zbraniecki on the behavior change

components/calendar/src/any_calendar.rs
```diff
@@ -711,6 +711,18 @@ pub enum AnyCalendarKind {
...
+    /// # Example
+    ///
+    /// ```
+    /// use icu::calendar::preferences::{CalendarAlgorithm, CalendarPreferences};
+    /// use icu::locale::locale;
+    ///
+    /// assert_eq!(CalendarPreferences::from(&locale!("und")).resolved_algorithm(), CalendarAlgorithm::Gregory);
+    /// assert_eq!(CalendarPreferences::from(&locale!("und-US-u-ca-hebrew")).resolved_algorithm(), CalendarAlgorithm::Hebrew);
+    /// assert_eq!(CalendarPreferences::from(&locale!("und-AF")).resolved_algorithm(), CalendarAlgorithm::Persian);
+    /// assert_eq!(CalendarPreferences::from(&locale!("und-US-u-rg-thxxxx")).resolved_algorithm(), CalendarAlgorithm::Buddhist);
```
sffc, line 724: Suggestion: add a test with both `-u-ca` and `-u-rg` (as you suggested I do with `-u-fw` and `-u-rg` in WeekInformation)

Issue #7362, opened 2025-12-30
Should invalid region and subdivision extension values cause an error?


sffc:
There are a number of ways they can be invalid: syntax, mismatched region, invalid subdivision for region, etc. Perhaps the most common examples will be things like `en-US-u-sd-tx` or `en-US-u-rg-DE` which are both incorrect.

Currently, we simply drop these subtags if they are invalid. Should we do something else instead? If so, what?

- Most functions that consume these subtags, via Preferences, return only a DataError. Should we expose this via DataError somehow?
- Should we print a warning to the log if the log is enabled?
- Should we make the conversion from Locale to Preferences be fallible? (in 3.0 with possible 2.x migration path)

A valid position to hold is that so few people manually write Unicode extension keywords that we should assume they are well-formed and just close this issue. However, it seems from previous discussions that we care deeply about informing users when they might be doing something wrong that impacts i18n correctness.

CC @zbraniecki @robertbastian

vamsi5410:
can i work on this

zbraniecki:
What's the value of rejecting such tags over accepting them and having a global or per-domain order of data resolution (for example "always take rg first, then region")?

vamsi5410:
@zbraniecki I think the main value of rejecting these tags is that it avoids guessing what the user meant. If we accept them and apply a fixed rule like “always take rg first, then region”, we might silently pick the wrong value without the user realizing there’s a problem.

Rejecting them makes the ambiguity visible and encourages fixing the input instead of hiding the mistake behind implicit precedence rules. That feels safer and easier to reason about in the long run, especially for i18n, where small mistakes can lead to subtle but incorrect behavior.

zbraniecki:
> I think the main value of rejecting these tags is that it avoids guessing what the user meant.

How do we know what user meant?

> If we accept them and apply a fixed rule like “always take rg first, then region”, we might silently pick the wrong value without the user realizing there’s a problem.

Correct, which can be seen as a better outcome than an alternative.

> Rejecting them makes the ambiguity visible and encourages fixing the input instead of hiding the mistake behind implicit precedence rules.

How does it make it visible? What is the UX of a user that interacts with a computer system that produced such locale identifiers, passed it to an ICU4X and ICU4X rejected it?

> That feels safer and easier to reason about in the long run, especially for i18n, where small mistakes can lead to subtle but incorrect behavior.

I agree. My concern is that:
a) This type of error is not actionable from the user perspective
[...]

PR #7377, opened 2026-01-06
Add `LocalePreferences::from_locale_strict`

robertbastian:
We already have this on preferences generate by the macro.

#6628 

#7362

### DIFF
--- a/components/locale_core/src/preferences/locale.rs
+++ b/components/locale_core/src/preferences/locale.rs
@@ -60,58 +60,7 @@ impl Default for LocalePreferences {
 
 impl From<&crate::Locale> for LocalePreferences {
     fn from(loc: &crate::Locale) -> Self {
-        let subdivision = if let Some(sd) = loc
-            .extensions
-            .unicode
-            .keywords
-            .get(&RegionalSubdivision::UNICODE_EXTENSION_KEY)
-        {
-            if let Ok(sd) = RegionalSubdivision::try_from(sd) {
-                Some(sd)
-            } else {
-                None
-            }
-        } else {
-            None
-        };
-
-        let region = if let Some(sd) = subdivision {
-            if let Some(region) = loc.id.region {
-                // Discard the subdivison if it doesn't match the region
-                Some(RegionalSubdivision(SubdivisionId {
-                    region,
-                    suffix: if sd.region == region {
-                        sd.suffix
-                    } else {
-                        SubdivisionSuffix::UNKNOWN
-                    },
-                }))
-            } else {
-                // Use the subdivision's region if there's no region
-                Some(sd)
-            }
-        } else {
-            loc.id.region.map(|region| {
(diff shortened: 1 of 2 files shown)


components/locale_core/src/preferences/locale.rs
```diff
@@ -118,15 +140,15 @@ impl From<LocalePreferences> for crate::Locale {
...
+                region: prefs.region.map(|sd| sd.region),
                 variants: prefs
                     .variant
                     .map(Variants::from_variant)
                     .unwrap_or_default(),
             },
             extensions: {
                 let mut extensions = crate::extensions::Extensions::default();
-                if let Some(sd) = prefs.subdivision {
+                if let Some(sd) = prefs.region.filter(|sd| !sd.suffix.is_unknown()) {
```
sffc, line 100: Observation: this means `en-US-u-sd-uszzzz` will roundtrip from Locale->LP->Locale as `en-US` which is fine and good.

components/locale_core/src/preferences/locale.rs
```diff
@@ -182,15 +205,15 @@ impl LocalePreferences {
         if let Some(script) = other.script {
             self.script = Some(script);
         }
-        if let Some(region) = other.region {
-            self.region = Some(region);
+        if let Some(sd) = other.region {
+            // Use the other region if it's different, or if it has a subdivision
+            if !sd.suffix.is_unknown() || Some(sd.region) != self.region.map(|sd| sd.region) {
```
sffc, line 225: Thought/Suggestion: I think this is the same as

```rust
if !sd.suffix.is_unknown() || self.region.map(|r| r.suffix.is_unknown()).unwrap_or(true)
```

Not sure which is clearer.

components/locale_core/src/preferences/locale.rs
```diff
@@ -284,10 +299,45 @@ mod tests {
...
+                // values though.
+                input: "!en-001-u-sd-001",
+                language_priority: "en-001",
+                region_priority: "en-001",
+            },
         ];
         for test_case in test_cases.iter() {
-            let locale = Locale::try_from_str(test_case.input).unwrap();
-            let prefs = LocalePreferences::from(&locale);
+            let prefs = if let Some(locale) = test_case.input.strip_prefix("!") {
```
sffc, line 332: Praise: I like how you wrote this test

components/locale_core/src/preferences/locale.rs
Manishearth: Curious: do we have use cases for people turning LPs into locales? Mostly for debugging?

(So I can better review things that affect this model in the future)

sffc: Serializing constructor options perhaps? Maybe for ECMA-402? But yes I agree that LP->Locale is not a hot path

PR #7380, opened 2026-01-06
Rename subdivision types

robertbastian:
The LDML `unicode_subdivision_id` expression contains both a region and a subdivision, yet it's just called `SubdivisionId` here. We use it as a region+subdivision type in multiple places, so a more accurate name would be `RegionAndSubdivision`. The current `SubdivisionSuffix`, which is the actual subdivision, can then be `Subdivision`.

### DIFF
--- a/components/locale_core/src/data.rs
+++ b/components/locale_core/src/data.rs
@@ -220,12 +220,12 @@ impl DataLocale {
         Ok(())
     }
 
-    fn region_and_subdivision(&self) -> Option<unicode_ext::SubdivisionId> {
+    fn region_and_subdivision(&self) -> Option<unicode_ext::RegionAndSubdivision> {
         self.subdivision
-            .and_then(|s| unicode_ext::SubdivisionId::try_from_str(s.as_str()).ok())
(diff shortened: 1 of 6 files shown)


robertbastian:
We don't generally follow LDML naming, or align with ICU.

sffc:
> We don't generally follow LDML naming, or align with ICU.

But, we do?

- Semantic skeletons are in LDML (we added them there) and the names of things were influenced by ICU/CLDR feedback
- We use LDML names for all the Unicode extension keywords and basically everything in the icu_locale crate
- You have cited LDML before when trying to name things like `related_iso_year` and `Weekday`

robertbastian:
> * You have cited LDML before when trying to name things like `related_iso_year` and `Weekday`

And yet we still call it `related_iso` because you prefer that.

sffc:
> > * You have cited LDML before when trying to name things like `related_iso_year` and `Weekday`
> 
> And yet we still call it `related_iso` because you prefer that.

We call it `related_iso` because we are already shipping it, and because Temporal is another spec we use as precedent.

PR #7384, opened 2026-01-07
Add buffer ctors for properties APIs

Manishearth:
Fixes the issue found in https://github.com/unicode-org/icu4x/discussions/7374

This was an oversight when we did new properties.

## Changelog

icu_properties: Add buffer provider constructors for all property APIs
 - New methods: `CodePointMapData::try_new_with_buffer_provider()`, `CodePointSetData::try_new_with_buffer_provider()`, `EmojiSetData::try_new_with_buffer_provider()`

### DIFF
--- a/components/properties/src/code_point_map.rs
+++ b/components/properties/src/code_point_map.rs
@@ -40,6 +40,18 @@ impl<T: TrieValue> CodePointMapData<T> {
         CodePointMapDataBorrowed::new()
     }
 
+    #[cfg(feature = "serde")]
+    #[doc = icu_provider::gen_buffer_unstable_docs!(BUFFER, Self::new)]
+    pub fn try_new_with_buffer_provider(
+        provider: &(impl icu_provider::buf::BufferProvider + ?Sized),
+    ) -> Result<Self, DataError>
+    where
+        T: EnumeratedProperty + for<'a> serde::Deserialize<'a>,
+    {
+        use icu_provider::buf::AsDeserializingBufferProvider;
+        Self::try_new_unstable(&provider.as_deserializing())
+    }
+
     #[doc = icu_provider::gen_buffer_unstable_docs!(UNSTABLE, Self::new)]
     pub fn try_new_unstable(
         provider: &(impl DataProvider<T::DataMarker> + ?Sized),
--- a/components/properties/src/code_point_set.rs
+++ b/components/properties/src/code_point_set.rs
@@ -40,6 +40,15 @@ impl CodePointSetData {
         CodePointSetDataBorrowed::new::<P>()
     }
 
+    #[cfg(feature = "serde")]
+    #[doc = icu_provider::gen_buffer_unstable_docs!(BUFFER, Self::new)]
+    pub fn try_new_with_buffer_provider<P: BinaryProperty>(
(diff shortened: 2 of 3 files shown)


components/properties/src/code_point_map.rs
```diff
@@ -40,6 +40,18 @@ impl<T: TrieValue> CodePointMapData<T> {
         CodePointMapDataBorrowed::new()
     }
 
+    #[cfg(feature = "serde")]
+    #[doc = icu_provider::gen_buffer_unstable_docs!(BUFFER, Self::new)]
+    pub fn try_new_with_buffer_provider(
+        provider: &(impl icu_provider::buf::BufferProvider + ?Sized),
+    ) -> Result<Self, DataError>
+    where
+        T: EnumeratedProperty + for<'a> serde::Deserialize<'a>,
```
Manishearth, line 49: the Deserialize impl shouldn't be necessary.

Manishearth: Never mind; CPTs use T's serde impl in human readable mode.

sffc: Isn't there a macro that generates the whole impl for you?

Manishearth: We have generics here.

That's probably how we messed up in the first place, because we couldn't use the macro.

robertbastian (commented):
this was not an oversight, it was intentional. properties data markers are more stable than other crates, because each data struct is an enum that we can evolve by adding new variants. this means that we will never need fallback code in a buffer constructor, so we don't need buffer constructors.

Manishearth:
I don't see the relationship between buffer providers and data struct format fallback. Data struct format fallback is something we _can_ do with buffer providers (and don't need to do with baked), but that's a property of custom providers in general, and it's not their raison d'être.

There are at least two use cases for buffer providers in properties. Firstly, a low-memory application may choose to load properties on demand. I think this is a pretty marginal use case since the deserialization code has its own cost, but ... possible.

The other one is tailoring/private use: the OP in https://github.com/unicode-org/icu4x/discussions/7374 cared about that use case in particular.

We should have a stable ctor that allows one to load properties off of a buffer provider. Currently you need to use `try_new_unstable`.

If we are 100% sure that the marker will never change then we could design `try_new_from_provider` instead (which has `try_new_unstable`'s bounds) but this would be a deviation from how we do data elsewhere and would need to be appropriately documented with examples for use with a buffer provider (since it's nontrivial to figure out you need to use AsDeserializing).

robertbastian:
In any case this is a new API so let's discuss this in the TC.

Manishearth:
Fair enough.

sffc:
I favor adding `with_buffer_provider` here because of consistency and discoverability. It does not _hurt_ to have the constructor, even if you subscribe to the line of reasoning presented by @robertbastian

robertbastian:
> If we are 100% sure that the marker will never change then we could design `try_new_from_provider` instead (which has `try_new_unstable`'s bounds) but this would be a deviation from how we do data elsewhere and would need to be appropriately documented with examples for use with a buffer provider (since it's nontrivial to figure out you need to use AsDeserializing).

This is a good observation. The unstable constructors for CPT and CPM are actually stable and we could rename them.

PR #7386, opened 2026-01-07
Clean up neo datagen

Manishearth:
I found the code organization to be pretty straightforward, I mostly just renamed it into names.rs and skeletons.rs.

There is some glue pattern code in names.rs, which was proving a bit more involved to move around.

There are no changes here other than code being moved around and reformatted (with duplicate imports being cleaned up), except in the last  commit which renames some internal APIs.

Progress on https://github.com/unicode-org/icu4x/issues/5613

That issue is about datagen and I think this is more or less sufficient for datagen, but it would be nice to also clean up the datetime provider code. Might try that as well.


provider/source/src/datetime/names.rs
```diff
@@ -74,7 +74,7 @@ const GLUE_PATTERN_KEY_LENGTHS: &[&DataMarkerAttributes] = &[
 ];
 
 impl SourceDataProvider {
-    fn load_neo_key<M: DataMarker>(
+    fn load_datetime_key<M: DataMarker>(
```
robertbastian: also replace `key` by `marker`

provider/source/src/datetime/semantic_skeletons.rs
```diff
@@ -2,14 +2,27 @@
 // called LICENSE at the top level of the ICU4X source tree
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
```
sffc, line 4: Suggestion: rename this file to `available_formats.rs`, and rename `neo_skeleton.rs` to something like `semantic_skeletons.rs`. Then the file names correspond to LDML section headers.

provider/source/marker
robertbastian, line 1: ?

PR #7387, opened 2026-01-08
Don't hallucinate patterns

robertbastian:
If there's no pattern for a plural category, it's not correct to fill in an empty string. Instead, it should fall back to the `other` value.

### DIFF
--- a/components/experimental/src/compactdecimal/formatter.rs
+++ b/components/experimental/src/compactdecimal/formatter.rs
@@ -755,4 +755,13 @@ mod tests {
             assert_writeable_eq!(result10T, case.expected10T, "{:?}", case);
         }
     }
+
+    #[test]
+    fn regression_7387() {
+        let formatter =
+            CompactDecimalFormatter::try_new_short(locale!("ar").into(), Default::default())
+                .unwrap();
+
+        assert_writeable_eq!(formatter.format_i64(3_000_000), "3\u{a0}مليون");
+    }
 }


eggrobin:
Well, this is exciting. ICU4X currently formats 1000000 in Filipino compact-short as 1. It might be worth adding a test for that. (EDIT: I see you added the Arabic counterpart where 3000000 came out as 3, I suppose that’s good enough.)

Apparently from the comment I was thinking that missing data for an exponent+plural case in CLDR (really CLDR-JSON) meant that non-compact formatting should be used. Even if that were the case, 1 would be wrong.

But those gaps in CLDR-JSON really don’t mean that at all. Instead, they are a result of partial resolution of inheritance at multiple levels in CLDR.

Italian
```
					<pattern type="1000000" count="one">↑↑↑</pattern>
					<pattern type="1000000" count="other">0 Mln</pattern>
```
becomes
```
					<pattern type="1000000" count="one">0 Mln</pattern>
					<pattern type="1000000" count="other">0 Mln</pattern>
```
and then
```
              "1000000-count-one": "0 Mln",
              "1000000-count-other": "0 Mln",
```
so Italian compact formatting is dense and works.

But Filipino
```
					<pattern type="1000000" count="one">↑↑↑</pattern>
					<pattern type="1000000" count="other">↑↑↑</pattern>
					<pattern type="10000000" count="one">↑↑↑</pattern>
					<pattern type="10000000" count="other">↑↑↑</pattern>
					<pattern type="100000000" count="one">↑↑↑</pattern>
					<pattern type="100000000" count="other">↑↑↑</pattern>
					<pattern type="1000000000" count="one">↑↑↑</pattern>
					<pattern type="1000000000" count="other">0B</pattern>
					<pattern type="10000000000" count="one">↑↑↑</pattern>
```
becomes
```
			<decimalFormatLength type="short">
				<decimalFormat>
					<pattern type="1000000000" count="one">0B</pattern>
					<pattern type="1000000000" count="other">0B</pattern>
```
(with the million patterns inherited from root omitted altogether), and in CLDR-JSON the patterns from root are supplied but not replicated into all relevant plural cases, thus
```
              "1000000-count-other": "0M",
              "10000000-count-other": "00M",
              "100000000-count-other": "000M",
              "1000000000-count-one": "0B",
              "1000000000-count-other": "0B",
```
So the data from CLDR-JSON requires plural case fallback in these weird edge cases: the change is correct.

Maybe CLDR-JSON should provide dense output, this clearly confused me.

robertbastian:
Filed https://unicode-org.atlassian.net/browse/CLDR-19193 for this.

srl295:
> the patterns from root are supplied but not replicated into all relevant plural cases

No. Root does not have those cases.

robertbastian:
> > the patterns from root are supplied but not replicated into all relevant plural cases
> 
> No. Root does not have those cases.

You can interpret it that way. It's also valid to interpret it the way Robin did.

eggrobin:
The survey tool is quite helpful here (in fact I used it to figure out what was meant to happen when reviewing this earlier), as it explains the three different paths.

> [Value: 0M](https://st.unicode.org/cldr-apps/v#/fil/Compact_Decimal_Formatting/1139c2015e8c8227)
This item is inherited from another field in the Root locale.

(This is neither in the `production` XML files nor in CLDR-JSON.)

> [Value: 0M](https://st.unicode.org/cldr-apps/v#/fil/Compact_Decimal_Formatting/76d709105676a7e5)
This item is inherited from the root locale.

(This does not show up in the `production` XML files, but shows up in CLDR-JSON.)

> [Value: 0B](https://st.unicode.org/cldr-apps/v#/fil/Compact_Decimal_Formatting/5adf16db99e818e)
This item is inherited from another field in this locale.

(This shows up in the `production` XML files as well as CLDR-JSON.)

srl295:
> The survey tool is quite helpful here (in fact I used it to figure out what was meant to happen when reviewing this earlier), as it explains the three different paths.
> 

In fact you can click the down arrow next to the words "info panel" and it will explain in more detail with explicit paths.

PR #7388, opened 2026-01-08
Pack compact decimal data

robertbastian:
Using `SinglePlaceholderPattern` and `PluralElements` reduces data size and stack size.


components/experimental/src/compactdecimal/provider.rs
```diff
@@ -37,117 +38,68 @@ icu_provider::data_marker!(
...
-    pub patterns: ZeroMap2d<'data, i8, Count, PatternULE>,
+    pub patterns:
+        VarZeroVec<'data, VarTupleULE<u8, PluralElementsPackedULE<SinglePlaceholderPattern>>>,
 }
 
-icu_provider::data_struct!(CompactDecimalPatternData<'_>, #[cfg(feature = "datagen")]);
+impl CompactDecimalPatternData<'_> {
+    /// The pattern `0`, which is used for low magnitudes and omitted from the data struct.
+    pub const PLURAL_PATTERN_0: &'static PluralElementsPackedULE<SinglePlaceholderPattern> =
+        unsafe { PluralElementsPackedULE::from_bytes_unchecked(&[0, 1]) };
```
sffc, line 75: Nit: If you need this unsafe code, please put it in the same file as PluralElementsPackedULE. You can call it `PlaceholderNoAffixes` or something.

components/experimental/src/compactdecimal/provider.rs
```diff
@@ -37,117 +38,68 @@ icu_provider::data_marker!(
...
 #[cfg_attr(feature = "datagen", derive(serde::Serialize, databake::Bake))]
 #[cfg_attr(feature = "datagen", databake(path = icu_experimental::compactdecimal::provider))]
 #[yoke(prove_covariance_manually)]
 pub struct CompactDecimalPatternData<'data> {
-    /// A map keyed on log10 of the CLDR `type` attribute and the CLDR `count` attribute.
+    /// A map keyed on log10 of the CLDR `type` attribute.
     #[cfg_attr(feature = "serde", serde(borrow))]
-    pub patterns: ZeroMap2d<'data, i8, Count, PatternULE>,
+    pub patterns:
+        VarZeroVec<'data, VarTupleULE<u8, PluralElementsPackedULE<SinglePlaceholderPattern>>>,
```
sffc, line 69: Observation: you are using a VarZeroVec as a map? Any particular reason you did this instead of a ZeroMap? Stack size?

robertbastian: we need to find the next smaller magnitude, it's not an exact lookup

components/experimental/src/compactdecimal/provider.rs
robertbastian: this is a PluralElements containing a specific pattern

sffc: Right, and the specific pattern is `"{0}"` right? That seems general enough to be in that crate where PluralElementsPackedULE is defined. I don't like unsafe code at a distance.

robertbastian: `PluralElements`/`PluralElementsPackedULE` is a container. I don't want to add constants for every possible concrete type. `icu_plurals` doesn't even depend on `icu_pattern`.

Issue #4241, opened 2023-10-30
Compact decimal should retain 2 significant digits for small numbers


sffc:
This was found from @sven-oly's conformance test suite.

If the input to `CompactDecimalFormatter::format_fixed_decimal` is a fractional number less than 1, we should retain 2 significant digits according to the ECMA-402 specification (citation available upon request). This is highlighted in the conformance test suite. Currently we are just rounding off the trailing fractional digits completely.

https://github.com/unicode-org/icu4x/blob/582e46cd5bac851d25eea6188e5641ad99ce8a78/experimental/compactdecimal/src/compactdecimal.rs#L492

Manishearth:
Triage: remains in 2.2 as long as compact decimal is a headline item

robertbastian:
@sffc I would like to request that citation. Should it round to two significant digits, or to the hundredths place?

sffc:
> [@sffc](https://github.com/sffc) I would like to request that citation. Should it round to two significant digits, or to the hundredths place?

It's in the following AO which is fairly confusing since it handles digit counts for all units and notations

https://tc39.es/ecma402/#sec-setnumberformatdigitoptions

The section that matters for compact notation is

```
    a. Set intlObj.[[MinimumFractionDigits]] to 0.
    b. Set intlObj.[[MaximumFractionDigits]] to 0.
    c. Set intlObj.[[MinimumSignificantDigits]] to 1.
    d. Set intlObj.[[MaximumSignificantDigits]] to 2.
    e. Set intlObj.[[RoundingType]] to more-precision.
```

PR #7389, opened 2026-01-08
Round fractional compact decimals to two significant digits

robertbastian:
#4241


Manishearth (approved):
This _seems_ correct, but it might be better to wait for Shane to supply his citation and be sure about the spec.

robertbastian:
all of these are covered by current tests

PR #7393, opened 2026-01-09
Create SinglePlaceholderPattern::PASS_THROUGH

sffc:
It can be used to slightly simplify the code in #7391

### DIFF
--- a/components/experimental/src/compactdecimal/provider.rs
+++ b/components/experimental/src/compactdecimal/provider.rs
@@ -97,9 +97,7 @@ fn validate_plural_pattern_0_map() {
         CompactDecimalPatternData::PLURAL_PATTERN_0,
         &*encode_varule_to_box(&PluralElements::new((
             FourBitMetadata::try_from_byte(0).unwrap(),
-            SinglePlaceholderPattern::try_from_str("{0}", Default::default())
-                .unwrap()
-                .as_ref()
+            SinglePlaceholderPattern::PASS_THROUGH
         )))
     );
 }


components/pattern/src/lib.rs
```diff
@@ -112,6 +112,25 @@ mod private {
...
+    /// use icu_pattern::SinglePlaceholderPattern;
+    /// use writeable::assert_writeable_eq;
+    ///
+    /// assert_writeable_eq!(
+    ///     SinglePlaceholderPattern::PASS_THROUGH.interpolate(["hello, world!"]),
+    ///     "hello, world!"
+    /// );
+    /// ```
+    pub const PASS_THROUGH: &'static SinglePlaceholderPattern =
+        SinglePlaceholderPattern::from_ref_store_unchecked("\x01");
```
robertbastian, line 136: add a test

robertbastian (commented):
I'm neutral on adding this as a constant. I think using unsafe in compact decimal is the pragmatic solution, because putting this in a const ULE collection will be disproportionately complex

sffc:
> I'm neutral on adding this as a constant.

I'll add @Manishearth to review the API addition in icu_pattern.

> I think using unsafe in compact decimal is the pragmatic solution, because putting this in a const ULE collection will be disproportionately complex

This PR does nothing in the compact decimal component.

sffc:
> This PR does nothing in the compact decimal component.

Well actually now it does. But it doesn't change any machinery, it just simplifies a call site.

robertbastian:
> This PR does nothing in the compact decimal component

if you're not doing this for compact decimal case I'm going to say let's not add APIs without use cases

components/pattern/src/lib.rs
sffc: done (there was already a test but now I test it in ULE space)

Manishearth:
API seems acceptable, agree that we should add it with a use case, but I think it's a decent API to have.

sffc:
I'm merging this because I think it is a useful API because it is the most commonly useful default value, which should be constructible without having to unwrap anything.

Note: Default::default() is impled generically on Pattern but returns an empty pattern with no placeholder.

PR #7394, opened 2026-01-09
Add FixedLengthVarULE and user in icu_plurals

sffc:
See #7391
Depends on #7399

I was making a PR that got a bit bigger than I wanted, so I split this out into a standalone change.

We finally have `VarZeroCow`, a heap-or-reference container for a VarULE. Another reasonable container to want is what I'm proposing in this PR, a `FixedLengthVarULE`, which stores the VarULE on the stack with a fixed number of bytes.

This can be used to compose ULEs and VarULEs in const contexts, as I show with the PluralElementsPackedULE example.

### DIFF
--- a/components/plurals/src/provider.rs
+++ b/components/plurals/src/provider.rs
@@ -482,7 +482,7 @@ where
         core::mem::transmute(bytes)
     }
 
-    /// Creates a singleton [`PluralElementsPackedULE`] in a const context.
+    /// Creates a [`PluralElementsPackedULE`] with an "other" variant in a const context.
     ///
     /// Const parameters:
     ///
@@ -508,7 +508,7 @@ where
     /// let value = "hello, world!"; // 13 bytes long
     /// let metadata = FourBitMetadata::try_from_byte(11).unwrap();
     /// let inner_ule = SizedVarULEBytes::<13, str>::try_from_encodeable(value).unwrap();
-    /// let plural_ule = PluralElementsPackedULE::new_singleton_mn::<13, 14>(inner_ule, metadata);
+    /// let plural_ule = PluralElementsPackedULE::new_mn::<13, 14>(inner_ule, metadata);
     /// let rules = PluralRules::try_new(locale!("en").into(), Default::default()).unwrap();
     ///
     /// assert_eq!(plural_ule.as_varule().get(0.into(), &rules).1, "hello, world!");
@@ -530,7 +530,7 @@ where
     /// use zerovec::ule::SizedVarULEBytes;
     ///
     /// const plural_ule: SizedVarULEBytes<1, PluralElementsPackedULE<str>> =
-    ///     PluralElementsPackedULE::new_singleton_mn::<0, 1>(SizedVarULEBytes::EMPTY_STR, FourBitMetadata::zero());
+    ///     PluralElementsPackedULE::new_mn::<0, 1>(SizedVarULEBytes::EMPTY_STR, FourBitMetadata::zero());
     ///
     /// let rules = PluralRules::try_new(locale!("en").into(), Default::default()).unwrap();
     ///
@@ -540,14 +540,14 @@ where
     /// ```
     ///
     /// [generic_const_exprs]: https://doc.rust-lang.org/beta/unstable-book/language-features/generic-const-exprs.html#generic_const_exprs
-    pub const fn new_singleton_mn<const M: usize, const N: usize>(
+    pub const fn new_mn<const M: usize, const N: usize>(
         input: SizedVarULEBytes<M, V>,
         metadata: FourBitMetadata,
     ) -> SizedVarULEBytes<N, PluralElementsPackedULE<V>> {
         #[allow(clippy::panic)] // for safety, and documented
         if N != M + 1 {
             panic!(concat!(
-                "new_singleton_mn: N (",
+                "new_mn: N (",
                 stringify!(N),
                 ") != 1 + M (",
                 stringify!(M),
--- a/components/plurals/src/provider.rs
+++ b/components/plurals/src/provider.rs
@@ -500,30 +500,37 @@ where
     ///
     /// ```
     /// use icu::plurals::provider::PluralElementsPackedULE;
+    /// use icu::plurals::provider::FourBitMetadata;
     /// use icu::plurals::PluralRules;
     /// use icu::locale::locale;
     /// use zerovec::ule::SizedVarULEBytes;
     ///
     /// let value = "hello, world!"; // 13 bytes long
+    /// let metadata = FourBitMetadata::try_from_byte(11).unwrap();
     /// let inner_ule = SizedVarULEBytes::<13, str>::try_from_encodeable(value).unwrap();
-    /// let plural_ule = PluralElementsPackedULE::new_singleton_mn::<13, 14>(inner_ule);
+    /// let plural_ule = PluralElementsPackedULE::new_singleton_mn::<13, 14>(inner_ule, metadata);
     /// let rules = PluralRules::try_new(locale!("en").into(), Default::default()).unwrap();
     ///
     /// assert_eq!(plural_ule.as_varule().get(0.into(), &rules).1, "hello, world!");
     /// assert_eq!(plural_ule.as_varule().get(1.into(), &rules).1, "hello, world!");
     /// assert_eq!(plural_ule.as_varule().get(2.into(), &rules).1, "hello, world!");
+    ///
+    /// assert_eq!(plural_ule.as_varule().get(0.into(), &rules).0, metadata);
+    /// assert_eq!(plural_ule.as_varule().get(1.into(), &rules).0, metadata);
+    /// assert_eq!(plural_ule.as_varule().get(2.into(), &rules).0, metadata);
     /// ```
     ///
     /// In a const context:
     ///
     /// ```
     /// use icu::plurals::provider::PluralElementsPackedULE;
+    /// use icu::plurals::provider::FourBitMetadata;
     /// use icu::plurals::PluralRules;
     /// use icu::locale::locale;
     /// use zerovec::ule::SizedVarULEBytes;
     ///
     /// const plural_ule: SizedVarULEBytes<1, PluralElementsPackedULE<str>> =
-    ///     PluralElementsPackedULE::new_singleton_mn::<0, 1>(SizedVarULEBytes::EMPTY_STR);
+    ///     PluralElementsPackedULE::new_singleton_mn::<0, 1>(SizedVarULEBytes::EMPTY_STR, FourBitMetadata::zero());
     ///
     /// let rules = PluralRules::try_new(locale!("en").into(), Default::default()).unwrap();
     ///
@@ -535,6 +542,7 @@ where
     /// [generic_const_exprs]: https://doc.rust-lang.org/beta/unstable-book/language-features/generic-const-exprs.html#generic_const_exprs
     pub const fn new_singleton_mn<const M: usize, const N: usize>(
         input: SizedVarULEBytes<M, V>,
+        metadata: FourBitMetadata,
     ) -> SizedVarULEBytes<N, PluralElementsPackedULE<V>> {
         #[allow(clippy::panic)] // for safety, and documented
         if N != M + 1 {
@@ -556,10 +564,10 @@ where
             remainder[i] = input.as_bytes()[i];
             i += 1;
         }
-        // First byte = 0 for a singleton
-        *start = 0;
+        // First byte = 0...mmmm for a singleton
+        *start = metadata.get();
         // Safety: bytes are a valid representation of this type:
-        // 1. The first byte is 0 which indicates a singleton
+        // 1. The first bit is 0 which indicates a singleton
         // 2. The remainder is a valid V by invariant of the input parameter
         unsafe { SizedVarULEBytes::new_unchecked(bytes) }
     }
--- a/components/plurals/src/provider.rs
+++ b/components/plurals/src/provider.rs
@@ -31,7 +31,7 @@ use zerovec::ule::EncodeAsVarULE;
 use zerovec::ule::UleError;
 use zerovec::ule::VarULE;
 use zerovec::ule::ULE;
-use zerovec::ule::{AsULE, ConstStackVarULE};
+use zerovec::ule::{AsULE, SizedVarULEBytes};
 use zerovec::VarZeroSlice;
 
 pub mod rules;
@@ -502,10 +502,10 @@ where
     /// use icu::plurals::provider::PluralElementsPackedULE;
     /// use icu::plurals::PluralRules;
     /// use icu::locale::locale;
-    /// use zerovec::ule::ConstStackVarULE;
+    /// use zerovec::ule::SizedVarULEBytes;
     ///
     /// let value = "hello, world!"; // 13 bytes long
-    /// let inner_ule = ConstStackVarULE::<13, str>::try_from_encodeable(value).unwrap();
+    /// let inner_ule = SizedVarULEBytes::<13, str>::try_from_encodeable(value).unwrap();
     /// let plural_ule = PluralElementsPackedULE::new_singleton_mn::<13, 14>(inner_ule);
     /// let rules = PluralRules::try_new(locale!("en").into(), Default::default()).unwrap();
     ///
@@ -520,10 +520,10 @@ where
     /// use icu::plurals::provider::PluralElementsPackedULE;
     /// use icu::plurals::PluralRules;
     /// use icu::locale::locale;
-    /// use zerovec::ule::ConstStackVarULE;
+    /// use zerovec::ule::SizedVarULEBytes;
     ///
-    /// const plural_ule: ConstStackVarULE<1, PluralElementsPackedULE<str>> =
-    ///     PluralElementsPackedULE::new_singleton_mn::<0, 1>(ConstStackVarULE::EMPTY_STR);
+    /// const plural_ule: SizedVarULEBytes<1, PluralElementsPackedULE<str>> =
+    ///     PluralElementsPackedULE::new_singleton_mn::<0, 1>(SizedVarULEBytes::EMPTY_STR);
     ///
     /// let rules = PluralRules::try_new(locale!("en").into(), Default::default()).unwrap();
     ///
@@ -534,8 +534,8 @@ where
     ///
     /// [generic_const_exprs]: https://doc.rust-lang.org/beta/unstable-book/language-features/generic-const-exprs.html#generic_const_exprs
     pub const fn new_singleton_mn<const M: usize, const N: usize>(
-        input: ConstStackVarULE<M, V>,
-    ) -> ConstStackVarULE<N, PluralElementsPackedULE<V>> {
+        input: SizedVarULEBytes<M, V>,
+    ) -> SizedVarULEBytes<N, PluralElementsPackedULE<V>> {
         #[allow(clippy::panic)] // for safety, and documented
         if N != M + 1 {
             panic!(concat!(
@@ -561,7 +561,7 @@ where
         // Safety: bytes are a valid representation of this type:
         // 1. The first byte is 0 which indicates a singleton
         // 2. The remainder is a valid V by invariant of the input parameter
-        unsafe { ConstStackVarULE::new_unchecked(bytes) }
+        unsafe { SizedVarULEBytes::new_unchecked(bytes) }
     }
 
     /// Returns a tuple with:
--- a/components/plurals/src/provider.rs
+++ b/components/plurals/src/provider.rs
@@ -508,7 +508,7 @@ where
     /// let value = "hello, world!"; // 13 bytes long
     /// let metadata = FourBitMetadata::try_from_byte(11).unwrap();
     /// let inner_ule = SizedVarULEBytes::<13, str>::try_from_encodeable(value).unwrap();
-    /// let plural_ule = PluralElementsPackedULE::new_mn::<13, 14>(metadata, inner_ule);
+    /// let plural_ule = PluralElementsPackedULE::new_mn::<_, 14>(metadata, inner_ule);
     /// let rules = PluralRules::try_new(locale!("en").into(), Default::default()).unwrap();
     ///
     /// assert_eq!(plural_ule.as_varule().get(0.into(), &rules), (metadata, "hello, world!"));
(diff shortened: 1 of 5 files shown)


robertbastian (commented):
I don't see the problem this solves.

sffc:
> This can be used to compose ULEs and VarULEs in const contexts, as I show with the PluralElementsPackedULE example.

It can also be used to create a string or zerovec on the stack, without any additional dependencies.

components/plurals/src/provider.rs
```diff
@@ -482,6 +482,69 @@ where
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use icu::plurals::provider::PluralElementsPackedULE;
+    /// use icu::plurals::PluralRules;
+    /// use icu::locale::locale;
+    /// use zerovec::ule::FixedLengthVarULE;
+    ///
+    /// let value = "hello, world!"; // 13 bytes long
```
Manishearth, line 508: issue: this isn't a motivating example: this doesn't work in `const` anyway (since there's `try_from_encodeable`)

(It's an example, but not a motivating one)

Manishearth (changes requested):
I don't really understand the motivation: why is `VarZeroCow<'static>` insufficient? What is the benefit of having these values on the stack instead of static memory?

As far as I can tell, anything you ought to be able to do with this type in `const` should be doable in non-const as well.

Manishearth:
Ah, I understand. This lets you copy the data into a new, stack-owned buffer, which you can't do with static data. Hm

components/plurals/src/provider.rs
sffc: It would work in const if I replaced `try_from_encodeable` with `new_unchecked`, I just wanted to not have unsafe code in the docs test

Manishearth: Yeah, that's fine, I was just saying that the example itself wasn't motivating the new zerovec API.

But I have a clearer idea of what's going on now.

components/plurals/src/provider.rs
```diff
@@ -482,6 +482,69 @@ where
...
+    /// let plural_ule = PluralElementsPackedULE::new_singleton_mn::<13, 14>(inner_ule);
+    /// let rules = PluralRules::try_new(locale!("en").into(), Default::default()).unwrap();
+    ///
+    /// assert_eq!(plural_ule.as_varule().get(0.into(), &rules).1, "hello, world!");
+    /// assert_eq!(plural_ule.as_varule().get(1.into(), &rules).1, "hello, world!");
+    /// assert_eq!(plural_ule.as_varule().get(2.into(), &rules).1, "hello, world!");
+    /// ```
+    ///
+    /// [generic_const_exprs]: https://doc.rust-lang.org/beta/unstable-book/language-features/generic-const-exprs.html#generic_const_exprs
+    pub const fn new_singleton_mn<const M: usize, const N: usize>(
```
robertbastian: I don't like these APIs where you have to guess the output.

robertbastian (changes requested):

components/plurals/src/provider.rs
```diff
@@ -482,6 +482,69 @@ where
...
+        #[allow(clippy::unwrap_used)] // the bytes are nonempty because N > 0
+        let (start, remainder) = bytes.split_first_mut().unwrap();
+        // TODO(1.87): use copy_from_slice
+        let mut i = 0;
+        #[allow(clippy::indexing_slicing)] // both remainder and input are length M
+        while i < M {
+            remainder[i] = input.as_bytes()[i];
+            i += 1;
+        }
+        *start = 0;
```
Manishearth: issue: mention "first byte = 0 for a singleton" for redundancy

Manishearth (commented):
I'd like to hear more about the use case for `new_singleton_mn` in a const context, since as far as I can tell that would need unsafe code anyway.

I think I see use cases for the new type: being able to create a self-contained stack VarULE is useful.

I think the type name is somewhat misleading. Perhaps ConstStackVarULE? or ConstOwnedVarULE?

I think the type is relatively niche and we should name it in a way that reflects that, and document it as such. Stuffing it in the `ule` module sounds good, otherwise. In the long term I'd like to organize the mess that is that module.

components/plurals/src/provider.rs
```diff
@@ -482,6 +482,87 @@ where
...
+    /// In a const context:
+    ///
+    /// ```
+    /// use icu::plurals::provider::PluralElementsPackedULE;
+    /// use icu::plurals::PluralRules;
+    /// use icu::locale::locale;
+    /// use zerovec::ule::FixedLengthVarULE;
+    ///
+    /// const plural_ule: FixedLengthVarULE<1, PluralElementsPackedULE<str>> =
+    ///     PluralElementsPackedULE::new_singleton_mn::<0, 1>(FixedLengthVarULE::EMPTY_STR);
```
Manishearth: clever

sffc:
I added a const example with no unsafe.

In general this is useful for making default/empty VarULEs in const contexts that are composed of inner ULE types.

Manishearth:
> In general this is useful for making default/empty VarULEs in const contexts that are composed of inner ULE types.

Right, but where are we hoping to do that? I'm still missing parts of the story here I think.

sffc:
> > In general this is useful for making default/empty VarULEs in const contexts that are composed of inner ULE types.
> 
> Right, but where are we hoping to do that? I'm still missing parts of the story here I think.

I want to use this in call sites such as

```rust
    pub const PLURAL_PATTERN_0: &'static PluralElementsPackedULE<SinglePlaceholderPattern> =
        unsafe { PluralElementsPackedULE::from_bytes_unchecked(&[0, 1]) };
```

in order to remove the unsafe. @robertbastian may or may not agree with that particular call site but it serves as a useful illustration.

components/plurals/src/provider.rs
sffc: I don't fully understand your comment, but this is in an unstable module.

Manishearth:
Hmm. I guess this gets more useful if we have consts on various VarULE types for common patterns.

I do worry that this is a fair amount of unsafe code to get rid of some easily-tested unsafe, though. Are we planning on using this const pattern a lot?

An alternative potential route might be to define a macro that can define VarULE consts from unchecked bytes, where it also generates a test. This macro becomes unsafe to use, BUT its safety invariant is mostly "if the test passes this macro usage is fine".

Not saying that's _better_, but making sure we have considered the design space.

robertbastian:
I honestly think for that call site this would be a regression. The current code is unsafe, but it's a single expression, similar to hundreds that we have in baked data, and there's a test that it's correct. This PR adds a lot of weird APIs and unsafe code just to make that one expression safe.

Manishearth:
Overall I've found the defaults story in VarULE to be unsatisfying, so "more consts" is definitely an interesting proposition. BUT I'm not fully convinced we should do it this way. Unsure.

components/plurals/src/provider.rs
robertbastian: you have to know `M` and `N` for this not to panic

sffc:
Note: the use case for this type of helper will only grow, because the composition stuff (VarTuple) is fairly new and we are using it more and more.

components/plurals/src/provider.rs
```diff
@@ -482,6 +482,88 @@ where
...
+    ///
+    /// let rules = PluralRules::try_new(locale!("en").into(), Default::default()).unwrap();
+    ///
+    /// assert_eq!(plural_ule.as_varule().get(0.into(), &rules).1, "");
+    /// assert_eq!(plural_ule.as_varule().get(1.into(), &rules).1, "");
+    /// assert_eq!(plural_ule.as_varule().get(2.into(), &rules).1, "");
+    /// ```
+    ///
+    /// [generic_const_exprs]: https://doc.rust-lang.org/beta/unstable-book/language-features/generic-const-exprs.html#generic_const_exprs
+    pub const fn new_singleton_mn<const M: usize, const N: usize>(
```
robertbastian: "singleton" is the wrong term, we don't use that with plurals. this method is equivalent to [`PluralElements::new`](https://docs.rs/icu/latest/icu/plurals/struct.PluralElements.html#method.new), so it should just be `new`

components/plurals/src/provider.rs
```diff
@@ -482,6 +482,88 @@ where
...
+        input: ConstStackVarULE<M, V>,
+    ) -> ConstStackVarULE<N, PluralElementsPackedULE<V>> {
+        #[allow(clippy::panic)] // for safety, and documented
+        if N != M + 1 {
```
robertbastian, line 546: issue: this should be a const assertion, not a runtime panic

sffc:
> I honestly think for that call site this would be a regression. The current code is unsafe, but it's a single expression, similar to hundreds that we have in baked data, and there's a test that it's correct. This PR adds a lot of weird APIs and unsafe code just to make that one expression safe.

It's not just about that one expression. It's about that class of expressions, which I claim is a class that will get bigger over time.

components/plurals/src/provider.rs
```diff
@@ -482,6 +482,88 @@ where
...
+        #[allow(clippy::unwrap_used)] // the bytes are nonempty because N > 0
+        let (start, remainder) = bytes.split_first_mut().unwrap();
+        // TODO(1.87): use copy_from_slice
+        let mut i = 0;
+        #[allow(clippy::indexing_slicing)] // both remainder and input are length M
+        while i < M {
+            remainder[i] = input.as_bytes()[i];
+            i += 1;
+        }
+        // First byte = 0 for a singleton
```
robertbastian: this is not "for a singleton" but "for a `other` case"

and this 0 represents a `PluralCategoryAndMetadata`, so you should use that type to construct it

utils/zerovec/src/ule/fixed_length.rs
```diff
@@ -0,0 +1,107 @@
...
+///
+/// let container = ConstStackVarULE::<13, str>::try_from_encodeable("hello, world!").unwrap();
+///
+/// assert_eq!(&*container, "hello, world!");
+///
+/// // Returns an error if the container is not the correct size:
+/// ConstStackVarULE::<20, str>::try_from_encodeable("hello, world!").unwrap_err();
+/// ```
+#[derive(Copy, Clone, PartialEq, Eq)]
+pub struct ConstStackVarULE<const N: usize, V: VarULE + ?Sized> {
```
robertbastian: issue: this is not generally on the stack. the point is that it's `Sized` so it *can* be on the stack. `SizedVarULE` sounds like a trait, maybe `SizedVarULEBytes`? `EncodedSizedVarULE`? `EncodedVarULEBytes`?

utils/zerovec/src/ule/fixed_length.rs
```diff
@@ -0,0 +1,107 @@
...
+    bytes: [u8; N],
+    _marker: PhantomData<V>,
+}
+
+impl<const N: usize, V: VarULE + ?Sized> ConstStackVarULE<N, V> {
+    /// Creates one of these from an [`EncodeAsVarULE`].
+    ///
+    /// Returns an error if the byte length in the container is not the correct length
+    /// for the encodeable object.
+    pub fn try_from_encodeable(input: impl EncodeAsVarULE<V>) -> Result<Self, UleError> {
```
robertbastian, line 53:
```suggestion
    pub fn try_from_encodeable(input: &impl EncodeAsVarULE<V>) -> Result<Self, UleError> {
```
?

sffc:
I also just want to point out:

```rust
    pub const PLURAL_PATTERN_0: &'static PluralElementsPackedULE<SinglePlaceholderPattern> =
        unsafe { PluralElementsPackedULE::from_bytes_unchecked(&[0, 1]) };
```

This code requires subject matter expertise in 3 different components: pattern (to know that 1 means "single placeholder pattern with a single placeholder"), plurals (to know that 0 means "singleton plural elements"), and compact (to know that this combination of pattern and plurals is suitable for `PLURAL_PATTERN_0`). Yes, the test helps a lot. But, you should be able to write a call site such as the following, and no extra test is required.

```rust
    pub const PLURAL_PATTERN_0: /* some type */ =
        PluralElementsPackedULE::new_singleton(SinglePlaceholderPattern::PASS_THROUGH_ULE);
```

(That line of code does two things that we don't have yet: `new_singleton`, which is gated on Rust feature `generic_const_exprs`, and `PASS_THROUGH_ULE`, because the ULE composition needs to use the correct types. This comment is not an opportunity to bikeshed either of those APIs.)

robertbastian (changes requested):
I still don't agree with this, but leaving a technical review

components/plurals/src/provider.rs
sffc: please suggest code that compiles with a const assertion.

robertbastian:
> This code requires subject matter expertise in 3 different components: pattern (to know that 1 means "single placeholder pattern with a single placeholder"), plurals (to know that 0 means "singleton plural elements"), and compact (to know that this combination of pattern and plurals is suitable for PLURAL_PATTERN_0)

No it doesn't. I wrote that code and don't know how these things are encoded. I just wrote the test and then put the bytes there that make the test pass. I think that's a valid and very readable approach.

> But, you should be able to write a call site such as the following, and no extra test is required

Yes! I would love that. But it's not feasible with current Rust/zerovec.

components/plurals/src/provider.rs
sffc: The word "singleton" is already used in this file multiple times, including in test function names.

The difference between singleton and non-singleton matters a great deal to PluralElementsPackedULE, because it results in two different encodings.

components/plurals/src/provider.rs
robertbastian: if you can't do it then we shouldn't have this yet

components/plurals/src/provider.rs
sffc: It is for a singleton. That is the term from the byte encoding definition.

sffc: Actually to be fully correct we should add in the `FourBitMetadata`.

utils/zerovec/src/ule/fixed_length.rs
sffc: I like `SizedVarULEBytes`

components/plurals/src/provider.rs
robertbastian: I thought the caller shouldn't have to know encoding details?

components/plurals/src/provider.rs
robertbastian: yes a `PackedPluralElementsULE<T>` corresponds to a `PluralElements<(FourBitMetadata, T)>`, so the constructor should take that tuple.

utils/zerovec/src/ule/fixed_length.rs
sffc: Not sure the recommend style here. It doesn't matter if the trait is implemented on a reference. @Manishearth ?

components/plurals/src/provider.rs
sffc: hm? We have lots of const code with runtime assertions because there isn't a good way to write const assertions.

components/plurals/src/provider.rs
sffc: done. (not as a tuple)

utils/zerovec/src/ule/fixed_length.rs
robertbastian: I don't think this is style choice. you only test this with a `&str` input, but say you have a `String`, you don't want to accidentally pass that by value.

components/plurals/src/provider.rs
robertbastian: all other APIs use `(FourBitMetadata, T)`, please at least change the order

components/plurals/src/provider.rs
```diff
@@ -482,6 +482,96 @@ where
...
+    /// use icu::locale::locale;
+    /// use zerovec::ule::SizedVarULEBytes;
+    ///
+    /// let value = "hello, world!"; // 13 bytes long
+    /// let metadata = FourBitMetadata::try_from_byte(11).unwrap();
+    /// let inner_ule = SizedVarULEBytes::<13, str>::try_from_encodeable(value).unwrap();
+    /// let plural_ule = PluralElementsPackedULE::new_singleton_mn::<13, 14>(inner_ule, metadata);
+    /// let rules = PluralRules::try_new(locale!("en").into(), Default::default()).unwrap();
+    ///
+    /// assert_eq!(plural_ule.as_varule().get(0.into(), &rules).1, "hello, world!");
```
robertbastian: nit: merge the assertions
```suggestion
    /// assert_eq!(plural_ule.as_varule().get(0.into(), &rules), ("hello, world!", metadata));
```

components/plurals/src/provider.rs
```diff
@@ -482,6 +482,96 @@ where
...
+    ///
+    /// ```
+    /// use icu::plurals::provider::PluralElementsPackedULE;
+    /// use icu::plurals::provider::FourBitMetadata;
+    /// use icu::plurals::PluralRules;
+    /// use icu::locale::locale;
+    /// use zerovec::ule::SizedVarULEBytes;
+    ///
+    /// const plural_ule: SizedVarULEBytes<1, PluralElementsPackedULE<str>> =
+    ///     PluralElementsPackedULE::new_singleton_mn::<0, 1>(SizedVarULEBytes::EMPTY_STR, FourBitMetadata::zero());
```
robertbastian: or better just integrate into the example above, this is very repetitive

```suggestion
    /// let plural_ule = const {
    ///     PluralElementsPackedULE::new_singleton_mn::<0, 1>(SizedVarULEBytes::EMPTY_STR, FourBitMetadata::zero()) };
```

components/plurals/src/provider.rs
robertbastian: this also doesn't need to be qualified as `_mn`; it's an unstable type so we can just change the generics if `generic_const_exprs` ever gets stabilised (I wouldn't hold my breath). so just `new`

components/plurals/src/provider.rs
sffc: Even the Rust standard library does runtime panics on const parameters. For example:

https://doc.rust-lang.org/std/primitive.slice.html#method.as_chunks

`slice::as_chunks` panics if the const parameter `N` is zero. There was a lengthy discussion on the proposal thread, starting here: https://github.com/rust-lang/rust/issues/74985#issuecomment-2770314101

components/plurals/src/provider.rs
sffc: I'd rather keep the `_mn` because it's a pain to debug cross-crate function naming issues when the function mostly stays the same but const parameters change a little. But I will remove `singleton`.

Manishearth:
> It's not just about that one expression. It's about that class of expressions, which I claim is a class that will get bigger over time.

I am not yet convinced that this is the case, but the use case seems sufficiently well outlined and I agree it is valid.. I would be in favor of us having more consts like this. As I said before, the default construction problem in varule is something I've wanted to see improvement on.

Regarding the MN stuff I'm overall okay with landing suboptimal API we know that we can clean up in future Rust. Though I don't know if we have a timeline for const expressions in generics?

sffc:
> Though I don't know if we have a timeline for const expressions in generics?

Well, it works in nightly, and https://github.com/rust-lang/rust/issues/76560 has a lot of upvotes. It seems like there are concrete next steps but no one is currently actively funding it.

sffc:
OK I added an ergonomic `to_sized_varule_bytes!` macro for construction, which even works with type hinting, and then migrated the compactdecimal call site to:

```rust
    pub const PLURAL_PATTERN_0: SizedVarULEBytes<
        2,
        PluralElementsPackedULE<SinglePlaceholderPattern>,
    > = PluralElementsPackedULE::new_mn(
        FourBitMetadata::zero(),
        to_sized_varule_bytes!(SinglePlaceholderPattern::PASS_THROUGH),
    );
```

Const variables always require a type, or else the type could be elided, except for the number 2.

Manishearth:
@sffc Thanks!

I think if we're open to using a macro we don't need the SizedVarULEBytes type at all. The macro can construct a const/static array, and then construct a second const/static that references it.

This is what I was referring to earlier:

> An alternative potential route might be to define a macro that can define VarULE consts from unchecked bytes, where it also generates a test. This macro becomes unsafe to use, BUT its safety invariant is mostly "if the test passes this macro usage is fine".

I can write more about this later.

sffc:
If we can't remove the unsafe, then a macro of that shape isn't worth doing in my opinion.

Manishearth:
Sorry, I should have been more explicit: I think using the design of your macro so far combined with my model, this can be done without callsite unsafe.

robertbastian:
FYI the unsafe code you're trying to replace does not exist anymore

Manishearth:
I've been thinking about the core `const` problem here more.

The angle I am coming at this from is that I want people to be able to use regular Rust DST mechanisms as much as possible. So, passing VarULE types around as `&V`, and occasionally `Box<V>`, `Rc<V>`, etc. In general, this seems aligned with Rust goals around how they intend to evolve DSTs.

The issue here is that `const` can do all kinds of stuff on the stack, but a const fn cannot really make a `static` at runtime and reference it. So we cannot return `&V` in `const`. We can return `[u8; N]` and have an unsafe-usable invariant that it is a valid `V`, or we can use a macro to make the `static`, or we can do this type of wrap thing.

To some extent, it makes sense why `const fn`s can't internally make a static yet: `const fn` is _optionally_ const: it must still be executable at runtime. Which means it still needs to follow the rule of runtime data not flowing into statics.

I asked some Rust lang/compiler people about this use case and it seems like there are two in-progress features that are ideal for this. One is const alloc (`#![feature(const_heap)]`):

```rust
#![feature(const_heap, const_trait_impl)]

const fn make_bytes(byte1: u8, byte2: u8) -> &'static [u8] {
    let mut bytes = Vec::with_capacity(2);
    bytes.push(byte1);
    bytes.push(byte2);
    bytes.const_make_global()
}
```

They also mentioned this as being supported by plans for [`comptime`](https://github.com/rust-lang/rust/pull/148820) (which is const-_only_, which won't have the "can't make statics at runtime" problem), but I don't have an equivalent code snippet yet since the code hasn't landed yet. I think it would probably also use `const_make_global`, which does work in `const` but is internally const-only (it's a no-op when called at runtime).

Not something we can use yet, but it's good to know of the ways in which these gaps are going to be filled.

sffc:
Discussion with @robertbastian: it seems better to use a proc macro that invokes EncodeAsVarULE. Most users of VarULE are already using `make_varule` so they already have a syn dependency.

I'm happy with such a solution, but I know @Manishearth wants to avoid adding more unsafe-generating proc macro code. wdyt?

Manishearth:
I need to page this discussion back in (and will respond when I do), but can you sketch out what the proc macro invocation would look like? I'm not clear on how the proc macro would have access to EncodeAsVarULE code.

I am open to some proc macro like solutions here but I don't see any workable ones at the moment.

Manishearth:
Alright. So I think my _primary_ concern here is that I think this moves us away from what I want zerovec's design to be like: I want us to rely on Rust DST stuff as much as possible; and I want Rust DST stuff to get better so this is easier.  I don't want this crate to be spending time patching gaps in Rust DST infrastructure, I prefer just not supporting something over that.

My secondary (but still major) concern is unsafety. I think "construct const VZV" is a useful use case, but the current API is basically still unsafe in const, _and_ it's a complex-enough generic unsafe API that makes it high-risk in my book.

While I generally disprefer proc macros over code for unsafe stuff, I think a proc macro might be fine depending on how complex it is. But I can't see how we can get a proc macro to work here.

PR #7397, opened 2026-01-09
Fix imports

robertbastian:


provider/source/src/datetime/semantic_skeletons/tests.rs
```diff
@@ -180,12 +174,13 @@ fn test_en_overlap_patterns() {
 ///
 /// See: https://unicode-org.atlassian.net/browse/CLDR-14993
 #[cfg(feature = "networking")]
-#[cfg(test)]
```
sffc, line 183: Why did you remove the `#[cfg(test)]`?

Manishearth: It's already on the module

Manishearth: It's already on the module

PR #7398, opened 2026-01-10
Add delegates/sffc/rust_versions_and_style.md

sffc:
This contains some principles that I'd like to be able to cite in a centralized location

### DIFF
--- a/documents/delegates/sffc/rust_versions_and_style.md
+++ b/documents/delegates/sffc/rust_versions_and_style.md
@@ -1,7 +1,7 @@
 Positions on Rust Version and Style
 ===================================
 
-These positions are held by sffc@ and not necesarilly the ICU4X Technical Committee.
+*These positions are held by sffc@ and not necesarilly the ICU4X Technical Committee.*
 
 ## MSRV upgrades should be a last resort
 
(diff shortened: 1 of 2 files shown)


documents/delegates/sffc/rust_versions_and_style.md
```diff
@@ -0,0 +1,30 @@
...
+In my experience, there are many, many reasons why systems use older versions of software. Some may be good, some may be bad, some may be lazy. But, ICU4X should try to "just work", regardless of the environment it is built on.
+
+At this stage in our lifecycle, the experience of our users matters more than the experience of our maintainers. To me, MSRV upgrades that improve individual call sites are not sufficient to justify an MSRV bump.
+
+There are of course reasons why we must sometimes bump the MSRV. Examples include:
+
+1. Rust introduces a new feature that we plan to use widely in our architecture (example: GATs)
+2. Rust introduces safe or const code that was _not possible_ to write as safe or const before
+3. Rust introduces non-panicky code that was _not possible_ to _compile_ into non-panicky code before
+
```
Manishearth, line 17: probably worth talking about dependencies as well. I think beyond a 6 release horizon (which is not uncommon) we rapidly hit the issue that getting patches from dependencies becomes harder and harder, and that's often the primary motivator.

We have very few dependencies for our main ICU4X code, but datagen is more profligate.

(Perhaps we should split out those MSRVs? We'd have to make our MSRV test even more robust about this, and have separate `features` tests)

PR #7401, opened 2026-01-12
Deprecate `LanguagePreferences` field accessors

robertbastian:
Callers should convert to `DataLocale` (choosing the fallback mode) and then access the fields.

## Changelog

icu_locale_core: Deprecate `LanguagePreferences` field accessors

### DIFF
--- a/components/calendar/src/preferences.rs
+++ b/components/calendar/src/preferences.rs
@@ -53,25 +53,33 @@ impl CalendarPreferences {
     /// assert_eq!(CalendarPreferences::from(&locale!("und-US-u-rg-thxxxx")).resolved_algorithm(), CalendarAlgorithm::Buddhist);
     /// ```
     pub fn resolved_algorithm(self) -> CalendarAlgorithm {
-        let region = self
-            .locale_preferences
-            .to_data_locale_region_priority()
-            .region;
-        let region = region.as_ref().map(|r| r.as_str());
+        use icu_locale_core::subtags::{region, Region};
+        const AE: Region = region!("AE");
+        const BH: Region = region!("BH");
+        const KW: Region = region!("KW");
+        const QA: Region = region!("QA");
+        const SA: Region = region!("SA");
+        const TH: Region = region!("TH");
+        const AF: Region = region!("AF");
+        const IR: Region = region!("IR");
+
         // This is tested to be consistent with CLDR in icu_provider_source::calendar::test_calendar_resolution
-        match self.calendar_algorithm {
-            Some(CalendarAlgorithm::Hijri(None)) => match region {
-                Some("AE" | "BH" | "KW" | "QA" | "SA") => {
-                    CalendarAlgorithm::Hijri(Some(HijriCalendarAlgorithm::Umalqura))
-                }
-                _ => CalendarAlgorithm::Hijri(Some(HijriCalendarAlgorithm::Civil)),
-            },
-            Some(a) => a,
-            None => match region {
-                Some("TH") => CalendarAlgorithm::Buddhist,
-                Some("AF" | "IR") => CalendarAlgorithm::Persian,
-                _ => CalendarAlgorithm::Gregory,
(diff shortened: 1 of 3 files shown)


components/locale_core/src/preferences/locale.rs
```diff
@@ -194,22 +201,15 @@ impl LocalePreferences {
...
     pub const fn language(&self) -> Language {
-        self.language
+        self.to_data_locale_language_priority().language
     }
 
     /// Preference of Region
-    ///
-    /// This respects the [`RegionOverride`], and should only be used
-    /// for [data that is region-specific](https://github.com/unicode-org/cldr/blob/main/common/supplemental/rgScope.xml).
+    #[deprecated(since = "2.2.0", note = "convert to `DataLocale` to access fields")]
```
zbraniecki, line 210: `question`: Why is the conversion to `DataLocale` (internal type) recommended over `LanguageIdentifier` or `Locale` (public type)?

zbraniecki (commented):
What is the driving motivator for this change? What problem is it meant to solve?

robertbastian:
Converting to DataLocale correctly applies the region and subdivision overrides. The region getter on the preferences does not. And then for consistency just do everything on DataLocale.

components/locale_core/src/preferences/locale.rs
robertbastian: DataLocale is a public type though? The use case for LanguagePreferences is integrating preferences with our data pipeline.

zbraniecki: Not only. The users may also want to use Local preferences before they feed it to our pipeline.

zbraniecki:
I still struggle to understand this change. 

The struct is a composition of preferences in its own merit. The user may want to build their bindings by constructing it, merging, inspecting, serializing, and branching logic based on the values in this naive struct.

When you construct a DataLocale out of it, then yes, you execute business logic to *interpret* fields and converge values from it, that makes sense.
But what is the value of preventing our users from being able to operate on the naked struct before they feed it to our pipeline?

robertbastian:
The region override field that is stored in `LocalePreferences` has a meaning that is defined by UTS 35. Clients should not go and interpret the fields in any other way. I view `DataLocale` as `ResolvedLanguagePreferences`.

robertbastian:
https://github.com/unicode-org/icu4x/pull/7376/ is an example where we shot ourselves in the foot with using raw fields on `LanguagePreferences`.

zbraniecki:
I'm aligned with that interpretation. I'm not aligned that there is no value in allowing users manage a non resolved one.
Construction of the unresolved one may be a multi step process and by making it immutable we force such users to introduce intermediate structures.
I prefer to allow users to work with unresolved until they decide to resolve them.

Saying that. I wouldn't mind encoding your mental model - explaining that DataLocale is ResolvedLocalePreferences and the conversion applies UTS 35 making the result meaningful to them in many scenarios. 
But I don't think it diminishes the value of being able to edit and read fields from unresolved prior to that step

zbraniecki:
Thanks for the example. I'd like to consider if there is a different way to shape the pit of success without blocking the mutability and introspection of the unresolved

robertbastian:
> But I don't think it diminishes the value of being able to edit and read fields from unresolved prior to that step

Currently only two of the 5 fields are readable anyway, and that doesn't seem to have been a problem for anyone.

zbraniecki:
Oh, I realize that in #5785 we changed the fields to not public. I even see my name signed under that drift but I kept my original mental model for the struct when holding this review. I'm sorry, I already lost the argument for `LocalePreferences` to be a naked struct - this PR is just following on that decision.
I'm aligned.

zbraniecki (approved):
lgtm

Manishearth:
I trust the three of you on this change and don't feel the need to review myself. Feel free to rerequest if you specifically want my review.

PR #7406, opened 2026-01-12
Cherry-pick #7396, #7402

robertbastian:
#7396, #7402

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,13 +9,14 @@ Several crates have had patch releases in the 2.1 stream:
         - Fix `icu_locale_core` dependency (unicode-org#7191)
     - (2.1.2)`icu_properties`
         - Fix some property constants (unicode-org#7269, unicode-org#7281, unicode-org#7284)
+        - Add conversions for `unicode_bidi::BidiClass` (unicode-org#7272)
         - Add `IndicConjunctBreak` (unicode-org#7280)
         - Add conversion between `icu::properties::props::Script` and `icu::locale::subtags:Script` (unicode-org#7270)
-        - Add conversion between `icu::properties::props::BidiClass` and `unicode_bidi::BidiClass` (unicode-org#7272)
     - (2.1.2) `icu_segmenter`
         - Add non-complex line and word break constructors (unicode-org#7268)
-- Utils
-  - Retire the `icu_harfbuzz` crate. The `icu_properties` and `icu_normalizer` types now directly implement the `harfbuzz-traits`
+- Data model and providers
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -9,13 +9,14 @@ Several crates have had patch releases in the 2.1 stream:
...
     - (2.1.2)`icu_properties`
         - Fix some property constants (unicode-org#7269, unicode-org#7281, unicode-org#7284)
+        - Add conversions for `unicode_bidi::BidiClass` (unicode-org#7272)
         - Add `IndicConjunctBreak` (unicode-org#7280)
         - Add conversion between `icu::properties::props::Script` and `icu::locale::subtags:Script` (unicode-org#7270)
-        - Add conversion between `icu::properties::props::BidiClass` and `unicode_bidi::BidiClass` (unicode-org#7272)
     - (2.1.2) `icu_segmenter`
         - Add non-complex line and word break constructors (unicode-org#7268)
-- Utils
-  - Retire the `icu_harfbuzz` crate. The `icu_properties` and `icu_normalizer` types now directly implement the `harfbuzz-traits`
```
Manishearth, line 18: question: don't we want this somewhere?

CHANGELOG.md
```diff
@@ -9,13 +9,14 @@ Several crates have had patch releases in the 2.1 stream:
...
+- Data model and providers
+    - (2.1.2) `icu_provider_source`, `icu_locale_data`, `icu_datetime_data`, `icu_experimental_data`
+        - Update to CLDR 48.1 (unicode-org#7396)
```
Manishearth, line 19: nit: include the property stuff in the changelog

CHANGELOG.md
robertbastian: it's in the unreleased-changes section on main, but it's not on this branch and hasn't been released

CHANGELOG.md
robertbastian: what properties stuff?

CHANGELOG.md
Manishearth: Fair enough.

CHANGELOG.md
Manishearth: `icu_provider_source` now can generate the indic property data markers that we forgot to release last time.


Normally we don't list new markers under `icu_provider_source` changelogs: it's implicit. However, this is a case where we forgot to do it, leading to broken upgrades for clients. Since changelogs are primarily a method of communicating with users and helping them understand what changed / why something might have broken, I think we should explicitly list the fix for something we broke.

robertbastian: yeah this is the cherry-pick though. we should have added it in https://github.com/unicode-org/icu4x/pull/7402 so it's on main

PR #7411, opened 2026-01-12
Compartmentalise decimal formatting code

robertbastian:
A `Signed<UnsignedDecimal>` is now formatted as a `FormattedSign<FormattedUnsignedDecimal>`. This way, code can be reused better between compact formatting, currency formatting, etc.

## Changelog: N/A

This is a simple refactor

### DIFF
--- a/components/experimental/src/compactdecimal/provider.rs
+++ b/components/experimental/src/compactdecimal/provider.rs
@@ -77,7 +77,7 @@ where
     }
 }
 
-#[cfg(feature = "datagen")]
+#[cfg(feature = "serde")]
 impl<'de, 'data, P: PatternBackend<Store = str>> serde::Deserialize<'de>
     for CompactPatterns<'data, P>
 where


sffc (commented):
I need a tour of your plans here, in order to know how I should review this. There are a bunch of doc hidden APIs. Are those intended to be on the stabilization track? Or are they hidden for testing or something?

robertbastian:
everything that is doc(hidden) should be pub(crate) once we move compact into this crate

components/decimal/src/format.rs
```diff
@@ -4,64 +4,64 @@
...
 use writeable::Writeable;
 
 /// An intermediate structure returned by [`DecimalFormatter`](crate::DecimalFormatter).
 /// Use [`Writeable`][Writeable] to render the formatted decimal to a string or buffer.
 #[derive(Debug, PartialEq, Clone)]
-pub struct FormattedDecimal<'l> {
-    pub(crate) value: &'l Decimal,
+pub struct FormattedDecimal<'l>(pub(crate) FormattedSign<'l, FormattedUnsignedDecimal<'l>>);
+
+#[doc(hidden)]
```
sffc: Here and elsewhere:

```suggestion
#[doc(hidden)] // TODO(#xxxx): make this pub(crate)
```

components/decimal/src/format.rs
```diff
@@ -4,64 +4,64 @@
...
+/// Building block for formatted numbers
+#[derive(Debug, PartialEq, Clone)]
+pub struct FormattedSign<'l, T> {
+    pub(crate) value: T,
+    pub(crate) sign: Option<(Part, &'l str, &'l str)>,
+}
+
+impl<T: Writeable> Writeable for FormattedSign<'_, T> {
+    fn write_to_parts<S: PartsWrite + ?Sized>(&self, sink: &mut S) -> core::fmt::Result {
```
sffc, line 43: Praise: cool way to write this logic

components/decimal/src/provider.rs
```diff
@@ -268,27 +269,29 @@ icu_provider::data_struct!(
...
+            crate::parts::PLUS_SIGN,
             self.strings.plus_sign_prefix(),
             self.strings.plus_sign_suffix(),
         )
     }
-    /// Return thhe decimal separator
-    pub fn decimal_separator(&self) -> &str {
-        self.strings.decimal_separator()
+    /// Return the decimal separator
+    pub fn decimal_separator(&self) -> (Part, &str) {
```
sffc: Question: why do you need to abstract out the `Part`?

components/experimental/src/compactdecimal/formatter.rs
```diff
@@ -252,59 +235,6 @@ impl CompactDecimalFormatter {
...
-    /// #    Default::default(),
-    /// # ).unwrap();
-    /// assert_writeable_eq!(short_english.format_i64(999_499), "999K");
-    /// assert_writeable_eq!(short_english.format_i64(999_500), "1M");
-    /// assert_writeable_eq!(short_english.format_i64(1650), "1.6K");
-    /// assert_writeable_eq!(short_english.format_i64(1750), "1.8K");
-    /// assert_writeable_eq!(short_english.format_i64(1950), "2K");
-    /// assert_writeable_eq!(short_english.format_i64(-1_172_700), "-1.2M");
-    /// ```
-    pub fn format_i64(&self, value: i64) -> FormattedCompactDecimal<'_> {
```
sffc, line 303: Question/Issue: Why are you deleting this convenience function?

It's common to format integers with compact decimal notation. I think I prefer to keep the convenience function.

components/experimental/src/compactdecimal/formatter.rs
```diff
@@ -508,19 +438,24 @@ impl CompactDecimalFormatter {
...
+
+        self.decimal_formatter.format_sign(
+            significand.sign,
+            plural_map
+                .get(
+                    PluralOperands::from_significand_and_exponent(&significand.absolute, exponent),
+                    &self.plural_rules,
+                )
+                .1
+                .interpolate([self
```
sffc, line 416: Praise: I'm glad that we can compose writeable patterns like this.

components/experimental/src/compactdecimal/formatter.rs
robertbastian: because we don't have it on `DecimalFormatter`, in fact we explicitly decided against it: https://github.com/unicode-org/icu4x/issues/5953#issuecomment-2639423609

components/experimental/src/compactdecimal/formatter.rs
robertbastian: yes, but I think `interpolate` should return a concrete type

components/decimal/src/provider.rs
robertbastian: ah, in a previous iteration the sign methods had multiple call sites, but the parts should only be defined in one. but with `FormattedSign` I can revert this

components/experimental/src/compactdecimal/formatter.rs
sffc: I feel CompactDecimalFormatter is different. Integers are one of the main use cases.

components/experimental/src/compactdecimal/formatter.rs
sffc: I'm okay with that, it's more just that I didn't want to over-commit to the shape of the return type. If we think we're ready to "stabilize" the signature, we can make it return a concrete type.

components/experimental/src/compactdecimal/formatter.rs
robertbastian: removed functions can easily be added. added functions cannot be easily removed. let's discuss at graduation

robertbastian:
@sffc @younies I'm really blocked on this, I'd appreciate a timely review.

younies (approved):
you need to fix the test cases

```
Baked data needs to be updated. Please run `cargo make bakeddata`
```

components/experimental/src/compactdecimal/formatter.rs
```diff
@@ -668,8 +583,75 @@ impl CompactDecimalFormatter {
...
+}
+
+impl<'a, P: PatternBackend<Store = str>> CompactPatterns<'a, P> {
+    pub(crate) fn get_pattern_and_significand(
+        &'a self,
+        value: &UnsignedDecimal,
+        rules: &PluralRules,
+    ) -> (Option<&'a Pattern<P>>, UnsignedDecimal)
+    where
+        <P as PatternBackend>::Store: zerovec::ule::VarULE,
```
sffc: Issue: Check your bounds. You have two bounds on `Store` (here and in the impl), and it's not clear you need either of them

components/experimental/src/compactdecimal/provider.rs
```diff
@@ -53,53 +53,152 @@ icu_provider::data_marker!(
...
+        = &'b [()]
+    where
+        Self: 'b;
+    fn maybe_as_encodeable<'b>(&'b self) -> Option<Self::EncodeableStruct<'b>> {
+        None
+    }
+}
+
+#[cfg(feature = "datagen")]
+impl<P: PatternBackend<Store = str>> CompactPatterns<'static, P> {
```
sffc, line 116: Issue: You shouldn't ever need to put a bound on `Store` which is internal/hidden

robertbastian: please clean this up if you think you know how, defining this type has taken way too much of my time already

components/experimental/src/compactdecimal/formatter.rs
robertbastian: ah this bound is old, from when the store was not fixed to str

sffc (approved):
ok if you land this I'll try fiddling with the bounds. If you needed these bounds to make code compile then there might be a bug in Pattern.

PR #7413, opened 2026-01-12
Correctly parse ethiopic-amete-alem alias

Manishearth:
`ethiopic-amete-alem` currently parses to `CalendarAlgorithm::Ethiopic`, which is incorrect and causes V8 to fail a test262 test. This is because our subtag parsing ignores subtags after the first for enum keywords, unless we have explicitly defined the enum keyword as having variants.

This fixes that.

### DIFF
--- a/components/locale_core/src/preferences/extensions/unicode/keywords/calendar.rs
+++ b/components/locale_core/src/preferences/extensions/unicode/keywords/calendar.rs
@@ -4,9 +4,10 @@
 
 #![allow(non_snake_case)]
 
-use crate::extensions::unicode::Value;
+use crate::extensions::unicode::{value, Value};
 use crate::preferences::extensions::unicode::enum_keyword;
-use crate::subtags::Subtag;
+use crate::preferences::extensions::unicode::errors::PreferencesParseError;
+use crate::subtags::{subtag, Subtag};
 
 enum_keyword!(
     /// Hijri Calendar sub-type
@@ -23,9 +24,18 @@ enum_keyword!(
         Rgsa
 });
 
-fn value_is_ethioaa(v: &Value) -> bool {
-s.as_subtags_slice() == &[subtag!("ethiopic"), subtag!("amete"), subtag!("alem")]
+/// Handles aliases present in `v`. If found, returns a CalendarAlgorithm, else returns None
+fn handle_aliases(v: &Value) -> Option<CalendarAlgorithm> {
+    if *v == value!("islamicc") {
+        return Some(CalendarAlgorithm::Hijri(Some(
+            HijriCalendarAlgorithm::Civil,
+        )));
+    } else if v.as_subtags_slice() == [subtag!("ethiopic"), subtag!("amete"), subtag!("alem")] {
+        return Some(CalendarAlgorithm::Ethioaa);
+    }
+    None
 }
+
 enum_keyword!(
     /// A Unicode Calendar Identifier defines a type of calendar.
     ///
@@ -68,7 +78,7 @@ enum_keyword!(
         ("persian" => Persian),
         /// Republic of China calendar
         ("roc" => Roc)
-}, "ca", s, if *s == value!("islamicc") { return Ok(Self::Hijri(Some(HijriCalendarAlgorithm::Civil))); } else if value_is_ethioaa(s) { return Ok(Self::Ethioaa) });
+}, "ca", s, if let Some(s) = handle_aliases(s) { return Ok(s) });
 
 #[test]
 fn test_calendar_aliases() {


components/locale_core/src/preferences/extensions/unicode/keywords/calendar.rs
```diff
@@ -21,6 +23,25 @@ enum_keyword!(
...
+    };
+    const ALEM: Subtag = match Subtag::try_from_str("alem") {
+        Ok(s) => s,
+        _ => panic!("valid subtag"),
+    };
+
+    v.subtag_count() == 3
+        && v.get_subtag(0) == Some(&ETHIOPIC)
+        && v.get_subtag(1) == Some(&AMETE)
+        && v.get_subtag(2) == Some(&ALEM)
```
robertbastian:
```suggestion
s.as_subtags_slice() == &[subtag!("ethiopic"), subtag!("amete"), subtag!("alem")]
```

robertbastian:
This is correct per https://github.com/unicode-org/cldr/blob/468272dc1ccbdd1ade875130fecadb5cf870de5a/common/bcp47/calendar.xml#L17

Please also add the other missing alias, `gregorian`.

robertbastian:
Actually, `gregorian` is not valid subtag syntax, so that one we cannot support. LDML allows not supporting it [though](https://www.unicode.org/reports/tr35/#BCP47_T_Extension:~:text=It%20is%20strongly,gregorian%22%20(for%20%22gregory%22).):

> It is strongly recommended that all API methods accept all possible aliases for keywords and types, but generate the canonical form. For example, "ar-u-ca-islamicc" would be equivalent to "ar-u-ca-islamic-civil" on input, but the latter should be output. The one exception is where an alias would only be well-formed with the old syntax, such as "gregorian" (for "gregory").

Manishearth:
Huh, I was searching the docs for the `subtag` macro and couldn't find it. We need to document it better.

robertbastian:
it's probably due to the macro being `#[doc(hidden)]` and then re-exported. I think it used to work though...

components/locale_core/src/preferences/extensions/unicode/keywords/calendar.rs
```diff
@@ -63,4 +69,31 @@ enum_keyword!(
...
+fn test_calendar_aliases() {
+    fn test(s: &str, expectation: CalendarAlgorithm) {
+        let v = Value::try_from_str(s).unwrap();
+        let parsed = CalendarAlgorithm::try_from(&v).unwrap();
+        assert_eq!(parsed, expectation, "{s} must parse to {expectation:?}");
+    }
+
+    test("ethiopic-amete-alem", CalendarAlgorithm::Ethioaa);
+    test("ethiopic", CalendarAlgorithm::Ethiopic);
+    test("ethiopic-foobar", CalendarAlgorithm::Ethiopic);
```
sffc, line 93: Is this correct? Or should we match this to None?

All I see in UTS 35 is:

> Note: Some calendar types are represented by two subtags. In such cases, the first subtag specifies a generic calendar type and the second subtag specifies a calendar algorithm variant. The CLDR uses generic calendar types (single subtag types) for tagging data when calendar algorithm variations within a generic calendar type are irrelevant. For example, type "islamic" is used for specifying Islamic calendar formatting data for all Islamic calendar types, including "islamic-civil" and "islamic-umalqura".

components/locale_core/src/preferences/extensions/unicode/keywords/calendar.rs
```diff
@@ -21,6 +23,10 @@ enum_keyword!(
         Rgsa
 });
 
+fn value_is_ethioaa(v: &Value) -> bool {
+    v.as_subtags_slice() == [subtag!("ethiopic"), subtag!("amete"), subtag!("alem")]
```
robertbastian: nit: inline this now?

components/locale_core/src/preferences/extensions/unicode/keywords/calendar.rs
Manishearth: We don't have a None. This is the current behavior.

components/locale_core/src/preferences/extensions/unicode/keywords/calendar.rs
```diff
@@ -63,4 +69,31 @@ enum_keyword!(
         ("persian" => Persian),
         /// Republic of China calendar
         ("roc" => Roc)
-}, "ca", s, if *s == value!("islamicc") { return Ok(Self::Hijri(Some(HijriCalendarAlgorithm::Civil))); });
+}, "ca", s, if *s == value!("islamicc") { return Ok(Self::Hijri(Some(HijriCalendarAlgorithm::Civil))); } else if value_is_ethioaa(s) { return Ok(Self::Ethioaa) });
```
robertbastian: nit: apply some formatting here

components/locale_core/src/preferences/extensions/unicode/keywords/calendar.rs
robertbastian: We have an `Err` though

components/locale_core/src/preferences/extensions/unicode/keywords/calendar.rs
Manishearth: I'll do better, I'll move the alias handling out completely

components/locale_core/src/preferences/extensions/unicode/keywords/calendar.rs
Manishearth: Oh, sure.

This isn't the current behavior of _any_ of these, though. If we want to support this, I think that's a different change. That needs us to tweak how the macro works.

Currently `buddhist-foobar` also parses to buddhist, etc. This isn't new behavior I introduced in my PR.

Issue #3881, opened 2023-08-17
[Date]TimeFormatter constructors don't take hour12 / hourCycle override


hsivonen:
ECMA-402 has `hour12` and `hourCycle` overrides for the locale's hour cycle. AFAICT, ICU4X `[Date]TimeFormatter` constructors don't have API surface for this override. For ECMA-402 compat, ICU4X should have API surface for this.

hsivonen:
Maybe `hourCycle` is available via `-u-hc-`? Still leaves `hour12` (which happens to be what appears in various preference UIs).

hsivonen:
`hourCycle` is in `icu_datetime::options::components::Bag`.

sffc:
`-u-hc-` works.

I don't think we have `hour12` except that you can mostly get that via `-u-hc-` but you need to choose between h11/h12 or h23/h24 instead of letting the locale choose.

hsivonen:
Letting the locale affect `hour12` expansion to `hourCycle` seems required for ECMA-402 compliance.

AFAICT, by the time the `[Date]TimeFormatter` constructor has done the provider-level resolution of the locale data, there doesn't appear to be a way for the application code to modify the instance to perform `hour12` resolution in application code.

hsivonen:
macOS Ventura and Gnome expose a boolean system pref for this topic, so `hour12` might be relevant to support to be able to honor system preferences. I don't have test code that I'd know would behave according to the system boolean pref semantics. However, from reading the ECMA-402 spec, I'm a bit surprised at how the ECMA-402 semantics are supposed to work. AFAICT, `hour12=false` for `en-US` would resolve to `h24`. Do users actually want that result instead of `h23`? (As a user of the en-US locale for untranslated strings with 24-hour clock enabled, I don't want `h24`. I guess I'll need to make a TODO item of observing the system clock at midnight.)

anba:
> However, from reading the ECMA-402 spec, I'm a bit surprised at how the ECMA-402 semantics are supposed to work.

The current spec is incorrect. There's a PR to fix this &rarr; <https://github.com/tc39/ecma402/pull/758>.

hsivonen:
> > However, from reading the ECMA-402 spec, I'm a bit surprised at how the ECMA-402 semantics are supposed to work.
> 
> The current spec is incorrect. There's a PR to fix this → [tc39/ecma402#758](https://github.com/tc39/ecma402/pull/758).

[...]

PR #7414, opened 2026-01-13
Add Clock12 and Clock24 hour cycles

sffc:
#3881

https://unicode-org.atlassian.net/browse/CLDR-18894

This can be used to implement `hour12` in ECMA-402. CC @hsivonen @nekevss

## Changelog

icu_datetime: Add support for Clock12 and Clock24 hour cycles

icu_locale_core: Add parsing for `-u-hc-c12` and `-u-hc-c24`

### DIFF
--- a/components/datetime/src/provider/fields/symbols.rs
+++ b/components/datetime/src/provider/fields/symbols.rs
@@ -593,6 +593,10 @@ field_type!(
 );
 
 impl Hour {
+    /// Picks the best [`Hour`] given the preferences.
+    ///
+    /// If `-u-hc-h**` is specified, that field is used. Otherwise, the field is selected
+    /// based on the locale, using `-u-hc-c**`, if present, as a hint.
     pub(crate) fn from_prefs(prefs: DateTimeFormatterPreferences) -> Option<Self> {
         let field = match prefs.hour_cycle? {
             HourCycle::H11 => Self::H11,
@@ -603,13 +607,16 @@ impl Hour {
                     .locale_preferences
                     .to_data_locale_region_priority()
                     .region;
-                // TODO(#7415): Test that this stays consistent with CLDR
+                // TODO(#7415): Test that this fallback stays consistent with CLDR.
+                // This is derived from timeData. Currently the only region using
+                // something other than h12/h23 is JP.
                 const JP: Region = region!("JP");
                 match region {
                     Some(JP) => Self::H11,
                     _ => Self::H12,
                 }
             }
+            // Note: ICU4X does not support H24.
             HourCycle::Clock24 => Self::H23,
             _ => unreachable!(),
         };


sffc:
Filed #7415 for a follow-up for testing against CLDR.

sffc:
Ping @unicode-org/icu4x-owners

components/datetime/src/provider/fields/components.rs
```diff
@@ -77,6 +77,9 @@ pub struct Bag {
     pub time_zone_name: Option<TimeZoneName>,
 
     /// An override of the hour cycle.
+    ///
+    /// TODO: This should probably not be the preferences HourCycle type. It directly sets the
```
Manishearth, line 81: question: how can this TODO be stably resolved? Can we file an issue about this?

components/datetime/src/provider/pattern/hour_cycle.rs
```diff
@@ -146,7 +146,12 @@ pub(crate) fn naively_apply_preferences(
                 length,
             }) = item
             {
-                let candidate_field = fields::Hour::from_hour_cycle(hour_cycle);
+                let candidate_field = match hour_cycle {
+                    HourCycle::H11 => fields::Hour::H11,
+                    HourCycle::H12 => fields::Hour::H12,
+                    HourCycle::H23 => fields::Hour::H23,
+                    _ => unreachable!(),
```
Manishearth, line 153: observation: panic in datagen-only code. perfectly ok.

components/datetime/src/provider/fields/symbols.rs
```diff
@@ -590,13 +593,115 @@ field_type!(
...
+                    .to_data_locale_region_priority()
+                    .region;
+                // TODO(#7415): Test that this stays consistent with CLDR
+                const JP: Region = region!("JP");
+                match region {
+                    Some(JP) => Self::H11,
+                    _ => Self::H12,
+                }
+            }
+            HourCycle::Clock24 => Self::H23,
```
Manishearth, line 620: nit: comment that H24 is basically deprecated (or something)

components/datetime/src/provider/fields/symbols.rs
```diff
@@ -590,13 +593,115 @@ field_type!(
...
             HourCycle::H12 => Self::H12,
             HourCycle::H23 => Self::H23,
+            HourCycle::Clock12 => {
+                let region = prefs
+                    .locale_preferences
+                    .to_data_locale_region_priority()
+                    .region;
+                // TODO(#7415): Test that this stays consistent with CLDR
+                const JP: Region = region!("JP");
+                match region {
```
Manishearth, line 614: nit: comment explaining that we maintain a manual table here for $reasons, and link to the data source

components/datetime/src/provider/fields/components.rs
sffc: This is in the `provider` module which is unstable. It's not really a thing that can be tracked in an issue because it's not really actionable. It is a note for anyone who wishes to rearchitect the classical components code.

Manishearth: Can this be rewritten so that that's clearer? TODO sounds like an actionable issue.

"If this code is being rearchitected, consider ..."

robertbastian, line 81: nit: this should not be a doc comment

PR #7416, opened 2026-01-13
Correctly using AppendItems for date/time/zone glue patterns

cloud-machine-111:
Logic to handle #7107!

Summary copied from the issue's thread:

- Added append_items field to Dates struct in ca.rs (nested within DateTimeFormats) so that the corresponding field can be parsed from the CLDR's calendar data JSONs. append_items is currently a hashmap to prioritize flexible key parsing.
- Updated datetimepattern_convert function in names.rs to use locale-specific "Timezone" append pattern instead of hardcoded strings (e.g. "{0} {2}".)
- The original hardcoded strings are used as fallbacks if the "Timezone" key isn't found in data.date_time_formats.append_items.

## Changelog

icu_datetime: Use appropriate date/time/zone glue patterns based on appendItems


provider/source/src/datetime/names.rs
```diff
@@ -505,12 +505,42 @@ fn datetimepattern_convert(
...
-            "{1} {2}"
+            let formatted_date = match length {
+                PatternLength::Long => &data.date_formats.long,
+                PatternLength::Medium => &data.date_formats.medium,
+                PatternLength::Short => &data.date_formats.short,
+            };
+
+            let pattern = pattern_anchor.insert(formatted_date.get_pattern().to_string());
+
+            let zone_fallback = String::from("{1} {2}"); // this was the previous fallback string.
```
robertbastian: don't allocate a `String` when a `&'static str` does the job

provider/source/src/datetime/names.rs
```diff
@@ -505,12 +505,42 @@ fn datetimepattern_convert(
...
+            let zone_pattern = data
```
sffc: I think the appendItems use `{0}` for "the thing to append something to" and `{1}` for "the thing being appended". You need to change that to our meaning, `{0}` for "date", `{1}` for "time", and `{2}` for "time zone".

+1 about running cargo make testdata or adding some other tests. I expect this to impact data but I can't see it without tests.

provider/data/datetime/fingerprints.csv
```diff
@@ -24724,6 +24768,15 @@ datetime/patterns/glue/v1, ms/mdt, -> ak/mdt
 datetime/patterns/glue/v1, ms/mdtz, -> ak/mdtz
 datetime/patterns/glue/v1, ms/sdt, -> ak/mdt
 datetime/patterns/glue/v1, ms/sdtz, -> ak/mdtz
+datetime/patterns/glue/v1, my/ldtz, 39B, 16B, 286f33256e782563
```
sffc, line 24771: Note: the default is

```
						<appendItem request="Timezone">{0} {1}</appendItem>
```

and `my` overrides it to be

```
					<appendItems>
						<appendItem request="Timezone">{1} {0}</appendItem>
					</appendItems>
```

(time zone at start rather than end)

So this seems like it is compliant with CLDR data.

provider/data/datetime/fingerprints.csv
```diff
@@ -24549,7 +24576,15 @@ datetime/patterns/glue/v1, hi/mdtz, -> ak/mdtz
 datetime/patterns/glue/v1, hi/sdt, -> ak/mdt
 datetime/patterns/glue/v1, hi/sdtz, -> ak/mdtz
 datetime/patterns/glue/v1, hr/ldt, -> bs/ldt
-datetime/patterns/glue/v1, hr/ldtz, -> bs/ldtz
+datetime/patterns/glue/v1, hr/ldtz, 51B, 28B, 4946bc1cf6992ea2
```
sffc, line 24579: Note: the pattern here is

```
					<appendItems>
						<appendItem request="Timezone">{0} ({1})</appendItem>
					</appendItems>
```

which is not the default, so it makes sense that there is a new data payload for this locale.

provider/source/src/datetime/names.rs
```diff
@@ -496,35 +496,25 @@ fn datetimepattern_convert(
...
+    // Note: We default to atTime here (See https://github.com/unicode-org/conformance/issues/469)
+    let at_time = data
+        .datetime_formats_at_time
+        .get_pattern(length)
+        .get_pattern();
 
-    let pattern = pattern.parse().expect("failed to parse pattern");
+    let pattern = match glue_type {
+        GlueType::DateTime => at_time.parse(),
+        GlueType::DateZone => append_tz.interpolate_to_string(["{1}", "{2}"]).parse(),
```
sffc, line 513: Praise: Clever way to use the Pattern API to change the placeholder numbers.

sffc:
It would be nice to add a test for this in a locale with a non-default time zone appendItem pattern, but I'm convinced that this is the correct bug fix so you can add the test in a follow-up. I'll merge now so it doesn't bitrot.

provider/source/src/cldr_serde/ca.rs
```diff
@@ -190,6 +190,35 @@ pub(crate) struct DateTimeFormats {
...
     pub(crate) available_formats: AvailableFormats,
+    #[serde(rename = "appendItems")]
+    pub(crate) append_items: AppendItems,
+}
+
+#[derive(PartialEq, Debug, Deserialize, Clone)]
+#[allow(dead_code)]
+pub struct AppendItems {
+    #[serde(rename = "Day")]
+    pub(crate) day: String,
```
sffc, line 201: Suggestion (optional): use a `Pattern` here instead of a `String` to avoid the parsing/allocation later on.

I'd like you to do this in a follow-up since the `Pattern` API may or may not be easy to use with `Deserialize` and I don't want to block this PR from landing.

cloud-machine-111:
> It would be nice to add a test for this in a locale with a non-default time zone appendItem pattern, but I'm convinced that this is the correct bug fix so you can add the test in a follow-up. I'll merge now so it doesn't bitrot.

Following up, if I were to write a test for nondefault AppendItem patterns, would it be appropriate to import 1-2 new locales (e.g. 'my', 'zh') into the test provider (`provider/source/src/tests/data.rs`)? It seems from [documentation](https://github.com/unicode-org/icu4x/blob/main/provider/source/tests/locales.rs.data) that the number of imported test locales is kept small intentionally, but currently the test provider doesn't import data from any locales whose [fingerprints change](https://github.com/unicode-org/icu4x/pull/7416/changes/BASE..6f9aecd14d3ad5bce16325da2a4b3c43763a2d7a) after the AppendItems update. Please let me know if there's another preferred approach!

Locales that changed in the diff: cv, eu, fa_AF, fa, hr, mn, my, sq, und, zh

sffc:
Please add zh for only the datetime component. You can include the data files you need here:

https://github.com/unicode-org/icu4x/blob/main/tools/make/download-repo-sources/globs.rs.data

sffc:
Although actually if you add the test in `icu_datetime`, you already have all the locales available. You only need to add it to globs.rs.data if you are testing specific datagen behavior. I just wanted to see evidence of these patterns being used in practice, so a test that utilizes these patterns in `icu_datetime` is sufficient.

cloud-machine-111:
> Although actually if you add the test in `icu_datetime`, you already have all the locales available. You only need to add it to globs.rs.data if you are testing specific datagen behavior. I just wanted to see evidence of these patterns being used in practice, so a test that utilizes these patterns in `icu_datetime` is sufficient.

Ah, I see why. Thank you :)

PR #7423, opened 2026-01-14
refactor: consolidate region extraction logic in SourceDataProvider

younies:
# Description

This commit moves the existing `extract_region` method and renames it to `extract_or_infer_region` to streamline region extraction from `DataLocale`.

> NOTE
> The method can also be reused by currency or other data providers that require region information or region inference.

### DIFF
--- a/provider/source/src/cldr_cache.rs
+++ b/provider/source/src/cldr_cache.rs
@@ -247,11 +247,7 @@ impl CldrCache {
 
         let mut lang_id = LanguageIdentifier::from((locale.language, locale.script, locale.region));
         let _ = self.extended_locale_expander()?.maximize(&mut lang_id);
-        lang_id.region.ok_or_else(|| {
-            DataErrorKind::InvalidRequest
-                .into_error()
-                .with_debug_context(&lang_id)
-        })
+        Ok(lang_id.region.unwrap())
     }
 
     /// Computes the likely script-based locale group for a given locale.


provider/source/src/cldr_cache.rs
```diff
@@ -226,6 +230,62 @@ impl CldrCache {
...
+    /// # Example
+    ///  - "en-US" -> "US"
+    ///  - "en" -> "US"
+    #[cfg(feature = "experimental")]
+    pub(crate) fn extract_or_infer_region(&self, locale: &DataLocale) -> Result<Region, DataError> {
+        if let Some(region) = locale.region {
+            return Ok(region);
+        }
+
+        let mut lang_id = LanguageIdentifier::from(locale.language);
```
robertbastian: issue: you need to consider the script, so that `zh-Hant` resolves to `TW`, not to `CN`

provider/source/src/cldr_cache.rs
```diff
@@ -226,6 +230,62 @@ impl CldrCache {
...
+        let _ = self.extended_locale_expander()?.maximize(&mut lang_id);
+        lang_id.region.ok_or_else(|| {
+            DataErrorKind::InvalidRequest
+                .into_error()
+                .with_debug_context(&lang_id)
+        })
```
robertbastian: it's fine to `unwrap` because `maximize` guarantees that there's a region

provider/source/src/cldr_cache.rs
younies: very good note,  done

provider/source/src/cldr_cache.rs
younies: I agree with you and I prefer having a layer of security if `maximize` has any failure in the future.

robertbastian: it's not a layer of security. you'll get a `DataError` when you unwrap this later, which is going to be very hard to track down to this function, and is unactionable for clients. an unwrap here gives you all the context

younies: okay, done

robertbastian (approved):

provider/source/src/cldr_cache.rs
```diff
@@ -227,6 +231,25 @@ impl CldrCache {
...
+    /// # Example
+    ///  - "en-US" -> "US"
+    ///  - "en" -> "US"
+    #[cfg(feature = "experimental")]
+    pub(crate) fn extract_or_infer_region(&self, locale: &DataLocale) -> Result<Region, DataError> {
+        if let Some(region) = locale.region {
+            return Ok(region);
+        }
+
+        let mut lang_id = LanguageIdentifier::from((locale.language, locale.script, locale.region));
```
sffc, line 248: Issue: use `locale.id` to get the LanguageIdentifier.

provider/source/src/cldr_cache.rs
```diff
@@ -227,6 +231,25 @@ impl CldrCache {
...
+        let _ = self.extended_locale_expander()?.maximize(&mut lang_id);
+        Ok(lang_id.region.unwrap())
```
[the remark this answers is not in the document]
sffc: +1, I think it's possible for expand to not return a region if there isn't data for it.

This should be uncommon, so if you want to panic, it's _fine_, but probably at least include a message with the faulty locale to help make debugging easier.

provider/source/src/cldr_cache.rs
robertbastian: `DataLocale` doesn't have an `id` field

provider/source/src/cldr_cache.rs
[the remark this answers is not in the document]
younies: we were discussing this here: https://github.com/unicode-org/icu4x/pull/7423#discussion_r2691663258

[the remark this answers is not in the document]
younies: we can use also the gemini suggestion too

younies:
@sffc : I have merged the PR because I need the method in my next PR, if you are insisting on replacing the `unwrap` with a `DataError`, I can do it in another PR

provider/source/src/cldr_cache.rs
sffc: ok, then call `Locale::from` and then get the `id` field?

PR #7425, opened 2026-01-14
FIx overflow in `PluralOperands` constructor

robertbastian:
Split from #7411


sffc (approved):
I see, so previously it was possible to have exp_i16 be like -100 which would cause mag_high to be say -100 while mag_low would be -18 which is garbage.

robertbastian:
no, the input exponent is unsigned.

PR #7429, opened 2026-01-15
Organize neo datetime provider code

Manishearth:
Fixes https://github.com/unicode-org/icu4x/issues/5613

I used the same organization as in https://github.com/unicode-org/icu4x/pull/7386: `names` and `semantic_skeleta`. Glue patterns go in `names` for now, but they're easy to move.

## Changelog (N/A)

Refactoring of internal provider code.


sffc (commented):
The stuff in components/datetime/src/provider/semantic_skeletons.rs is not really about semantic skeleta. It's just a bunch of data marker attributes, for looking up names as well as patterns. I think the file is a misnomer.

Glue patterns definitely don't belong in a file named names.rs. It's about as far as you can get away from names in datetime.

Manishearth:
Pure rebase over Rob's PR, no changes.

@sffc The glue key is a minor thing that can be moved later, I did this to be consistent with the choice we made in provider/source. We could just have glue.rs in both, in provider/source the code was very closely related so it stayed together.

Happy to rename semantic_skeleton.rs if you have a suggestion. Things can also be further organized.

I have a preference for merging this as is since this _splits_ the module. After a split like this, further renames are much easier to do.

sffc:
Needs another rebase.

Manishearth:
Done

components/datetime/src/provider/mod.rs
```diff
@@ -137,34 +138,34 @@ pub const MARKERS: &[DataMarkerInfo] = &[
     time_zones::TimezoneNamesSpecificLongV1::INFO,
     time_zones::TimezoneNamesSpecificShortV1::INFO,
     time_zones::TimezoneNamesEssentialsV1::INFO,
-    neo::DatetimeNamesWeekdayV1::INFO,
-    neo::DatetimeNamesDayperiodV1::INFO,
-    neo::DatetimePatternsGlueV1::INFO,
+    names::DatetimeNamesWeekdayV1::INFO,
+    names::DatetimeNamesDayperiodV1::INFO,
+    names::DatetimePatternsGlueV1::INFO,
```
sffc, line 143: I thought we decided to not put `DatetimePatternsGlueV1` in the `names` namespace

Manishearth: I haven't had a workday since your feedback.

Please refer to my previous comment about wishing to land this so I can be incremental about this: splitting the file is still a large step and makes further changes easier.

Manishearth: Also, we didn't "decide", you provided feedback and I gave a reason why I went with that choice, and we've not discussed it further.

I am happy to change it I just haven't really thought about where it ought to go. Suggestions welcome.

Manishearth: Also, we didn't "decide", you provided feedback and I gave a reason why I went with that choice, and we've not discussed it further.

I am happy to change it I just haven't really thought about where it ought to go. Suggestions welcome.

PR #7433, opened 2026-01-15
Add more ULE `const` constructors for integers and floats

robertbastian:
Split from #7323 

## Changelog

zerovec: Add more `const` constructors for converting ULE types to integers and floats
  - New methods: `RawBytesULE::as_signed_int(), as_float()`

### DIFF
--- a/utils/zerovec/src/ule/plain.rs
+++ b/utils/zerovec/src/ule/plain.rs
@@ -60,8 +60,10 @@ impl<const N: usize> From<[u8; N]> for RawBytesULE<N> {
 }
 
 macro_rules! impl_numbers_with_raw_bytes_ule {
-    ($unsigned:ty, $signed:ty, $($float:ty,)? $size:literal) => {
-        impl RawBytesULE<$size> {
+    ($unsigned:ty, $signed:ty $(, $float:ty)?) => {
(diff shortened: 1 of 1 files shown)


utils/zerovec/src/ule/plain.rs
```diff
@@ -208,6 +208,8 @@ impl AsULE for u8 {
 // EqULE is true because u8 is its own ULE.
 unsafe impl EqULE for u8 {}
 
+impl_const_constructors!(u8, 1);
```
Manishearth: thought: i bet we can get rid of this size parameter and use const stuff, eventually. Would make these macros a bit more safe.

robertbastian: done

Manishearth (approved):

robertbastian:
I don't see the issue. `RawBytesULE::<n>::as_signed` and `RawBytesULE::<n>::as_float` have concrete return types and very clear names. How does one accidentally "get back the wrong type"?

Also why is unsigned the "safest" type?!

Manishearth:
The APIs seem quite clear to me. I understand that one might accidentally convert to a RBU from unsigned and convert to signed, forgetting the original, but that's already true in the other direction. This is a lower level API, people shouldn't be using it that much without taking care.

sffc:
if you have a use case for these additional functions, it's _fine_

PR #7436, opened 2026-01-15
Add UCD source to datagen

opnuub:
@sffc 
Modified:
- provider/icu4x-datagen/src/main.rs
- provider/source/src/lib.rs
- provider/source/src/segmenter/mod.rs (to import unihan.rs)
- provider/source/src/segmenter/unihan.rs (to use LiteMap<codepoint, int8> via test)
- provider/source/src/source.rs (added UnihanCache with proprocessing)
- tools/make/download-repo-sources/globs.rs.data (added filename IRGSources.txt)
- tools/make/download-repo-sources/src/main.rs


tools/make/download-repo-sources/src/main.rs
```diff
@@ -254,6 +269,25 @@ fn main() -> eyre::Result<()> {
         })
         .collect::<Vec<_>>()
         .join(",\n                        ");
+    #[allow(dead_code)]
```
sffc: Shouldn't be dead code?

```suggestion
```

tools/make/download-repo-sources/src/main.rs
```diff
@@ -254,6 +269,25 @@ fn main() -> eyre::Result<()> {
...
+            let path = path.replace('\\', "/");
+            format!(r#"("{path}", include_bytes!("../../tests/data/unihan/{path}").as_slice())"#)
+        })
+        .collect::<Vec<_>>()
+        .join(",\n                        ");
+    let irg_path = out_root.join("tests/data/unihan/Unihan_IRGSources.txt");
+    fs::write(
+        &irg_path,
+        fs::read_to_string(&irg_path)?
+            .lines()
```
sffc: Nit: Since this file is very big, do the more efficient line-by-line reading and writing, as described in

https://doc.rust-lang.org/rust-by-example/std_misc/file/read_lines.html

For writing, you can write each line to the output file using `writeln!`

tools/make/download-repo-sources/src/main.rs
```diff
@@ -254,6 +269,25 @@ fn main() -> eyre::Result<()> {
...
+            .filter(|l| l.contains("kRSUnicode") || l.starts_with('#'))
```
sffc: Praise: The filtering for just kRSUnicode seems fine. It keeps the file manageable in size.

provider/source/src/source.rs
```diff
@@ -484,3 +485,52 @@ impl TzdbCache {
             .map_err(|&e| e)
     }
 }
+
+#[derive(Debug)]
+#[allow(dead_code)]
```
sffc, line 491: Please remove all dead_code attributes; they shouldn't be necessary

provider/source/src/source.rs
```diff
@@ -484,3 +485,52 @@ impl TzdbCache {
...
+                candidate = first_part;
+            }
+            let radical_str = if let Some(idx) = candidate.find('.') {
+                &candidate[..idx]
+            } else {
+                candidate
+            };
+            let clean_str = radical_str.replace('\'', "");
+            if let Ok(value) = clean_str.parse::<u8>() {
+                map.insert(codepoint, IRGValue { value });
```
sffc: Suggestion: use `map.try_append()` instead, since I think these should be in ascending order of code point value. It is much more efficient than `map.insert()`.

provider/source/src/source.rs
```diff
@@ -484,3 +485,52 @@ impl TzdbCache {
...
+#[derive(Debug)]
+pub(crate) struct UnihanCache {
+    #[allow(dead_code)]
+    pub(crate) root: AbstractFs,
+}
+
+impl UnihanCache {
+    #[allow(dead_code)]
+    pub(crate) fn irg_sources(&self) -> Result<LiteMap<char, IRGValue>, DataError> {
+        let raw_content = self.root.read_to_string("Unihan_IRGSources.txt")?;
```
sffc: Issue: You should add a `OnceLock` field and use `get_or_init` like the other cache functions, in order to avoid computing this map multiple times.

provider/source/src/source.rs
```diff
@@ -484,3 +485,52 @@ impl TzdbCache {
...
+            let mut candidate = parts[2].trim();
+            if let Some(first_part) = candidate.split_whitespace().next() {
+                candidate = first_part;
+            }
+            let radical_str = if let Some(idx) = candidate.find('.') {
+                &candidate[..idx]
+            } else {
+                candidate
+            };
+            let clean_str = radical_str.replace('\'', "");
```
sffc: Question: What do the quotes mean? In general, how should this type of data be interpreted:

```
U+4333	kRSUnicode	120.19
U+4334	kRSUnicode	120.19
U+4335	kRSUnicode	120.21
U+4336	kRSUnicode	120'.3
U+4337	kRSUnicode	120'.5
U+4338	kRSUnicode	120'.4
U+4339	kRSUnicode	120'.5
U+433A	kRSUnicode	120'.6
U+433B	kRSUnicode	120'.6
U+433C	kRSUnicode	120'.7
U+433D	kRSUnicode	120'.8
U+433E	kRSUnicode	120'.9
U+433F	kRSUnicode	120'.9
U+4340	kRSUnicode	120'.10
U+4341	kRSUnicode	120'.13
U+4342	kRSUnicode	121.3
U+4343	kRSUnicode	121.4
U+4344	kRSUnicode	121.5
U+4345	kRSUnicode	121.5
U+4346	kRSUnicode	121.5
U+4347	kRSUnicode	121.5 121.4
U+4348	kRSUnicode	121.5
U+4349	kRSUnicode	121.5
```

provider/source/src/source.rs
```diff
@@ -484,3 +485,52 @@ impl TzdbCache {
...
+            }
+            let parts: Vec<&str> = line.trim().split('\t').collect();
+            if parts[1] != "kRSUnicode" {
+                continue;
+            }
+            let codepoint = parts[0]
+                .strip_prefix("U+")
+                .and_then(|hex| u32::from_str_radix(hex, 16).ok())
+                .and_then(char::from_u32)
+                .unwrap_or('\u{4E00}');
```
sffc: Nit: I think you should be more strict. If the file isn't the format you expect, either return a `DataError` or panic. It's datagen so panicking is okay.

provider/source/src/segmenter/unihan.rs
```diff
@@ -0,0 +1,36 @@
...
+
+        // let sentence = "我爱中文";
+
+        // for char in sentence.chars() {
+        //     if let Some(irg_value) = irg_map.get(&char) {
+        //         println!("Char: {} (U+{:04X}) => IRG Value: {:.1}", char, char as u32, irg_value.value);
+        //     } else {
+        //         println!("Char: {} (U+{:04X}) => Not found in IRG sources", char, char as u32);
+        //     }
+        // }
```
sffc: Nit: Delete old commented code that you don't plan to use.

provider/source/src/lib.rs
```diff
@@ -268,6 +283,18 @@ impl SourceDataProvider {
         }
     }
 
+    #[cfg(feature = "networking")]
+    pub fn with_unihan_for_tag(self, tag: &str) -> Self {
```
sffc, line 309: Thought: Not sure if it should be called "tag" or something else like "release" or "version". I sent an email.

provider/source/src/lib.rs
```diff
@@ -118,6 +120,8 @@ impl SourceDataProvider {
     /// The segmentation LSTM model tag that has been verified to work with this version of `SourceDataProvider`.
     pub const TESTED_SEGMENTER_LSTM_TAG: &'static str = "v0.1.0";
 
+    pub const TESTED_UNIHAN_TAG: &'static str = "latest";
```
sffc: Issue: This should be a stable version number such as 16.0.0

provider/source/src/lib.rs
```diff
@@ -268,6 +283,18 @@ impl SourceDataProvider {
         }
     }
 
+    #[cfg(feature = "networking")]
+    pub fn with_unihan_for_tag(self, tag: &str) -> Self {
+        Self {
+            unihan_paths: Some(Arc::new(UnihanCache {
+                root: AbstractFs::new_from_url(format!(
+                    "https://www.unicode.org/Public/UCD/{tag}/ucd/Unihan.zip"
```
sffc: I think the URL should not have the extra "UCD" in it, because we want the directory where you can put in all the different version tags.

provider/icu4x-datagen/src/main.rs
```diff
@@ -159,6 +159,17 @@ struct Cli {
     #[cfg(feature = "provider")]
     icuexport_root: Option<PathBuf>,
 
+    #[arg(long, value_name = "TAG", default_value = "latest")]
+    #[arg(help = "Download Unihan data from unicode.org.")]
+    #[cfg_attr(not(feature = "networking"), arg(hide = true))]
+    #[cfg(feature = "provider")]
+    unihan_tag: String,
```
robertbastian: thought: should this be `unicode_tag`? Unihan is versioned with Unicode

provider/source/src/lib.rs
robertbastian: this needs to be:
```suggestion
                    "https://unicode.org/Public/{tag}/ucd/Unihan.zip"
```
your URL only works with `latest`

provider/source/src/source.rs
```diff
@@ -484,3 +485,52 @@ impl TzdbCache {
...
+
+#[derive(Debug)]
+pub(crate) struct UnihanCache {
+    #[allow(dead_code)]
+    pub(crate) root: AbstractFs,
+}
+
+impl UnihanCache {
+    #[allow(dead_code)]
+    pub(crate) fn irg_sources(&self) -> Result<LiteMap<char, IRGValue>, DataError> {
```
robertbastian: issue: LiteMap is nice for small code size and no-std environments, but it has a limited API and is probably less performant than std maps. in `icu_provider_source`, we usually use `HashMap`, and `BTreeMap` when sorted iteration is important.

sffc: I suggested LiteMap because the file is sorted in ascending code point order so it's efficient to build a LiteMap from it. (see other comment about using `try_append` instead of `insert`)

provider/icu4x-datagen/src/main.rs
sffc: I was thinking that Unihan and UCD would be two different sources, since Unihan is an extra zip file separate from the UCD.

I have an email thread about this, too.

robertbastian: two different zip files yes, but we probably don't want to allow version skew

sffc: Yeah `--ucd-tag` is probably the right thing to do, and it'll be used for both data sources (Unihan and a future non-Unihan UCD).

But, we should keep `--unihan-root`.

provider/source/src/source.rs
opnuub: I think because only the test is running these functions so for now I'll have to use dead code to suppress these warnings

provider/source/src/source.rs
opnuub: Quotes mean that they considered the same radical ID, (120' and 120 are referring to the same thing), but one is the simplified version and the other is the traditional version. By the model's design both should be deemed the same

PR #7439, opened 2026-01-16
Add tests for low-frequency suffixes in Dense ZeroTrie

IshaanXCoder:
This PR addresses #7302 

Implements suffix-frequency pruning to exclude rare suffixes from the dense matrix representation of ZeroAsciiDenseSparse2dTrie.

### DIFF
--- a/utils/zerotrie/tests/dense_test.rs
+++ b/utils/zerotrie/tests/dense_test.rs
@@ -82,12 +82,6 @@ fn test_basic() {
     let data_as_map = strings_to_2d_btreemap(BASIC_DATA, "/");
     let dense =
         ZeroAsciiDenseSparse2dTrieOwned::try_from_btree_map_str(&data_as_map, b'/').unwrap();
-
-    assert_eq!(
-        format!("{:?}", dense.as_borrowed()),
-        "ZeroAsciiDenseSparse2dTrieBorrowed { primary: ZeroTrieSimpleAscii { store: [196, 97, 101, 102, 105, 4, 8, 12, 114, 128, 47, 128, 110, 129, 47, 133, 114, 130, 47, 137, 116, 47, 73, 84, 141] }, suffixes: ZeroTrieSimpleAscii { store: [197, 65, 70, 73, 83, 85, 2, 4, 10, 12, 85, 128, 82, 129, 194, 82, 84, 1, 130, 131, 65, 132, 194, 75, 83, 1, 133, 134] }, dense: ZeroVec([65535, 0, 1, 65535, 2, 3, 4, 0, 1, 65535, 65535, 65535, 2, 3, 65535, 0, 65535, 65535, 1, 2, 3]), suffix_count: 7, delimiter: 47 }"
-    );
-
     check_data(&data_as_map, dense.as_borrowed(), true);
     let non_existent_data_as_map = strings_to_2d_btreemap(NON_EXISTENT_BASIC_DATA, "/");
     check_data(&non_existent_data_as_map, dense.as_borrowed(), false);


utils/zerotrie/src/dense.rs
```diff
@@ -171,6 +171,9 @@ impl<'a> ZeroAsciiDenseSparse2dTrieBorrowed<'a> {
             cursor.step(*b);
         }
         let row_index = cursor.take_value();
+        if cursor.is_empty() {
+            return row_index;
+        }
```
sffc: Issue: This is an unnecessary change that I believe is wrong.

utils/zerotrie/src/dense.rs
```diff
@@ -193,16 +196,21 @@ impl<'a> ZeroAsciiDenseSparse2dTrieBorrowed<'a> {
             return None;
         };
         let suffix_count = usize::from(self.suffix_count);
-        let Some(offset) = self.dense.get(suffix_count * row_index + column_index) else {
-            // The row and column indexes should be in-range
-            debug_assert!(false);
-            return None;
+        let index = row_index
```
sffc: Nit: Using checked_add and checked_mul is beneficial, but not related to the subject of this PR. Please split those changes to a separate PR.

utils/zerotrie/src/builder/dense.rs
```diff
@@ -185,24 +188,54 @@ impl ZeroAsciiDenseSparse2dTrieOwned {
...
+            .saturating_add(PERCENT_DENOMINATOR - 1)
+            .saturating_div(PERCENT_DENOMINATOR);
+        let min_prefixes = core::cmp::max(2, computed_min);
+        let mut dense_candidates: Vec<(&str, usize)> = suffix_prefix_count
+            .iter()
+            .filter(|(_, &count)| count >= min_prefixes)
+            .map(|(&suffix, &count)| (suffix, count))
+            .collect();
+
+        // If none meet the threshold, fallback to picking top-K by frequency.
```
sffc: If none meet the threshold, then I think we should just leave the dense matrix empty.

IshaanXCoder:
Hey @sffc thanks for the review, i've tested the code (didnt push the tests)

utils/zerotrie/src/dense.rs
IshaanXCoder: Alright I'll remove this

IshaanXCoder:
Hey @sffc , i've commited the changes, PTAL.

utils/zerotrie/src/dense.rs
IshaanXCoder: sure, i've created #7442 , working their on this.

sffc:
CC @s3arthak who previously worked on this in #7307

utils/zerotrie/src/dense.rs
```diff
@@ -193,16 +193,20 @@ impl<'a> ZeroAsciiDenseSparse2dTrieBorrowed<'a> {
...
```
sffc: Please remove the checked math from this PR. You should decouple these PRs.

utils/zerotrie/tests/dense_suffix_filter_test.rs
```diff
@@ -0,0 +1,149 @@
+// This file is part of ICU4X. For terms of use, please see the file
```
sffc: You added the tests to dense_test.rs but didn't remove dense_suffix_filter_test.rs

utils/zerotrie/tests/dense_suffix_filter_test.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    // Non-existent lookups
+    assert_eq!(trie.get("en", "FR"), None);
+    assert_eq!(trie.get("nonexistent", "US"), None);
+}
+
+#[test]
+fn test_empty_dense_when_no_suffix_meets_threshold() {
+    // Create a dataset where NO suffix appears in enough prefixes to meet threshold.
+    // Dense matrix should be empty, all lookups use sparse path.
```
sffc: As suggested in https://github.com/unicode-org/icu4x/pull/7443#discussion_r2702663225:

> Suggestion: assert the `Debug` output of the trie, which encodes the values that went into sparse and the values that went into dense.

IshaanXCoder:
Hey @sffc , made the changes as per the review, PTAL.

utils/zerotrie/src/dense.rs
```diff
@@ -193,16 +193,16 @@ impl<'a> ZeroAsciiDenseSparse2dTrieBorrowed<'a> {
...
+        let Some(offset) = self.dense.get(index) else {
             debug_assert!(false);
             return None;
         };
         if offset == DenseType::MAX {
             // There is an entry in the dense matrix but it is a None value
             return None;
         }
-        Some(usize::from(offset) + row_value_offset)
+        usize::from(offset).checked_add(row_value_offset)
```
sffc: Again, please don't change this file in this PR. This PR should only be changing the builder, as well as adding tests.

utils/zerotrie/tests/dense_test.rs
```diff
@@ -175,3 +175,127 @@ fn test_dense_sparse_window_selection() {
...
+    // Each suffix appears in only one prefix (0% overlap)
+    data.entry("a").or_default().insert("s1", 1);
+    data.entry("b").or_default().insert("s2", 2);
+    data.entry("c").or_default().insert("s3", 3);
+
+    let trie = ZeroAsciiDenseSparse2dTrieOwned::try_from_btree_map_str(&data, b'/').unwrap();
+
+    assert_eq!(
+        format!("{:?}", trie.as_borrowed()),
+        "ZeroAsciiDenseSparse2dTrieBorrowed { primary: ZeroTrieSimpleAscii { store: [195, 97, 98, 99, 4, 8, 47, 115, 49, 129, 47, 115, 50, 130, 47, 115, 51, 131] }, suffixes: ZeroTrieSimpleAscii { store: [115, 195, 49, 50, 51, 1, 2, 128, 129, 130] }, dense: ZeroVec([]), suffix_count: 3, delimiter: 47 }"
```
sffc, line 246: This debug impl test is good since it shows that everything goes into the suffixes trie. Please add the debug impl test to the other cases in order to assert how many things go to dense vs sparse.

IshaanXCoder: Alright

utils/zerotrie/src/dense.rs
```diff
@@ -193,8 +193,8 @@ impl<'a> ZeroAsciiDenseSparse2dTrieBorrowed<'a> {
             return None;
         };
         let suffix_count = usize::from(self.suffix_count);
-        let Some(offset) = self.dense.get(suffix_count * row_index + column_index) else {
-            // The row and column indexes should be in-range
+        let index = row_index * suffix_count + column_index;
+        let Some(offset) = self.dense.get(index) else {
```
sffc: Sorry, why do you keep changing this file?

IshaanXCoder: I was trying out something else, mistakenly pushed this. reverting it, really sorry

IshaanXCoder: Yes, reverted it, my bad!

IshaanXCoder:
Sure i'll try to make a PR by tonight :)

IshaanXCoder:
@sffc just to be clear if we're on the same page, 
I'll Implement the suffix-frequency filtering in the builder and write tests for it and make a new PR, right?

Also should i wait to get #7439 #7443 merged and then make a PR to avoid merge conflicts in dense.rs
Other option being i'll start working on it in a new file and once above 2 PRs are merged, i'll add them to dense.rs itself

utils/zerotrie/tests/dense_test.rs
```diff
@@ -82,6 +82,12 @@ fn test_basic() {
     let data_as_map = strings_to_2d_btreemap(BASIC_DATA, "/");
     let dense =
         ZeroAsciiDenseSparse2dTrieOwned::try_from_btree_map_str(&data_as_map, b'/').unwrap();
+
+    assert_eq!(
+        format!("{:?}", dense.as_borrowed()),
```
robertbastian: issue: using the debug string representation for testing is brittle, it's not meant to be stable and makes refactoring a pain

robertbastian (changes requested):

utils/zerotrie/tests/dense_test.rs
IshaanXCoder: Made the commit, are we good to go?

robertbastian: I don't want you to delete the assertion. Can you find another way to check this?

IshaanXCoder: Tried a few ways, but they're leading to failures, the issue is that the fields are private in `ZeroAsciiDenseSparse2dTrieBorrowed`

```
let trie = dense.as_borrowed();

assert_eq!(trie.suffix_count, 7);
assert_eq!(trie.delimiter, b'/');

let num_of_rows = data_as_map.len();

assert_eq!(
    trie.dense.len(),
    num_of_rows * trie.suffix_count as usize,
    "dense matrix size must equal input rows × suffix_count"
);

assert_eq!(
    trie.dense.len() % trie.suffix_count as usize,
    0,
    "dense storage must be divisible by suffix_count"
);

```

To use these, I would need to expose these fields via #[doc(hidden)]

robertbastian: If you make this a unit test instead of an integration test you can access the fields. You are checking internal representation after all, not public API.

IshaanXCoder: Okay i'll write unit tests for all the tests currently in `dense_test.rs`, right? or some specific ones only?

IshaanXCoder:
@sffc are we good to merge this?

sffc (approved):

sffc:
I think Robert's design would be better but not better enough to not land the tests

PR #7440, opened 2026-01-17
Pattern should always be VarULE

sffc:
Follow-up from #7411

## Changelog: N/A

internal to experimental code


components/pattern/src/common.rs
```diff
@@ -83,7 +83,10 @@ pub trait PatternBackend: crate::private::Sealed + 'static + core::fmt::Debug {
 
     /// The unsized type of the store required for this backend, usually `str` or `[u8]`.
     #[doc(hidden)] // TODO(#4467): Should be internal
+    #[cfg(not(feature = "zerovec"))]
     type Store: ?Sized + PartialEq + core::fmt::Debug;
+    #[cfg(feature = "zerovec")]
+    type Store: ?Sized + PartialEq + core::fmt::Debug + zerovec::ule::VarULE;
```
robertbastian, line 90: observation: making a bound tighter based on a feature is not generally ok, but this trait is sealed

components/experimental/src/compactdecimal/provider.rs
```diff
@@ -65,7 +65,7 @@ pub struct CompactPatterns<'a, P: PatternBackend>(
 );
 
 #[cfg(feature = "datagen")]
-impl<'data, P: PatternBackend> serde::Serialize for CompactPatterns<'data, P>
+impl<'data, P: PatternBackend<Store = str>> serde::Serialize for CompactPatterns<'data, P>
```
robertbastian: serde should only need `P::Store: Serialize/Deserialize`

sffc: For Serialize, it looks like it works to write `where Pattern<P>: Serialize`.

For Deserialize, on this code:

```rust
impl<'de, 'data, P: PatternBackend> serde::Deserialize<'de>
    for CompactPatterns<'data, P>
where
    'de: 'data,
    Pattern<P>: serde::Deserialize<'de>,
```

I'm getting this error:

```
error[E0275]: overflow evaluating the requirement `&icu_pattern::Pattern<_>: displaynames::provider::_::_serde::Deserialize<'_>`
  |
  = help: consider increasing the recursion limit by adding a `#![recursion_limit = "256"]` attribute to your crate (`icu_experimental`)
  = note: required for `&icu_pattern::Pattern<_>` to implement `displaynames::provider::_::_serde::Deserialize<'_>`
  = note: 125 redundant requirements hidden
  = note: required for `&icu_pattern::Pattern<_>` to implement `displaynames::provider::_::_serde::Deserialize<'_>`

For more information about this error, try `rustc --explain E0275`.
```

sffc: It works if the bound is on `Box<Pattern<P>>`.

sffc:
I could alternatively remove the requirement that Store always implement VarULE and instead ask clients to write `where Pattern<P>: VarULE`.

sffc:
> I could alternatively remove the requirement that Store always implement VarULE and instead ask clients to write `where Pattern<P>: VarULE`.

I tried doing this but then `where Pattern<P>: zerovec::ule::VarULE` is required everywhere that `CompactPatterns` is referenced, including in derives, which don't currently carry these `where` clauses.

components/experimental/src/compactdecimal/provider.rs
```diff
@@ -78,11 +78,10 @@ where
 }
 
 #[cfg(feature = "serde")]
-impl<'de, 'data, P: PatternBackend<Store = str>> serde::Deserialize<'de>
-    for CompactPatterns<'data, P>
+impl<'de, 'data, P: PatternBackend> serde::Deserialize<'de> for CompactPatterns<'data, P>
 where
     'de: 'data,
-    P::PlaceholderKeyCow<'data>: serde::Deserialize<'de>,
+    alloc::boxed::Box<Pattern<P>>: serde::Deserialize<'de>,
```
robertbastian, line 88: issue: this impl should be able to exist in no-alloc, but that bound won't work. does the bound `&'data Pattern<P>: Deserialize<'de>` work?

sffc: I think these bounds are equivalent. We haven't ever supported serde in no-alloc mode. human-readable allocates, and machine-readable uses the store directly without needing this impl.

robertbastian: We very much support serde in no-alloc mode. `icu_provider_blob` is no-alloc, and e.g. `icu_decimal/serde` is as well. It would be stupid to block future no-alloc efforts on something we can trivially fix here while we're familiar with these bounds.

sffc: `icu_decimal/serde` serializes VarZeroCows which are alloc-gated

```rust
#[cfg(all(feature = "serde", feature = "alloc"))]
impl<'a, 'de: 'a, V: VarULE + ?Sized> serde::Deserialize<'de> for VarZeroCow<'a, V>
where
    Box<V>: serde::Deserialize<'de>,
```

some intermediate dependency is probably turning on zerovec/alloc.

robertbastian: huh, they shouldn't be.

robertbastian (commented):
please merge main, there's a `Default` impl now as well

components/experimental/src/compactdecimal/provider.rs
sffc: I mean, most human-readable deserialization requires alloc, and we aren't cfging the impls to handle human readable only when an alloc feature is turned on. We rely on DCE to eliminate the human readable code.

components/experimental/src/compactdecimal/provider.rs
```diff
@@ -64,7 +64,7 @@ pub struct CompactPatterns<'a, P: PatternBackend>(
     pub VarZeroVec<'a, VarTupleULE<u8, PluralElementsPackedULE<Pattern<P>>>>,
 );
 
-impl<P: PatternBackend<Store = str>> Default for CompactPatterns<'_, P> {
+impl<P: PatternBackend> Default for CompactPatterns<'_, P> {
```
robertbastian, line 65: is this derivable now?

components/experimental/src/compactdecimal/provider.rs
sffc: Please note that

```rust
alloc::boxed::Box<Pattern<P>>: serde::Deserialize<'de>,
```

requires

```rust
    B: PatternBackend<Store = str>,
    B::PlaceholderKeyCow<'data>: Deserialize<'de>,
```

so this is a no-op. It's just putting the bound on the thing that actually matters.

robertbastian: yes, but the latter doesn't use an `alloc` type, and could be supported in no-alloc in the future in an `icu_pattern`-only change. any bound involving `Box` inherently can't, so we'll have to update this bound then.

components/experimental/src/compactdecimal/provider.rs
sffc: not sure. I think it would result in a `P: Default` bound which is not correct.

components/experimental/src/compactdecimal/provider.rs
sffc: But, if the `icu_pattern` impl is changed in such a way that it no longer requires these two bounds, such as relaxing the constraint on `Store = str`, this crate won't automatically benefit from the change.

sffc:
CC @Manishearth because there might be stuff in here you'd like to see, but maybe not, so I'll merge tomorrow unless I hear otherwise

components/experimental/src/compactdecimal/provider.rs
Manishearth: At the moment `Deserialize` requires alloc, I think. We could in theory make it no alloc but having shifting trait bounds like that is messy. We are nowhere close to achieving that.

PR #7443, opened 2026-01-17
Use checked arithmetic in dense ZeroTrie

### LINKED ISSUE #7442: Add checked arithmetic to prevent integer overflow in dense trie lookups
## Description

The `ZeroAsciiDenseSparse2dTrieBorrowed::get()` method currently uses unchecked arithmetic for matrix index calculations and value additions. This could potentially cause integer overflow on edge cases or with malformed data.

### Current Code

**File**: [utils/zerotrie/src/dense.rs](file:///Users/ishaan/Developer/icu4x/utils/zerotrie/src/dense.rs)

```rust
let suffix_count = usize::from(self.suffix_count);
let Some(offset) = self.dense.get(suffix_count * row_index + column_index) else {
    // The row and column indexes should be in-range
    debug_assert!(false);
    return None;
};
// ...
Some(usize::from(offset) + row_value_offset)
```

### Proposed Changes

1. **Use checked arithmetic for matrix indexing**:
```rust
let index = row_index
    .checked_mul(suffix_count)
    .and_then(|v| v.checked_add(column_index))?;
let Some(offset) = self.dense.get(index) else {
    unreachable!(
        "matrix index out of bounds: row={}, col={}, suffix_count={}",
        row_index, column_index, suffix_count
    );
};
```

2. **Use checked arithmetic for final value calculation**:
```rust
usize::from(offset).checked_add(row_value_offset)
```

IshaanXCoder:
Closes #7442 
Adds overflow protection to `ZeroAsciiDenseSparse2dTrieBorrowed::get()` using checked arithmetic operations.

### Changes

**File**: utils/zerotrie/src/dense.rs

1. **Matrix indexing** now uses `checked_mul()` and `checked_add()`
2. **Final value calculation** uses `checked_add()` which perevents overflow when adding offset to row_value_offset

## Changelog: N/A

This is a new unreleased type.

### DIFF
--- a/utils/zerotrie/src/dense.rs
+++ b/utils/zerotrie/src/dense.rs
@@ -209,7 +209,15 @@ impl<'a> ZeroAsciiDenseSparse2dTrieBorrowed<'a> {
             // There is an entry in the dense matrix but it is a None value
             return None;
         }
-        usize::from(offset).checked_add(row_value_offset)
+        let Some(result) = usize::from(offset).checked_add(row_value_offset) else {
+            debug_assert!(
+                false,
+                "overflow: offset={}, row_value_offset={}",
+                offset, row_value_offset
+            );
+            return None;
+        };
+        Some(result)
     }
 
     /// Borrows the structure for encoding into [`zerovec`].


utils/zerotrie/tests/dense_suffix_filter_test.rs
```diff
@@ -0,0 +1,149 @@
...
+// called LICENSE at the top level of the ICU4X source tree
+// (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
+#![cfg(all(feature = "alloc", feature = "dense"))]
+
+use std::collections::BTreeMap;
+use zerotrie::dense::ZeroAsciiDenseSparse2dTrieOwned;
+
+#[test]
+fn test_dense_suffix_count_with_low_frequency_suffixes() {
```
sffc: Did you intend to include these tests in this PR or the other PR?

utils/zerotrie/src/dense.rs
```diff
@@ -193,16 +193,20 @@ impl<'a> ZeroAsciiDenseSparse2dTrieBorrowed<'a> {
...
         let suffix_count = usize::from(self.suffix_count);
-        let Some(offset) = self.dense.get(suffix_count * row_index + column_index) else {
-            // The row and column indexes should be in-range
-            debug_assert!(false);
-            return None;
+        let index = row_index
+            .checked_mul(suffix_count)
+            .and_then(|v| v.checked_add(column_index))?;
+        let Some(offset) = self.dense.get(index) else {
+            unreachable!(
```
sffc: Please use `debug_assert!(false` as before, instead of `unreachable!`, so that we don't have a panic compiled into the code.

utils/zerotrie/tests/dense_suffix_filter_test.rs
```diff
@@ -0,0 +1,149 @@
+// This file is part of ICU4X. For terms of use, please see the file
```
sffc: Please put these tests in `dense_test.rs` and use `check_data` for the assertions.

utils/zerotrie/tests/dense_suffix_filter_test.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    // Each suffix appears in only one prefix (0% overlap)
+    data.entry("a").or_default().insert("s1", 1);
+    data.entry("b").or_default().insert("s2", 2);
+    data.entry("c").or_default().insert("s3", 3);
+
+    let trie = ZeroAsciiDenseSparse2dTrieOwned::try_from_btree_map_str(&data, b'/').unwrap();
+    let trie = trie.as_borrowed();
+
+    // With empty dense matrix, all lookups should still work via sparse path
```
sffc: Suggestion: assert the `Debug` output of the trie, which encodes the values that went into sparse and the values that went into dense.

utils/zerotrie/tests/dense_suffix_filter_test.rs
IshaanXCoder: yess these were to be in the previous PR, my bad

IshaanXCoder:
Hey @sffc i can see there's 2 fail in the check (semver fails after 8m), looking at the logs, it is due to a pre-existing issue, they're in icu_calendar and icu_locale_core.

IshaanXCoder:
@sffc made the required changes, PTAL

utils/zerotrie/tests/dense_suffix_filter_test.rs
sffc: The extra tests are still here. Please move them to the other PR

utils/zerotrie/src/dense.rs
```diff
@@ -193,16 +193,22 @@ impl<'a> ZeroAsciiDenseSparse2dTrieBorrowed<'a> {
...
-            debug_assert!(false);
+        let index = row_index
+            .checked_mul(suffix_count)
+            .and_then(|v| v.checked_add(column_index))?;
+        let Some(offset) = self.dense.get(index) else {
+            debug_assert!(
+                false,
+                "matrix index out of bounds: row={}, col={}, suffix_count={}",
+                row_index, column_index, suffix_count
+            );
```
sffc: Yeah, we should combine the branches into one instead of silently using `?`.

```suggestion
        let Some(offset) = row_index
            .checked_mul(suffix_count)
            .and_then(|v| v.checked_add(column_index))
            .and_then(|index| self.dense.get(index))
        else {
            debug_assert!(
                false,
                "matrix index out of bounds: row={}, col={}, suffix_count={}",
                row_index, column_index, suffix_count
            );
```

Manishearth: Hmm? That doesn't change the branch structure at all. Both of these have the same branches. the combinators will branch to produce the right `Option`, whereas the `?` will branch to produce early returns. THe optimizer will likely see these as exactly the same, but even if it doesn't, it's the same number of branches.

sffc: It triggers the `debug_assert` if the math overflows.

Manishearth: Oh, sure. Yeah, we should do that. Branchwise it's not very different, but that's fine.

utils/zerotrie/src/dense.rs
```diff
@@ -193,16 +193,23 @@ impl<'a> ZeroAsciiDenseSparse2dTrieBorrowed<'a> {
...
             return None;
         };
         if offset == DenseType::MAX {
             // There is an entry in the dense matrix but it is a None value
             return None;
         }
-        Some(usize::from(offset) + row_value_offset)
+        usize::from(offset).checked_add(row_value_offset)
```
sffc: Thought: Maybe this one should debug assert, too

sffc (approved):

utils/zerotrie/src/dense.rs
IshaanXCoder: Alright 👍

IshaanXCoder:
@sffc please review the PRs, made the changes

PR #7444, opened 2026-01-18
ZeroTrie const builder now use mut pointers

### LINKED ISSUE #7084: Improve ZeroTrie const builder to use mut pointers
Rust 1.83 lets us finally use mut pointers in const. This could make the ZeroTrie const builder more efficient.

psg-19:
Fixes #7084 

Previously, the const builder used a functional style where every modification returned a new instance of the builder. This change allows for in place mutation, significantly improving efficiency and compile-time performance for large tries by avoiding these expensive copies.

## Changelog: N/A


sffc:
Thanks for the PR! I'll take a look soon.

PR #7445, opened 2026-01-18
replaced normalizer benchmark input with test corpora text

### LINKED ISSUE #6881: Replace normalizer bench data with data from https://github.com/unicode-org/test-corpora/
We should replace normalizer test data with data from https://github.com/unicode-org/test-corpora/ . However, `test-corpora` contains (X)HTML, so it's necessary to make sure that the data that gets used is paragraph-level (as opposed to various headers) text without markup.

To the extend multiple paragraphs/lines from `test-corpora` are combined, the combination should not introduce line breaks or ASCII spaces into languages that wouldn't use spaces normally at the point of combination.

SATVIKsynopsis:
Fixes #6881 

Replaces one normalizer benchmark input with plain text content
derived from` unicode-org/test-corpora`.


components/normalizer/benches/data/en.txt
```diff
@@ -0,0 +1 @@
...eytimah| |yan| |kutbeytal| |rakitak|.***| |CUGAAYSO| |QIMMIS| |GUTENBERG| |EBOOK| |KAXXA| |GATSBY| |***Kaxxa| |GatsbieddeF.| |Iskot| |FitzgeraldAddatinoh| |QarwaliAnuIIIIIIVVVIVIIVIIIIXQagitakwohuhZeldaTokkeek| |dahab| |koofiyat| |haysit|,| |toh| |tet| |tasgayye| |koo| |tekkek|;Atu| |fayyal| |kaqitte| |koo| |tekkek|,| |teetih| |kaadu| |kaqit|,is| |weqtam| |fanah| |"| |kacnoyta|,| |dahab|-|koofiya|,| |fayyale| |kacnoyta|,Anu| |koo| |aallem| |faxximta|!|”Toomas| |parki| |d'|inviliyersAnuyok| |qunxa| |mariino| |kee| |mango| |gadamsite| |liggidittet| |yabba| |dagoo| |fayu| |yoh| |yeceeh| |too| |anu| |inni| |mesenkacat| |fakkiimeh| |suge| |tohuk| |qemmissa| |haanam|."|Faxe| |way| |atu| |faxe| |num| |sadak| |sugte| |wak|,|"| |usuk| |yoh| |warse|,| |"|bas| |kassit| |ta| |baadal| |tan| |ummatta| |inkih| |atu| |luk| |sugte| |tuxxiq| |alle| |waytam|.|"Usuk| |mangom| |maxacinna|,| |takkay| |immay| |nanu| |umman| |way| |qaadah| |ane| |wayta| |angaaraw| |abneh| |sugneeh|,| |anu| |edde| |radeh| |usuk| |tohuk| |muxxi| |iyyam| |faxam|.Tohih| |taagah|,| |anu| |inkih| |tan| |cokmitte| |reserve| |abak|,| |mango| |curious| |natures| |yoh| |fakkiimeeh|,| |kaadu| |veteran| |bores| |lih| |angaaraw| |le|.
```
sffc: Please at least at the top of the file the source path within test-corpora, such as

```
# Source within unicode-org/test-corpora: gutenberg/Fitzgerald-64317/out/google/txt/es/7069925737426509887_64317-h-1.htm.txt
```

SATVIKsynopsis:
I added the source file header containing source and path at the top.

sffc:
The source header looks good. Please don't concatenate the lines together, though. Also I think you should use the txt file rather than the breakpoints file, since we are testing normalizer, not segmenter.

SATVIKsynopsis:
I have updated to use the english gutenberg `.txt` source and not breakpoints, added proper source header, and stripped headers before benchmarking.

components/normalizer/benches/data/en.txt
```diff
@@ -0,0 +1,4 @@
...
...nny it’ll seem, sending presents to one’sown feet! And how odd the directions will look!Alice’s Right Foot, Esq.,Hearthrug,near the Fender,           (with Alice’s love).Oh dear, what nonsense I’m talking!”Just then her head struck against the roof of the hall: in fact she was nowmore than nine feet high, and she at once took up the little golden key andhurried off to the garden door.Poor Alice! It was as much as she could do, lying down on one side, to lookthrough into the garden with one eye; but to get through was more hopeless thanever: she sat down and began to cry again.“You ought to be ashamed of yourself,” said Alice, “a great girl like you,”(she might well say this), “to go on crying in this way! Stop this moment, Itell you!” But she went on all the same, shedding gallons of tears, until therewas a large pool all round her, about four inches deep and reaching half downthe hall.After a time she heard a little pattering of feet in the distance, and shehastily dried her eyes to see what was coming. It was the White Rabbitreturning, splendidly dressed, with a pair of white kid gloves in one hand anda large fan in the other: he came trotting along in a great hurry, muttering to
```
hsivonen: This has metadata as part of the test content: "Alice’s Adventures in Wonderland | Project GutenbergCHAPTER II.The Pool of Tears".

Also, this takes non-ASCII quotes for English. The test corpus happens to use non-ASCII quotes here, but for benchmarking purposes it's useful to keep English as ASCII-only.

Also, this changes only one instance of English test content as opposed to changing all the test content under this directory.

To find text that for other Latin-script languages of interest is sufficiently accented, I suggest for each language starting with the third paragraph of the first chapter, i.e. the paragraph that in English reads:

"There was nothing so very remarkable in that; nor did Alice think it so
very much out of the way to hear the Rabbit say to itself, “Oh dear! Oh
dear! I shall be late!” (when she thought it over afterwards, it occurred to
her that she ought to have wondered at this, but at the time it all seemed
quite natural); but when the Rabbit actually took a watch out of its
waistcoat-pocket, and looked at it, and then hurried on, Alice started to
her feet, for it flashed across her mind that she had never before seen a
rabbit with either a waistcoat-pocket, or a watch to take out of it, and
burning with curiosity, she ran across the field after it, and fortunately was
just in time to see it pop down a large rabbit-hole under the hedge."

...and in the case of English, replacing the non-ASCII quotation marks with ASCII quotation marks. (It's fine to have French use Latin1 guillemets.)

SATVIKsynopsis:
I have updated `en.txt` to start from the third paragraph of the first chapter as suggested using ascii quotes only and removed metadata from the content.

components/normalizer/benches/composing_normalizer_nfc.rs
```diff
@@ -33,10 +35,6 @@ fn normalizer_bench_data() -> [BenchDataContent; 15] {
     let nfkc_normalizer = ComposingNormalizerBorrowed::new_nfkc();
     let nfkd_normalizer = DecomposingNormalizerBorrowed::new_nfkd();
 
-    let content_latin: (&str, &str) = (
-        "TestNames_Latin",
-        &strip_headers(include_str!("./data/TestNames_Latin.txt")),
```
sffc, line 38: Is it possible to delete `TestNames_Latin.txt` or is still used somewhere else?

SATVIKsynopsis: It is still used by multiple benchmarks (`decomposing_normalizer_nfd.rs`, `nfkd.rs`, `canonical_*` and collator benches) so I think it should be kept.

sffc:
I'm sorry, but this PR does not fix the issue, and we've spent more time iterating on this PR than making forward progress. I left a comment on the issue being more clear about what the deliverables should be in order to close the issue. If you wish to continue, please open a new PR.

PR #7449, opened 2026-01-19
Improve Iran sightings example

robertbastian:
Use a `LazyLock` to source the data from https://github.com/roozbehp/qamari

I dropped the invariant that a `HijriYear` is 354 or 355 days long - this invariant was not used by our code, and some years in https://github.com/roozbehp/qamari have 353 days.

## Changelog: N/A

example code

### DIFF
--- a/components/calendar/src/cal/east_asian_traditional.rs
+++ b/components/calendar/src/cal/east_asian_traditional.rs
@@ -835,31 +835,43 @@ impl ToExtendedYear for EastAsianTraditionalYear {
 impl EastAsianTraditionalYear {
     /// Creates [`EastAsianTraditionalYear`] from the given parts.
     ///
-    /// `start_day` is the date for the first day of the year, see [`Date::to_rata_die`]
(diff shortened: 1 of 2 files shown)


components/calendar/src/cal/east_asian_traditional.rs
```diff
@@ -835,31 +835,43 @@ impl ToExtendedYear for EastAsianTraditionalYear {
...
+    ) -> Option<Self> {
+        let mut month_lengths = [false; 13];
+
+        let mut i = 0;
+        #[allow(clippy::indexing_slicing)]
+        while i < 12 + leap_month.is_some() as usize {
+            match month_starts[i + 1].to_i64_date() - month_starts[i].to_i64_date() {
+                29 => month_lengths[i] = false,
+                30 => month_lengths[i] = true,
+                _ => return None,
```
Manishearth, line 861: thought: returning a Result would be good to think about before stabilization, but not now.

Manishearth (approved):

PR #7452, opened 2026-01-20
Move some lints to the workspace

robertbastian:
Based on #7451

https://github.com/unicode-org/icu4x/issues/5974

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -340,11 +340,22 @@ inherits = "bench"
 debug = true
 
 [workspace.lints]
+clippy.collection_is_never_read = "warn"
+clippy.dbg_macro = "warn"
+clippy.debug_assert_with_mut_call = "warn"
 clippy.exhaustive_enums = "deny"
 clippy.exhaustive_structs = "deny"
+clippy.fn_to_numeric_cast_any = "warn"
+clippy.infinite_loop = "warn"
+clippy.large_stack_arrays = "warn"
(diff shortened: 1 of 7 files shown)


ffi/freertos/src/lib.rs
```diff
@@ -27,12 +27,12 @@
...
 //! This crate has a build script that will attempt to detect the nightly version and configure
 //! things appropriately, where possible. Older nightlies will end up setting the
-//! `--cfg needs_alloc_error_handler` flag: if using a custom build system and a nightly from
+//! `--cfg icu4x_needs_alloc_error_handler` flag: if using a custom build system and a nightly from
 //! 2022 or earlier, please set this flag.
 
 #![allow(clippy::upper_case_acronyms)]
 #![cfg_attr(
-    all(target_os = "none", needs_alloc_error_handler),
+    all(target_os = "none", icu4x_needs_alloc_error_handler),
```
Manishearth: note: this is a separate change. might be worth doing this as a separate PR so it can have a nice changelog entry.

Cargo.toml
```diff
@@ -339,6 +339,56 @@ lto = true
 inherits = "bench"
 debug = true
 
+[workspace.lints]
```
sffc, line 342: There are more lints here than in the boilerplate; why?

robertbastian: more lints = better code

PR #7455, opened 2026-01-20
Deprecate `AstronomicalSimulation`

robertbastian:
#7320

## Changelog

icu_calendar: Deprecate `AstronomicalSimulation` (use `UmmAlQura` or a custom calendar instead)

### DIFF
--- a/components/calendar/src/cal/hijri.rs
+++ b/components/calendar/src/cal/hijri.rs
@@ -146,9 +146,16 @@ pub trait Rules: Clone + Debug + crate::cal::scaffold::UnstableSealed {
 /// Currently, this uses simulation results published by the KACST,
 /// making it identical to [`UmmAlQura`].
 ///
+/// In previous versions, this type used arithmetic published by E. M. Reingold, S. K. Shaukat,
+/// et al.[^1]. Since it now uses the KACST simulations, it is deprecated in favor of [`UmmAlQura`].
+/// If you wish to use the previous behavior, use the [`calendrical_calculations`] crate directly;
+/// for an example, see [`tests/reingold.rs`](https://github.com/unicode-org/icu4x/blob/main/components/calendar/tests/reingold.rs).
+///
 /// The precise behavior of this calendar may change in the future if:
 /// - We decide to tweak the precise astronomical simulation used
 /// - We decide to expand or reduce the range where we are using the astronomical simulation.
+///
+/// [^1]: See [`calendrical_calculations::islamic::observational_islamic_from_fixed`]
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


robertbastian:
It has a note "use `UmmAlQura`", I don't know what else to say really

components/calendar/src/cal/hijri.rs
```diff
@@ -151,9 +151,12 @@ pub trait Rules: Clone + Debug + crate::cal::scaffold::UnstableSealed {
 /// - We decide to expand or reduce the range where we are using the astronomical simulation.
```
sffc, line 156: Please add to the docs,

```rust
/// In previous versions, this type used methods published by E. M. Reingold, S. K. Shaukat, et al.[^1]
/// Since this type now uses the KACST simulations, it is deprecated in favor of [`UmmAlQura`].
/// If you still wish to use the Reingold methods, use the [`calendrical_calculations`] crate directly.
/// For an example, see `tests/reingold.rs`
///
/// [^1]: See [calendrical_calculations::islamic::observational_islamic_from_fixed]
```

components/calendar/src/cal/hijri.rs
```diff
@@ -146,14 +146,24 @@ pub trait Rules: Clone + Debug + crate::cal::scaffold::UnstableSealed {
...
+/// for an example, see [`tests/reingold.rs`](https://github.com/unicode-org/icu4x/blob/main/components/calendar/tests/reingold.rs).
+///
 /// The precise behavior of this calendar may change in the future if:
 /// - We decide to tweak the precise astronomical simulation used
 /// - We decide to expand or reduce the range where we are using the astronomical simulation.
+///
+/// [^1]: See [`calendrical_calculations::islamic::observational_islamic_from_fixed`]
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 #[non_exhaustive]
+#[deprecated(since = "2.2.0", note = "use `UmmAlQura`")]
```
sffc, line 161:
```suggestion
#[deprecated(since = "2.2.0", note = "use `UmmAlQura` or `calendrical_calculations`")]
```

sffc (approved):

components/calendar/src/cal/hijri.rs
robertbastian: `UmmAlQura` is a type implementing the same trait which can be swapped out. `calendrical_calculations` is a whole crate which needs rewiring, so I don't want to suggest this in the compiler warning

sffc: Then the note should say "read the docs of the type for migration instructions"

robertbastian: they can always do that if they know they don't want UAQ

PR #7456, opened 2026-01-20
Safely handle panics in the replace_cart callback

### LINKED ISSUE #7431: `Yoke::wrap_cart_in_box` would have UB if `Box::new` panics
As far as I'm aware, `Box::new` is permitted to panic (see [`handle_alloc_error`](https://doc.rust-lang.org/std/alloc/fn.handle_alloc_error.html)). Therefore, `Yoke::wrap_cart_in_box` can violate the safety contract of `Yoke::replace_with_cart`.

# Possible Solution
Instead of the current implementation:
```rust
    pub unsafe fn replace_cart<C2>(self, f: impl FnOnce(C) -> C2) -> Yoke<Y, C2> {
        Yoke {
            // Safety note: the safety invariant of this function guarantees that
            // the data that the yokeable references has its ownership (if any)
            // transferred to the new cart before self.cart is dropped.
            yokeable: self.yokeable,
            cart: f(self.cart),
        }
    }
```
I think it would suffice to do:
```rust
    pub unsafe fn replace_cart<C2>(self, f: impl FnOnce(C) -> C2) -> Yoke<Y, C2> {
        let yokeable = ManuallyDrop::new(self.yokeable);
        let cart = f(self.cart);
        Yoke {
            // Safety note: the safety invariant of this function guarantees that
            // the data that the yokeable references has its ownership (if any)
            // transferred to the new cart before self.cart is dropped, unless
            // `f` panics, in which case the above `ManuallyDrop` ensures that
            // the yokeable is leaked (preventing any UB from dropping the
            // yokeable after its cart).
            yokeable: yokeable.into_inner(),
            cart,
        }
    }
```
The safety docs would also have to be updated (this strictly loosens the restrictions placed on users, so it should not be breaking), and it should instead be noted that if `f` panics, then the yokeable is leaked (and the panic is propagated).

Manishearth:
Fixes #7431

## Changelog

yoke: Handle potential UB around panicky callbacks in `replace_cart`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -136,6 +136,7 @@ Fully filled in up to 30c187f4b7
     - `yoke`, `yoke_derive`: `0.8.1 -> ???`
         - impl common traits (Display, PartialEq/Eq, PartialOrd/Ord) (#7400)
         - derive: Allow trait bounds in `where` clauses (unicode-org#7230)
+        - Safely handle panics in the `replace_cart` callback, additionally fixing OOM safety issue in `wrap_cart_in_*` (unicode-org#7456)
     - `zerofrom`: No change (`0.1.6`)
     - `zerotrie`: `0.2.3 -> ???`
         - Add `ZeroAsciiDenseSparse2dTrie` for more efficient storage of data keys with many attributes (unicode-org#7264, unicode-org#7304, unicode-org#7305)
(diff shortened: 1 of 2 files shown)


Manishearth:
Whoops, didn't push after fixing the build

sffc (approved):

PR #7459, opened 2026-01-20
More clippy

robertbastian:
#5974

## Changelog: N/A

### DIFF
--- a/tools/make/codegen/templates/properties_names.rs.jinja
+++ b/tools/make/codegen/templates/properties_names.rs.jinja
@@ -112,8 +112,8 @@ pub mod ffi {
         #[diplomat::rust_link(icu::properties::props::GeneralCategoryGroup, Struct)]
         #[diplomat::attr(auto, constructor)]
         #[cfg(feature = "compiled_data")]
-        pub fn create() -> Box<GeneralCategoryNameToGroupMapper> {
-            Box::new(GeneralCategoryNameToGroupMapper(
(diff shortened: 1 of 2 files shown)


Manishearth (approved):

components/calendar/src/cal/buddhist.rs
```diff
@@ -90,11 +90,11 @@ impl Date<Buddhist> {
     /// assert_eq!(date_buddhist.month().ordinal, 1);
     /// assert_eq!(date_buddhist.day_of_month().0, 2);
     /// ```
-    pub fn try_new_buddhist(year: i32, month: u8, day: u8) -> Result<Date<Buddhist>, RangeError> {
+    pub fn try_new_buddhist(year: i32, month: u8, day: u8) -> Result<Self, RangeError> {
         ArithmeticDate::from_year_month_day(year, month, day, &AbstractGregorian(BuddhistEra))
             .map(ArithmeticDate::cast)
             .map(BuddhistDateInner)
-            .map(|i| Date::from_raw(i, Buddhist))
+            .map(|i| Self::from_raw(i, Buddhist))
```
sffc, line 97: I think I don't agree with this lint as a universal thing to enable. I frequently use the actual type name instead of `Self` because it is more clear for readers of the code. Similarly to how I sometimes use `Type::default()` instead of `Default::default()` or `Type::from()` instead of `.into()`

PR #7460, opened 2026-01-20
Enable `clippy::unused_qualifications`

robertbastian:
Which is more like `unncessary_qualifications`.

This was unfortunately manual work.

#5974

### DIFF
--- a/components/calendar/src/cal/hebrew.rs
+++ b/components/calendar/src/cal/hebrew.rs
@@ -151,7 +151,7 @@ impl DateFieldsResolver for Hebrew {
 
     fn reference_year_from_month_day(
         &self,
-        month: types::Month,
+        month: Month,
         day: u8,
     ) -> Result<Self::YearInfo, EcmaReferenceYearError> {
         // December 31, 1972 occurs on 4th month, 26th day, 5733 AM
@@ -186,7 +186,7 @@ impl DateFieldsResolver for Hebrew {
     fn ordinal_from_month(
         &self,
         year: Self::YearInfo,
-        month: types::Month,
+        month: Month,
         options: DateFromFieldsOptions,
     ) -> Result<u8, MonthCodeError> {
         let is_leap_year = year.keviyah.is_leap();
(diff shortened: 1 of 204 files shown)


components/calendar/src/cal/hebrew.rs
```diff
@@ -151,7 +151,7 @@ impl DateFieldsResolver for Hebrew {
 
     fn reference_year_from_month_day(
         &self,
-        month: types::Month,
+        month: Month,
```
sffc, line 154: Nit: This seems like a good opportunity to be consistent between `types::Month` and `Month`; you are changing some files one way and other files the other way

sffc (commented):
There are so many changes here in so many different files that I'm not going to nitpick. I don't really see the reason to do this but also don't see the reason not to do this.

components/calendar/src/cal/hebrew.rs
robertbastian: I've tried to stay consistent with the majority of each file

sffc:
> I don't really see the reason to do this but also don't see the reason not to do this.

The reason not to do this is that it will cause merge conflicts across our 65 open PRs. There's not really a "good time" to do these changes that touch hundreds of code files. We usually save it for things that the WG feels are high-value, like fixing up constructors and Cargo feature docs and unstable modules.

robertbastian:
The reason to do this is that I spent more than an hour cleaning this up.

robertbastian:
> The reason not to do this is that it will cause merge conflicts across our 65 open PRs. 

The changes are extremely local, it's just stripping paths from types. If this produces a merge conflict, any merge editor will be able to resolve that automatically.

robertbastian:
please review

utils/resb/src/binary/deserializer.rs
```diff
@@ -1249,19 +1242,18 @@ fn read_u32(input: &[u8]) -> Result<(u32, &[u8]), BinaryDeserializerError> {
     // only fail if the slice is the wrong size.
     #[expect(clippy::unwrap_used)]
     let bytes = input
-        .get(0..core::mem::size_of::<u32>())
```
sffc, line 1252: I tend to prefer always qualifying free functions in imported modules, such as `mem::size_of::<T>`

robertbastian: Same, but I guess `drop`, `size_of` and `align_of` are exceptions to the rule because they are in the prelude. We didn't qualify `drop` anywhere.

PR #7461, opened 2026-01-20
Revert "More clippy"

sffc:
Reverts unicode-org/icu4x#7459 before it bitrots so we can discuss this.

## Changelog: N/A


Manishearth:
Is the primary objection to these lints being enforced or the changes due to these lints? I was happy to approve these because it is easy to undo the lint enforcement and overall they seem positive.

(I know the objection to some extent is churn, but then reverting does not particularly undo that cost)

robertbastian:
I don't agree with this but it's easy to reproduce later.

sffc:
> Is the primary objection to these lints being enforced or the changes due to these lints?

Primarily the changes due to these lints, but also that enforcing the lint would require either a lot of changes or a lot of suppressions.

> I was happy to approve these because it is easy to undo the lint enforcement and overall they seem positive.

Acknowledged. They seemed weekly negative to me. I might warm up to the changes.

> (I know the objection to some extent is churn, but then reverting does not particularly undo that cost)

It does, because code based on main before this PR and after this PR no longer has conflicts.

robertbastian:
There was one conflicting line in all open PRs since early November.

PR #7469, opened 2026-01-22
Move glue patterns to semantic_skeleta

Manishearth:
Lefotver from https://github.com/unicode-org/icu4x/pull/7429


components/datetime/src/provider/mod.rs
```diff
@@ -140,7 +140,6 @@ pub const MARKERS: &[DataMarkerInfo] = &[
     time_zones::TimezoneNamesEssentialsV1::INFO,
     names::DatetimeNamesWeekdayV1::INFO,
     names::DatetimeNamesDayperiodV1::INFO,
-    names::DatetimePatternsGlueV1::INFO,
     DatetimePatternsTimeV1::INFO,
```
sffc, line 140: Now it's weird that `DatetimePatternsTimeV1` is in a different module than `DatetimePatternsGlueV1`. Maybe they should both go in `semantic_skeletons`.

sffc (approved):

components/datetime/src/provider/mod.rs
Manishearth: Should all of `packed_patterns` go there?

Manishearth:
Rebasing to fix main

components/datetime/src/provider/mod.rs
sffc: I don't think `packed_pattern.rs` should be changed, but the markers could be exported from somewhere more appropriate like `semantic_skeletons` if you like

Manishearth: Yeah that's what I'm suggesting: changing the whole reexport rather than just one key.

Manishearth: Yeah that's what I'm suggesting: changing the whole reexport rather than just one key.

PR #7473, opened 2026-01-22
Fix semver CI

robertbastian:
It broke itself by using the stable toolchain but not being compatible with today's stable toolchain (https://github.com/obi1kenobi/cargo-semver-checks/issues/1526)

This makes it use the toolchain from `rust-toolchain.toml`, so it won't be broken by the passage of time again.


robertbastian:
Force-merging as this breaks main and all PR CI.

PR #4996, opened 2024-06-03
Add icu_preferences util

zbraniecki:

PR #7477, opened 2026-01-22
Match preferences on full value

robertbastian:
Currently it's only a prefix match, i.e. `-u-ca-gregorian-backward` will still parse as `CalendarAlgorithm::Gregorian`, which is probably not intended.

## Changelog

icu_locale_core: Match preferences on full value. This removes fallback behavior from e.g. `u-ca-gregory-foobar` to `-u-ca-gregory`. See [CLDR-19229](https://unicode-org.atlassian.net/browse/CLDR-19229).

### DIFF
--- a/components/locale_core/src/preferences/extensions/unicode/macros/enum_keyword.rs
+++ b/components/locale_core/src/preferences/extensions/unicode/macros/enum_keyword.rs
@@ -258,7 +258,7 @@ mod tests {
 
         let v = unicode::Value::from_str("foo").unwrap();
         let dk = DummyKeyword::try_from(&v);
-        assert!(dk.is_err());
+        dk.unwrap_err();
 
         assert_eq!(DummyKeyword::Standard.as_str(), "standard");
     }
@@ -287,7 +287,7 @@ mod tests {
 
         let v = unicode::Value::from_str("foo").unwrap();
         let dk = DummyKeyword::try_from(&v);
-        assert!(dk.is_err());
+        dk.unwrap_err();
 
         assert_eq!(DummyKeyword::Standard.as_str(), "standard");
     }
@@ -316,7 +316,7 @@ mod tests {
 
         let v = unicode::Value::from_str("foo").unwrap();
         let dk = DummyKeyword::try_from(&v);
-        assert!(dk.is_err());
+        dk.unwrap_err();
 
         let v = unicode::Value::from_str("sub-standard").unwrap();
         let dk = DummyKeyword::try_from(&v).unwrap();
@@ -329,9 +329,8 @@ mod tests {
         assert_eq!(unicode::Value::from(dk), v);
 
         let v = unicode::Value::from_str("sub-foo").unwrap();
-        let dk = DummyKeyword::try_from(&v).unwrap();
-        assert_eq!(dk, DummyKeyword::Sub(None));
-        assert_eq!(unicode::Value::from(dk), unicode::value!("sub"));
+        let dk = DummyKeyword::try_from(&v);
+        dk.unwrap_err();
 
         assert_eq!(
             DummyKeyword::Sub(Some(DummySubKeyword::Rare)).as_str(),


Manishearth:
I am also open to the WG discussing this and deciding we want this behavior regardless of CLDR ambiguities.

I think @zbraniecki @hsivonen @sffc may have thoughts.

sffc:
I think the current behavior is better but we should consult with CLDR

robertbastian:
I'm pretty sure this PR implements CLDR behaviour. CLDR defines only exact matches as allowed values, not arbitrary suffixes. ECMA-402 also does not recognise `u-ca-buddhist-foo`.

Manishearth:
Yeah, I can't find anything in CLDR or BCP47 that says that they _should_ be interpreted as having superset relations. The only hint is the existence of `islamic` as a subtag.

We used to use exact matching here and maybe a year or so ago we switched over to having `Islamic(Option<...>)` as the parsing scheme.

Manishearth:
Looking at the history here the new preferences API in https://github.com/unicode-org/icu4x/pull/4996 made this choice, but our old prefs handling did not. cc @zbraniecki might have just been a mistake?

sffc:
What I meant is that it seems like the current behavior is better, and we should consult with CLDR to put it in the spec.

Manishearth:
Hmmm. I'm not convinced of that: it has already caused one bug, with `ethiopic-amete-alem` (which shouldn't have been falling back to `ethiopic`).

I think it _marginally_ makes sense for calendars and doesn't really make sense for other things.

But open to letting CLDR have a discussion here.

sffc:
It makes sense for something like `buddhist-trad` or `japanese-trad` to fall back to `buddhist` or `japanese` if the implementation doesn't yet support the subtype.

robertbastian:
But it doesn't, because that will be a different calendar. `islamic-umalqura` should not fall back to `islamic` if the implementation doesn't support the subtype, because they are different calendars.

Manishearth:
Right, I think it narrowly makes sense for calendars and not many of the others, and even then it doesn't really make sense for calendars: getting a different calendar as a "fallback", especially one which produces different dates (not just eras) is probably not good.

If CLDR commits to naming these keys such that fallback works that way, perhaps, but I'm not even sure that that's currently the case for calendars.

sffc:
| Locale | What Shane thinks it should mean |
|---|---|
| fr-u-ca-buddhist-trad | Buddhist calendar, traditional variant if available, else variant inferred from locale |
| fr-u-ca-buddhist | Buddhist calendar, variant inferred from locale |
| fr | Calendar inferred from locale |

It's how fallback works. You drop subtags until you find a match.

Note: If `buddhist-trad` were ever added, I would expect `buddhist-proleptc` to also be added.

This is Shane's opinion. We should schedule it for a CLDR discussion. I do not find the spec language intentional enough that we should assume one behavior over another. In the same discussion, we should resolve the issue about what an unqualified `islamic` is supposed to mean.

robertbastian:
> It's how fallback works. You drop subtags until you find a match.

Nope. The value of a `-u-` extension is atomic, even if it's composed of multiple subtags. You don't drop half of it.

robertbastian:
Actually CLDR defines [`valueType`s](https://www.unicode.org/reports/tr35/tr35.html#Unicode_Locale_Extension_Data_Files:~:text=transform%20extension%20%22x0%22.-,valueType,-The%20valueType%20attribute), which are either `single`, `incremental`, `multiple`, or `any`. `incremental` and `multiple` are used to distinguish between a subtag list that defines a set of values, and a subtag list that defines a single value. `ca` is the only key that is defined as `incremental`, because it's the only key that has values comprised of multiple subtags. `dx` and `kr` are defined as `multiple` – ~we don't implement that correctly~ (edit: we implement `dx` correctly and we don't implement `kr`).

Note that the explanation for `incremental` explicitly says:

> Thus `ca-islamic-umalqura` is valid. However, `ca-gregory-japanese` is not valid, because "gregory-japanese" is not listed as a type.

So, CLDR has thought about this, and the spec explicitly requires matching on the whole value.

sffc:
We discussed this in the CLDR Design WG meeting. Without considering all implications, the initial reaction of @macchiati and @richgillam was that calendars should fall back using suffix stripping.

I opened a CLDR issue: https://unicode-org.atlassian.net/browse/CLDR-19229

Manishearth:
@sffc Thanks! That seems to make sense, and as Markus noted it should be calendar, only, which was my primary contention about blocking this (I'm fine with this behavior for calendars, I'm very wary about extending it elsewhere)

sffc:
Follow-up: https://github.com/unicode-org/icu4x/issues/7546

Issue #6064, opened 2025-02-03
Handle Negative Sub Pattern in Currency Pattern


younies:
The currency patterns can be as follows:
`"standard": "‏#,##0.00 ¤;‏-#,##0.00 ¤",`

Right now, we neglect the negative part. We need to investigate its usage and how to treat it.

Manishearth:
We're considering currency to be a 2.3 headliner. If we stop considering this at such we can move this to 2.x.

JustFly1984:
Cross-linking work in progress: https://github.com/unicode-org/icu4x/pull/7884

This PR advances **optional negative `;` subpatterns** for currency patterns: `currency_pattern_selection` classifies positive and negative arms together with a merge heuristic when they disagree, adds unit coverage for the merge helper, and narrows **short compact currency** logging so warnings fire on mismatch or parse failure rather than on every negative subpattern. MessageFormat / dimension paths are updated with the new datagen behavior where relevant.

**Partial:** per in-repo notes, further work may still include per-polarity pattern selection in data, additional formatter parity, and golden updates beyond what this PR lands. Worth keeping the issue open until that remaining scope is agreed and tracked.

PR #7482, opened 2026-01-23
Simplify currency code

robertbastian:
Also fixes negative formatting ("$-12" -> "-$12").

### DIFF
--- a/components/experimental/src/dimension/provider/currency/essentials.rs
+++ b/components/experimental/src/dimension/provider/currency/essentials.rs
@@ -144,6 +144,8 @@ pub struct CurrencyPatternConfig {
 }
 
 impl<'a> CurrencyEssentials<'a> {
+    // Returns the currency placeholder value for the currency,
+    // and the currency pattern for the given width and currency
     pub(crate) fn name_and_pattern(
         &'a self,
         width: Width,


components/experimental/src/dimension/provider/currency/essentials.rs
```diff
@@ -73,10 +76,10 @@ pub struct CurrencyEssentials<'data> {
         feature = "serde",
         serde(
             borrow,
-            deserialize_with = "icu_pattern::deserialize_option_borrowed_cow::<DoublePlaceholder, _>"
+            deserialize_with = "icu_pattern::deserialize_borrowed_cow::<DoublePlaceholder, _>"
         )
     )]
-    pub standard_alpha_next_to_number_pattern: Option<Cow<'data, DoublePlaceholderPattern>>,
+    pub standard_alpha_next_to_number_pattern: Cow<'data, DoublePlaceholderPattern>,
```
younies, line 82: what will happen if the `standard_alpha_next_to_number_pattern` is not exist ? it is not `Option` anymore ?

components/experimental/src/dimension/provider/currency/essentials.rs
```diff
@@ -139,3 +142,41 @@ pub struct CurrencyPatternConfig {
     /// If the value is `None`, this means that the narrow pattern does not have a place holder.
     pub narrow_placeholder_value: Option<PlaceholderValue>,
 }
+
+impl<'a> CurrencyEssentials<'a> {
+    pub(crate) fn name_and_pattern(
```
younies, line 149: can you add a comment to the function ?

younies: I feel this function is more related to the formatter

provider/source/src/currency/essentials.rs
```diff
@@ -399,19 +385,9 @@ fn test_basic() {
...
     assert_writeable_eq!(
-        ar_eg_payload
-            .standard_pattern
-            .as_ref()
-            .unwrap()
-            .interpolate((3, "$")),
+        ar_eg_payload.standard_pattern.interpolate((3, "$")),
         "\u{200f}3\u{a0}$"
     );
-    // TODO(#6064)
```
younies, line 409: why removing the todo ?

components/experimental/src/dimension/provider/currency/essentials.rs
robertbastian: it seems to exist for all locales

components/experimental/src/dimension/provider/currency/essentials.rs
robertbastian: it's used by multiple formatters, and it is sensitive to the deduplication within the data struct, so I think it's best to encapsulate it here

provider/source/src/currency/essentials.rs
robertbastian: because I removed the assert because the field is not optional anymore

components/experimental/src/dimension/provider/currency/essentials.rs
robertbastian: it looks like there was an issue with CLDR 46 data for ar-EG, which probably led you to believe that these fields are optional. in CLDR 47 and CLDR 48 they are present in all locales, and I would consider it a CLDR bug if they were not

components/experimental/src/dimension/provider/currency/essentials.rs
younies: I meant to move it to the formatter (maybe a helper function), but no strong push to be there.

Just remember to add a comment.

provider/source/src/currency/essentials.rs
younies: Okay, the bug is about the negative part: https://github.com/unicode-org/icu4x/issues/6064, I think it should not be there in the first place.

components/experimental/src/dimension/provider/currency/essentials.rs
robertbastian: i have added a comment

PR #7485, opened 2026-01-23
Add constructors with `Month` for lunisolar calendars

robertbastian:
#7043

Need to discuss naming.

## Changelog

icu_calendar: Add constructors with `Month` for lunisolar calendars
 - New methods: `Date::try_new_korean_traditional()`, `Date::try_new_chinese_traditional()`, `Date::try_new_hebrew_v2()`


components/calendar/src/error.rs
```diff
@@ -63,6 +65,45 @@ pub enum DateError {
     UnknownMonthCode(MonthCode),
 }
 
+/// Error type for date creation via [`Date::try_new_from_codes`].
+///
+/// [`Date::try_new_from_codes`]: crate::Date::try_new_from_codes
+#[derive(Debug, Copy, Clone, PartialEq, Display)]
+#[non_exhaustive]
+pub enum LunisolarRangeError {
```
Manishearth, line 73: We should consider this error's design to be a part of https://github.com/unicode-org/icu4x/issues/7010 (a discussion which _must_ happen before 2.2 anyway), so we can land this as is and bikeshed errors later.

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -260,6 +260,28 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
         Ok(ArithmeticDate::new_unchecked(year, month, day))
     }
 
+    pub(crate) fn try_from_ymd_lunisolar(
```
Manishearth, line 263: thought: would be nice to share code here if possible

components/calendar/src/cal/east_asian_traditional.rs
```diff
@@ -482,14 +483,40 @@ impl Rules for Korea {
...
+            .map(ChineseDateInner)
+            .map(|inner| Date::from_raw(inner, calendar))
+    }
+}
+
+impl<A: AsCalendar<Calendar = KoreanTraditional>> Date<A> {
+    /// This method uses an ordinal month, which is probably not what you want.
+    ///
+    /// Use [`Date::try_new_korean_traditional`]
+    #[deprecated(since = "2.1.0", note = "use `Date::try_new_korean_traditional`")]
```
Manishearth, line 519: observation: curious that `try_new_dangi_with_calendar` wasn't deprecated, but the `chinese` version was. Probably a mistake, since we should have deprecated both at once when we removed the calendar argument?

components/calendar/src/cal/hebrew.rs
```diff
@@ -347,13 +348,38 @@ impl Calendar for Hebrew {
...
+    ) -> Result<Date<Hebrew>, LunisolarRangeError> {
+        ArithmeticDate::try_from_ymd_lunisolar(year, month, day, &Hebrew)
+            .map(HebrewDateInner)
+            .map(|inner| Date::from_raw(inner, Hebrew))
+    }
+
+    /// This method uses an ordinal month, which is probably not what you want.
+    ///
+    /// Use [`Date::try_new_hebrew_fixed`]
+    #[deprecated(since = "2.1.0", note = "use `Date::try_new_hebrew_fixed`")]
```
Manishearth, line 382: thought: less enthused about this deprecation. the other two had reasons to already be deprecated, but this one has the right name

components/calendar/src/cal/hebrew.rs
```diff
@@ -347,13 +348,38 @@ impl Calendar for Hebrew {
...
+    /// let date = Date::try_new_hebrew_fixed(5782, Month::new(6), 7)
+    ///     .expect("Failed to initialize Date instance.");
+    ///
+    /// assert_eq!(date.era_year().year, 5782);
+    /// // Adar I
+    /// assert_eq!(date.month().number(), 6);
+    /// assert_eq!(date.month().is_formatting_leap(), true);
+    /// assert_eq!(date.day_of_month().0, 7);
+    /// ```
+    pub fn try_new_hebrew_fixed(
```
Manishearth, line 369: issue: `_fixed` is definitely confusing, we should bikeshed this. i find the other names to be fine.

components/calendar/src/error.rs
Manishearth: Though I'm not opposed to expediting that discussion. It need sto happen.

components/calendar/src/calendar_arithmetic.rs
robertbastian: share with what? this shares the code between all lunisolar calendars

components/calendar/src/cal/east_asian_traditional.rs
robertbastian: it was, scroll up in the diff

components/calendar/src/calendar_arithmetic.rs
Manishearth: Share with `from_era_year_month_code_day`, I think?

Manishearth: I guess it's not super straightforward since both operate on a resolved yearinfo rather than a year.

components/calendar/src/cal/hebrew.rs
sffc: Using ordinal months in Hebrew is a huge footgun. This should have been deprecated long ago

components/calendar/src/error.rs
sffc: You know how I feel about nesting the RangeError but we can land this and bikeshed later

PR #7486, opened 2026-01-23
Add `Month` field to `DateFields`

robertbastian:
#7149 

## Changelog: N/A

Will be covered in graduation PR

### DIFF
--- a/components/calendar/src/calendar_arithmetic.rs
+++ b/components/calendar/src/calendar_arithmetic.rs
@@ -283,10 +283,13 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
             },
         };
 
+        // We're returning these error early so that we return structural type
+        // errors before range errors, see comment in the year code below.
         if fields.month_code.is_none() && fields.ordinal_month.is_none() && fields.month.is_none() {
-            // We're returning this error early so that we return structural type
-            // errors before range errors, see comment in the year code below.
             return Err(DateFromFieldsError::NotEnoughFields);
+        }
+        if fields.month_code.is_some() && fields.month.is_some() {
+            return Err(DateFromFieldsError::InconsistentMonth);
         }
 
         let mut valid_month = None;
--- a/components/calendar/src/types.rs
+++ b/components/calendar/src/types.rs
@@ -114,44 +114,32 @@ mod unstable {
         /// assert_eq!(year_info.extended_year, 2025);
         /// ```
         pub extended_year: Option<i32>,
+        /// The month representing a valid month in this calendar year.
+        ///
+        /// Only one of [`Self::month`] and [`Self::month_code`] may be set.
+        ///
+        /// For a full example, see [`Self::ordinal_month`].
+        pub month: Option<crate::types::Month>,
         /// The month code representing a valid month in this calendar year,
         /// as a UTF-8 string.
         ///
         /// See [`MonthCode`](crate::types::MonthCode) for information on the syntax.
         ///
-        /// # Examples
-        ///
-        /// To set the month code field, use a byte string:
-        ///
-        /// ```
-        /// use icu::calendar::types::DateFields;
-        ///
-        /// let mut fields = DateFields::default();
-        ///
-        /// // As a byte string literal:
-        /// fields.month_code = Some(b"M02L");
-        ///
-        /// // Using str::as_bytes:
-        /// fields.month_code = Some("M02L".as_bytes());
-        /// ```
+        /// Only one of [`Self::month`] and [`Self::month_code`] may be set.
         ///
         /// For a full example, see [`Self::ordinal_month`].
         pub month_code: Option<&'a [u8]>,
-        /// The month representing a valid month in this calendar year.
-        ///
-        /// For a full example, see [`Self::ordinal_month`].
-        pub month: Option<crate::types::Month>,
         /// See [`MonthInfo::ordinal`](crate::types::MonthInfo::ordinal).
         ///
-        /// If both this and [`Self::month_code`] are set, they must refer to
+        /// If both this and [`Self::month`]/[`Self::month_code`] are set, they must refer to
         /// the same month.
         ///
-        /// Note: using [`Self::month_code`] is recommended, because the ordinal month numbers
-        /// can vary from year to year, as illustrated in the following example.
+        /// Note: using [`Self::month`]/[`Self::month_code`] is recommended, because the ordinal
+        /// month numbers can vary from year to year, as illustrated in the following example.
         ///
         /// # Examples
         ///
-        /// Either `month_code`, `month`, or `ordinal_month` can be used in [`DateFields`], but they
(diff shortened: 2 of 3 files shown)


Manishearth:
Open question: should this _replace_ MonthCode in try_from_fields, or be an alternate input?

Replacing may have tricky spec implications for Temporal, due to validation order. On the other hand, having both feels clunky.

Haven't really thought about this yet or looked closely at the code, flagging this now as something to think about.

sffc:
From my recollection, adding `month` as an extra field was the compromise to bridge the gap between my position (DateFromFields should be an unvalidated bag of stuff, which Temporal needs) and Robert's position (we should use the nicer Rust types).

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -339,9 +342,18 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
             (Some(_), None) | (None, Some(_)) => return Err(DateFromFieldsError::NotEnoughFields),
         };
 
-        let month = match fields.month_code {
-            Some(month_code) => {
-                let validated = match valid_month_code {
+        let month = match (fields.month_code, fields.month) {
```
sffc, line 348: Issue: please return Err if both are set and they don't agree

components/calendar/src/error.rs
```diff
@@ -127,7 +127,7 @@ mod unstable {
         ///
         /// let mut fields = DateFields::default();
         /// fields.extended_year = Some(2000);
-        /// fields.month_code = Some(b"????");
+        /// fields.month_code = Some(b"sep");
```
sffc, line 130: Praise: better example

components/calendar/src/error.rs
```diff
@@ -231,7 +231,7 @@ mod unstable {
         /// fields.extended_year = Some(5784);
         ///
         /// let err = Date::try_from_fields(fields, Default::default(), Hebrew)
-        ///     .expect_err("month M06 is not the 6th month in leap year 5784");
+        ///     .expect_err("month 6 is not the 6th month in leap year 5784");
```
sffc: Nit: please use month codes when stringifying a Month

Manishearth:
I see. Fair enough.

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -339,9 +342,18 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
...
+            (_, Some(month)) => {
```
Manishearth, line 349: Can we document that:

  - `month` takes precedence over `month_code` and there is no related error handling?
  - and/or you are not supposed to use this API with both `month` and `month_code` set

Manishearth:
> Good, just want to reject when month and month code disagree

is it even worth it to perform that check? We could also reject when both are specified, period.

There's use cases for specifying both ordinal and structured months. OTOH month vs month_code are two ways of specifying the exact same info.

robertbastian:
I'd prefer just ignoring `month_code` if `month` is set, and documenting that.

robertbastian:
To be explicit, there are three options and I'd like to do 1:
1. if `month` is set, ignore `month_code`
2. if `month` and `month_code` are set, error
3. if `month` and `month_code` are set, parse `month_code` and error if it doesn't match `month`

Manishearth:
I'd prefer erroring but I'm not opposed to the others.

robertbastian:
Please be specific. 2 or 3?

sffc:
3 > 2 >> 1

Manishearth:
@robertbastian I prefer 2 and 3 over 1. If asked to pick, I would say 2 > 3 because I don't see a strong reason for anyone to need consistency checking between these two.

2 > 3 >> 1

sffc:
I prefer 3 over 2 because it is more like what we do with `month_code` and `ordinal_month`: we accept both and reject if they don't end up agreeing. I don't see why `month` should be treated differently: it is simply a third way of specifying the same thing.

Manishearth:
I think month_code and ordinal month are two ways of specifying different things that compute the same underlying data. They are not computable from each other. I can see _some_ reasons of specifying both, for example to check consistency.

On the other hand, month and month_code are two ways of specifying the same thing. Specifying both is largely a programmer error, in my view.

I'm not strongly opposed to adding code to check consistency here, I just think it's not actually _useful_ beyond being consistent. I'd rather just say "don't do this" and return InconsistentMonth.

robertbastian:
I'm 1 > 2 > 3 so I'll do 2.

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -283,13 +283,16 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
...
-        if fields.month_code.is_none() && fields.ordinal_month.is_none() {
-            // We're returning this error early so that we return structural type
-            // errors before range errors, see comment in the year code below.
+        // We're returning these error early so that we return structural type
+        // errors before range errors, see comment in the year code below.
+        if fields.month_code.is_none() && fields.ordinal_month.is_none() && fields.month.is_none() {
             return Err(DateFromFieldsError::NotEnoughFields);
         }
+        if fields.month_code.is_some() && fields.month.is_some() {
+            return Err(DateFromFieldsError::InconsistentMonth);
```
sffc, line 292: I don't love that we return `InconsistentMonth` when they may or may not be inconsistent

components/calendar/src/types.rs
```diff
@@ -114,51 +114,43 @@ mod unstable {
...
+        /// Only one of [`Self::month`] and [`Self::month_code`] may be set.
+        ///
+        /// For a full example, see [`Self::ordinal_month`].
+        pub month: Option<crate::types::Month>,
         /// The month code representing a valid month in this calendar year,
         /// as a UTF-8 string.
         ///
         /// See [`MonthCode`](crate::types::MonthCode) for information on the syntax.
         ///
-        /// # Examples
```
sffc, line 122: Nit: Keep the byte string example because it is not intuitive. You can also add an example to Month that is better.

robertbastian: I've never liked that example, because it's literally teaching Rust to our users, that's not what docs are for.

Now that `month_code` is not the main way to supply a month anymore, I think it's even less necessary.

components/calendar/src/calendar_arithmetic.rs
robertbastian: ok. the consensus is option 2 though. please don't block this

sffc: option 2 is "error", it didn't say specifically which error.

I'm okay deciding the error later, but before 2.2

components/calendar/src/types.rs
sffc: Most of these fields have examples.

robertbastian: And `month` and `month_code` both say to look at `ordinal_month` for an example. The "example" that `month_code` had explained how to get bytes from a str, it wasn't an example of any kind of logic

PR #7488, opened 2026-01-23
Enable `clippy::or-fun-call`

robertbastian:
Don't accidentally execute code in error paths.


sffc (approved):
These are very positive changes that we should enforce. Too bad you need `const { }` instead of automatically inlining const fns.

I remember seeing this type of warning before, but maybe not everywhere

PR #7489, opened 2026-01-23
Enable `clippy::branches-sharing-code`

robertbastian:
Keeps our code more readable. Barely any violations anyway.


sffc (approved):
This seems to align with our goal of reducing code size.

PR #7491, opened 2026-01-23
Enable `clippy::match-bool`

robertbastian:
It's a stylistic choice, *but* after it forced me to change the code in `ffi/ecma402/src/pluralrules.rs`, another lint fired to tell me to use `saturating_sub` instead. Matching on `bool` is probably not idiomatic enough to have lints consider this pattern, which is why I think it's worth enabling.


sffc (commented):
This is one where my initial reaction is, "mixed" on the call sites, and "weekly negative" on the enforcement moving forward.

sffc (changes requested):
Did you want me to take a stronger position? Then no, I think these changes are overall negative and I don't think we should force developers to conform to this lint moving forward.

PR #7492, opened 2026-01-23
Enable `clippy::unnecessary-wraps`

robertbastian:
Lints non-pub functions that unnecessarily return `Result`. This is something that often comes up in code reviews, and makes the fallibility of our code harder to follow.


sffc (approved):
I approve, this produces code where it is easier to reason about error cases, and we value narrow error types

PR #7497, opened 2026-01-23
Reorder DateDuration methods to group related methods (no API change)

Manishearth:
Don't know why it was arranged like that, but this makes the docs look neater.

I think this fixes https://github.com/unicode-org/icu4x/issues/7059 as it was the only issue I found during API review. WG members are encouraged to do their own API review if they want to double check.


Manishearth:
autosubmit=true

Issue #7462, opened 2026-01-21
Update behavior of MinuteOptional format in h23 patterns


sffc:
It seems like https://unicode-org.atlassian.net/browse/CLDR-19205 might be moving in the direction where the ":00" is dropped in h11/h12 but not in h23. We should update ICU4X to have this behavior.

Jayant-kernel:
I would like to work on this

sffc:
Please wait until CLDR has made a decision. Also, there is already a PR open for this.

sffc:
New CLDR issue: https://unicode-org.atlassian.net/browse/CLDR-19327

Manishearth:
Blocked on CLDR. Becomes `2.<next release>` when CLDR has a decision

PR #7501, opened 2026-01-24
Update behavior of MinuteOptional format in h23 patterns

IshaanXCoder:
## MinuteOptional Behavior Update

**Related issue** - #7462

**Description**
Set the `MinuteOptional` behavior with CLDR-19205 to improve 24-hour time formatting.

**Changes**
- Updated `MinuteOptional` logic to **preserve zero minutes** (eg, `15:00`) for 24 hour cycles 
- Retained existing behavior for 12 hour cycles(`h11`, `h12`) which continue to drop zero minutes (eg, `3 PM`)
- Refactored `neo.rs` to detect hour cycles during pattern selection
- Added tests in `minute_optional_test.rs`

### DIFF
--- a/components/datetime/src/provider/pattern/runtime/pattern.rs
+++ b/components/datetime/src/provider/pattern/runtime/pattern.rs
@@ -61,6 +61,7 @@ pub struct PatternMetadata(u8);
 
 impl PatternMetadata {
     pub(crate) const DEFAULT: PatternMetadata = Self::from_time_granularity(TimeGranularity::None);
+    const TIME_GRANULARITY_MASK: u8 = 0x07;
     const PREFER_KEEP_MINUTES_MASK: u8 = 0x08;
 
     #[inline]


IshaanXCoder:
@sffc PTAL whenever free :)

components/datetime/src/raw/neo.rs
```diff
@@ -345,7 +346,24 @@ impl TimePatternSelectionData {
     ) -> Option<TimePatternDataBorrowed<'_>> {
         let payload = self.payload.get_option()?;
         let time_precision = options.time_precision.unwrap_or_default();
-        let (variant, subsecond_digits) = input.resolve_time_precision(time_precision);
+
+        let prefer_keep_minutes = if matches!(time_precision, TimePrecision::MinuteOptional) {
```
sffc, line 350: I think this gets the job done!

But, calculating this in every format operation is going to be expensive. For a very similar case, we have an optimization: `PatternMetadata`. I think we should add this bit to `PatternMetadata` so that we don't need to compute it each time.

sffc (commented):
Thanks for the contribution. CLDR hasn't yet decided whether to adopt this behavior, so for now I'll keep the PR as a draft, but I gave you a suggestion on how to improve the PR in the meantime.

components/datetime/src/provider/pattern/runtime/pattern.rs
```diff
@@ -60,20 +61,47 @@ pub struct PatternMetadata(u8);
 
 impl PatternMetadata {
     pub(crate) const DEFAULT: PatternMetadata = Self::from_time_granularity(TimeGranularity::None);
+    const PREFER_KEEP_MINUTES_MASK: u8 = 0x08;
```
sffc, line 65:
```suggestion
    const TIME_GRANULARITY_MASK: u8 = 0x07;
    const PREFER_KEEP_MINUTES_MASK: u8 = 0x08;
```

components/datetime/src/provider/pattern/runtime/pattern.rs
```diff
@@ -60,20 +61,47 @@ pub struct PatternMetadata(u8);
...
 
     #[inline]
     pub(crate) fn time_granularity(self) -> TimeGranularity {
-        TimeGranularity::from_ordinal(self.0)
+        TimeGranularity::from_ordinal(self.0 & !Self::PREFER_KEEP_MINUTES_MASK)
```
sffc, line 69: Here and elsewhere:

```suggestion
        TimeGranularity::from_ordinal(self.0 & Self::TIME_GRANULARITY_MASK)
```

components/datetime/src/provider/pattern/runtime/pattern.rs
```diff
@@ -60,20 +61,47 @@ pub struct PatternMetadata(u8);
...
-        Self::from_time_granularity(time_granularity)
+        let mut time_granularity = TimeGranularity::None;
+        let mut prefer_keep_minutes = false;
+
+        for item in iter_items {
+            let item_granularity = TimeGranularity::from(item);
+            if item_granularity > time_granularity {
+                time_granularity = item_granularity;
+            }
+            if !prefer_keep_minutes {
```
sffc, line 90: Suggestion: Implement this more like TimeGranularity, with an enum in mod.rs with 2 variants and all the constructors and parsing code in that file.

We already have `CoarseHourCycle`. I suggest using that enum.

components/datetime/src/provider/pattern/runtime/pattern.rs
IshaanXCoder: Sure i'll fix the suggestions, also, there're some cicd failures despite checking them locally, I'll try to fix them too

PR #7502, opened 2026-01-25
fixed large-magnitude numbers being converted to PluralOperand

### LINKED ISSUE #2588: Add tests and/or fix large-magnitude numbers being converted to PluralOperands
From @markusicu:

> How do you handle conversions from `i128` and `u128` types when you only have a `u64` for the `i`, `v`, `f` operands?  We truncate the high digits, and there are no CLDR rules for which that truncation makes a difference. But we can double check that this is indeed the case.

I should note that this also happens in string and FixedDecimal conversions. We should make sure we have consistent behavior across the board.

CC @zbraniecki

psg-19:
Fixes #2588 

summary :- for the purpose of CLDR Plural Rules, we effectively only need the last 18 digits. So when the input is in `<u128>`, check if the input is less than `1e18` (`<u64>`), if it is so then directly take the input. But if it is greater than that then just do the modulo of the input with `1e18` and add that modulo in `1e18`, in this way the input is still in range of `<u64>` but preserving the modulo properties of the input which the Plural Rules need.

## Changelog

icu_plurals: Handle large-magnitude numbers being converted to PluralOperands


components/plurals/src/operands.rs
```diff
@@ -248,6 +269,10 @@ impl PluralOperands {
             i += dec.digit_at(magnitude - exp_i16) as u64;
         }
 
+        if *mag_range.end() + exp_i16 > 17 {
+            i += LIMIT_18_DIGITS;
```
robertbastian, line 273: I don't understand why you added this, isn't this whole PR based on the assumptions that plural rules are congruent modulo `LIMIT_18_DIGITS`?

components/plurals/src/operands.rs
```diff
@@ -231,9 +233,28 @@ macro_rules! impl_signed_integer_type {
...
 impl_signed_integer_type!(i8 i16 i32 i64 i128 isize);
 
+impl From<u128> for PluralOperands {
+    #[inline]
+    fn from(input: u128) -> Self {
+        let i = if input < LIMIT_18_DIGITS as u128 {
+            input as u64
+        } else {
+            (input % LIMIT_18_DIGITS as u128) as u64 + LIMIT_18_DIGITS
+        };
```
robertbastian, line 246: question: you do `+ LIMIT_18_DIGITS` so that `is_exactly_one` and `is_exactly_zero` behave correctly? a comment would have been nice

components/plurals/src/operands.rs
robertbastian: I see, it's for the same reason as in the `u128` code

psg-19: Without this, a huge number like `1e18 + 1` or `1e30` would result in i = 1 or i = 0. Adding the limit ensures they are treated as `large numbers` while still preserving their modulo properties.

components/plurals/src/operands.rs
psg-19: Yes thats exactly why I did `+ LIMIT_18_DIGITS`, will add a comment there

PR #7503, opened 2026-01-26
Start producing Meiji era only after Meiji 6

Manishearth:
See https://github.com/tc39/proposal-intl-era-monthcode/pull/102

We already have a note about this in our docs, but the calendar switchover happened in Meiji 6, which means dates specified as Meiji 1-5 are unambiguously lunisolar. We're already liberal in accepting eraYears that are not in range of the era, which means we can't (and shouldn't) do much about people plugging dates specified as Meiji 1-5, but we can avoid producing such dates.

Matching Temporal seems like the right call to me here.

If matching Temporal here is not desired this can in theory be patched over in temporal_rs and Firefox (@hsivonen), but I feel convinced by the Temporal discussions.

## Changelog

icu_calendar: Start producing Meiji era only after Meiji 6

### DIFF
--- a/components/calendar/src/cal/japanese.rs
+++ b/components/calendar/src/cal/japanese.rs
@@ -20,8 +20,7 @@ use tinystr::{tinystr, TinyAsciiStr};
 /// is uses Japanese eras instead of the Common Era.
 ///
 /// This implementation extends proleptically for dates before the calendar's creation
-/// in 6 Meiji (1873 CE).
-/// The Meiji era is used proleptically back to and including 1868-10-23, Gregorian eras are used before that.
+/// in 6 Meiji (1873 CE). Gregorian eras are used before that.
 ///
 /// This corresponds to the `"japanese"` [CLDR calendar](https://unicode.org/reports/tr35/#UnicodeCalendarIdentifier).
 ///
@@ -153,7 +152,7 @@ impl GregorianYears for &'_ Japanese {
     fn era_year_from_extended(&self, year: i32, month: u8, day: u8) -> types::EraYear {
         let date: EraStartDate = EraStartDate { year, month, day };
 
-        if let Some((start, code, idx)) = self.eras().find(|&(start, ..)| date >= start) {
+        let ret = if let Some((start, code, idx)) = self.eras().find(|&(start, ..)| date >= start) {
             types::EraYear {
                 era: code.resize(),
                 era_index: Some(idx),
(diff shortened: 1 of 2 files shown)


components/calendar/src/cal/japanese.rs
```diff
@@ -163,7 +162,22 @@ impl GregorianYears for &'_ Japanese {
...
+        // Special case: Meiji 1 to 5 are treated as CE.
+        //
+        // This calendar was adopted in Meiji 6 (1873), but Meiji started
+        // on 1868-10-23, so for Meiji 1 - 5 the lunisolar calendar was stil
+        // in use. In order to not produce incorrect pre-Meiji 6
+        // dates, we force the Gregorian calendar there.
+        //
+        // https://tc39.es/proposal-intl-era-monthcode/#table-calendar-types
+        // https://github.com/tc39/proposal-intl-era-monthcode/issues/86
+        if ret.era == "meiji" && ret.year < 6 {
```
Manishearth, line 176: This could also be handled by updating the condition in the loop above, but this felt cleaner to me. Not particularly attached to the specific implementation.

hsivonen (approved):
LGTM. Thank you!

provider/source/src/calendar/eras.rs
```diff
@@ -185,7 +185,13 @@ fn test_calendar_eras() {
         for (idx, (_, era)) in data.eras.iter().enumerate() {
             let (in_era_iso, not_in_era_iso) = match (era.start, era.end) {
                 (Some(start), None) => {
-                    let start = Date::try_new_iso(start.year, start.month, start.day).unwrap();
+                    let year = if era.code == "meiji" {
+                        // Meiji starts roundtripping only after Meiji 6
+                        start.year + 5
```
sffc, line 190: I don't understand this but it's in a test

Manishearth: It's testing roundtripping by checking if "era epoch date" roundtrips to the same era, which is a property that is true for all eras except Meiji since its start date is not its epoch date.

robertbastian: this is why I'd prefer to wait for CLDR to add data for this

robertbastian:
I would have liked this change to be coordinated with CLDR. Unilaterally implementing this will put us out of alignment with ICU, again.

There is some discussion in https://unicode-org.atlassian.net/browse/CLDR-11400?focusedCommentId=184108

PR #7504, opened 2026-01-27
Add Unihan source to datagen

opnuub:
Replaces #7436

Modified:

provider/icu4x-datagen/src/main.rs
provider/source/src/lib.rs
provider/source/src/segmenter/mod.rs (to import unihan.rs)
provider/source/src/segmenter/unihan.rs (to use LiteMap<codepoint, int8> via test)
provider/source/src/source.rs (added UnihanCache with proprocessing)
tools/make/download-repo-sources/globs.rs.data (added filename IRGSources.txt)
tools/make/download-repo-sources/src/main.rs

## Changelog

icu4x-datagen: Add `--ucd-tag` and `--unihan-root` flags

icu_provider_source: Allow specifying UCD/Unihan source
 - New methods: `SourceDataProvider::with_unihan_for_tag()`, `SourceDataProvider::is_missing_unihan_error()`


Manishearth (approved):
Generally seems correct. ucd_tag/unihan_root also seems like the right naming (and this was discussed over email IIRC), though we may eventually also want ucd_root.

Would want Shane to sign off since he's been more involved in this project.

sffc (approved):
Nice work!

provider/icu4x-datagen/src/main.rs
```diff
@@ -172,6 +172,17 @@ struct Cli {
     #[cfg(feature = "provider")]
     icuexport_root: Option<PathBuf>,
 
+    #[arg(long, value_name = "TAG", default_value = "17.0.0")]
```
sffc, line 175: Oh, one thing I noticed: the default value should be "latest"... but also "latest" means something special for UCD. Let's discuss in #7510

provider/icu4x-datagen/src/main.rs
```diff
@@ -424,6 +435,10 @@ fn run(cli: Cli) -> eyre::Result<()> {
             );
         } else if SourceDataProvider::is_missing_segmenter_lstm_error(e) {
             eyre::bail!("Segmentation LSTM data is required for this invocation, set --segementer-lstm-path or --segementer-lstm-tag");
+        } else if SourceDataProvider::is_missing_unihan_error(e) {
+            eyre::bail!(
+                "Unihan data is required for this invocation, set --unihan-root or --unihan-tag"
```
robertbastian, line 440:
```suggestion
                "Unihan data is required for this invocation, set --unihan-root or --ucd-tag"
```

provider/source/src/source.rs
```diff
@@ -485,3 +486,61 @@ impl TzdbCache {
...
+        self.irg_cache
+            .get_or_init(|| {
+                let raw_content = self.root.read_to_string("Unihan_IRGSources.txt")?;
+                let mut map = LiteMap::new();
+                for line in raw_content.lines() {
+                    if line.starts_with('#') || line.trim().is_empty() {
+                        continue;
+                    }
+                    let parts: Vec<&str> = line.trim().split('\t').collect();
+                    if parts[1] != "kRSUnicode" {
```
robertbastian, line 515: this panics if the line only contains one column

provider/source/src/source.rs
```diff
@@ -485,3 +486,61 @@ impl TzdbCache {
...
+                        continue;
+                    }
+                    let codepoint = parts[0]
+                        .strip_prefix("U+")
+                        .and_then(|hex| u32::from_str_radix(hex, 16).ok())
+                        .and_then(char::from_u32)
+                        .expect("Invalid Unihan codepoint format");
+
+                    let mut candidate = parts[2].trim();
```
robertbastian, line 524: this panics if the line only contains two columns

provider/source/src/source.rs
```diff
@@ -485,3 +486,61 @@ impl TzdbCache {
...
+                    if let Some(first_part) = candidate.split_whitespace().next() {
+                        candidate = first_part;
+                    }
+                    let radical_str = if let Some(idx) = candidate.find('.') {
+                        &candidate[..idx]
+                    } else {
+                        candidate
+                    };
```
robertbastian, line 532:
```suggestion
                    let radical_str = candidate.split('.').next().unwrap_or(candidate);
```

provider/source/src/source.rs
```diff
@@ -485,3 +486,61 @@ impl TzdbCache {
...
+                    let clean_str = radical_str.replace('\'', "");
```
robertbastian, line 533: do the ticks not have a meaning?

tools/make/download-repo-sources/src/main.rs
```diff
@@ -254,6 +269,24 @@ fn main() -> eyre::Result<()> {
...
+    let irg_path = out_root.join("tests/data/unihan/Unihan_IRGSources.txt");
+    let file = File::open(&irg_path)?;
+    let reader = io::BufReader::new(file);
+    let filtered_content: String = reader
+        .lines()
+        .map_while(Result::ok)
+        .filter(|l| l.contains("kRSUnicode") || l.starts_with('#'))
+        .collect::<Vec<_>>()
+        .join("\n");
+    fs::write(&irg_path, filtered_content)?;
```
robertbastian, line 289: don't modify files for testdata

provider/source/src/source.rs
opnuub: According to the official Unihan specification:

> Lines beginning with # are comment lines used to provide the header and footer.
> Each of the remaining lines is one entry, with three, tab-separated fields:
> the Unicode Scalar Value, the property name, and the value for the property for the given Unicode Scalar Value.

Source: https://www.unicode.org/reports/tr38/#Unihan.zip

So the file is effectively:
- comment lines starting with `#`
- or data lines with exactly 3 tab-separated fields

This code explicitly skips comment lines, so anything that reaches the split is expected to be a valid data line (i.e. 3 columns).

provider/source/src/source.rs
opnuub: Yes, the tick means traditional version of the radical (3' is zh-hant, 3 is zh-hans). I’m ignoring it here because both represent the same underlying radical. In practice this helped the han segmenter model, since removing equivalent variants reduced noise.

tools/make/download-repo-sources/src/main.rs
opnuub: It reduces the file size from ~12MB to ~2MB while preserving the Unihan format (comments + valid 3-column entries. I'm happy to write it as a derived file instead if we want to keep the raw download intact.

provider/source/src/source.rs
robertbastian: might be worth a comment

provider/source/src/source.rs
robertbastian: yes, but it panics if the file does not conform to the spec. we want to raise `DataError`s in such cases, not panic

sffc: I think we aren't super consistent in datagen about panic vs DataError for malformed data files.

PR #7507, opened 2026-01-27
Expose the UTS 46 virama check using UTS 46 data

hsivonen:
This reduces the optimized wasm size of dependents that don't use icu_normalizer or icu_collator for other purposes by 16.7 KB.

This does not break compat with the current `idna_adapter` but this allows for a new version of `idna_adapter` that uses this and then provides the 16.7 KB size saving.

Let's get this into ICU4X 2.2.

## Changelog

icu_normalizer: Expose the UTS 46 virama check using UTS 46 data

### DIFF
--- a/components/normalizer/src/uts46.rs
+++ b/components/normalizer/src/uts46.rs
@@ -62,6 +62,25 @@ impl Uts46MapperBorrowed<'static> {
 }
 
 impl Uts46MapperBorrowed<'_> {
+    /// Returns `true` iff the canonical combining class of `c` is 9 (Virama).
+    ///
+    /// This method uses the UTS 46 data and does not add a dependency on NFD
+    /// data like `CanonicalCombiningClassMapBorrowed` does.
+    #[inline]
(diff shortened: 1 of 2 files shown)


hsivonen:
I should mention that testing is end-to-end using `cargo test` is the `rust-url` repo with the dependency chain changed to local path deps.

Manishearth:
Clever trick.

components/normalizer/src/uts46.rs
```diff
@@ -62,6 +62,25 @@ impl Uts46MapperBorrowed<'static> {
...
+    #[inline]
+    pub fn is_virama(&self, c: char) -> bool {
+        let trie_val = self
+            .normalizer
+            .decomposing_normalizer
+            .decompositions
+            .trie
+            .get(c);
+        if crate::trie_value_has_ccc(trie_val) {
+            (trie_val as u8) == 9
```
robertbastian, line 78: issue: magic number

```suggestion
            (trie_val as u8) == ccc!(Virama, 9).to_icu4c_value()
```

Issue #567, opened 2021-03-23
Implement generalized decimal pattern parsing


sffc:
In #541, I use a temporary decimal pattern parsing algorithm as a placeholder for the main one that we will need to write in order to handle the full spectrum of CLDR data.

This algorithm will need to:

- Handle quoted literals in prefix and suffix
- Support currency symbols, including at the decimal separator (https://github.com/unicode-org/cldr/pull/1137)
- Support grouping digits in the fraction

I think we should consider following a model similar to that in DateTimeFormat, where we have a public `Pattern` type, which is serializable and allowed to live in the data struct, with a human-readable representation for JSON and a machine-readable representation for Bincode (and rkyv).

gregtatum:
> I think we should consider following a model similar to that in DateTimeFormat, where we have a public Pattern type, which is serializable and allowed to live in the data struct, with a human-readable representation for JSON and a machine-readable representation for Bincode (and rkyv).

👍

sffc:
Dependency of #275

sffc:
@younies is going to take a look at this.

sffc:
Needed for currency. Assigning to @robertbastian

PR #7508, opened 2026-01-27
Implement generalized LDML decimal pattern parsing

afrdbaig7:
#567 

This PR adds a generalized decimal pattern parser to icu_decimal in line with Unicode LDML number format specifications. It introduces DecimalPattern and DecimalPatternItem abstractions with support for standard LDML pattern symbols, quoted literals, and automatic merging of adjacent literals. The change includes unit tests covering common and edge cases and exports the pattern module behind the alloc feature, providing a reusable foundation for spec-compliant decimal formatting in ICU4X.


sffc:
Hi, thanks for the contribution! This does not fix #7121, which is for a different component. Could you link the correct issue number?

afrdbaig7:
> Hi, thanks for the contribution! This does not fix #7121, which is for a different component. Could you link the correct issue number?

You’re right — that was my mistake. This change does not address #7121.
The correct issue is #567, and I’ve updated the PR description accordingly.
Thanks for catching that.

afrdbaig7:
> I think this PR is attempting to solve the problem of decimal pattern parsing being overly specific in datagen. The change should land in the datagen component and should replace the hand-written match statement over decimal patterns there.

Thanks for the review — that clarification helps. I agree the core issue is the hand-written matching in provider/source/src/decimal/decimal_pattern.rs, and that it should be replaced with a more general solution.
I initially placed the parser in components/decimal based on the issue discussion (and by analogy with DateTimePattern), but I’m happy to adjust the approach to better fit the current architecture. My priority is to remove the brittle logic in datagen in the most appropriate way.

Would you prefer that I move the parsing logic directly into provider/source for now, or keep it in components/decimal and update datagen to consume it? I’m happy to follow whichever direction you think is more appropriate.

sffc:
Datetime Pattern is different because we use it at runtime. We don't use the decimal patterns at runtime; we convert them to another format. So they should live entirely in the datagen crates.

provider/source/src/decimal/decimal_pattern.rs
```diff
@@ -35,39 +117,142 @@ impl FromStr for DecimalSubPattern {
...
+        let body_end = body_end.unwrap_or(body_start);
+
+        // Extract prefix (all literals before body)
+        let prefix: String = items[..body_start]
+            .iter()
+            .filter_map(|item| {
+                if let DecimalPatternItem::Literal(s) = item {
+                    Some(s.as_str())
+                } else {
+                    None
```
robertbastian: I think we need to error if there's a non-literal in the prefix

provider/source/src/decimal/decimal_pattern.rs
```diff
@@ -35,39 +117,142 @@ impl FromStr for DecimalSubPattern {
...
+                        matches!(
+                            item,
+                            DecimalPatternItem::MandatoryDigit | DecimalPatternItem::OptionalDigit
+                        )
+                    })
+                    .count() as u8
+            } else {
+                digits_after_last
             };
+
```
robertbastian, line 259: reject if there are three grouping separators

provider/source/src/decimal/decimal_pattern.rs
```diff
@@ -35,39 +117,142 @@ impl FromStr for DecimalSubPattern {
...
+                if let DecimalPatternItem::Literal(s) = item {
+                    Some(s.as_str())
+                } else {
+                    None
+                }
+            })
+            .collect();
+
+        // Analyze the body to extract grouping and fraction info
+        let body_items = &items[body_start..=body_end];
```
robertbastian, line 187: I think the body should contain only `DecimalSeparator`, `GroupingSeparator`, `MandatoryDigit` and `OptionalDigit`. However your code allows arbitrary items, which it should probably reject?

you kind of run into this below where you count only digits, but there you ignore other items instead of erroring

provider/source/src/decimal/decimal_pattern.rs
```diff
@@ -35,39 +117,142 @@ impl FromStr for DecimalSubPattern {
...
+        };
+
+        // Calculate fraction digits from the fractional part
+        let (min_fraction_digits, max_fraction_digits) = if let Some(pos) = decimal_pos {
+            let fraction_items = &body_items[pos + 1..];
+            let mandatory: u8 = fraction_items
+                .iter()
+                .filter(|item| matches!(item, DecimalPatternItem::MandatoryDigit))
+                .count() as u8;
+            let optional: u8 = fraction_items
```
robertbastian, line 287: issue: this allows weird mixes of mandatory and optional digits. it should validate that it's a number of mandatory digits followed by a number of optional digits

afrdbaig7:
@robertbastian Thanks for the detailed review 

I’ve addressed the points you mentioned and updated the implementation to:
Return an error if the prefix or suffix contains any non-literal items (so things like currency symbols must be quoted)
Reject patterns with more than two grouping separators
Validate that the body contains only digit placeholders and valid separators
Ensure that mandatory digits come before optional digits in the fraction part
I’ve pushed an updated commit with these fixes. Please let me know if anything still looks off or if you’d like this validation handled differently.

components/locale_core/src/subtags/variants.rs
```diff
@@ -119,6 +119,58 @@ impl Variants {
...
+    /// use icu::locale::subtags::{variant, Variants};
+    ///
+    /// let mut variants = Variants::new();
+    /// assert!(variants.push(variant!("posix")));
+    /// assert!(!variants.push(variant!("posix"))); // Already present
+    /// assert!(variants.push(variant!("macos")));
+    /// assert_eq!(variants.to_string(), "macos-posix");
+    /// ```
+    #[cfg(feature = "alloc")]
+    pub fn push(&mut self, variant: Variant) -> bool {
```
sffc: Issue: Please put the variants changes in a separate PR.

afrdbaig7: Sorry about that I’ve removed the unrelated changes  and the PR now only focuses on decimal pattern parsing.

afrdbaig7:
> You have test failures because this handles the currency sign as a non-literal, which is correct, however the currency pattern parsing code needs to be updated. You can reproduce the test failure with this test: `provider/source/src/currency/essentials.rs:350:10`

Thanks  I’ve updated the prefix and suffix handling so that currency symbols and other affix characters are preserved properly. The tests are passing now.
Really appreciate you pointing that out still learning and trying to get these edge cases right

provider/source/src/decimal/decimal_pattern.rs
```diff
@@ -155,44 +155,34 @@ impl FromStr for DecimalSubPattern {
...
         let body_end = body_end.unwrap_or(body_start);
 
-        // Validate prefix: all items before body must be literals
+        // Validate and extract prefix: must be literals or affix symbols (¤, %, ‰, +, -)
+        let mut prefix = String::new();
         for item in &items[..body_start] {
-            if !matches!(item, DecimalPatternItem::Literal(_)) {
-                return Err(Error::InvalidAffixItem);
+            match item {
+                DecimalPatternItem::Literal(s) => prefix.push_str(s),
```
robertbastian, line 162: as a next step we should update the currency pattern parsing code to use `DecimalPatternItem` so we can remove these string replacements again

PR #7513, opened 2026-01-27
Move formatting code from Month back to MonthInfo

sffc:
One way to implement the suggestion in #7149

### DIFF
--- a/components/calendar/src/calendar_arithmetic.rs
+++ b/components/calendar/src/calendar_arithmetic.rs
@@ -558,7 +558,9 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
                 return true;
             }
         } else {
-            let target_month = cal.month_info_from_ordinal(target.year(), target.month()).value;
+            let target_month = cal
+                .month_info_from_ordinal(target.year(), target.month())
+                .value;
             if month != target_month {
                 let ordering = month.cmp_lexicographic(target_month);
                 if sign > 0 {
@@ -634,7 +636,9 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
         let y0 =
             cal.year_info_from_extended(duration.add_years_to(parts.year().to_extended_year()));
         // 1. If CompareSurpasses(_sign_, _y0_, _parts_.[[MonthCode]], _parts_.[[Day]], _calDate2_) is *true*, return *true*.
-        let base_month = cal.month_info_from_ordinal(parts.year(), parts.month()).value;
+        let base_month = cal
+            .month_info_from_ordinal(parts.year(), parts.month())
+            .value;
         if Self::compare_surpasses_lexicographic(sign, y0, base_month, parts.day(), cal_date_2, cal)
         {
             return true;


components/calendar/src/calendar_arithmetic.rs
```diff
@@ -192,8 +192,15 @@ pub(crate) trait DateFieldsResolver: Calendar {
     ///
     /// The default impl is for non-lunisolar calendars!
     #[inline]
-    fn month_from_ordinal(&self, _year: Self::YearInfo, ordinal_month: u8) -> Month {
-        Month::new_unchecked(ordinal_month, types::LeapStatus::Normal)
+    fn month_info_from_ordinal(&self, _year: Self::YearInfo, ordinal_month: u8) -> MonthInfo {
```
robertbastian, line 195: issue: right now this is a clean method that converts between ordinal month and month. you're making it into something that computes a full output struct with multiple codes, deprecated fields, and even the input. this is not an improvement

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -551,7 +558,9 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
                 return true;
             }
         } else {
-            let target_month = cal.month_from_ordinal(target.year(), target.month());
+            let target_month = cal
```
robertbastian, line 561: issue: all three call sites in this module only access the `.value` field, which shows that the method really should be `month_from_ordinal`

components/calendar/src/types.rs
```diff
@@ -669,7 +644,6 @@ pub struct MonthInfo {
     pub standard_code: MonthCode,
 
     /// The [`Month::formatting_code()`] of [`Self::value`].
-    #[deprecated(since = "2.2.0", note = "use `value.formatting_code()")]
     pub formatting_code: MonthCode,
```
robertbastian, line 647: issue: one of the reasons why we deprecated this is because this is a mutable field and the `MonthCode` type has no invariants. it's totally valid and possible for a `MonthInfo` with `formatting_code: MonthCode("boo")` to exist. With a getter like `value.formatting_code()` we don't have that problem

components/calendar/src/calendar_arithmetic.rs
sffc: It's a non-goal to beautify the internal trait DateFieldsResolver at this point. It is a messy trait. I could have added a new trait function to keep `month_from_ordinal` standalone but decided it wasn't a good ROI.

components/calendar/src/types.rs
sffc: I favored the smaller-touch approach since we already have this API in 2.1, but I'm willing to add a boolean field `is_formatting_leap` to the struct if you block on this.

components/calendar/src/calendar_arithmetic.rs
sffc: The function is inline so this should optimize out. I'm not looking to beautify the internal trait.

components/calendar/src/calendar_arithmetic.rs
robertbastian: it's not really messy. also there's a difference between improving and not making worse

components/calendar/src/calendar_arithmetic.rs
robertbastian: it's not about beauty, it's about writing readable maintainable code

components/calendar/src/types.rs
Manishearth: I am also in favor of the boolean field (it was my original suggestion)

components/calendar/src/calendar_arithmetic.rs
sffc: I don't consider reading a field from a struct that contains more than one field to be less readable or maintainable.

components/calendar/src/calendar_arithmetic.rs
sffc: I guess I disagree that this makes the trait worse

components/calendar/src/types.rs
sffc: I'm willing to do this but I won't put more time into this PR until I get consensus that it will be merged

Manishearth:
We moved leap status out of Month, so this is done

PR #7514, opened 2026-01-27
Implement spec behavior around date arithmetic

### LINKED ISSUE #7454: Implement updated date difference algorithm
Need to implement the change proposed and approved in https://github.com/tc39/proposal-intl-era-monthcode/pull/101.

sffc:
Fixes #7454

Maybe we want to pull out the cmp_lexicographic on Month, but I'll leave it private for now.

## Changelog: N/A

Will be covered by arithmetic stabilization PR.

### DIFF
--- a/components/calendar/src/calendar_arithmetic.rs
+++ b/components/calendar/src/calendar_arithmetic.rs
@@ -502,7 +502,7 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
     }
 
     /// Implements the Temporal abstract operation `CompareSurpasses` based on month code
-    #[allow(clippy::collapsible_else_if)] // to match the spec
+    #[allow(clippy::collapsible_if, clippy::collapsible_else_if)] // to match the spec
     fn compare_surpasses_lexicographic(
         sign: i64,
         year: C::YearInfo,
@@ -551,7 +551,7 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
     }
 
     /// Implements the Temporal abstract operation `CompareSurpasses` based on month ordinal
-    #[allow(clippy::collapsible_else_if)] // to match the spec
+    #[allow(clippy::collapsible_if, clippy::collapsible_else_if)] // to match the spec
     fn compare_surpasses_ordinal(
         sign: i64,
         year: C::YearInfo,


components/calendar/src/calendar_arithmetic.rs
```diff
@@ -501,6 +501,73 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
         Self::new_unchecked(resolved_year, resolved_month, resolved_day)
     }
 
+    fn compare_surpasses_lexicographic(sign: i64, year: C::YearInfo, month: Month, day: u8, target: &Self, cal: &C) -> bool {
```
Manishearth: nit: please document + link to spec text when available

components/calendar/src/types.rs
```diff
@@ -464,6 +465,19 @@ pub(crate) enum LeapStatus {
     FormattingLeap,
 }
 
+impl LeapStatus {
+    fn cmp_lexicographic(self, other: LeapStatus) -> Ordering {
```
robertbastian, line 469:
```suggestion
    fn cmp_as_code(self, other: LeapStatus) -> Ordering {
```

components/calendar/src/types.rs
```diff
@@ -599,6 +613,33 @@ impl Month {
...
+    pub(crate) fn cmp_lexicographic(self, other: Self) -> Ordering {
+        self.number
+            .cmp(&other.number)
+            .then_with(|| self.leap_status.cmp_lexicographic(other.leap_status))
+    }
+}
+
+#[test]
+fn test_cmp_lexicographic() {
+    let months_in_order = [
```
robertbastian, line 626: nit: you're not testing `::FormattingLeap`

components/calendar/tests/arithmetic.rs
```diff
@@ -207,6 +207,10 @@ fn test_tricky_leap_months() {
         Date::try_new_from_codes(None, year, month.code(), day, Hebrew).unwrap()
     }
 
+    fn chinese_date(year: i32, month: Month, day: u8) -> Date<ChineseTraditional> {
+        Date::try_new_from_codes(None, year, month.code(), day, ChineseTraditional::new()).unwrap()
+    }
```
robertbastian, line 212: this is `Date::try_new_chinese_traditional`, that PR landed at the same time but we should clean up

components/calendar/src/types.rs
sffc: "lexicographic" comes from the spec

components/calendar/src/types.rs
sffc: because I want to get rid of `::FormattingLeap` with a passion

#7513

components/calendar/tests/arithmetic.rs
sffc: https://github.com/unicode-org/icu4x/pull/7548

components/calendar/src/types.rs
robertbastian: because the spec operates on strings. this is not a string

also, can we please stop assuming that everyone breathes the Temporal spec?

components/calendar/src/types.rs
robertbastian: you should still test it because you wrote code handle it

components/calendar/src/types.rs
sffc: This particular section of code is literally written to align line-for-line, bug-for-bug with Temporal

robertbastian: This particular section of code? `types.rs`?

Manishearth: This is a private function written for the benefit of code where we have basically deferred to Temporal completely. I think it's _fine_ to match the spec naming here. Ideally it would be documented more.

I also think it's fine to have more broadly applicable naming.

PR #7515, opened 2026-01-28
added missing CLDR skeleton patterns

### LINKED ISSUE #7428: Use Hv, EH, and other new skeletons in datagen
See https://unicode-org.atlassian.net/browse/CLDR-18539

CLDR added these patterns but I think we don't read them yet in datagen.

psg-19:
Fixes #7428 
Summary :- added [new datetime  skeletons](https://github.com/unicode-org/icu4x/blob/main/provider/source/tests/data/cldr/cldr-dates-full/main/und/ca-gregorian.json#L299-L353) patterns to SUPPORTED_STRING_SKELETONS.

### DIFF
--- a/components/datetime/src/dynamic.rs
+++ b/components/datetime/src/dynamic.rs
@@ -479,6 +479,9 @@ impl_attrs! {
         (ATTR_T, STR_T, "j"),
         (ATTR_T12, STR_T12, "h"),
         (ATTR_T24, STR_T24, "h0"),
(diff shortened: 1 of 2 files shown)


psg-19:
> This doesn't utilize these patterns. It just adds some metadata for them.

yes I guess what I did just adds some test coverage, not actual utilization 😅

robertbastian:
un-draft once CI is passing

PR #7518, opened 2026-01-28
exposed is_in_leap_year for Date

### LINKED ISSUE #5654: Add Temporal CalendarInLeapYear to FFI
`is_in_leap_year` is only [exposed](https://github.com/unicode-org/icu4x/blob/52d7c0b304121b0412e5359401b354f4b4f9cfa5/ffi/capi/src/date.rs#L138-L143) for `IsoDate`, but not for `Date`. 

See also #3963.

psg-19:
Fixes #5654 
exposed `is_in_leap_year` for `Date` in [date.rs](https://github.com/unicode-org/icu4x/blob/main/ffi/capi/src/date.rs)

## Changelog

icu_capi: Expose `Date::is_in_leap_year()`


psg-19:
Yeah sure, will do it soon.

psg-19:
> I see, the issue is that the fn is exported on IsoDate but not Date.
> 
> Would you mind checking that there are no other functions on IsoDate that might be missing on Date?

Hi @sffc , I found `week_of_year()` to be missing on Date, but we need to convert to `ISO` first using `to_calendar(ISO)` method, then call `week_of_year()` because there is no `week_of_year()` for `icu_calendar::Date`. Is it a good idea to convert calenders like this ? 
Plus the return type of this function would be `IsoWeekOfYear` .

sffc:
Not having `week_of_year` on Date might be expected; we only support week formatting in the ISO calendar.

PR #7519, opened 2026-01-28
Expose BCP-47 locale variant APIs over FFI

### LINKED ISSUE #6671: FFI for locale variants
There are getters and setters for language/script/region but it appears there is currently no way to access the locale variants from FFI: https://github.com/unicode-org/icu4x/blob/ed457c605c06a07101b20c68d0b1ca20656ee5b5/components/locale_core/src/langid.rs#L88

I need this to implement https://github.com/tc39/ecma402/pull/960.

afrdbaig7:
This PR adds FFI support for accessing and mutating BCP-47 locale variants so that non-Rust consumers (C/C++/JS/Dart) can work with variant subtags in a safe and consistent way.
It exposes a small and focused set of Locale APIs to query, add, remove, and clear variants. Mutation operations are implemented by operating on a temporary collection, followed by sorting and deduplication before reconstructing the internal representation, ensuring canonical BCP-47 ordering is preserved.
The change is intentionally scoped only to locale variants and does not alter existing Locale semantics. Unit tests are included to cover add/remove behavior, duplicate handling, canonical sorting, invalid variant input, and bounds checking.
This work is intended to improve feature parity for FFI consumers while keeping the API surface minimal and maintainable.

Closes #6671.

## Changelog

icu_capi: Expose BCP-47 locale variant APIs over FFI
 - New methods: `variants()`, `variant_count()`, `variant_at()`, `has_variant()`, `remove_variant()`, `clear_variants()` on `Locale`

icu_locale_core: Add locale variant mutation APIs
  - New methods: `Variants::push()`, `Variants::remove()`


ffi/capi/src/locale_core.rs
```diff
@@ -199,3 +273,55 @@ impl ffi::Locale {
         (&self.0).into()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_variants_basic() {
```
Manishearth: issue: I don't think these tests add much. If you want to test FFI, write an FFI test under examples/cpp or examples/npm or examples/dart.

ffi/capi/src/locale_core.rs
```diff
@@ -149,6 +149,80 @@ pub mod ffi {
...
+        }
+
+        /// Returns the number of variants in this [`Locale`].
+        #[diplomat::rust_link(icu::locale::Locale::id, StructField)]
+        #[diplomat::attr(auto, getter)]
+        pub fn variant_count(&self) -> usize {
+            self.0.id.variants.len()
+        }
+
+        /// Writes the variant at the given index to `write`.
```
Manishearth: issue: here and elsewhere, "gets" or "returns", not "writes". The Write infrastructure is largely internal to how FFI works, not exposed to the users of most FFI backends.

ffi/capi/src/locale_core.rs
```diff
@@ -149,6 +149,80 @@ pub mod ffi {
...
+        ///
+        /// Does nothing if the variant is already present.
+        #[diplomat::rust_link(icu::locale::Locale::id, StructField)]
+        pub fn add_variant(&mut self, s: &DiplomatStr) -> Result<(), LocaleParseError> {
+            let variant = icu_locale_core::subtags::Variant::try_from_utf8(s)?;
+            let mut variants = self.0.id.variants.to_vec();
+            if let Err(i) = variants.binary_search(&variant) {
+                variants.insert(i, variant);
+                self.0.id.variants =
+                    icu_locale_core::subtags::Variants::from_vec_unchecked(variants);
```
Manishearth: nit: document the invariant here and why it is upheld

(the invariant is that the vector must be sorted and deduplicated)

ffi/capi/src/locale_core.rs
afrdbaig7: Thanks, that makes sense. I agree these tests don’t really validate the FFI surface. I’ll remove them and add proper FFI-level tests under examples/cpp to exercise the API from a real consumer’s perspective.

ffi/capi/src/locale_core.rs
afrdbaig7: Got it — I’ll update the doc comments to use “gets” / “returns” instead of “writes” so they reflect the user-facing behavior rather than the internal FFI mechanics.

ffi/capi/src/locale_core.rs
afrdbaig7: Got it. I'll add a comment explaining the invariant. Something like:
// Safety: `Variants::from_vec_unchecked` expects the input to be sorted and
// deduplicated. We preserve this invariant by starting from the existing
// canonical variant list and inserting the new variant via `binary_search`.
// Inserting at the returned index keeps the list sorted, and we only insert
// when the variant is not already present, so duplicates cannot occur.

Manishearth: This is not a safety invariant, it should not use the word "safety"

afrdbaig7: thanks. I'll use a regular comment since this is a correctness invariant, not a safety one:
// Invariant: `from_vec_unchecked` requires sorted and deduplicated input.
// - binary_search maintains sorted order via correct insertion index.
// - Insert only on Err (not found) ensures no duplicates.

ffi/capi/src/locale_core.rs
```diff
@@ -149,6 +149,82 @@ pub mod ffi {
...
+            icu_locale_core::subtags::Variant::try_from_utf8(s)
+                .map(|v| self.0.id.variants.contains(&v))
+                .unwrap_or(false)
+        }
+
+        /// Adds a variant to the [`Locale`].
+        ///
+        /// Returns an error if the variant string is invalid. Does nothing if the variant is already present.
+        #[diplomat::rust_link(icu::locale::Locale::id, StructField)]
+        pub fn add_variant(&mut self, s: &DiplomatStr) -> Result<(), LocaleParseError> {
```
robertbastian: issue: this is a lot of logic to live in the FFI layer. also, this operation is not currently available in Rust. please add this to the Rust `Variants` type as `Variants::push` or something, and then call that from here

ffi/capi/src/locale_core.rs
```diff
@@ -149,6 +149,82 @@ pub mod ffi {
...
+            let variant = icu_locale_core::subtags::Variant::try_from_utf8(s)?;
```
robertbastian, line 202: issue: this clones the vector only to be overwritten later. if you move this code to `icu_locale_core` you'll have access to the internal field, which you can then mutate without cloning.

ffi/capi/src/locale_core.rs
```diff
@@ -149,6 +149,82 @@ pub mod ffi {
...
+                    icu_locale_core::subtags::Variants::from_vec_unchecked(variants);
+            }
+            Ok(())
+        }
+
+        /// Removes a specific variant from the [`Locale`].
+        ///
+        /// Returns an error if the variant string is invalid. Does nothing if the variant is not present.
+        #[diplomat::rust_link(icu::locale::Locale::id, StructField)]
+        pub fn remove_variant(&mut self, s: &DiplomatStr) -> Result<(), LocaleParseError> {
```
robertbastian: also move this to `icu_locale_core`

robertbastian: this is not a parsing operation, so the error shouldn't be `LocaleParseError`. you should either return `()` or `bool`

ffi/capi/src/locale_core.rs
```diff
@@ -149,6 +149,82 @@ pub mod ffi {
             Ok(())
         }
 
+        // --- Variants ---
+
+        /// Returns a string representation of the [`Locale`] variants.
+        #[diplomat::rust_link(icu::locale::Locale::id, StructField)]
```
robertbastian: issue: wrong `rust_link`
```suggestion
        #[diplomat::rust_link(icu::locale::Variants, Struct)]
```

also everywhere below

ffi/capi/src/locale_core.rs
```diff
@@ -149,6 +149,82 @@ pub mod ffi {
...
+        }
+
+        /// Returns the number of variants in this [`Locale`].
+        #[diplomat::rust_link(icu::locale::Locale::id, StructField)]
+        #[diplomat::attr(auto, getter)]
+        pub fn variant_count(&self) -> usize {
+            self.0.id.variants.len()
+        }
+
+        /// Returns the variant at the given index, or `None` if the index is out of bounds.
```
robertbastian: nit: don't say `None` in docs, as different language have different mechanism for an absent value

```suggestion
        /// Returns the variant at the given index, or nothing if the index is out of bounds.
```

afrdbaig7:
@robertbastian  Thanks a lot for the detailed review. I’ve gone through the feedback and made the following updates:

Moved the mutation logic into Variants::push and Variants::remove in icu_locale_core, so the FFI layer now just calls into these APIs directly.
Switched to ShortBoxSlice::insert / remove to avoid the unnecessary clone.
Changed remove_variant to return bool instead of LocaleParseError, since error propagation wasn’t really needed there.
Fixed the rust_link attributes to correctly point to icu::locale::Variants.
Updated the doc comment wording from “None” to “nothing”.
Please let me know if there’s anything I missed or if you’d like this structured differently.

robertbastian:
you need to run `cargo make diplomat-gen`

afrdbaig7:
> you need to run `cargo make diplomat-gen`

Ran cargo make diplomat-gen and force-pushed the updated changes.
Apologies for missing the bindings earlier that was an oversight on my end. I’ve double-checked everything now to make sure the generated files are correctly included and aligned with the current changes.
Thanks for pointing it out. Please let me know if there’s anything else I should improve  happy to fix it right away.

robertbastian:
Please don't force push once the review process has started. I have no way to see a diff to the last reviewed state now

afrdbaig7:
> Please don't force push once the review process has started. I have no way to see a diff to the last reviewed state now

Sorry about that — I realize force-pushing during review makes it harder to track changes. I’ll keep future commits additive so the diff stays clear and easier to review. Thanks for the patience, and I appreciate the feedback.

robertbastian:
you did it again

afrdbaig7:
> you did it again

Hi, I’m really sorry about this. I made the same mistake again by squashing and force-pushing, and I understand now how that breaks the review flow and makes things harder for you. That wasn’t careful on my part.
I should have stuck to additive commits and preserved the history properly. I won’t repeat this again — from now on I’ll only push incremental changes so the review context stays intact.
Thanks for your patience, and sorry again for the inconvenience.

sffc:
Kotlin CI is failing:

```
Error:  /Users/runner/work/icu4x/icu4x/ffi/mvn/src/main/kotlin/src/main/kotlin/org/unicode/icu4x/Locale.kt: (299, 30) Destructuring of type 'org.unicode.icu4x.GCSlice' requires operator function 'component1()'.
```

afrdbaig7:
> Kotlin CI is failing:
> 
> ```
> Error:  /Users/runner/work/icu4x/icu4x/ffi/mvn/src/main/kotlin/src/main/kotlin/org/unicode/icu4x/Locale.kt: (299, 30) Destructuring of type 'org.unicode.icu4x.GCSlice' requires operator function 'component1()'.
> ```

The Kotlin CI failure seems to be coming from the diplomat Kotlin codegen step. The error mentions GCSlice, but I don’t see that anywhere in the bindings generated on my local setup.
I’m wondering if CI might be using a different diplomat version than what I have locally, which could explain the mismatch. I’ll try to check the versions and see if I can reproduce the issue on my side. Please let me know if there’s anything specific I should look into.

robertbastian:
I think it's just Kotlin being broken, please annotate the failing methods with 

```rust
#[diplomat::attr(kotlin, disable)]
```

and regenerate to remove the Kotlin bindings.

PR #7521, opened 2026-01-28
Use pinned Dart SDK in nightly CI for now

Manishearth:
Until https://github.com/orgs/dart-lang/projects/99 completes (~2 months) this CI will keep failing and the only mitigation will be pinning to unreleased Git hashes of `record_use` which we'd rather not do.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -336,7 +336,12 @@ jobs:
     - uses: dart-lang/setup-dart@0a8a0fc875eb934c15d08629302413c671d3f672 # v1.6.5
       with:
         # Keep in sync with artifacts-build.yml
-        sdk: ${{ github.event_name == 'schedule' && github.event.schedule == '0 14 * * *' && 'dev' || '3.11.0-46.0.dev' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test.yml
```diff
@@ -336,7 +336,12 @@ jobs:
     - uses: dart-lang/setup-dart@0a8a0fc875eb934c15d08629302413c671d3f672 # v1.6.5
       with:
         # Keep in sync with artifacts-build.yml
-        sdk: ${{ github.event_name == 'schedule' && github.event.schedule == '0 14 * * *' && 'dev' || '3.11.0-46.0.dev' }}
+        sdk: 3.11.0-46.0.dev
```
robertbastian, line 339: if the first falling build is `3.12.0-62.0.dev`, we should keep testing on the previous one. also keeping the nightly logic will make it easier to reenable this in the future.

```suggestion
        sdk: ${{ github.event_name == 'schedule' && github.event.schedule == '0 14 * * *' && '3.12.0-49.0.dev' || '3.11.0-46.0.dev' }}```

PR #7522, opened 2026-01-29
Testing date-time-zone formatting for locales with nondefault appendItems

cloud-machine-111:
Follow-up on #7416. Adds testing data for 'zh' and 'cv' locales, which have non-default timezone append patterns under the Gregorian calendar.

### DIFF
--- a/components/datetime/tests/fixtures/tests/components_with_zones.json
+++ b/components/datetime/tests/fixtures/tests/components_with_zones.json
@@ -170,7 +170,7 @@
             }
         }
     },
-{
+    {
         "description": "Full date time example with long generic time zone. For locale with non-default timezone appendItem (zh, {1}{0} rather than {1} {0}).",
         "input": {
             "value": "2020-01-21T08:25:07.000+05:00",


sffc (approved):
Thank you!

As far as I can tell, the output agrees with the spec and the data, but it produces unintelligible results in Chinese. I filed CLDR issues about this:

https://unicode-org.atlassian.net/browse/CLDR-19226
https://unicode-org.atlassian.net/browse/CLDR-19227

components/datetime/tests/fixtures/tests/components_with_zones.json
```diff
@@ -169,5 +169,63 @@
...
+            }
+        },
+        "output": {
+            "values": {
+                "zh-u-hc-h23": "GMT+52020年1月21日星期二 08:25:07"
+            }
+        }
+    },
+    {
+        "description": "Full date time example with short specific time zone. For locale with non-default timezone appendItem (cv, {0} {1} rather than {1} {0}).",
```
robertbastian, line 203: are you sure this is `{0} {1}`? it looks like `{1} ({0})` to me, i.e. the same order as in english (date, time, zone), but with parentheses

cloud-machine-111: From what I can tell, test behavior for both locales matches CLDR data, but my comments were misleading. This was entirely my bad!

To my understanding, `{0} {1}` means the zone comes last, since CLDR uses `{1}` to mean "thing to append to `{0}`", and in this case, `{1}` is "timezone". ([Reference](https://github.com/unicode-org/icu4x/pull/7416#discussion_r2687702190))

So [zh-gregorian](https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-dates-full/main/zh/ca-gregorian.json) and [cv-gregorian](https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-dates-full/main/cv/ca-gregorian.json) correctly use CLDR patterns, but I'd accidentally switched the order of the default appendItem in both descriptions (should be `{0} {1}`, not `{1} {0}`) and left out the parentheses that _do_ exist the `cv` appenditem.

I'll make a follow-up updating descriptions as follows:
zh: "Full date time example with long generic time zone. For locale with non-default timezone appendItem (zh, {1}{0} rather than {0} {1})."

cv: "Full date time example with short specific time zone. For locale with non-default timezone appendItem (cv, {0} ({1}) rather than {0} {1})."

PR #7527, opened 2026-01-30
Add date arithmetic tests

mega123-art:
Implementation of comprehensive date arithmetic coverage tests as suggested in #7215.
image

### DIFF
--- a/components/calendar/src/tests/arithmetic.rs
+++ b/components/calendar/src/tests/arithmetic.rs
@@ -4,11 +4,11 @@
 
 use crate::options::{DateAddOptions, DateDifferenceOptions, Overflow};
 use crate::types::{DateDuration, DateDurationUnit};
-use crate::{AsCalendar, Date, Iso};
+use crate::{AsCalendar, Date};
 
 super::test_all_cals!(
     #[ignore]
-    fn test_arithmetic_exhaustive<C: Calendar>(cal: Ref<C>) {
+    fn test_arithmetic<C: Calendar>(cal: Ref<C>) {
         fn new_duration(years: i32, months: i32, weeks: i32, days: i32) -> DateDuration {
             let is_negative = years < 0 || months < 0 || weeks < 0 || days < 0;
             // Verify no mixed signs
@@ -29,12 +29,12 @@ super::test_all_cals!(
         let mut durations = Vec::new();
 
         // Check +/- 65 days
-        for i in -65i32..=65 {
+        for i in -65..=65 {
             durations.push(new_duration(0, 0, 0, i));
         }
 
         // Check +/- 30 months
-        for i in -30i32..=30 {
+        for i in -30..=30 {
             durations.push(new_duration(0, i, 0, 0));
         }
 
@@ -45,7 +45,7 @@ super::test_all_cals!(
         }
 
         // Check +/- 10 years
-        for i in -10i32..=10 {
+        for i in -10..=10 {
             durations.push(new_duration(i, 0, 0, 0));
         }
 
@@ -72,8 +72,9 @@ super::test_all_cals!(
         let start_rd = start_date.to_rata_die();
         let end_rd = end_date.to_rata_die();
 
-        let mut add_options = DateAddOptions::default();
-        add_options.overflow = Some(Overflow::Constrain);
+        let add_options = DateAddOptions {
+            overflow: Some(Overflow::Constrain),
+        };
 
         for rd_offset in 0..=(end_rd - start_rd) {
             let date = Date::from_rata_die(start_rd + rd_offset, cal);
@@ -88,29 +89,30 @@ super::test_all_cals!(
                     diff_options.largest_unit = Some(DateDurationUnit::Days);
                 }
 
-                let added_date = match date.clone().try_added_with_options(*duration, add_options) {
-                    Ok(d) => d,
-                    Err(_) => {
+                let added_date = date
+                    .try_added_with_options(*duration, add_options)
+                    .unwrap_or_else(|_| {
                         panic!(
                             "Failed to add duration {:?} to date {:?} in calendar {:?}",
-                            duration, date, cal.as_calendar().debug_name()
-                        );
-                    }
-                };
+                            duration,
+                            date,
+                            cal.as_calendar().debug_name()
+                        )
+                    });
 
-                let calculated_duration =
-                    match date.try_until_with_options(&added_date, diff_options) {
-                        Ok(d) => d,
-                        Err(_) => {
-                            panic!(
-                                "Failed to calculate difference between {:?} and {:?} in calendar {:?}",
-                                date, added_date, cal.as_calendar().debug_name()
-                            );
-                        }
-                    };
+                let calculated_duration = date
+                    .try_until_with_options(&added_date, diff_options)
+                    .unwrap_or_else(|_| {
+                        panic!(
+                            "Failed to calculate difference between {:?} and {:?} in calendar {:?}",
+                            date,
+                            added_date,
+                            cal.as_calendar().debug_name()
+                        )
+                    });
 
                 // Round-trip check
-                let added_back = date.clone()
+                let added_back = date
                     .try_added_with_options(calculated_duration, add_options)
                     .unwrap();
                 assert_eq!(
--- a/components/calendar/src/tests/mod.rs
+++ b/components/calendar/src/tests/mod.rs
@@ -2,10 +2,10 @@
 // called LICENSE at the top level of the ICU4X source tree
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
+mod arithmetic;
 mod continuity_test;
 mod exhaustive;
 mod extrema;
-mod arithmetic_exhaustive;
 mod not_enough_fields;
 
 macro_rules! test_all_cals {


robertbastian (commented):
please use the `test_all_cals!` macro. it decouples these tests from `AnyCalendar` (and actually tests all calendars), and also generates one test function per calendar so they can fail more granularly

mega123-art:
@robertbastian please check again i made those suggested changes

components/calendar/src/tests/arithmetic_exhaustive.rs
```diff
@@ -0,0 +1,125 @@
...
+        let end_date = Date::try_new_iso(2004, 12, 31).unwrap();
+        let start_rd = start_date.to_rata_die();
+        let end_rd = end_date.to_rata_die();
+
+        let mut add_options = DateAddOptions::default();
+        add_options.overflow = Some(Overflow::Constrain);
+
+        for rd_offset in 0..=(end_rd - start_rd) {
+            let iso_date = Date::from_rata_die(start_rd + rd_offset, Iso);
+            let date = iso_date.to_calendar(cal);
```
robertbastian: no need to go through ISO

```suggestion
            let date = Date::from_rata_die(start_rd + rd_offset, cal);
```

components/calendar/src/tests/arithmetic_exhaustive.rs
```diff
@@ -0,0 +1,125 @@
...
+                                "Failed to calculate difference between {:?} and {:?} in calendar {:?}",
+                                date, added_date, cal.as_calendar().debug_name()
+                            );
+                        }
+                    };
+
+                // Round-trip check
+                let added_back = date.clone()
+                    .try_added_with_options(calculated_duration, add_options)
+                    .unwrap();
```
robertbastian: can't you just compare `duration` with `calculated_duration`? this would avoid doing another `try_added_with_options`

components/calendar/src/tests/arithmetic_exhaustive.rs
```diff
@@ -0,0 +1,125 @@
...
+
+                let added_date = match date.clone().try_added_with_options(*duration, add_options) {
+                    Ok(d) => d,
+                    Err(_) => {
+                        panic!(
+                            "Failed to add duration {:?} to date {:?} in calendar {:?}",
+                            duration, date, cal.as_calendar().debug_name()
+                        );
+                    }
+                };
```
robertbastian:
```suggestion
                let added_date = match date.clone().try_added_with_options(*duration, add_options).unwrap_or_else(|_| 
                panic!(
                    "Failed to add duration {:?} to date {:?} in calendar {:?}",
                    duration, date, cal.as_calendar().debug_name()
                ));
```

components/calendar/src/tests/arithmetic_exhaustive.rs
```diff
@@ -0,0 +1,120 @@
...
+                let added_date = match date.clone().try_added_with_options(*duration, add_options).unwrap_or_else(|_| 
+                panic!(
+                    "Failed to add duration {:?} to date {:?} in calendar {:?}",
+                    duration, date, cal.as_calendar().debug_name()
+                ));
+
+                let calculated_duration =
+                    match date.try_until_with_options(&added_date, diff_options) {
+                        Ok(d) => d,
+                        Err(_) => {
```
robertbastian: this is also `unwrap_or_else`

components/calendar/src/tests/arithmetic.rs
```diff
@@ -0,0 +1,120 @@
...
+
+            for duration in &durations {
+                let mut diff_options = DateDifferenceOptions::default();
+                if duration.years != 0 {
+                    diff_options.largest_unit = Some(DateDurationUnit::Years);
+                } else if duration.months != 0 {
+                    diff_options.largest_unit = Some(DateDurationUnit::Months);
+                } else {
+                    diff_options.largest_unit = Some(DateDurationUnit::Days);
+                }
```
robertbastian, line 90:
```suggestion
                diff_options.largest_unit = Some(if duration.years != 0 {
                    DateDurationUnit::Years
                } else if duration.months != 0 {
                    DateDurationUnit::Months
                } else {
                    DateDurationUnit::Days
                });
```

robertbastian: thought: we should define `DateDuration::largest_unit`

components/calendar/src/tests/arithmetic_exhaustive.rs
```diff
@@ -0,0 +1,120 @@
...
+                months: months.unsigned_abs(),
+                weeks: weeks.unsigned_abs(),
+                days: days.unsigned_abs() as u64,
+            }
+        }
+
+        let mut durations = Vec::new();
+
+        // Check +/- 65 days
+        for i in -65i32..=65 {
```
robertbastian: nit: the integer types will be inferred from the signature of `new_duration`


```suggestion
        for i in -65..=65 {
```

mega123-art:
@robertbastian sorry again just can u give a final check i made all changes as suggested
it is my first pr here

components/calendar/src/tests/arithmetic_exhaustive.rs
mega123-art: The round-trip check date + calculated_duration == added_date is necessary because DateDuration equality is structural, not semantic.

For example, adding "13 months" might be semantically equivalent to adding "1 year and 1 month", or "12 months and 30 days" depending on the calendar and calculation. The until calculation returns a normalized duration (e.g., using Days if requested, or maximizing Years /Months), which might not strictly equal the original input duration struct, even though they represent the same passage of time.

By adding the calculated duration back to the start date, we verify that the calculated duration is effectively correct, which is the goal of the test.

robertbastian: hmm, I think it should test both then.

> For example, adding "13 months" might be semantically equivalent to adding "1 year and 1 month", or "12 months and 30 days"

But you solve this by setting `diff_options.largest_unit`, no?

mega123-art: largest_unit ensures we get "13 months" instead of "1 year 1 month", but it doesn't force "13 months" when the day arithmetic clamps.

Example from the test failure:

Start: 2000-01-29
Input: 13 months
Added Date: 2001-02-28 (Clamped from non-existent Feb 29/30/31)
Calculated Diff: 12 months, 30 days (because Jan 29 to Feb 28 is not a full month).
So 13 months != 12 months, 30 days structurally, causing assert_eq! to fail. The round-trip check date + calculated_duration == added_date confirms that 12 months, 30 days is strictly correct for that specific date pair.

mega123-art:
> it's a shame that we need to `#[ignore]` these. how long do these tests take?

I benchmarked the tests without #[ignore]. The full suite takes ~19 seconds in parallel (and >50s single-threaded). Since this significantly slows down the default cargo test loop, I believe it's best to keep them ignored.

mega123-art:
> oh that's unfortunate. I hope we can improve performance and unignore this at some point.
> 
> one last thing, please don't call this test "exhaustive", because it's not, it only checks a few dates and a few durations

Makes sense. I'll rename the test file to 
arithmetic.rs (or similar) since it's testing a sampled range rather than being truly exhaustive.
then will be merged? or any changes required

robertbastian:
> then will be merged? or any changes required

looks good but you need to fix CI

mega123-art:
> > then will be merged? or any changes required
> 
> looks good but you need to fix CI

yaa right i saw the fmt and clippy one 
i am on it 
thanks a lot sir

components/calendar/src/tests/arithmetic.rs
```diff
@@ -0,0 +1,126 @@
...
+// called LICENSE at the top level of the ICU4X source tree
+// (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
+use crate::options::{DateAddOptions, DateDifferenceOptions, Overflow};
+use crate::types::{DateDuration, DateDurationUnit};
+use crate::{AsCalendar, Date};
+
+super::test_all_cals!(
+    #[ignore]
+    fn test_arithmetic<C: Calendar>(cal: Ref<C>) {
```
sffc, line 11: Nit: Don't put the whole test function in the macro. Define the function outside the macro and use the macro to call the function. Why: various tools like fmt, IDEs, rust-analyzer, etc., don't work as well when code is inside of a macro.

mega123-art: so should i appply these suggested changes or foucus on some other issue

sffc: Consider this feedback on the PR that didn't make it in before @robertbastian merged it. Please make a follow-up PR.

mega123-art: [https://github.com/unicode-org/icu4x/pull/7552](url)
made this follow up pr sir
thanks

robertbastian: > Why: various tools like fmt, IDEs, rust-analyzer, etc., don't work as well when code is inside of a macro.

This is not true for rust-analyzer, which is what every IDE uses. We use this pattern for all tests that use `test_all_cals!`

Issue #7526, opened 2026-01-29
Land normalizer perf improvements


hsivonen:
I've developed multiple normalizer perf improvements to get the normalizer into a state where replacing ICU4C normalizer in Gecko would be appropriate.

It makes sense to review this as one big thing and land them in parts like this:

1. Land `icu_collections` changes; publish to crates.io.
2. Publish new versions of `utf8_iter` and `utf16_iter` depending on the previous to crates.io.
3. Land `icu_normalizer` and `icu_collator` changes that depend on both the above items.

Manishearth:
- @hsivonen This involves changes to data structs. Could I have help from @Manishearth for writing the needed code?
- @Manishearth I could take a look when you're ready
- @sffc How do we feel about making this a 2.2 headline?
- @hsivonen With a mid-March timeline, chances are this could get in, but not sure.
- @Manishearth Headline means that the release driver needs to track it.
- @zbraniecki We could highlight it, too.
- @sffc Henri was saying that this is a multi-step process. We don't want to ship a release with half the normalizer improvements but not the other half.
- @Manishearth I don't want to increase the release driver burden. Henri should just post in Slack when he's doing this work.
- @hsivonen Ok

Manishearth:
This is not going to make it for 2.2, and is milestone-non-blocking. Hopefully 2.3, will retain the milestone-non-blocking unless we decide this is crucial for that release.

sffc:
Doesn't look like this is going to make it for 2.3, but let's try hard for 2.4, because then we'll need to start integrating Unicode 18, and I don't want to make big merge conflicts.

Manishearth:
I would prefer to treat this as nice-to-have for 2.4, but let @hsivonen make the call on whether he wishes to prioritize landing it and avoiding PR bitrot, or allowing for bitrot and having time to do this.


I do not actually expect the Unicode 18 update to cause much PR bitrot for the strategy we have chosen. The current PRs already need major refactors to work with the proposed model.

So: we should have this be in the release milestone as an encouraging, but non-blocking, and "try really hard" is up to Henri's own tradeoffs IMO.

Issue #7450, opened 2026-01-20
Do not mark lead jamo as potential passthrough


hsivonen:
In principle, a lead jamo has the characteristics that make it eligible to be marked as potential NFC passthrough. However, in practice a lead jamo is likely to be followed by a V jamo. In that case, attempting the NFC fast path is just waste.

Even though doing so would violate a documented "if and only if", it appears possible, without an actual data semver break, to mark lead jamo as non-round-trippable (set bit 30 to 1 in the trie value).

We should pursue doing that.

hsivonen:
I have code for this is my upcoming megapatch.

Issue #6838, opened 2025-08-12
Experiment with trie access by UTF-8 bytes for str normalization


hsivonen:
The code point trie is designed to distribute the bits in the fast case in a way that fits the UTF-8 bit distribution. Specifically, in the fast case, the inner access is by 6 bits, which are already the low 6 bits of the last UTF-8 byte.

On the `str` normalization fast path, we know that UTF-8 is well-formed, so we could decide the UTF-8 sequence length from the UTF-8 lead byte without validation. Experiment with having trie getters for each UTF-8 sequence length, querying the trie by UTF-8 bytes on the fast path, and computing a `char` only if we fall off the fast path.

(This should work better if the trie type is not dynamic but always assumed to be fast.)

PR #7528, opened 2026-01-30
Review normalizer changes

hsivonen:
For #7526.

This PR is meant for review, not for landing. See #7526 for the landing mechanics. CI won't pass, since this has local path references to `utf8_iter` and `utf16_iter`.

This PR is missing the code for constructing with old-style canonical composition data.

This PR requires the decomposition data to use a fast trie, unless the `serde` feature is used. What does this mean for struct versioning?

The numeric passthrough bounds in the data are pretty useless. Since all characters below U+0300 are assigned, their normalization characteristics won't change per Unicode policies, so we might as well not have passthrough bounds in data and could set a single flag in the constructors to remember if we have K or non-K form, and then choose hard-code bounds based on that flag.

The bounds are now only used for slice-mode UTF-16 and in the `str` case in the composing normalizer. The latter now makes it a safety issue that the bounds must be in the two-byte UTF-8 range. UTS 46 only uses the iterator mode, so it can have a lower bound (that's useless for optimization).


Manishearth:
Do you have links to the discussion where we first worked out the architecture for this?

components/normalizer/src/lib.rs
```diff
@@ -344,65 +379,374 @@ fn in_inclusive_range16(u: u16, start: u16, end: u16) -> bool {
...
-    if v < HANGUL_V_COUNT {
-        let l = u32::from(starter).wrapping_sub(HANGUL_L_BASE);
-        if l < HANGUL_L_COUNT {
-            let lv = l * HANGUL_N_COUNT + v * HANGUL_T_COUNT;
-            // Safe, because the inputs are known to be in range.
-            return Some(unsafe { char::from_u32_unchecked(HANGUL_S_BASE + lv) });
+#[derive(Debug)]
+pub(crate) enum CanonicalCompositionsPayload {
+    Current(DataPayload<NormalizerNfcV2>),
+    #[cfg(feature = "serde")]
```
Manishearth, line 385: observation: this seems like a good way to do this

components/normalizer/src/lib.rs
```diff
@@ -344,65 +379,374 @@ fn in_inclusive_range16(u: u16, start: u16, end: u16) -> bool {
...
-            let lvt = lv + (u32::from(second) - HANGUL_T_BASE);
-            // Safe, because the inputs are known to be in range.
-            return Some(unsafe { char::from_u32_unchecked(HANGUL_S_BASE + lvt) });
+    }
+}
+
+#[derive(Debug, Copy, Clone)]
+pub(crate) enum CanonicalCompositionsBorrowed<'data> {
+    Current(&'data CanonicalCompositionsNew<'data>),
+    #[cfg(feature = "serde")]
```
Manishearth, line 406: observation: same

components/collections/src/codepointtrie/cptrie.rs
```diff
@@ -579,14 +585,23 @@ impl<'trie, T: TrieValue> CodePointTrie<'trie, T> {
         );
 
         let bit_prefix = (code_point as usize) >> FAST_TYPE_SHIFT;
+        let bit_suffix = (code_point & FAST_TYPE_DATA_MASK) as usize;
+        self.get_bit_prefix_suffix_assuming_fast_index(bit_prefix, bit_suffix)
+    }
+
+    #[inline(always)]
+    unsafe fn get_bit_prefix_suffix_assuming_fast_index(
```
Manishearth, line 593: nit: document invariant

components/collections/src/codepointtrie/cptrie.rs
```diff
@@ -697,6 +712,97 @@ impl<'trie, T: TrieValue> CodePointTrie<'trie, T> {
...
+    /// Returns the value that is associated with `ascii` in this [`CodePointTrie`].
+    ///
+    /// # Safety
+    ///
+    /// `ascii` must be less than 128.
+    #[inline(always)]
+    pub unsafe fn get7(&self, ascii: u8) -> T {
+        debug_assert!(ascii < 128);
+        debug_assert!((ascii as usize) < self.data.len());
+        // SAFETY: Length of `self.data` checked in the constructor.
```
Manishearth, line 734: nit: ideally, this should be done with `data` documented as having an invariant, and anything that sets or mutates data says "`data` invariant upheld: length >= 128", so here you can just reference the field invariant

Manishearth (commented):
This overall appears to be the correct approach. I'll probably do a more careful unsafe review when you have a PR that just does the collections changes.


I _think_ this is already doing the thing I wanted it to do around the two data payloads when we had the discussion; I'm not sure if there's anything left for me to help with

hsivonen:
The changes here address these open issues at least:

* #2431 
* #7450 
* #6925 
* #6838 
* #5819 
* #2384 (not measured but WONTFIXed)
* #1848

hsivonen:
The previously-discussed plan was that the requirement for the NFD and NFKD tries to be always fast unless the `serde` feature is used would require incrementing the marker version to V2.

I think https://github.com/unicode-org/icu4x/issues/7665 can be fixed by taking an currently always-set bit and making it not set signify the possibily to stay on the fast patch across a single so marked non-starter. This would be compatible with past data, just the past data would not get the fastest path.

So I think we should bundle this landing and a fix for https://github.com/unicode-org/icu4x/issues/7665 into the same ICU4X release to make the V2 aspect for the decomposition trie cover both changes.

Manishearth:
Seems fine. What is the path to actually turning this into landable code? From my end, the architecture seems fine, I just need smaller PRs that I can review for safety/etc.

hsivonen:
> Seems fine. What is the path to actually turning this into landable code? From my end, the architecture seems fine, I just need smaller PRs that I can review for safety/etc.

Let's get #7768 through review and onto crates.io, and then the normalizer and collator PRs become smaller and also runnable in CI.

PR #7531, opened 2026-01-31
Only apply RD range when parsing from_str

Manishearth:
Currently, we have soft per-calendar year ranges (±1M years, ±500K after #7533) that affect per-calendar constructors, and a hard internal RD range that is the minimum range required to support all of the soft ranges.

It is possible to get out of the soft range by e.g. converting dates between calendars. This is by design.

The soft year range works for all of the constructors from codes/etc. However, _string_ constructors work on dates like `1970-01-01[u-ca=islamic-tbla]`, where the date is specified as an ISO date, and annotated as being interpreted in a particular calendar.

This is the primary serialization format for such dates, as specified in IXDTF and Temporal.

Unfortunately, in the current implementation, the ISO year range is applied to this serialization format, regardless of the calendar chosen. If a date is in range for `islamic-tbla`, but out of range for ISO, it cannot be deserialized from its serialized form.

Instead, we should just apply the R.D. range when calling `from_str`. An alternate route is to apply the calendar-specific year range when deserializing from a calendar, however this results in dates that are theoretically possible to construct but will not deserialize.

Based on https://github.com/unicode-org/icu4x/pull/7533

### DIFF
--- a/components/calendar/src/cal/iso.rs
+++ b/components/calendar/src/cal/iso.rs
@@ -5,7 +5,7 @@
 use crate::cal::abstract_gregorian::{
     impl_with_abstract_gregorian, AbstractGregorian, GregorianYears,
 };
-use crate::calendar_arithmetic::ArithmeticDate;
+use crate::calendar_arithmetic::{ArithmeticDate, DateFieldsResolver};
 use crate::error::UnknownEraError;
 use crate::{types, Date, DateError, RangeError};
 use tinystr::tinystr;
@@ -77,6 +77,55 @@ impl Date<Iso> {
             .map(ArithmeticDate::cast)
             .map(IsoDateInner)
             .map(|i| Date::from_raw(i, Iso))
+    }
+
+    /// Construct ISO dates without an ISO year range check
+    ///
+    /// This allows for construction in cases where interchange is needed, e.g. from IXDTF
+    #[cfg(feature = "ixdtf")]
+    pub(crate) fn try_new_iso_without_range_check(
+        year: i32,
+        month: u8,
+        day: u8,
+    ) -> Result<Date<Iso>, RangeError> {
+        use crate::calendar_arithmetic::VALID_RD_RANGE;
+
+        // Month must be 1-12
+        if !(1..=12).contains(&month) {
+            return Err(RangeError {
+                field: "month",
+                value: i32::from(month),
+                min: 1,
+                max: 12,
+            });
+        }
+
+        // Day must be valid for the month and year
+        let days_in_month = AbstractGregorian::<IsoEra>::days_in_provided_month(year, month);
+        if !(1..=days_in_month).contains(&day) {
+            return Err(RangeError {
+                field: "day",
+                value: i32::from(day),
+                min: 1,
+                max: i32::from(days_in_month),
+            });
+        }
+
+        let date = Date::from_raw(
+            IsoDateInner(ArithmeticDate::new_unchecked(year, month, day)),
+            Iso,
+        );
+        let rd = date.to_rata_die();
+        if !VALID_RD_RANGE.contains(&rd) {
+            return Err(RangeError {
+                field: "RataDie",
+                value: rd.to_i64_date() as i32,
+                min: VALID_RD_RANGE.start().to_i64_date() as i32,
+                max: VALID_RD_RANGE.end().to_i64_date() as i32,
+            });
+        }
+        // Construct using new_unchecked, as we are explicitly not checking the year range
+        Ok(date)
     }
 }
 
--- a/components/calendar/src/ixdtf.rs
+++ b/components/calendar/src/ixdtf.rs
@@ -95,7 +95,11 @@ impl<A: AsCalendar> Date<A> {
         calendar: A,
     ) -> Result<Self, ParseError> {
         let date_record = ixdtf_record.date.ok_or(ParseError::MissingFields)?;
-        let iso = Date::try_new_iso(date_record.year, date_record.month, date_record.day)?;
+        let iso = Date::try_new_iso_without_range_check(
+            date_record.year,
+            date_record.month,
+            date_record.day,
+        )?;
 
         if let Some(ixdtf_calendar) = ixdtf_record.calendar {
             if let Some(expected_calendar) = calendar.as_calendar().calendar_algorithm() {
@@ -114,5 +118,39 @@ impl<A: AsCalendar> Date<A> {
             }
         }
         Ok(iso.to_calendar(calendar))
+    }
+}
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::cal::{Hijri, HijriTabularEpoch, HijriTabularLeapYears};
+
+    #[test]
+    fn test_parse_out_of_iso_year_range() {
+        // This test ensures that dates with large years, representable in a calendar
+        // like Islamic-tabular, can roundtrip through the IXDTF string format.
+        // It specifically verifies that the `Date::try_new_iso_without_range_check`
+        // function, used internally for IXDTF parsing, does not perform ISO year
+        // range validity checks (e.g., against VALID_YEAR_RANGE).
+
+        // Double check that +501000-01-01 is outside of the valid year range for ISO
+        assert!(Date::try_new_iso(501000, 1, 1).is_err());
+
+        // This date is within the RD range but outside of ISO's valid year range
+        // as verified above. It should parse.
+        //
+        // We picked Hijri here since Hijri lunar years are *much* shorter than solar years
+        // and the difference is more stark.
+        let valid_date_string = "+501000-01-01[u-ca=islamic-tbla]";
+        let calendar =
+            Hijri::new_tabular(HijriTabularLeapYears::TypeII, HijriTabularEpoch::Thursday);
+
+        Date::try_from_str(valid_date_string, calendar)
+            .expect("Should parse date string successfully");
+
+        // This date is outside of the valid year range, and should fail to parse
+        let invalid_date_string = "+501912-01-01[u-ca=islamic-tbla]";
+
+        Date::try_from_str(invalid_date_string, calendar).expect_err("Should be out of range");
     }
 }


Manishearth:
Needs https://github.com/unicode-org/icu4x/pull/7533 to land first.

robertbastian:
I don't see what the actual issue you're trying to fix is. The test you have written could be written with the current ranges as well.

It *looks* like you're trying to align IXDTF parsing to the constructor-documented year ranges. I don't think those should be aligned, we arbitrarily chose the constructor year ranges so that they are easy to document for the constructors. They are not invariants of the `Date` type. All valid IXDTF strings, i.e. up to 6 digit years, are representable under the RD invariant.

Manishearth:
Sorry, the PR description is from when this was still a draft, so it doesn't describe what this PR does, it describes why this PR doesn't work :) That's fixed now that I wrote https://github.com/unicode-org/icu4x/pull/7533

I updated the PR description.

> It looks like you're trying to align IXDTF parsing to the constructor-documented year ranges.

No, I'm aligning it to the RD range, which is an internal invariant of the Date type.

> I don't think those should be aligned, we arbitrarily chose the constructor year ranges so that they are easy to document for the constructors. 

I agree that IXDTF parsing should not be aligned with constructor documented year ranges, and I'm glad to hear you feel the same.

Currently the _ISO year range_ is applied to `try_from_str` regardless of the calendar, which violates that in a very confusing way. That's the impetus behind this entire change, it is silly for the ISO year range to crop up when doing calendar specific operations. It makes a _little_ sense because the serialization format is ISO, but not really.

components/calendar/src/cal/iso.rs
```diff
@@ -78,6 +78,55 @@ impl Date<Iso> {
...
+
+        // Month must be 1-12
+        if !(1..=12).contains(&month) {
+            return Err(RangeError {
+                field: "month",
+                value: i32::from(month),
+                min: 1,
+                max: 12,
+            });
+        }
```
robertbastian, line 101:
```suggestion
        range_check(month, "month", 1..=12)?;
```

components/calendar/src/cal/iso.rs
```diff
@@ -78,6 +78,55 @@ impl Date<Iso> {
...
+        // Day must be valid for the month and year
+        let days_in_month = AbstractGregorian::<IsoEra>::days_in_provided_month(year, month);
+        if !(1..=days_in_month).contains(&day) {
+            return Err(RangeError {
+                field: "day",
+                value: i32::from(day),
+                min: 1,
+                max: i32::from(days_in_month),
+            });
+        }
```
robertbastian, line 112:
```suggestion
       range_check(day, "day", 1..=AbstractGregorian::<IsoEra>::days_in_provided_month(year, month))?;
```

components/calendar/src/cal/iso.rs
```diff
@@ -78,6 +78,55 @@ impl Date<Iso> {
...
+        );
+        let rd = date.to_rata_die();
+        if !VALID_RD_RANGE.contains(&rd) {
+            return Err(RangeError {
+                field: "RataDie",
+                value: rd.to_i64_date() as i32,
+                min: VALID_RD_RANGE.start().to_i64_date() as i32,
+                max: VALID_RD_RANGE.end().to_i64_date() as i32,
+            });
+        }
```
robertbastian, line 126: I don't think it makes sense to throw a range error over RD, this is not actionable for callers.

Given that IXDTF is written as Gregorian dates, I'm tempted to restrict it to the same range as `Date::try_new_gregorian` after all

components/calendar/src/cal/iso.rs
```diff
@@ -78,6 +78,55 @@ impl Date<Iso> {
             .map(IsoDateInner)
             .map(|i| Date::from_raw(i, Iso))
     }
+
+    /// Construct ISO dates without an ISO year range check
+    ///
+    /// This allows for construction in cases where interchange is needed, e.g. from IXDTF
+    #[cfg(feature = "ixdtf")]
+    pub(crate) fn try_new_iso_without_range_check(
```
robertbastian, line 86: please inline this to the IXDTF parser, it will make it easier to reason about error cases

components/calendar/src/cal/iso.rs
```diff
@@ -78,6 +78,55 @@ impl Date<Iso> {
...
+        year: i32,
+        month: u8,
+        day: u8,
+    ) -> Result<Date<Iso>, RangeError> {
+        use crate::calendar_arithmetic::VALID_RD_RANGE;
+
+        // Month must be 1-12
+        if !(1..=12).contains(&month) {
```
robertbastian, line 94: The IXDTF parser already returns a `ParseError` here:

https://github.com/unicode-org/icu4x/blob/50912ea1d660faded388fd77dd5db87c98774554/utils/ixdtf/src/parsers/datetime.rs#L294

components/calendar/src/cal/iso.rs
```diff
@@ -78,6 +78,55 @@ impl Date<Iso> {
...
+                field: "month",
+                value: i32::from(month),
+                min: 1,
+                max: 12,
+            });
+        }
+
+        // Day must be valid for the month and year
+        let days_in_month = AbstractGregorian::<IsoEra>::days_in_provided_month(year, month);
+        if !(1..=days_in_month).contains(&day) {
```
robertbastian, line 105: The IXDTF parser already returns a `ParseError` here:
https://github.com/unicode-org/icu4x/blob/50912ea1d660faded388fd77dd5db87c98774554/utils/ixdtf/src/parsers/datetime.rs#L226

components/calendar/src/ixdtf.rs
```diff
@@ -116,3 +120,37 @@ impl<A: AsCalendar> Date<A> {
...
 }
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::cal::{Hijri, HijriTabularEpoch, HijriTabularLeapYears};
+
+    #[test]
+    fn test_parse_out_of_iso_year_range() {
+        // This test ensures that dates with large years, representable in a calendar
+        // like Islamic-tabular, can roundtrip through the IXDTF string format.
```
robertbastian, line 131: there is no round-tripping going on in this test

components/calendar/src/cal/iso.rs
```diff
@@ -78,6 +78,55 @@ impl Date<Iso> {
...
+                max: i32::from(days_in_month),
+            });
+        }
+
+        let date = Date::from_raw(
+            IsoDateInner(ArithmeticDate::new_unchecked(year, month, day)),
+            Iso,
+        );
+        let rd = date.to_rata_die();
+        if !VALID_RD_RANGE.contains(&rd) {
```
robertbastian, line 119: so if you don't change the year range to 500000 (and update the RD range to stay tight), you don't need this check. The IXDTF parser allows at most 999999, which is within the RD range.

Manishearth:
> I agree with the second commit, but not with the first.
> 
> I think what you're trying to do here is:
> 
> * you're tightening the year range, because
> * you're tightening the RD range, because
> * you want to add IXDTF serialization in the future, and that requires an RD range smaller than -999999-01-01..999999-12-31
> 
> Is that accurate? The explanations/justifications I've seen all seem backward


No, that's not accurate. We already have IXDTF  here, there's no "future" thing here. IXDTF as a serialization format is specced by Temporal. We don't have serde impls but it is philosophically a serialization format and we should support it in a non-buggy way.

My path on this issue is that I was investigating our date ranges for https://github.com/unicode-org/icu4x/issues/7207, and realized that the current impl applies the ISO year range for `try_from_str`, which seemed silly to me.

I went to fix it (this PR), and realized that it was untestable because our RD range is not actually parseable. This I consider to be a bigger bug, so I went ahead and proposed a fix to that and based this PR on top of it.

There are multiple ways of doing the first commit, we should discuss that on the other issue. The first commit does not drive this one.

My core motivator here is I do not want any of our dates to be non-serializable.

robertbastian:
> realized that it was untestable because our RD range is not actually parseable

It's "untestable" because there is currently no bug. Currently every valid IXDTF string is a valid `Date`. With this change, you are turning some IXDTF strings into errors. I don't see the point of that.

> realized that the current impl applies the ISO year range for try_from_str, which seemed silly to me.

It's silly that it checks it, because the parser already guarantees that we're in range. Let's just remove the checks and replace them by a comment.

robertbastian:
#7545

Manishearth:
> It's "untestable" because there is currently no bug.

Yes, but there's the underlying bug that all of our valid dates are not roundtrippable through IXDTF.

Again, I noticed what I seemed to be a bug in the code, and while attempting to test it, discovered what I considered to be a _worse_ bug, fixed it, and then fixed the original bug.

You're right that the bug this fixes is not present in the current code, but it is not present due to what I consider to be a worse bug.

robertbastian:
> You're right that the bug this fixes is not present in the current code, but it is not present due to what I consider to be a worse bug.

Let's do this one step at a time then, with clear justifications for each step.

You keep talking about round tripping, and I keep saying we don't have IXDTF serialization.

Manishearth:
> Let's do this one step at a time then, with clear justifications for each step.

I am? I made two PRs, one depending on the other? That's ... how that works.

I'm frustrating with your insistence on debating this PR outside of the context of the PR it is based on. It's based on another PR. That's how that works. Yes, this "fix" is not necessary in the current code. I discovered that via a somewhat circituous route, but before anyone saw these PRs I set them up to depend on each other in that way.

It's a waste of time to try and debate the merits of this PR while assuming hte other one won't land. That's not what this PR was written for, and I agree that without the other PR we shouldn't land this. 

> You keep talking about round tripping, and I keep saying we don't have IXDTF serialization.

Huh, you're right. We only have deserialization, it is [temporal_rs which implements serialization](https://docs.rs/temporal_rs/latest/temporal_rs/struct.PlainDate.html#method.to_ixdtf_string).

That's fair, I crossed my wires between the two codebases in understanding what we do and don't have.

We _probably_ should support IXTDF serialization, though.

sffc:
I'm removing myself from reviewer because I don't want to spend time reviewing things that we might revert or change again once we decide on the policy.

PR #7532, opened 2026-01-31
Add ZonedTime and use it in Combo docs

sffc:
First part of #5937
Last part of #6242

This PR was written with the help of Gemini CLI with some manual tweaks, which you can see in the commit history.

## Changelog

icu_time: Add ZonedTime for representing and formatting times with zones
 - New type: `ZonedTime`

### DIFF
--- a/components/datetime/src/combo.rs
+++ b/components/datetime/src/combo.rs
@@ -128,8 +128,7 @@ use crate::scaffold::*;
 /// use icu::datetime::NoCalendarFormatter;
 /// use icu::locale::locale;
 /// use icu::time::zone::iana::IanaParser;
-/// use icu::time::ZonedDateTime;
-/// use icu_time::ZonedTime;
+/// use icu::time::ZonedTime;
 /// use writeable::assert_writeable_eq;
 ///
 /// let formatter = NoCalendarFormatter::try_new(
@@ -138,7 +137,7 @@ use crate::scaffold::*;
 /// )
 /// .unwrap();
 ///
-/// let zoned_time = ZonedDateTime::try_strict_from_str(
+/// let zoned_time = ZonedTime::try_strict_from_str(
 ///     "15:44:00-07:00[America/Los_Angeles]",
 ///     IanaParser::new(),
 /// )
--- a/components/time/src/ixdtf.rs
+++ b/components/time/src/ixdtf.rs
@@ -4,6 +4,8 @@
 
 #[cfg(feature = "unstable")]
 use crate::zone::ZoneNameTimestamp;
+#[cfg(feature = "unstable")]
+use crate::ZonedTime;
 use crate::{
     zone::{iana::IanaParserBorrowed, models, InvalidOffsetError, UtcOffset},
     DateTime, Time, TimeZoneInfo, ZonedDateTime,
@@ -615,7 +617,7 @@ impl<A: AsCalendar> ZonedDateTime<A, TimeZoneInfo<models::Full>> {
 }
 
 #[cfg(feature = "unstable")]
-impl crate::ZonedTime<UtcOffset> {
+impl ZonedTime<UtcOffset> {
     /// Create a [`ZonedTime`] from an RFC 9557 string.
     ///
     /// This function is "strict": the string should have only an offset and no named time zone.
@@ -635,7 +637,7 @@ impl crate::ZonedTime<UtcOffset> {
 }
 
 #[cfg(feature = "unstable")]
-impl crate::ZonedTime<TimeZoneInfo<models::AtTime>> {
+impl ZonedTime<TimeZoneInfo<models::AtTime>> {
     /// Create a [`ZonedTime`] from an RFC 9557 string.
     ///
     /// This function is "strict": the string should have only a named time zone and no offset.


components/datetime/src/scaffold/get_field.rs
```diff
@@ -286,6 +287,59 @@ where
     }
 }
 
+impl<Z> UnstableSealed for ZonedTime<Z> {}
+
+impl<Z> GetField<Hour> for ZonedTime<Z> {
+    fn get_field(&self) -> Hour {
+        self.time.hour
```
robertbastian, line 297: Delegate to `Time`'s impl. For `.zone` you do this.

```suggestion
        self.time.get_field()
```

components/datetime/src/combo.rs
```diff
@@ -123,29 +123,36 @@ use crate::scaffold::*;
...
+/// let iana_parser = IanaParser::new();
+/// let time_zone_id = iana_parser.parse("America/Los_Angeles");
+/// let offset = "-0700".parse::<UtcOffset>().unwrap();
+///
+/// // Create a `TimeZoneInfo` with a timestamp to correctly resolve the time zone name.
+/// let time_zone_info = time_zone_id
+///     .with_offset(Some(offset))
+///     .with_zone_name_timestamp(ZoneNameTimestamp::far_in_future());
+///
+/// let zoned_time = ZonedTime {
```
robertbastian: issue: is there no IXDTF string representation for this? we should not launch this type without parsing.

if not, I think it's more readable to parse a `ZonedDateTime` and then throw away the date, than to assemble this from parts. this aligns more with what clients will actually do, they have a ZDT but only want to display a ZT

ffi/capi/tests/missing_apis.txt
```diff
@@ -42,3 +42,4 @@ icu::collator::CollatorBorrowed::write_sort_key_utf8_to#FnInStruct
 icu::properties::script::HarfbuzzScriptData#Struct
 icu::properties::script::HarfbuzzScriptData::new#FnInStruct
 icu::properties::script::HarfbuzzScriptDataBorrowed#Struct
+icu::time::ZonedTime#Struct
```
robertbastian: fix?

sffc:
My main motivation for this PR was testing gemini-cli. I don't intend to spend more time polishing this API by adding string parsing and FFI. If that is a problem, I can mark it as `unstable`.

components/datetime/src/scaffold/get_field.rs
sffc: I do it for `.zone` because zone is generic. All the other impls in this file read the fields directly.

sffc:
> I can mark it as `unstable`.

^ I was asking for a reply to this.

robertbastian (changes requested):
> > I can mark it as `unstable`.
> 
> ^ I was asking for a reply to this.

I don't want to add half-baked APIs just because you want to play around with Gemini. Either we do this correctly or we don't.

robertbastian:
> if not, I think it's more readable to parse a ZonedDateTime and then throw away the date, than to assemble this from parts. this aligns more with what clients will actually do, they have a ZDT but only want to display a ZT

sffc:
A half-baked API is better landed on main than sitting in a PR.

sffc:
I intend to push some code changes later when I'm back on the correct computer, but I'm still awaiting further discussion on whether `ZonedTime` without string and FFI is motivated to land, perhaps with an `unstable` feature.

If we land it, we can keep the issue open for string and FFI.

sffc:
> > if not, I think it's more readable to parse a ZonedDateTime and then throw away the date, than to assemble this from parts. this aligns more with what clients will actually do, they have a ZDT but only want to display a ZT

It appears that RFC 9557 only defines suffixes as annotating a date-time, not a standalone date or time

https://www.rfc-editor.org/rfc/rfc9557.html#name-abnf

But, I think we allow annotations on dates?

robertbastian:
We allow annotations on dates.

ISO-8601 allows times to be combined with offsets as well: https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators

So we should parse the strings `12:34:56+01:00`, `12:34:56Z`, and probably `12:34:56+01:00[Europe/Zurich]` into `ZonedTime`.

robertbastian:
> A half-baked API is better landed on main than sitting in a PR.

It's better to merge *code* than to keep PRs open. However I disagree with your statement when it comes to public APIs. I don't want to get into a situation where a non-trivial part of our API surface is unstable just because we wanted to get PRs merged.

sffc:
It's better to have unstable APIs because subsequent PRs to get them closer to graduation are smaller and bite-size.

components/datetime/src/combo.rs
```diff
@@ -123,31 +123,39 @@ use crate::scaffold::*;
...
 ///     locale!("en-US").into(),
 ///     T::medium().with_zone(SpecificLong),
 /// )
 /// .unwrap();
 ///
 /// let zdt = ZonedDateTime::try_strict_from_str(
-///     "2024-10-18T15:44-0700[America/Los_Angeles]",
-///     Gregorian,
+///     "2024-10-18T15:44:00-07:00[America/Los_Angeles]",
+///     Iso,
```
robertbastian: unnecessary change

components/time/src/types.rs
```diff
@@ -408,3 +408,53 @@ impl ZonedDateTime<Iso, UtcOffset> {
...
+/// # }
+/// ```
+///
+/// See the docs on `NoCalendarFormatter` for more information and examples.
+///
+/// ✨ *Enabled with the `unstable` Cargo feature.*
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+#[allow(clippy::exhaustive_structs)] // this type is stable
+#[cfg(feature = "unstable")]
+pub struct ZonedTime<Z> {
```
robertbastian, line 502: issue: this needs to be reexported in `datetime::input`

robertbastian (changes requested):
> It's better to have unstable APIs because subsequent PRs to get them closer to graduation are smaller and bite-size.

This PR does the trivial work of adding the `ZonedTime` type and its 7 `GetField` implementations. The real work for adding this type has always been adding a `ZonedTimeFormatter` on FFI, and IXDTF parsing.

The issue is not in the milestone, so nobody is going to do the actual work for this release. As I said before, I don't want to have a half-baked (more like tenth-baked) API, even unstably. So far our unstable APIs have been things that we consider done, but we want to fine-tune for one release. This is not close to done.

This also does not fix the issue like you say in the PR description.

After rereading through the issue, I also still have concerns with this type overall.

components/time/src/types.rs
Manishearth: It's fine to not while this is unstable IMO.

Manishearth (approved):
I think this is sufficiently motivated (see the discussion on the issue). I'm weakly negative on adding unmotivated experimental stuff (I think that is a recipe for "too much unstable"), but this has use cases, and we've had this issue open and in various milestones for a while. So no problem with landing it as unstable from my side. API looks mostly like what I expect.

sffc:
> This PR does the trivial work of adding the `ZonedTime` type and its 7 `GetField` implementations. The real work for adding this type has always been adding a `ZonedTimeFormatter` on FFI, and IXDTF parsing.

If users want this type today, they can write these 7 impls manually in userland, but I think the impls should live in the icu4x components.

> The issue is not in the milestone

The WG previously said: "OK to add this type. Cool if it makes 2.0, OK if it doesn't." It didn't make it in 2.0, so then the issue got punted to Priority Backlog. This PR is the first step to solve the issue.

> so nobody is going to do the actual work for this release.

We should keep the issue open.

> As I said before, I don't want to have a half-baked (more like tenth-baked) API, even unstably. So far our unstable APIs have been things that we consider done, but we want to fine-tune for one release. This is not close to done.

I disagree strongly on a multiple dimensions here:

1. The current code is good. There are additional APIs that need to be added, sure, but the current code is code that will need to land either way, and it is code that needs to land first.
2. This API is basically done for Rust users. Additional convenience APIs may be added, but the examples demonstrate that this solves real problems for Rust users.
3. As a WG we are exceedingly nitpicky about things. I'm not saying this is bad, but the unstable feature is the escape hatch to let us make technical progress despite nitpicks.

> This also does not fix the issue like you say in the PR description.

I will fix the PR description.

> After rereading through the issue, I also still have concerns with this type overall.

Please comment on the issue with any new information that should make us revisit the WG conclusion.

Manishearth:
> If users want this type today, they can write these 7 impls manually in userland, but I think the impls should live in the icu4x components.



Adding to this, our GetField traits are there for people to use them if they really need, BUT they're unstable and probably should remain as unstable power user APIs, so we should create stable happy paths for use cases we find.

robertbastian:
> Please comment on the issue with any new information that should make us revisit the WG conclusion.

I have already done this in the issue since the WG decision.

robertbastian:
Turns out we already have the `ZonedTimeFormatter` on FFI, and adding the IXDTF parsing wasn't too complicated.

robertbastian:
I'm fine with merging this now because it's complete, I'm still not convinced that the type is not a footgun and should be stabilised.

PR #7533, opened 2026-01-31
Reduce year validity range to 500,000

Manishearth:
We discovered that 7-digit years do not work in IXDTF. Reducing to 500k, instead, which has a decent buffer between the Temporal lower bound and the IXDTF upper bound.

900k would also probably be acceptable if we want to reduce slowly.

### DIFF
--- a/components/calendar/src/calendar_arithmetic.rs
+++ b/components/calendar/src/calendar_arithmetic.rs
@@ -19,7 +19,18 @@ use core::hash::{Hash, Hasher};
 use core::ops::RangeInclusive;
 
 /// This is checked by constructors. Internally we don't care about this invariant.
-pub const VALID_YEAR_RANGE: RangeInclusive<i32> = -1_000_000..=1_000_000;
+///
+/// The values selected here affect `VALID_RD_RANGE`, which we do care about internally.
+/// If changing this, this should be changed first, and then `test_validity_ranges` can
+/// be used to figure out the new RD range.
+///
+/// Some things to consider when changing this:
+///
+/// - Temporal's IXDTF only supports up to 6-digit years (ISO 8601 requires expanded years to
+///   specify a digit count, Temporal [specifies](https://tc39.es/proposal-temporal/#sec-temporal-iso8601grammar)
+///   it as six digits)
+/// - The Temporal range is 271821-04-20 to +275760-09-13.
+pub const VALID_YEAR_RANGE: RangeInclusive<i32> = -500_000..=500_000;
 
 /// This is a fundamental invariant of `ArithmeticDate` and by extension all our
 /// date types. Because this range slightly exceeds the [`VALID_YEAR_RANGE`], only
@@ -27,8 +38,10 @@ pub const VALID_YEAR_RANGE: RangeInclusive<i32> = -1_000_000..=1_000_000;
 /// constructor actually uses this, but for clamping instead of erroring.
 // This range is the tightest possible range that includes all valid years for all
 // calendars, this is asserted in [`test_validity_ranges`].
+//
+// This range is -505503-04-10 to 501911-12-31
 pub const VALID_RD_RANGE: RangeInclusive<RataDie> =
-    RataDie::new(-367256444)..=RataDie::new(365940477);
+    RataDie::new(-184631444)..=RataDie::new(183319227);
 
 // Invariant: VALID_RD_RANGE contains the date
 #[derive(Debug)]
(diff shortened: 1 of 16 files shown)


Manishearth:
So, looking deeper into this, it's interesting.

Temporal [specifies it as 4 digits, or 6 digits with a sign](https://tc39.es/proposal-temporal/#prod-DateYear). This is what we implement. This matches [older JS-spec requirements](https://tc39.es/ecma262/#sec-expanded-years).

Temporal describes its spec as [basically being an ISO 8601/RFC 9557 hybrid](https://tc39.es/proposal-temporal/#sec-temporal-iso8601grammar). There's some discussion about this [here](https://github.com/tc39/proposal-temporal/issues/2843#issuecomment-2118555765).

-------


[ISO 8601](https://www.loc.gov/standards/datetime/iso-tc154-wg5_n0038_iso_wd_8601-1_2016-02-16.pdf) has a concept of "expanded representations"

> If, by agreement, expanded representations are used, the formats shall be as specified below. The interchange parties shall agree the additional number of digits in the time element year. In the examples below it has been agreed to expand the time element year with two digits

Basically, you can use + or - before a year to say it is "expanded", and then "expanded" years MUST have a fixed number of digits (if it's 6, you specify a 5-digit year with a leading `+0`, etc).

IXDTF ([RFC 9557](https://datatracker.ietf.org/doc/rfc9557/), building on [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339)) itself doesn't specify anything about expanded years, though it does basically say it's building on ISO 8601 as a profile for it.

-------

So, this is a Temporal spec thing, which is conformant to but not specified by ISO 8601. Our ixdtf parser implements the Temporal variant.

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -19,16 +19,29 @@ use core::hash::{Hash, Hasher};
...
+///
+/// The values selected here affect `VALID_RD_RANGE`, which we do care about internally.
+/// If changing this, this should be changed first, and then `test_validity_ranges` can
+/// be used to figure out the new RD range.
+///
+/// Some things to consider when changing this:
+///
+/// - Temporal's IXDTF only supports up to 6-digit years (ISO 8601 requires expanded years to
+///   specify a digit count, Temporal [specifies](https://tc39.es/proposal-temporal/#sec-temporal-iso8601grammar)
+///   it as six digits)
```
robertbastian, line 31: I don't think this is relevant to our internal range. The string range *can* be smaller than what we can represent.

Manishearth: This is listed under "some things to consider". It is relevant, and should absolutely be considered. We are actively considering it now.

If someone wishes to change these limits again they can consider it again. I don't see why this should not be left around as a breadcrumbs.

Manishearth:
> Ok. But this still doesn't required our non-IXDTF Date constructors to be limited to 500'000 years.

The IXDTF string is the reliable serialization format for all of our dates. I think it's an important property that we can actually serialize all of our dates.

The 100000 limit was arbitrarily chosen as "larger than Temporal". This is also arbitrarily chosen but it maintains more properties. That seems good to me. If we support a date we should support as many operations as possible.

------


If I had my druthers we would actually have extremely narrow date ranges for most calendars. We should not have had to spend so much time in ICU4X and Temporal debating what the behavior of the Chinese calendar outside of our data range should be: the calendar should just _end_ there. But choices made in Temporal forced our hand here.

So while we support them, I think these dates are basically pointless and "reducing the range" has no downside (given that these are pointless) provided we don't reduce past the Temporal range.

sffc:
We should always roundtrip through the string constructors because they are for serialization. If the string constructors have an upper year limit, then we should keep all Dates within that upper limit.

components/calendar/src/calendar_arithmetic.rs
robertbastian: This whole PR seems to be motivated by us having IXDTF *serialization*. We don't have IXDTF serialization, and there's no open issue for adding it. Is this something we're actually planning?

sffc:
> If IXDTF supports 6-digit years, we should as well. A limit of 500000 only supports half of them, so we should set the limit at 999999

This is a departure from the previous strategy of setting the limit based on calendar years. We can't set a limit of 999999 because it is possible to create dates in calendars like ROC and Julian that map to ISO years greater than that. It is exactly what this PR is trying to fix.

sffc:
Alternatively, although it would be a departure from the current strategy, I would be okay setting the calendar year limits at +/- 500,000 and the RD limits at -999999-01-01 to +999999-12-31.

robertbastian:
I think we need to have a holistic discussion around the ranges, including arithmetic. I don't think this PR should be merged in any shape until we have figured that out.

> I would be okay setting the calendar year limits at +/- 500,000 and the RD limits at -999999-01-01 to +999999-12-31.

Again, this is only required for IXDTF *serialization*. We don't have IXDTF serialization.

Manishearth:
> Alternatively, although it would be a departure from the current strategy, I would be okay setting the calendar year limits at +/- 500,000 and the RD limits at -999999-01-01 to +999999-12-31.

Yeah, not a fan, because I think that the model we have settled on with soft year ranges and a hard RD range is a really nice one.



> If IXDTF supports 6-digit years, we should as well. A limit of 500000 only supports half of them, so we should set the limit at 999999

Temporal's IXDTF variant supports 6 digit years, and it only supports them to ±270000. IXDTF and ISO in general, as mentioned, don't specifically provide a number for this.

I think "ICU4X can deserialize dates it serializes" plus "ICU4X can deserialize and support all Temporal dates" are sufficient properties here. I do not think "ICU4X can deserialize all dates supported by the grammar it parses" is a useful property since nobody is producing those.

I think this is the best route especially when you factor in Temporal client needs.

sffc:
> > Alternatively, although it would be a departure from the current strategy, I would be okay setting the calendar year limits at +/- 500,000 and the RD limits at -999999-01-01 to +999999-12-31.
> 
> Yeah, not a fan, because I think that the model we have settled on with soft year ranges and a hard RD range is a really nice one.

Hm? This retains soft year range and hard RD range, just changes the way the RD range is derived.

Manishearth:
Hmmm, I guess so? I kind of like that the hard range is "tight".

Manishearth:
I am warming up to the idea of a wider RD range. I think if we set the RD range to all parseable Temporal dates, and then update https://github.com/unicode-org/icu4x/pull/7545/ to check the RD range.

A thing that helped change my mind is that my current plan requires #7531 to introduce a "field: RataDie" for the range error, which is kind of ugly and confusing. Whereas if our RD range is all dates parseable by the `ixdtf` crate[^1], then we can just treat this as a `field: year` error since the bounds in ISO will be pure year bounds.

I still _like_ Robert's `validity_range` test and how it checks for tightness. But I'm okay giving that up.

 [^1]: Which, as I've said before, is an arbitrary range obtained from an implementation of _part_ of Temporal. It's an artifact of how we have organized our software.

Manishearth:
New plan: https://github.com/unicode-org/icu4x/issues/7076#issuecomment-3879635822

PR #7537, opened 2026-02-02
`enum_keyword!` and `struct_keyword!` are not actually public API

robertbastian:


Manishearth:
It is public API?

https://unicode-org.github.io/icu4x/rustdoc/icu_locale_core/preferences/extensions/unicode/macro.struct_keyword.html


And released: https://docs.rs/icu_locale_core/latest/icu_locale_core/preferences/extensions/unicode/macro.struct_keyword.html

maybe it should not be? We could hide it as a legacy API that was exposed by mistake if that is our decision.

robertbastian:
I checked in the metacrate and it's not there 🤦 

https://unicode-org.github.io/icu4x/rustdoc/icu/locale/preferences/extensions/unicode/index.html

Manishearth:
macros, macros, macros :upside_down_face:

robertbastian:
do we need to reexport in the meta crate with `#[macro_use]`?

Manishearth:
Yeah, probably. This may also just be a rust bug.

PR #7540, opened 2026-02-02
Avoid tagging Hangul syllable in collation data

### LINKED ISSUE #6856: Investigate why collation tailorings for languages whose characters stay < U+1000 get highStart = 0xd800
If you do `rg 'highStart = 0xd800' -t toml` in the icuexportdata directory, you find just about every non-search collation tailoring for languages whose characters are supposed to stay below U+1000.

This is bad, because it means that in the case of small tries (which make sense for these tailorings) fallback to root for characters in the U+1000 to U+D7FF range becomes unnecessarily slow. Ideally, `highStart` for these languages would be no higher than `0x1000` in which case the fallback would never take the slowest path.

`highStart = 0xd800` suggests that the collation tailoring builder either assigns a special value to the reserved character immediately before the surrogate range (or, less likely, unconditionally hoists the whole Hangul Jamo Extended-B block to tailoring).

Figure out what's going on and whether the collation builder could be made not to do this without breaking anything.

hsivonen:
Hangul syllables are already tagged in normalization data.

This makes various collation tailorings smaller and the fallback to root faster.

Closes #6856

(This depends on #7539, so chances are that tests fail until that PR lands to `main` and `main` is merged into this branch.)

## Changelog

icu_collator: (Optimization) Avoid tagging Hangul syllable in collation data

### DIFF
--- a/provider/source/src/collator/mod.rs
+++ b/provider/source/src/collator/mod.rs
@@ -228,8 +228,8 @@ impl IterableDataProviderCached<CollationTailoringV1> for SourceDataProvider {
 fn rebuild_data(trie: CodePointTrie<u32>) -> CodePointTrie<u32> {
     #[cfg(not(any(feature = "use_wasm", feature = "use_icu4c")))]
     {
+        let _ = trie;
         unreachable!("Should have errored out earlier");
-        trie
     }
     #[cfg(any(feature = "use_wasm", feature = "use_icu4c"))]
     {


hsivonen:
Computing the highest code point with non-default trie value is so far only a micro optimization, but as a follow-up, I intend to use that information for deciding whether to upgrade the trie type to fast.

robertbastian:
So this saves 131216-131040=176B or 0.13% of data?

edit: ah I only looked at the root, this saves 34kB for tailorings, 3.5%

hsivonen:
> So this saves 131216-131040=176B or 0.13% of data?

I didn't do the math. This is not primarily about making the data smaller but about making fallback to root faster in the small trie mode for languages whose tailored characters stay below the fast-access bound for small tries.

provider/source/src/collator/mod.rs
```diff
@@ -213,13 +215,44 @@ impl IterableDataProviderCached<CollationTailoringV1> for SourceDataProvider {
...
+fn rebuild_data(trie: CodePointTrie<u32>) -> CodePointTrie<u32> {
+    let default_value = trie.get('\u{10FFFF}');
+    let error_value = trie.get32(u32::MAX);
+    let mut values = vec![default_value; char::MAX as usize];
+
+    for i in 0..(char::MAX as u32) {
+        values[i as usize] = trie.get32(i);
+    }
+    #[expect(clippy::needless_range_loop)]
+    for i in 0xAC00..=0xD7A3 {
```
Manishearth: nit: please include comments as to what's going on here.

hsivonen: Done.

PR #7541, opened 2026-02-02
Add more variants to `CodePointTrieBuilderData`

robertbastian:
Avoids creating the massive vector of all code points.

## Changelog

icu_codepointtrie_builder: Add more ways of specifying code point trie data.
 - New variants: `CodePointTrieBuilderData::ByCodePoint`, `CodePointTrieBuilderData::Map` 
 - New trait impl: `Debug for CodePointTrieBuilderData`

### DIFF
--- a/components/collections/codepointtrie_builder/src/lib.rs
+++ b/components/collections/codepointtrie_builder/src/lib.rs
@@ -84,6 +84,11 @@
 //! [`CodePointTrie`]: icu_collections::codepointtrie::CodePointTrie
 //! [`UMutableCPTrie`]: (https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/umutablecptrie_8h.html#ad8945cf34ca9d40596a66a1395baa19b)
 
+use core::ops::RangeInclusive;
+use std::collections::HashMap;
+use std::panic::RefUnwindSafe;
+use std::panic::UnwindSafe;
+
 use icu_collections::codepointtrie::TrieType;
 use icu_collections::codepointtrie::TrieValue;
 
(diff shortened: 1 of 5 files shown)


components/collections/codepointtrie_builder/src/lib.rs
```diff
@@ -102,13 +107,31 @@ mod native;
...
 pub enum CodePointTrieBuilderData<'a, T> {
     /// A list of values for each code point, starting from code point 0.
     ///
     /// For example, the value for U+0020 (space) should be at index 32 in the slice.
     /// Index 0 sets the value for the U+0000 (NUL).
     ValuesByCodePoint(&'a [T]),
+    /// A closure that returns a value for a code point.
+    ///
+    /// This is called for every code point in the given range
+    ByCodePoint(
```
Manishearth, line 119: suggestion: ByCodePointCallback?

no preference, just throwing it out there

provider/source/src/properties/bidi.rs
```diff
@@ -40,41 +40,52 @@ impl DataProvider<PropertyEnumBidiMirroringGlyphV1> for SourceDataProvider {
...
+                                log::trace!(
+                                    "Missing mirroring glyph: U+{cp:X}: {}",
+                                    char::try_from_u32(cp).unwrap()
+                                );
+                                None
+                            } else {
+                                Some(m)
+                            }
+                        },
+                        paired_bracket_type: match bpt_lookup.get(&(bpt_trie.get32(cp))).copied() {
```
Manishearth, line 64: nit: can we move the code for `mirroring_glyph` and `paired_bracket_type` out of this struct literal?

components/collections/codepointtrie_builder/src/lib.rs
```diff
@@ -102,13 +107,31 @@ mod native;
...
+        RangeInclusive<u32>,
```
Manishearth, line 120: observation: This only lets you specify one range, which is fine because the function below can always return None

e.g. when dealing with surrogates

sffc:
This is one of the oldest issues https://github.com/unicode-org/icu4x/issues/1884

But it had been closed as fixed before. Thanks for finally actually fixing it.

PR #7543, opened 2026-02-02
Consider numbering system in compact formatter 

robertbastian:
## Changelog

icu_decimal: (Experimental) Consider numbering system in compact formatter

### DIFF
--- a/components/decimal/src/preferences.rs
+++ b/components/decimal/src/preferences.rs
@@ -60,22 +60,3 @@ prefs_convert!(
     CompactDecimalFormatterPreferences,
     icu_plurals::PluralRulesPreferences
 );
-
-use icu_provider::prelude::*;
-
-impl DecimalFormatterPreferences {
-    pub(crate) fn nu_id<'a>(
-        &'a self,
-        locale: &'a DataLocale,
-    ) -> Option<DataIdentifierBorrowed<'a>> {
-        self.numbering_system
-            .as_ref()
-            .map(|s| s.as_str())
-            .map(|nu| {
-                DataIdentifierBorrowed::for_marker_attributes_and_locale(
-                    DataMarkerAttributes::from_str_or_panic(nu),
-                    locale,
-                )
-            })
-    }
-}
--- a/components/decimal/src/provider.rs
+++ b/components/decimal/src/provider.rs
@@ -539,3 +539,20 @@ pub(crate) fn load_with_fallback<'a, M: DataMarker>(
 
     Err(DataErrorKind::InvalidRequest.into_error())
 }
+
+impl crate::DecimalFormatterPreferences {
+    pub(crate) fn nu_id<'a>(
+        &'a self,
+        locale: &'a DataLocale,
+    ) -> Option<DataIdentifierBorrowed<'a>> {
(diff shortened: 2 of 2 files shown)


provider/core/src/data_provider.rs
```diff
@@ -20,6 +20,39 @@ where
     /// Returns [`Ok`] if the request successfully loaded data. If data failed to load, returns an
     /// Error with more information.
     fn load(&self, req: DataRequest) -> Result<DataResponse<M>, DataError>;
+
+    #[doc(hidden)]
+    fn load_with_fallback<'a>(
```
Manishearth: question: is there a use case for overriding this?

if not, can this instead be a free function? I'd rather not add trait methods if we can help it.

sffc: I would prefer this being a free function, and would prefer it to be public and documented. I do not want to add an internal trait method to our most important/pervasive trait.

components/decimal/src/lib.rs
```diff
@@ -143,6 +143,21 @@ define_preferences!(
     }
 );
 
+impl DecimalFormatterPreferences {
+    #[doc(hidden)]
+    pub fn nu_id<'a>(&'a self, locale: &'a DataLocale) -> Option<DataIdentifierBorrowed<'a>> {
```
sffc: Issue: This is not important enough to justify a cross-crate internal API which we've established are not zero-cost due to version shift and migration issues.

If you think this is a common enough pattern to need this abstraction, please make a function on `NumberingSystem` (perhaps generated for all of the preference enums), `DataIdentifierBorrowed`, a free function in `icu_provider`, etc.

sffc (changes requested):

components/decimal/src/lib.rs
robertbastian: > which we've established are not zero-cost due to version shift and migration issues.

Have we? We have tilde dependencies here

> please make a function on NumberingSystem

making it a function on `NumberingSystem` is only half as useful, because the preferences contain `Option<NumberingSystem>`, so you'd have to map everywhere

> (perhaps generated for all of the preference enums)

using the preference as a marker attribute is not something that is correct for all preferences

> DataIdentifierBorrowed, a free function in icu_provider, etc.

This is specific to using the numbering system as an marker attribute

Manishearth: This type of ~ dep hidden API is a miniscule cost for google3-like clients.

Our decision to not use `~` deps for `icu_provider` and `icu_locale_core` covers most of the actual cost here: being able to piecemeal update "scaffolding" and "utils" first is useful, being able to piecemeal update individual ICU4X components less so.

So hidden APIs _that we expect to change_ in `icu_provider` and `icu_locale_core` are non-zero-cost. Hidden APIs that we don't really expect to change are fine (and there are mitigation strategies if we _need_ to change them). Hidden APIs on component crates are also mostly fine.

There is a slight cost of messing up versioning when doing releases. We have made this mistake before but for `icu_provider`, not a component crate.

robertbastian: it's crate-internal after #7565

PR #7544, opened 2026-02-02
Document the limits on the Date type

sffc:
We keep talking past each other when it comes to date ranges, so I wrote down some invariants of what I think we should be trying to implement.

### DIFF
--- a/components/calendar/src/date.rs
+++ b/components/calendar/src/date.rs
@@ -113,7 +113,8 @@ impl<C> Deref for Ref<'_, C> {
 ///
 /// 1. Dates can always be converted between calendars
 /// 2. Dates can always be round-tripped through [`RataDie`]
-/// 3. Every possible Date can be constructed from an RFC 9557 string
+/// 3. Every date that is possible to acquire through usage of ICU4X APIs, including calendar
+///    conversion and arithmetic, can be constructed from an RFC 9557 string
 ///
 /// # Examples
 ///


components/calendar/src/date.rs
```diff
@@ -96,12 +96,26 @@ impl<C> Deref for Ref<'_, C> {
...
-/// from its fields via [`Self::try_new_from_codes()`], or can be constructed with one of the
-/// `new_<calendar>_date()` per-calendar methods (and then freely converted between calendars).
+/// # Limits
+///
+/// Constructors may impose limits on the largest or smallest date that they allow,
+/// subject to the following invariants:
+///
+/// 1. Dates can always be converted between calendars
+/// 2. Dates can always be round-tripped through [`RataDie`]
+/// 3. Every possible Date can be constructed from an RFC 9557 string
```
Manishearth: "Every date that is possible to acquire through usage of ICU4X APIs can be constructed from ..."

components/calendar/src/date.rs
```diff
@@ -96,12 +96,26 @@ impl<C> Deref for Ref<'_, C> {
...
+///
 /// This can work with wrappers around [`Calendar`] types,
 /// e.g. `Rc<C>`, via the [`AsCalendar`] trait.
 ///
-/// This can be constructed  constructed
-/// from its fields via [`Self::try_new_from_codes()`], or can be constructed with one of the
-/// `new_<calendar>_date()` per-calendar methods (and then freely converted between calendars).
+/// # Limits
+///
+/// Constructors may impose limits on the largest or smallest date that they allow,
```
Manishearth: nit: may impose documented limits

Though for the RD range I would prefer to not document the values, instead having the constructor say "an internal range which is at least the range of all possible dates that can be constructed"

OR we list a range and explicitly say it is subject to change.

components/calendar/src/date.rs
```diff
@@ -96,12 +96,26 @@ impl<C> Deref for Ref<'_, C> {
 ///
 /// **The primary definition of this type is in the [`icu_calendar`](https://docs.rs/icu_calendar) crate. Other ICU4X crates re-export it for convenience.**
 ///
+/// Options to create one of these:
+///
+/// 1. Generically from fields via [`Self::try_from_fields()`] or [`Self::try_new_from_codes()`]
+/// 2. With calendar-specific constructors, typically named `try_new_xyz()`
+/// 3. From a RFC 9557 string via [`Self::try_from_str()`]
+/// 4. From a [`RataDie`] via [`Self::from_rata_die()`]
```
Manishearth, line 104: also list arithmetic here? Maybe not because it's currently unstable.

Manishearth (approved):
I agree with these

components/calendar/src/date.rs
robertbastian: this is where I disagree. this phrasing means the the RD range needs to be smaller than the IXDTF range (-999999-01-01..=999999-12-31), making the Date -> RFC9557 conversion infallible (an operation which *we don't have*). I want the RD range to be larger than the IXDTF range, making the RFC-9557 -> Date conversion infallible (well, less fallible). I think from a user perspective this also makes more sense:

```suggestion
/// 3. Every valid RFC 9557 string can be parsed into a [`Date`]
```

Manishearth: The "IXDTF range" is not quite a thing, nor is it meaningful to talk about "valid RFC 9557 string"s when applying a date range.

As I mentioned before, the ISO specification allows for an "expanded date" but does not specify a number of digits.

The Temporal specification chooses a 6-digit expanded date format, which means it is able to _parse_ up to 999999. However, it also caps dates at  ~270k, so 999999-12-31 is not an actual valid value. There is no Temporal code where dates between ~270k and 999999 are meaningful.

There is no specification or software out there that will accept 999999-12-31 but not 1000000-01-01, except for ICU4X's ixdtf parser. Most things out there reject extended dates completely. If something supports extended dates, it picks its own number of digits. Temporal has picked 6 _and_ imposed further constraints on it.


So I don't see a reason to treat 999999-12-31 as a bound handed to us from above. The only place that it comes from is from a _partial_ reading of a particular standard.

sffc:
AFAICT, the 6-digit years are a Temporal grammar limitation, not an IXDTF or ISO-8601 limitation.

There are multiple variables, some of which we can tweak, some of which we cannot:

| Name | Limit | Governing authority |
|---|---|---|
| Temporal date range | Approximately +/- 200k ISO years | ECMA |
| Temporal syntax range | All 6-digit ISO years | ECMA |
| RFC 3339 range | 4-digit ISO years | IETF |
| ISO-8601 syntax range | ISO years bounded by a power of 10 | "interchange parties" |
| ICU4X constructor range | +/- 1e6 calendar years | ICU4X TC |
| ICU4X RD range | Derived from above | ICU4X TC |

Given that we care about interchange with Temporal, as an "interchange party" we should probably also pick 6-digit years in our syntax.

Manishearth:
And note that specifically the "temporal syntax range" comes from lifting a _part_ of the standard and implementing it as a separate crate. It's an artifact of code organization.

sffc:
My two cents (alluded in another thread):

I see @robertbastian's point about `try_from_str` accepting all the strings in its range.

But, I want for all ICU4X Dates to be represented as a string. Although we don't currently have serialization, it is something I see in scope.

This is why I prefer setting the RD range to be equal to the Temporal syntax range.

Then, the constructor ranges can be whatever they like for ease of documentation. +/- 500k years seems fine, or +/- 900k. We can't do +/- 1M in my model, though, because it would exceed the Temporal syntax range.

components/calendar/src/date.rs
```diff
@@ -96,12 +96,27 @@ impl<C> Deref for Ref<'_, C> {
 ///
 /// **The primary definition of this type is in the [`icu_calendar`](https://docs.rs/icu_calendar) crate. Other ICU4X crates re-export it for convenience.**
 ///
+/// Options to create one of these:
+///
+/// 1. Generically from fields via [`Self::try_from_fields()`] or [`Self::try_new_from_codes()`]
+/// 2. With calendar-specific constructors, typically named `try_new_xyz()`
```
robertbastian, line 102:
```suggestion
/// 2. With calendar-specific constructors, e.g. [`Self::try_new_chinese_traditional()`]
```

components/calendar/src/date.rs
robertbastian: arithmetic needs you to have a `Date` already, so it's not really creating one from scratch. Same with `to_calendar()`

components/calendar/src/date.rs
```diff
@@ -96,12 +96,27 @@ impl<C> Deref for Ref<'_, C> {
...
+/// 3. From a RFC 9557 string via [`Self::try_from_str()`]
+/// 4. From a [`RataDie`] via [`Self::from_rata_die()`]
+///
 /// This can work with wrappers around [`Calendar`] types,
 /// e.g. `Rc<C>`, via the [`AsCalendar`] trait.
```
robertbastian, line 107: nit: might want to remove this now. there's no point in wrapping any of our calendars at least, so this might be confusing

components/calendar/src/date.rs
```diff
@@ -96,12 +96,27 @@ impl<C> Deref for Ref<'_, C> {
...
 ///
-/// This can be constructed  constructed
-/// from its fields via [`Self::try_new_from_codes()`], or can be constructed with one of the
-/// `new_<calendar>_date()` per-calendar methods (and then freely converted between calendars).
+/// # Limits
+///
+/// Constructors may impose documented limits on the largest or smallest date that they allow,
+/// subject to the following invariants:
+///
+/// 1. Dates can always be converted between calendars
```
robertbastian, line 114:
```suggestion
/// 1. Dates can always be converted between calendars ([`Self::to_calendar`])
```

components/calendar/src/date.rs
```diff
@@ -96,12 +96,27 @@ impl<C> Deref for Ref<'_, C> {
...
+/// 2. Dates can always be round-tripped through [`RataDie`]
```
robertbastian, line 115:
```suggestion
/// 2. Dates can always be round-tripped through [`RataDie`] ([`Self::to_rata_die`], [`Self::from_rata_die()`]) 
```

components/calendar/src/date.rs
```diff
@@ -96,12 +96,27 @@ impl<C> Deref for Ref<'_, C> {
...
+/// 3. Every date that is possible to acquire through usage of ICU4X APIs, including calendar
```
robertbastian, line 116: don't say ICU4X

components/calendar/src/date.rs
```diff
@@ -96,12 +96,27 @@ impl<C> Deref for Ref<'_, C> {
...
+///    conversion and arithmetic, can be constructed from an RFC 9557 string
```
robertbastian, line 117: I don't understand why you changed this to be so complicated. `Date`s can obviously only be obtained from APIs in this crate, because its fields are private

```suggestion
/// 3. Every [`Date`] can be constructed from an RFC 9557 string
```

This is a technical point, I still don't agree with this invariant

Manishearth:
> Now for serialization: we don't have it. 

We've wanted it. I think @nekevss has been planning on contributing it upstream to `ixdtf` at some point. I assumed that had already happened, but it hadn't.

> This conceptually works, but practically doesn't because the ixdtf crate only supports 6-digit years

Right, but we can fix that by changing the limits.

components/calendar/src/date.rs
```diff
@@ -96,12 +96,27 @@ impl<C> Deref for Ref<'_, C> {
...
+/// 3. From a RFC 9557 string via [`Self::try_from_str()`]
+/// 4. From a [`RataDie`] via [`Self::from_rata_die()`]
+///
 /// This can work with wrappers around [`Calendar`] types,
 /// e.g. `Rc<C>`, via the [`AsCalendar`] trait.
 ///
-/// This can be constructed  constructed
-/// from its fields via [`Self::try_new_from_codes()`], or can be constructed with one of the
-/// `new_<calendar>_date()` per-calendar methods (and then freely converted between calendars).
+/// # Limits
```
robertbastian, line 109: I don't find it useful to document the limits on the type. Each constructor should document its limits

components/calendar/src/date.rs
robertbastian, line 114: this follows from `.to_calendar()` being infallible

components/calendar/src/date.rs
robertbastian, line 115: this follows from `from_rata_die`'s docs

components/calendar/src/date.rs
robertbastian, line 116: this could be documented on a future serialization function – right now it's an internal invariant that shouldn't be in the type's docs

nekevss:
> We've wanted it. I think @nekevss has been planning on contributing it upstream to `ixdtf` at some point. I assumed that had already happened, but it hadn't.

FWIW, this is true. I'm finally getting a little bit of time back now that I'm getting into the swing of things with the new job, so I can definitely prioritize upstreaming the serialization.

sffc:
See discussion in #7076

This PR is easier to rewrite than it is to update, so I am closing it, with an action on #7076 to make sure everything is sufficiently documented.

PR #7545, opened 2026-02-02
Remove range checks from `Date::try_from_str`

robertbastian:
They're still there, in the IXDTF parser.

## Changelog: N/A

Changelog will be written in #7533

### DIFF
--- a/components/calendar/src/ixdtf.rs
+++ b/components/calendar/src/ixdtf.rs
@@ -4,7 +4,9 @@
 
 use core::str::FromStr;
 
+use crate::calendar_arithmetic::VALID_RD_RANGE;
 use crate::{AsCalendar, Calendar, Date, Iso, RangeError};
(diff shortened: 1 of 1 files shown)


Manishearth (commented):
I would like us to fix https://github.com/unicode-org/icu4x/pull/7533 which would necessitate further changes here. I guess we could land this in the meantime (it doesn't change behavior) but I don't think that's a particularly helpful step until we resolve the other things.

robertbastian:
> I don't think that's a particularly helpful step until we resolve the other things.

I think it is, because it makes the ranges it's checking explicit. It also doesn't go through the ISO calendar or `ArithmeticDate` anymore, which is harder to reason about.

sffc:
I'm removing myself from reviewer on these PRs because I think they all might need to be changed once we finally decide on the holistic policy for date limits.

PR #7550, opened 2026-02-03
Style guide: List out caveats for cross-crate internal APIs

sffc:
We have been operating with this assumption for some time. I think we should write it down.

It is "suggested", not "required", giving leeway to authors and reviewers.

## Changelog: N/A

### DIFF
--- a/documents/process/style_guide.md
+++ b/documents/process/style_guide.md
@@ -287,11 +287,13 @@ See [this issue](https://github.com/unicode-org/rust-discuss/issues/15) for more
 
 ### Cross-crate internal APIs :: suggested
 
-Cross-crate internal APIs (public APIs marked `#[doc(hidden)]` for use by other ICU4X crates) allow sharing functionality between crates. If used, they must adhere to the following requirements:
+Cross-crate internal APIs (public APIs marked `#[doc(hidden)]` for use by other ICU4X crates) allow sharing functionality between crates. If used, they must adhere to the following guidelines:
 
-- **Action at a distance:** A change to an internal API in one crate can break assumptions in another. To mitigate this, cross-crate internal APIs should be documented and tested to the same level of scrutiny as public APIs.
-- **Piecewise integration:** When clients update one crate at a time, they may experience ephemeral build breakages if these APIs change. Taking care to not break users of these APIs is especially important in foundational crates like `icu_provider` and `icu_locale_core`, which do not use `~` dependencies and thus have higher upgrade hazards.
-- **Use outside ICU4X:** Exported APIs can be discovered and used by clients. To discourage this, name such APIs explicitly (e.g., including `unstable` or `internal` in the name).
+- **Test and document as if stable:** A change to an internal API in one crate can break assumptions in another ("action at a distance"). To mitigate this, cross-crate internal APIs should be documented and tested to a similar level of scrutiny as public APIs.
+- **Design for graceful upgrades:** When clients update one crate at a time, they may experience ephemeral build breakages if these APIs change. Spend time designing the API shape to reduce the potential that you need to change it in the next release, which can cause friction for clients, or such that it is easy to "duplicate" the API instead of editing the existing one. (Note: anything using a trait impl is trickier to duplicate.)
+    - For **core crates** (e.g., `icu_provider`, `icu_locale_core`) that are not used with `~` dependencies, internal APIs _must not_ be changed in breaking ways, since doing so violates Cargo semver. If you need to change something, make a new internal API and keep the old one working.
+    - For **other crates**, internal APIs _should not_ be changed in breaking ways, since non-Cargo clients may still upgrade the crates out-of-sync, but exceptions can be made on a case-by-case basis.
+- **Take steps to reduce usage outside ICU4X:** Exported APIs can be discovered and used by clients. To discourage this, name such APIs explicitly (e.g., including `unstable` or `internal` in the name).
 
 In other words, spend time on the design of these APIs, treating them with the same care as public APIs even if they are not yet stabilized.
 


documents/process/style_guide.md
```diff
@@ -285,6 +285,16 @@ One suggested situation in which public fields would be acceptable is for user-f
 
 See [this issue](https://github.com/unicode-org/rust-discuss/issues/15) for more.
 
+### Don't use cross-crate internal APIs :: suggested
+
+All public APIs, even those marked `#[doc(hidden)]`, add a cost to maintenance and integration:
+
+- Action at a distance: you might make a change to an internal API in one crate and all the tests pass only to discover that a client in another crate was assuming something else about that function's behavior.
+- Piecewise integration: when clients are in the process of updating crates, they may update one crate at a time, resulting in an ephemeral build breakage. Since we care about easing clients' integration processes, we should generally avoid causing this type of issue.
```
Manishearth: As the person who has mostly done these updates: I think our versioning policy around `icu_provider` and `icu_locale_core` already handles 99% of the problems here.

documents/process/style_guide.md
```diff
@@ -285,6 +285,16 @@ One suggested situation in which public fields would be acceptable is for user-f
 
 See [this issue](https://github.com/unicode-org/rust-discuss/issues/15) for more.
 
+### Don't use cross-crate internal APIs :: suggested
+
+All public APIs, even those marked `#[doc(hidden)]`, add a cost to maintenance and integration:
+
+- Action at a distance: you might make a change to an internal API in one crate and all the tests pass only to discover that a client in another crate was assuming something else about that function's behavior.
```
Manishearth: thought: IMO most of this is mitigated by proper documentation. We should do more of that, and have a decently high documentation requirement on these internal APIs.


"you might change an API and break assumptions" is true of public APIs as well.

documents/process/style_guide.md
```diff
@@ -285,6 +285,16 @@ One suggested situation in which public fields would be acceptable is for user-f
...
+- Piecewise integration: when clients are in the process of updating crates, they may update one crate at a time, resulting in an ephemeral build breakage. Since we care about easing clients' integration processes, we should generally avoid causing this type of issue.
+- Use outside ICU4X: if an API is exported, it is subject to being used by clients, whether intentionally or unintentionally.
+
+If you are tempted to add an internal cross-crate API, take a step back, identify the fundamental need, and consider introducing a properly documented and tested public API.
```
Manishearth: issue: I'm worried about pressure towards early-stabilizing public APIs because we need them for something else. I consider that to be a _major_ tradeoff, enough to usually make me prefer internal APIs.

Even if we `cfg(feature = unstable)` these APIs, they will need to be paired with an internal hidden API so that ICU4X crates can use them.

documents/process/style_guide.md
```diff
@@ -285,6 +285,16 @@ One suggested situation in which public fields would be acceptable is for user-f
 
 See [this issue](https://github.com/unicode-org/rust-discuss/issues/15) for more.
 
+### Cross-crate internal APIs :: suggested
+
+Cross-crate internal APIs (public APIs marked `#[doc(hidden)]` for use by other ICU4X crates) allow sharing functionality between crates. If used, they must adhere to the following requirements:
+
+- **Action at a distance:** A change to an internal API in one crate can break assumptions in another. To mitigate this, cross-crate internal APIs should be documented and tested to the same level of scrutiny as public APIs.
+- **Piecewise integration:** When clients update one crate at a time, they may experience ephemeral build breakages if these APIs change. Taking care to not break users of these APIs is especially important in foundational crates like `icu_provider` and `icu_locale_core`, which do not use `~` dependencies and thus have higher upgrade hazards.
```
Manishearth: I'd prefer to split these two suggestions of different strengths

It is a hard API requirement (not just especially important) that we do not break piecewise integration of icu_provider and icu_locale_core, since that is something you might even encouter via Cargo. 

It is a "nice to have" to avoid this for others. In practice I don't even think it's nice to have anymore; the `~` dep means most integrators just do it all at once.


So I think I would have this be two points:

 - `icu_provider`/`icu_locale_core` stability: These crates can be updated independently of ICU4X, since they do not use `~` deps. If adding internal-hidden APIs here, please understand that these APIs will have to remain for the entire ICU4X major release. They might be supplemented with better APIs, but they cannot be removed. Try to design such APIs with care such that (a) they are much less likely to ever need change and (b) if they need change, change can be effected by cleanly duplicating them.
 - Piecewise integration: While Cargo does not allow piecewise updating `~` deps, clients in non-Cargo ecosystems may do so and experience ephemeral build breakages. In general it is worth designing these APIs following the rules in the previous points where possible, though it is not a  strict requirement to never break these.

(I have not thought too deeply about precise wording, this is a suggestion)

documents/process/style_guide.md
```diff
@@ -285,6 +285,16 @@ One suggested situation in which public fields would be acceptable is for user-f
 
 See [this issue](https://github.com/unicode-org/rust-discuss/issues/15) for more.
 
+### Cross-crate internal APIs :: suggested
+
+Cross-crate internal APIs (public APIs marked `#[doc(hidden)]` for use by other ICU4X crates) allow sharing functionality between crates. If used, they must adhere to the following requirements:
```
Manishearth: n.b.

I think I've said this before, but there are a bunch of other reasons we use doc(hidden), specifically: macros and partially-public or sealed traits.

I am reading this entire section as if it only applies to things which are doc-hidden for cross crate purposes only. It may be worth explicitly calling out the other use cases as being not covered by this section as long as we do not expect them to be used cross-crate.

documents/process/style_guide.md
sffc: I feel like the text already says this. There's a general rule which is a suggestion, but following that rule is "especially important" for the core crates without tilde dependencies.

documents/process/style_guide.md
sffc: I think this already is stated fairly clearly. The title of the section is "Cross-crate internal APIs", and the doc(hidden) is only a paranthetical, and the paranthetical is well qualified ("for use by other ICU4X crates").

sffc: I tweaked the language slightly, to call out Cargo vs non-Cargo

documents/process/style_guide.md
sffc: I tweaked the language slightly, to call out Cargo vs non-Cargo

Manishearth: Yes, my issue is that as currently written it makes it seem like the suggestion is "ultimately, be careful about these APIs, and try to avoid breaking them, but it's fine if you have to", except that's not true for the core crates: it's not fine if you have to: you _must not_ break them.

There's a shared general guideline to follow around _designing_ these APIs: try to design them such that they will not need redesign, assuming that you won't have opportunities to change them, and such that if they do have to change it is easy to do it with a "duplicated" API instead of editing the existing one.

This is a rough guideline and it is possible to ignore it based on relevant case-specific tradeoffs.

The rules around _changing_ them, however, are different. For the core crates, you _must_ change them in compatible ways (usually with a "duplicate" API). For the other crates, you don't _have_ to, you should just try. Other tradeoffs may take part.

Not all API patterns are easily "upgraded": anything using a trait impl is trickier to "duplicate".


Maybe instead split this into two points: one point talks about the design guideline (which is shared), and one point talks about "graceful upgrades", which is what you do if you do have to upgrade. That point can describe itself as an immutable rule for core crates, and a guideline for other crates.

sffc:
ok is this landable now?

Manishearth (approved):

sffc:
thank you

sffc:
> I vehemently disagree with this and don't appreciate being left out of the discussion.

It certainly wasn't my intention to leave anyone out of this discussion. I requested review from @unicode-org/icu4x-owners, and the PR was open for 3 weeks to gather feedback (weeks when I believe the whole team was working).

sffc:
> You're basically putting semver stability constraints on hidden APIs between crates even when they use ~ dependencies. That defeats the whole purpose of using `#[doc(hidden)]` in the first place and adds considerable friction to work like currency formatting, which needs to use private, evolving APIs on the decimal formatter.

I agree; the PR is not intended to add significant friction. The only real "requirement" is that we have docs on the cross-crate functions, which is something we should have been doing anyway. This PR was written and revised with tilde dependencies as a primary consideration.

Is there a wording change that you feel would make this more clear?

Manishearth:
N.b. this is a "suggested" rule which means it can be ignored on a case by case basis, but of course it's better to be explicit instead of relying on that.

PR #7552, opened 2026-02-03
Refactor arithmetic tests to improve tooling support

mega123-art:
This pr adds tooling as suggested by @sffc

### DIFF
--- a/components/calendar/src/tests/arithmetic.rs
+++ b/components/calendar/src/tests/arithmetic.rs
@@ -1,3 +1,7 @@
+// This file is part of ICU4X. For terms of use, please see the file
+// called LICENSE at the top level of the ICU4X source tree
+// (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
 use crate::options::{DateAddOptions, DateDifferenceOptions, Overflow};
 use crate::types::{DateDuration, DateDurationUnit};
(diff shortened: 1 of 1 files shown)


robertbastian:
I don't think this is an improvement, the IDE issue that @sffc mentioned is non-existent, and this adds an extra level of redirection that makes it harder to follow the test

mega123-art:
> I don't think this is an improvement, the IDE issue that @sffc mentioned is non-existent, and this adds an extra level of redirection that makes it harder to follow the test

Ig u r right lets see what @sffc tells

sffc:
~Thanks! But, there is no deleted code. This is currently adding a new test without replacing an existing test.~

Fixed via merge.

robertbastian (changes requested):
the `test_all_cals!` macros is written in such a way that you make it wrap your test function. this is how all its invocations in this crate are written. this PR makes code paths harder to follow

sffc:
I checked and it seems `cargo fmt` works now in this type of macro invocation. I'm fairly certain it didn't always work. Haven't checked rust-analyzer.

I still find the refactor better since it reduces a level of indentation. I don't find it harder to follow the code path.

robertbastian:
> Haven't checked rust-analyzer.

I have

> I don't find it harder to follow the code path.

Because you're not currently debugging this

PR #7563, opened 2026-02-03
`icu_calendar` cleanups

robertbastian:
* Remove `Ref` usages from tests and docs now that all our calendars are `Copy`
* Migrate two tests to `test_all_cals!`
* clean up some references to HijriSimulatedMecca

### DIFF
--- a/components/calendar/fuzz/fuzz_targets/common.rs
+++ b/components/calendar/fuzz/fuzz_targets/common.rs
@@ -70,9 +70,6 @@ pub enum AnyCalendarKind {
     Hebrew,
     Indian,
     HijriTabularTypeIIFriday,
-    // Not needed by Temporal and has some bugs
-    // https://github.com/unicode-org/icu4x/issues/7049#issuecomment-3384358307
-    // HijriSimulatedMecca,
     HijriTabularTypeIIThursday,
     HijriUmmAlQura,
     Iso,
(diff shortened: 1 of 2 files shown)


components/calendar/benches/convert.rs
```diff
@@ -5,17 +5,17 @@
...
+fn bench_calendar<C: Copy + Calendar>(
     group: &mut BenchmarkGroup<WallTime>,
     name: &str,
     calendar: C,
 ) {
     let iso = Date::try_new_iso(2023, 8, 16).unwrap();
     group.bench_function(name, |b| {
         b.iter(|| {
-            let converted = black_box(iso).to_calendar(Ref(&calendar));
+            let converted = black_box(iso).to_calendar(calendar);
```
sffc, line 18: It seems unnecessary to change the bench to force Copy calendars? Just because our main calendars are Copy doesn't mean they always will be

components/calendar/src/cal/hijri.rs
```diff
@@ -741,15 +741,15 @@ fn computer_reference_years() {
     fn compute_hijri_reference_year<C>(
         ordinal_month: u8,
         day: u8,
-        cal: &C,
+        cal: C,
         year_info_from_extended: impl Fn(i32) -> C::YearInfo,
     ) -> C::YearInfo
     where
-        C: DateFieldsResolver,
+        C: DateFieldsResolver + Copy,
```
sffc, line 748: Nit: It's a test, so 🤷 , but I don't like changing Calendar bounds to Copy

components/calendar/benches/convert.rs
robertbastian: it's a test, we can write this however we like

robertbastian:
How is that a regression? It allows us to stop using `Ref` everywhere, which simplifies code

robertbastian:
> It seems like the main thing this PR does is adding Copy bounds to a bunch of calendar helpers and tests 

This is not the main thing this PR does. Please look again

sffc:
I'm fine with calendars trying to be Copy, but I don't want to build in Copy _requirements_ just because all our calendars happen to currently implement Copy.

Manishearth:
These are tests, however. That's not a requirement, that's at best a soft nudge, since it's potentially annoying to undo, to the order of maybe 5 minutes of work.

It's a requirement if it's in the public API.

Our tests rely on all kinds of things that "happen to be true" right now and would need to be updated in the future if those things change. I don't think that constitutes any kind of requirement.

sffc:
I agree, which is why I said:

> Nit: It's a test, so 🤷 , but I don't like changing Calendar bounds to Copy

Issue #3647, opened 2023-07-07
Finish compact decimal formatting


sffc:
We have formatting from integers. We need to add it from general FixedDecmial & f64 and add support for currencies.

sffc:
Note that there are two places where floating point numbers are rounded: first in the conversion of the floating point number to the shortest decimal string, and second when performing the user-visible rounding logic.

Discussion about double rounding:

- @sffc - Double-rounding is specified [here](https://tc39.es/ecma402/#sec-tointlmathematicalvalue) in EMCA-402. In ECMAScript, a way to rationalize this is that floating point numbers are seen as compact representations of decimal numbers, so this operation is simply a function in the 1:1 mapping.
- @eggrobin - In the general case, that rationalization is not applicable. But, in the case of number formatting in ICU4X, this is a definite edge case. We could say that we can change the behavior here in the future if desired. I don't want to tie ourselves to the ECMA double rounding standard.

Consensus: For now, implement compact rounding with the ECMA-style double rounding, but leave open the door to do it differently in the future.

LGTM: @eggrobin @sffc @skius @younies

younies:
is this fixed ?

sffc:
No, @robertbastian was working on this but it needs a graduation review.

Manishearth:
The current status of this code is that it is good enough for a tech preview. We should graduate next release but do not need to for 2.2

sffc:
For 2.3, I'd like to see the new CompactDecimalFormatter. It would also be nice to have the new DecimalFormatter. Both should be unstable for 2.3

If you make the new DecimalFormatter, also please do Nan/Infinity the way you think they should be

sffc:
Moving this to 2.4

Issue #275, opened 2020-09-27
`NumberFormatter` & co. scope (monolithic vs modular)


zbraniecki:
In ICU (and ECMA402) `NumberFormat` becomes the jack of all trades with formatting for numbers, currencies, measuring units, and so on. There's even a drive from Shane to incorporate Pluralization as a feature of a `NumberFormat`.

Shane justified it by saying that all number formatters take similar options and similar operations to tailor the data.

The cost of such approach is that it becomes trickier to modularize such crates and `NumberFormat` becomes actually a pretty large codebase on a very fundamental level that is required by basically everything and fragile DCE is the only hope to keep the overhead lower.

It is my impression that in ICU4X context we can bring that modularity back, and I believe Shane is intending to separate the part that operates on the number (like rounding, tailoring etc.) into `FixedDecimal` and similar helper structs and effectively removing the benefit of clustering all types of numerical operations in a single formatter.
@sffc can you share your take on this if I misunderstood you?

if that hypothesis is correct we have a way to get modular and lean `CurrencyFormatter`, `MeasureUnitFormatter`, `RuleBasedNumberFormatter`, `RelativeTimeFormatter`, `DurationFormatter` and so on into their own components and keep each one small without paying with an overhead when all of them are in use.

This issue has been filled to discuss that and verify if we're all on the same page about how we want to tackle that topic.

sffc:
I am of the opinion that crates are not the most effective way to go about modularization.  I have written in wrapper-layer.md that I believe we can use dead code elimination to achieve modularization in a much more effective way.

sffc:
> I believe Shane is intending to separate the part that operates on the number (like rounding, tailoring etc.) into FixedDecimal and similar helper structs and effectively removing the benefit of clustering all types of numerical operations in a single formatter.
> @sffc can you share your take on this if I misunderstood you?

FixedDecimal is intended as a type that preserves leading and trailing zeros on input and output of NumberFormat, which is an important feature we largely lack in 402.  Rounding operations cannot be split from NumberFormat because rounding depends on locale data for currencies, compact decimals, and measurement units.

sffc:
2020-12-04 discussion:

- We all agree that the FFI (logical) API layer should be as modular as possible. So this question is only about the ergonomic layer.
[...]

PR #7565, opened 2026-02-03
Reshape `CompactDecimalFormatter` and move it into `icu_decimal`

robertbastian:
#3647 

#5835

## Changelog

icu_decimal: Move `CompactDecimalFormatter` (experimental) to this crate

### DIFF
--- a/components/decimal/src/compact_formatter.rs
+++ b/components/decimal/src/compact_formatter.rs
@@ -9,7 +9,7 @@ use crate::{
     preferences::CompactDecimalFormatterPreferences, provider::*, DecimalFormatter,
 };
 use alloc::borrow::Cow;
-use fixed_decimal::{CompactDecimal, Decimal, UnsignedDecimal};
+use fixed_decimal::{Decimal, UnsignedDecimal};
 use icu_pattern::{Pattern, PatternBackend, SinglePlaceholder};
 use icu_plurals::PluralRules;
 use icu_provider::DataError;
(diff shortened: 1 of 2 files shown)


sffc:
Do you have a vision for which formatters live in which crates?

robertbastian:
Yes, the first row of the table in https://github.com/unicode-org/icu4x/issues/275#issuecomment-999811989 should live in `icu_decimal`. Maybe also the percent row, because the data for that is trivial.

I think currency and units we wanted to put in a "dimension" crate, but I'd also be happy with a currency and a units crate, there doesn't seem to be a lot of synergy between these.

Manishearth:
Defer to Shane's review. I like the code organization based on that table.

components/decimal/src/compact_formatter.rs
```diff
@@ -459,69 +406,69 @@ impl CompactDecimalFormatter {
...
+    /// /// contain the sequence of digits specified by the significand.
+    /// let one = 1.into();
     /// assert_writeable_eq!(
-    ///     long_french.format_compact_decimal(&a_thousand).unwrap(),
+    ///     long_french.format_with_exponent(&one, 3).unwrap(),
     ///     "mille"
     /// );
     /// ```
-    pub fn format_compact_decimal<'l>(
+    pub fn format_with_exponent<'l>(
```
sffc, line 466: I didn't realize until I was looking at the docs today that this deleted the `format_compact_decimal` function. We definitely need to discuss that. I feel very strongly that `CompactDecimal` is a type we ought to be able to represent since it is used as input not only to formatting but also plural selection. I remember you mentioned that you don't like the ownership semantics; that is something we should look at, but we should not do away with the type entirely.

components/decimal/src/compact_formatter.rs
```diff
@@ -236,71 +241,6 @@ impl CompactDecimalFormatter {
...
-    /// assert_writeable_eq!(short_english.format_f64(1650.0).unwrap(), "1.6K");
-    /// assert_writeable_eq!(short_english.format_f64(1750.0).unwrap(), "1.8K");
-    /// assert_writeable_eq!(short_english.format_f64(1950.0).unwrap(), "2K");
-    /// assert_writeable_eq!(
-    ///     short_english.format_f64(-1_172_700.0).unwrap(),
-    ///     "-1.2M"
-    /// );
-    /// ```
-    #[cfg(feature = "ryu")]
-    pub fn format_f64(
```
sffc, line 293: I also didn't realize this deleted `format_f64`.

components/decimal/src/compact_formatter.rs
```diff
@@ -375,35 +323,35 @@ impl CompactDecimalFormatter {
...
     ///     "-1.2M"
     /// );
     /// assert_writeable_eq!(
-    ///     short_english.format_fixed_decimal(&"0.2222".parse().unwrap()),
+    ///     short_english.format(&"0.2222".parse().unwrap()),
     ///     "0.22"
     /// );
     /// ```
-    pub fn format_fixed_decimal(&self, value: &Decimal) -> impl Writeable + Display + '_ {
+    pub fn format(&self, value: &Decimal) -> impl Writeable + Display + '_ {
```
sffc, line 354: I also didn't realize this made `format_fixed_decimal` into the default `format` function.

I'm not comfortable with that. `[Fixed]Decimal` is not intended to be used as the input to any formatter other than `[Fixed]DecimalFormatter`.

It's like how we don't allow a timestamp to be used as the input to a datetime formatter. We force users to do conversions externally, and then we just transparently localize. I feel strongly that "transparent format" is the correct design for a modern i18n library.

I don't remember why we had `format_fixed_decimal` on this type. Probably because there was something we couldn't do with the other format functions.

components/decimal/src/compact_formatter.rs
robertbastian: we've made this decision for `DecimalFormatter`, it also applies here

components/decimal/src/compact_formatter.rs
robertbastian: The problem is that the `CompactDecimal` type is an inherently localized type, because the exponent that is picked is locale-dependent. Note how this function, which is equivalent to `format_compact_decimal` is fallible, but `format` is infallible

PR #7566, opened 2026-02-03
Replace `experimental` features by `unstable` features

### LINKED ISSUE #6618: The `icu/experimental` feature should probably be `icu/unstable`
Unstable is the common name for Cargo features that expose semver unstable code. `cargo semver-checks` actually has special handling for it, and there doesn't seem to be a way to tell it that the API behind the `experimental` features is unstable.

~We might also want to get rid of the `experimental` module and expose the experimental components at the top level. This way, stabilisation will be smoother because paths won't change.~

robertbastian:
This blocks #7565 as those markers go from experimental to unstable, but those are two different features right now, which breaks the registry and datagen.

#6659

Fixes #6618 

## Changelog

General: Replace `experimental` features with `unstable` features

### DIFF
--- a/.github/workflows/artifacts-build.yml
+++ b/.github/workflows/artifacts-build.yml
@@ -647,7 +647,7 @@ jobs:
             include_data: false
     runs-on: ${{ matrix.os }}
     env:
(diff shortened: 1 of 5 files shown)


robertbastian:
in `icu_capi` I'm keeping both because `experimental` is also a component

Manishearth:
Broadly in favor of this change, have not looked at the code, relying on shane's review.

Issue #6647, opened 2025-05-29
Support deep links in WASM Demo


sffc:
The WASM demo page has a lot of features on it. I would like to be able to link to a specific demo and pre-populate the fields with seed data.

Mechanically, this can be done by adding a [hash query router](https://www.google.com/search?q=hash+query+router).

mega123-art:
@sffc is this issue still open?

sffc:
This issue is still open, yes. Also see https://github.com/rust-diplomat/diplomat/issues/1037

robertbastian:
Query routing now works: https://unicode-org.github.io/icu4x/wasm-demo/index.html#WeekInformation.firstWeekday?self_locale_name=de

The next step would be to generate the query based on the inputs. Maybe put a copy button on the right next to the output.

PR #7567, opened 2026-02-03
feat(web-demo): Support deep links in WASM Demo (#6647)

mega123-art:
This PR implements a hash query router for the WASM demo page as requested in #6647.

It enables:
- **Direct linking** to specific demos via URL hash (e.g., `#FixedDecimalFormatter_format`)
- **Pre-populating fields** with seed data via query parameters (e.g., `?value=123.45&grouping=auto`)

This feature improves the shareability of demo states.

## Changelog

demo: Support deep links to the WASM demo

### DIFF
--- a/tools/web-demo/src/js/index.mjs
+++ b/tools/web-demo/src/js/index.mjs
@@ -11,28 +11,28 @@ window.icu = window.top.icu = icu;
 
 // Renders all termini into the class="container" element
 Object.values(RenderInfo.termini).toSorted((a, b) => a.funcName < b.funcName ? -1 : 1).forEach((t) => {
-    let details = document.createElement("details");
-    let summary = document.createElement("summary");
-    summary.innerHTML = `<code>${t.funcName}</code>`;
-    details.appendChild(summary);
-    details.appendChild(document.createElement("br"));
-    details.appendChild(new TerminusRender(
-        RenderInfo.termini[t.funcName],
-        (el) => {
-            // Necessary for Prism to know the language to highlight for, and also
-            // to ensure CSS `white-space: pre-wrap` is applied from selector
-            el.classList.add("language-js");
-            el.textContent = beautify.js(el.textContent, {
-                indent_size: 2,
-                indent_char: " ",
-                break_chained_methods: true,
-                // brace_style: "collapse",
-                wrap_line_length: 45,
-            });
-            Prism.highlightElement(el);
-        },
-    ));
-    document.getElementsByClassName("container")[0].appendChild(details);
+	let details = document.createElement("details");
+	let summary = document.createElement("summary");
+	summary.innerHTML = `<code>${t.funcName}</code>`;
+	details.appendChild(summary);
+	details.appendChild(document.createElement("br"));
+	details.appendChild(new TerminusRender(
+		RenderInfo.termini[t.funcName],
+		(el) => {
+			// Necessary for Prism to know the language to highlight for, and also
+			// to ensure CSS `white-space: pre-wrap` is applied from selector
+			el.classList.add("language-js");
+			el.textContent = beautify.js(el.textContent, {
+				indent_size: 2,
+				indent_char: " ",
+				break_chained_methods: true,
+				// brace_style: "collapse",
+				wrap_line_length: 45,
+			});
+			Prism.highlightElement(el);
+		},
+	));
+	document.getElementsByClassName("container")[0].appendChild(details);
 });
 
 document.querySelector("#loading").hidden = true;
@@ -61,19 +61,18 @@ function updateFromHash() {
         details.scrollIntoView();
     }
 
-    // Optimization: If no query string, stop here.
+    const terminusParams = terminusRender.querySelector('terminus-params');
+    if (!terminusParams) return;
+
     if (!queryString) return;
 
     const params = new URLSearchParams(queryString);
-    const terminusParams = terminusRender.querySelector('terminus-params');
-    if (!terminusParams) return;
 
     for (const paramElement of terminusParams.children) {
         const nameSlot = paramElement.querySelector('[slot="param-name"]');
         if (!nameSlot) continue;
         const paramName = nameSlot.innerText;
 
-        // Optimization: Check existence and capability before calculating value
         if (params.has(paramName) && paramElement.setValue) {
             const valStr = params.get(paramName);
             const tagName = paramElement.tagName.toLowerCase();


mega123-art:
@sffc sir i have created pr please review it as u get time

tools/web-demo/src/js/index.mjs
```diff
@@ -37,3 +37,70 @@ Object.values(RenderInfo.termini).toSorted((a, b) => a.funcName < b.funcName ? -
...
+function updateFromHash() {
+    const hash = window.location.hash;
+    if (!hash || hash.length < 2) return;
+
+    const cleanHash = hash.substring(1);
+    const parts = cleanHash.split('?');
+    const funcName = parts[0];
+    const queryString = parts[1] || "";
+
+    if (!funcName) return;
```
robertbastian, line 50: is this actually possible? if `cleanHash` is non-null, then how can `cleanHash.split(anything)[0]` be null?

tools/web-demo/src/js/index.mjs
```diff
@@ -37,3 +37,70 @@ Object.values(RenderInfo.termini).toSorted((a, b) => a.funcName < b.funcName ? -
...
+                    valueToSet = parseInt(valStr, 10);
+                } else {
+                    valueToSet = valStr.codePointAt(0);
+                }
+            }
+            
+            if (paramElement.setValue) {
+                paramElement.setValue(valueToSet);
+                if (paramElement.inputElement) {
+                    paramElement.inputElement.dispatchEvent(new Event('input', { bubbles: true }));
```
robertbastian: thought: you probably only need to send one input event at the end of your loop, not for every parameter

tools/web-demo/src/js/index.mjs
```diff
@@ -37,3 +37,70 @@ Object.values(RenderInfo.termini).toSorted((a, b) => a.funcName < b.funcName ? -
...
 document.querySelector("#loaded").hidden = false;
+
+function updateFromHash() {
+    const hash = window.location.hash;
+    if (!hash || hash.length < 2) return;
+
+    const cleanHash = hash.substring(1);
+    const parts = cleanHash.split('?');
+    const funcName = parts[0];
+    const queryString = parts[1] || "";
```
robertbastian: if there is no query string, you should just return after scrolling to the function. there's no point in creating an empty `URLSearchParams` and iterating through all parameters
```suggestion
    const queryString = parts[1];
```

and then before line 69

```js
    if (!queryString) return;
    const params = new URLSearchParams(queryString);
```

tools/web-demo/src/js/index.mjs
```diff
@@ -37,3 +37,70 @@ Object.values(RenderInfo.termini).toSorted((a, b) => a.funcName < b.funcName ? -
...
+                if (valStr.startsWith('0x') || valStr.startsWith('U+')) {
+                    valueToSet = parseInt(valStr.replace(/^U\+/, '0x'), 16);
+                } else if (/^\d+$/.test(valStr)) {
+                    valueToSet = parseInt(valStr, 10);
+                } else {
+                    valueToSet = valStr.codePointAt(0);
+                }
+            }
+            
+            if (paramElement.setValue) {
```
robertbastian: no point in calculating the value if this if false. change line 74 to be

```js
        if (params.has(paramName) && paramElement.setValue) {
```

you can then also get rid of the `valueToSet` variable, and just call `paramElement.setValue` in every case

tools/web-demo/src/js/index.mjs
```diff
@@ -37,3 +37,70 @@ Object.values(RenderInfo.termini).toSorted((a, b) => a.funcName < b.funcName ? -
...
+            let valueToSet = valStr;
+
+            if (tagName === 'terminus-param-boolean') {
+                valueToSet = (valStr === 'true' || valStr === '1');
+            } else if (tagName === 'terminus-param-number') {
+                valueToSet = parseFloat(valStr);
+            } else if (tagName === 'terminus-param-codepoint') {
+                if (valStr.startsWith('0x') || valStr.startsWith('U+')) {
+                    valueToSet = parseInt(valStr.replace(/^U\+/, '0x'), 16);
+                } else if (/^\d+$/.test(valStr)) {
```
robertbastian: `\d` includes non-Latin digits

```suggestion
                } else if (/^[0-9]+$/.test(valStr)) {
```

tools/web-demo/src/js/index.mjs
```diff
@@ -37,3 +37,70 @@ Object.values(RenderInfo.termini).toSorted((a, b) => a.funcName < b.funcName ? -
...
+            const tagName = paramElement.tagName.toLowerCase();
+            let valueToSet = valStr;
+
+            if (tagName === 'terminus-param-boolean') {
+                valueToSet = (valStr === 'true' || valStr === '1');
+            } else if (tagName === 'terminus-param-number') {
+                valueToSet = parseFloat(valStr);
+            } else if (tagName === 'terminus-param-codepoint') {
+                if (valStr.startsWith('0x') || valStr.startsWith('U+')) {
+                    valueToSet = parseInt(valStr.replace(/^U\+/, '0x'), 16);
```
robertbastian: doesn't need to be a regex

```suggestion
                    valueToSet = parseInt(valStr.replace('U+', '0x'), 16);
```

tools/web-demo/src/js/index.mjs
```diff
@@ -37,3 +37,70 @@ Object.values(RenderInfo.termini).toSorted((a, b) => a.funcName < b.funcName ? -
...
+                } else if (/^\d+$/.test(valStr)) {
+                    valueToSet = parseInt(valStr, 10);
+                } else {
+                    valueToSet = valStr.codePointAt(0);
+                }
+            }
```
robertbastian, line 95: how does this handle enums?

tools/web-demo/src/js/index.mjs
mega123-art: You're right, parts[0]  will be an empty string "" rather than null/undefined if the hash starts with a ? (e.g. #?params). The check if (!funcName) catches that empty string case correctly. I've updated the code to address your other suggestions about early returns and optimizations.

tools/web-demo/src/js/index.mjs
mega123-art: It was handling them implicitly by falling through, but I've updated the logic to be more explicit. Enums (and regular strings) now fall into the final else block where setValue(valStr) is called directly

tools/web-demo/src/js/index.mjs
robertbastian: ah of course, the empty string if false 🤦

tools/web-demo/src/js/index.mjs
robertbastian: I was asking how it handles invalid enum *values*

tools/web-demo/src/js/index.mjs
```diff
@@ -63,6 +61,10 @@ function updateFromHash() {
         details.scrollIntoView();
     }
 
+    // Optimization: If no query string, stop here.
```
robertbastian: remove these "Optimization" comments please

tools/web-demo/src/js/index.mjs
```diff
@@ -63,6 +61,10 @@ function updateFromHash() {
...
+    if (!queryString) return;
+
+    const params = new URLSearchParams(queryString);
```
robertbastian, line 69: this can be below `if (!terminusParams) return;`

tools/web-demo/src/js/index.mjs
```diff
@@ -11,28 +11,28 @@ window.icu = window.top.icu = icu;
...
-				indent_char: " ",
-				break_chained_methods: true,
-				// brace_style: "collapse",
-				wrap_line_length: 45,
-			});
-			Prism.highlightElement(el);
-		},
-	));
-	document.getElementsByClassName("container")[0].appendChild(details);
+    let details = document.createElement("details");
```
robertbastian: please revert unrelated formatting changes. in fact, please change your code to 2-space indentation

tools/web-demo/src/js/index.mjs
mega123-art: If an invalid value is passed, the browser's <select> element simply ignores it, leaving the dropdown at its default selection (usually the first option). It fails silently/gracefully.

robertbastian: ok

tools/web-demo/src/js/index.mjs
mega123-art: ya right sorry for this

Manishearth:
If we like this design we should probably incorporate it into the upstream diplomat code. I don't fully understand the split in responsibilities here (@tyler-zeromatter?) but I bet most of this can be reused.

sffc:
I see the new function updateFromHash in index.bundle.js on https://unicode-org.github.io/icu4x/wasm-demo/index.html

How does this work? I see code that reads the query parameter, but no example of how to construct the query parameter. In other words, it solves only half the problem. My expectation is that the URL will be automatically updated when the fields on the site are changed, and then the user can share that URL to restore the state.

tyler-zeromatter:
@Manishearth would you clarify what you mean by split in responsibilities?

If we're talking about architecture for merging upstream, I think a good solution would be to update `rendering.mjs`' `ParameterTemplate` class to read a portion of URLSearchParams in some way (I think by being passed the relevant parameter from `TerminusParams`). Then when the value is changed, emit an event that `TerminusParams` will then use to update the URL.

robertbastian:
> I see the new function updateFromHash in index.bundle.js on https://unicode-org.github.io/icu4x/wasm-demo/index.html
> 
> How does this work? I see code that reads the query parameter, but no example of how to construct the query parameter. In other words, it solves only half the problem. My expectation is that the URL will be automatically updated when the fields on the site are changed, and then the user can share that URL to restore the state.

Please see my comment on the issue.

> If we like this design we should probably incorporate it into the upstream diplomat code. I don't fully understand the split in responsibilities here (@tyler-zeromatter?) but I bet most of this can be reused.

This is not upstreamable. Diplomat only generates JS code and data, the HTML is entirely up to the user. Hence this is fully in the realm of ICU4X.

tyler-zeromatter:
We do include a default renderer for generating HTML from JS, in https://github.com/rust-diplomat/diplomat/blob/main/tool/templates/demo_gen/default_renderer/rendering.mjs. I think this could easily be updated to provide deep links by default.

Manishearth:
Exactly, yes.

Manishearth:
> @Manishearth would you clarify what you mean by split in responsibilities?



In that there is ICU4X-side demogen code and Diplomat-side demogen code and I'm not quite sure what each does or how the responsibilities differ.

tyler-zeromatter:
The structure has changed since I last worked on this, but I think the rough outline is still the same:

1. `demo_gen` creates a `termini` JS object (from https://github.com/unicode-org/icu4x/blob/main/tools/web-demo/gen/index.mjs).
2. ICU4X provides an HTML template that it will use to take the `termini` object and convert it into HTML: https://github.com/unicode-org/icu4x/blob/main/tools/web-demo/public/index.html.
3. `demo_gen` also creates a "default renderer" that reads the `termini` JS object and the HTML template to populate the templates with the values from `termini`: https://github.com/unicode-org/icu4x/blob/main/tools/web-demo/gen/rendering/rendering.mjs
  - This can be overridden with your own renderer, but ICU4X uses the supplied `demo_gen` default renderer.
4. ICU4X calls into the default renderer with Javascript, customizing the style with SCSS and any further functionality: https://github.com/unicode-org/icu4x/blob/main/tools/web-demo/src/js/index.mjs

So in terms of the split, `demo_gen` provides the `termini` object (step 1) and a default renderer that you can call into (step 3).

ICU4X provides the template that the default renderer reads (step 2), and calls into it and adds custom code (step 4).

Let me know if that makes sense, or if you'd like me to clarify any of the steps.

Issue #7557, opened 2026-02-03
Add a Bengali collator benchmark


hsivonen:
The collator benchmarks we already have are rather questionable conceptually (is sorting a list of names the most relevant case?), but they are even worse in their language coverage.

Let's add a benchmark with Bengali name-like strings (something that's privacy-wise appropriate, so not real names of non-public figure persons) to expand the language coverage within the concept.

hsivonen:
Some of the items should share a non-trivial prefix. E.g. repeating some of the family names twice.

SATVIKsynopsis:
Thanks for opening this.

I recently worked on expanding language coverage in the normalizer benchmarks using test-corpora data, and I am happy to follow up here as well.

A Bengali benchmark makes sense given its user base and interesting collation or normalization characteristics. I can take this up in a follow up PR as suggested.

Please let me know if there are any specific constraints or preferences before I start.

hsivonen:
* While there should be some identical prefixes, the data shouldn't unrealistically over-do the number of identical prefixes.
* The data should use the Bengali script. (Clarifying this, because an ASCII PR was posted.)
* There should be some cases where in one string a starter decomposes into more than one starter and the first difference with another string is in the middle of the decomposition. (I.e. the first character of the decomposition exists as-is in the string being compared with.)
* There should be a couple of strings where the difference is between U+0994 vs. U+0982 vs. U+0983 vs. U+0981. I'm not sure if this can realistically appear in names. If not, perhaps the test data should be something other than names that's realistic data to sort.
* It would be useful to have a couple of strings where the difference is that one string has a character combination from the left side of an equals sign in the traditional collation order and the other string has the characters from the right side of the equals sign instead. See https://github.com/unicode-org/cldr/blob/main/common/collation/bn.xml . Again, I'm not sure how realistically this can occur in names.

sffc:
We're looking for a contributor familiar with Bangla to contribute the data.

SATVIKsynopsis:
Thanks for the note.

I am familiar with Bangla and can help contribute the benchmark data. Before starting, I just want to confirm whether there are any specific constraints or preferred data sources you would like me to follow.

I will take this up once my current open PR is finished.

hsivonen:
Since you are familiar with Bangla, I guess the best data source would be coming up with something inoffensive off the top of your head to avoid encumberances that data sources might entail. That's how I did #7578.

[...]

PR #7568, opened 2026-02-03
added  Bengali collator benchmark

psg-19:
Fixes: #7557 
This PR adds a new collator benchmark for the Bengali language to improve coverage.
The data also contains duplicate names as mentioned in the issue.


hsivonen:
This PR uses the Latin script, but the point is to exercise collation for the Bengali script. I've posted clarifications on the issue.

Issue #7558, opened 2026-02-03
Add a Greek collation benchmark


hsivonen:
The collator benchmarks we already have are rather questionable conceptually (is sorting a list of names the most relevant case?), but they are even worse in their language coverage.

Let's add a benchmark with Greek name-like strings (something that's privacy-wise appropriate, so not real names of non-public figure persons) to expand the language coverage within the concept.

Some of the items should share a non-trivial prefix. E.g. repeating some of the family names twice.

AliRana30:
@hsivonen I'd like to work on this issue. I'll create a benchmark with Greek name-like strings that include non-trivial prefixes (repeating family names) to expand the language coverage for collation benchmarks.

hsivonen:
For clarity:

* While there should be some identical prefixes, the data shouldn't unrealistically over-do the number of identical prefixes.
* The data should use the Greek script.
* There should be some cases where the first difference is between an unaccented letter and an accented version of the same letter.
* At least one of the cases from the previous point should have dialytika and tonos on one of the letters being compared (ΐ or ΰ). Not sure how realistically this appears in Greek family names, but this is interesting for exercising Greek-relevant code paths.
* There should be some cases where the first difference is immediately after an accented letter.

hsivonen:
Please don't post AI slop PRs for this; if the team decides on a differently-shaped workload, we can generate algorithmically interesting nonsense ourselves. A hand-crafted PR on a similar level of care as #7578 from someone who is literate in Greek would be welcome.

(One useful bit from #7587 is that Νικολαΐδης looks like a real family name with a letter that has both dialytika and tonos.)

sffc:
We're looking for a contributor familiar with Greek to contribute the data.

PR #7569, opened 2026-02-03
added greek collator benchmark

psg-19:
Fixes: #7558 
This PR adds a new collator benchmark for the Greek language to improve coverage.
The data also contains duplicate names as mentioned in the issue.


hsivonen:
I posted clarification on the issue about what kind of test data is useful. This isn't even in the Greek script.

PR #7573, opened 2026-02-03
fix-7562Add Polish collation benchmark

### LINKED ISSUE #7562: Add a Polish collation benchmark
While adding Polish probably wouldn't add much information to the benchmarks that I've already filed issues, it would be useful to have Polish in the dataset for the purpose of porting benchmarks to JavaScript for JS engines that use ICU4X. JS engines typically store string data as Latin1 or UTF-16, and sorting a list of Polish names is a realistic workload that will end up comparing Latin1 strings with UTF-16 Latin strings in a JS context.

Let's add a benchmark with Polish name-like strings (something that's privacy-wise appropriate, so not real names of non-public figure persons) to expand the language coverage within the concept. Let's make sure some of the names stay fully in the Latin1 range.

We could even simulate JS engine perf in Rust by encoding the strings as Latin1 or UTF-16 and putting them behind an either-or enum.

Jayant-kernel:
Adds Polish name-like strings benchmark for collation performance testing.

The benchmark uses common Polish surnames (Kowalski, Nowak, Wiśniewski, 
Wójcik, Kamiński, etc.) with various diacritical marks to provide 
interesting Latin1/UTF-16 performance comparisons.
Some surnames share prefixes (e.g., Kowalski/Kowalska, Nowak/Nowacki) 
to create non-trivial sorting scenarios as requested.

Fixes #7562


Jayant-kernel:
@psg-19 @hsivonen @echeran 
my PR is ready for review .

hsivonen (approved):
Thanks. Ideally, this would have more cases where the comparison terminates at an accent difference or the letter immediately before the first difference was accented, but this seems to get the job done.

(I'll post a follow-up to shuffle these.)

Jayant-kernel:
@hsivonen thanks for merging 
Sir could you please tell me a community channel for unicode because i couldnt find it .
thanks.

PR #7574, opened 2026-02-03
Use month number instead of ordinal month in formatting

robertbastian:
This is not a behaviour change but only because we don't have a test for these field lengths with calendars that use leap months. We should never display the ordinal month.

## Changelog

icu_datetime: Use month number instead of ordinal month in formatting


sffc:
^

PR #7575, opened 2026-02-03
Update pinned diplomat

Manishearth:
Mostly to update since all of the changes that have happened recently. Also want to make sure that `diplomat-gen` doesn't break MSRV CI.

Based on https://github.com/unicode-org/icu4x/pull/7576


Manishearth:
We already exclude `diplomat-gen`, wonderful

Manishearth:
@robertbastian The details are in the other PR, https://github.com/unicode-org/icu4x/pull/7576. Cargo will refuse to resolve crates that mention an edition that is unsupported by Cargo.

PR #7576, opened 2026-02-03
Update MSRV to 1.86

Manishearth:
Current Rust version is 1.93, so this is 6 or more rust versions behind (following the policy for "no questions asked" MSRV bumps). I am still trying to be conservative with the update, 1.85 is the first release with Edition 2024.

This allows crates in our workspace to depend on Edition 2024 Cargo.toml files. We don't particularly need anything from the edition, however the latest Askama version (which Diplomat uses) uses it. It has an MSRV of 1.88, but `diplomat-gen` is excluded from our MSRV test anyway, so that doesn't matter.

https://github.com/unicode-org/icu4x/pull/7575 is the PR that caused this need.

Update: bumped to 1.86 so we can pull in a newer wasmi, https://github.com/unicode-org/icu4x/pull/7299

## Changelog

General: Update MSRV to 1.86


robertbastian:
Actually, would you mind going to 1.86? I want to update `wasmi` because of the memory leak in #7299, and it has a 1.86 MSRV. Otherwise I can do it in another PR.

Manishearth:
I can given that this failed CI.

Cargo had to make that choice: the edition can impact crate resolution, which occurs early.

The choice that cargo could make to ameliorate this is doing a smarter filter before resolution and then only partially resolving, but it's a bit of a cyclic dependency and not how workspaces have been designed (nor how lock files work: they only need to be generated once and they are generated with any cargo command, you should not need `--all-targets`)

sffc:
I will have faith that this MSRV bump is well-motivated.

robertbastian:
workspace-level dependency resolution strikes again

Manishearth:
Updated to 1.86

Manishearth:
@robertbastian if your other investigations end up speeding it up without a newer wasmi, would you consider un-bumping this after this lands?

robertbastian:
I think it's worth being on wasmi v1 regardless, the previous version we used is 10+ major versions old.

PR #7577, opened 2026-02-04
Add Swedish collator benchmark #7559

### LINKED ISSUE #7559: Add a tailored Latin collator benchmark

Let's add a benchmark with e.g. Swedish name-like strings (something that's privacy-wise appropriate, so not real names of non-public figure persons) to expand the language coverage within the concept.

Some of the items should make the deciding difference be a difference between an ASCII letter and a letter that's considered a base letter in the tailoring but has a canonical decomposition in NFD. If we use Swedish for this, we should have deciding comparisons between a and ä, a and å, ä and å, and o and ö.

If we use Swedish for this, we could also get realistic common prefixes from Ericsson, Ericson, Eriksson, Erikson, and Erixon.

Kayd-06:
This PR adds a new benchmark for the Swedish locale to expand the language coverage of the collator benchmarks and test specific tailoring cases.

Fixes #7559

### DIFF
--- a/components/collator/benches/bench.rs
+++ b/components/collator/benches/bench.rs
@@ -132,6 +132,14 @@ pub fn collator_with_locale(criterion: &mut Criterion) {
             .rev()
             .collect::<Vec<&str>>(),
     );
+    let content_swedish: (&str, Vec<&str>) = (
+        "TestNames_Swedish",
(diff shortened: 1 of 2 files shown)


hsivonen:
Thanks, but I'm closing this, because I had already self-assigned the issue and have a more realistic workload in #7578.

PR #7578, opened 2026-02-04
Add a Swedish collation workload

hsivonen:
Obviously, the given names are actually irrelevant for benchmarking unless there's a duplicate family name.

This isn't about being truly statistically representative of Swedish names, but the point is to exercise collator behavior that's relevant to pretty much all Latin-script languages other than Vietnamese, German phonebook and the root collation-using languages while doing so in a reasonably realistic way as opposed to using completely unrealistic strings that would only exercise the interesting code paths.

The point of the two names starting with Sjö is to exercise the case of the identical prefix ending with a character that decomposes in NFD.

(Why Swedish? Of the languages that have the collation characteristics that I'm trying to exercise, I can generate fake names off the top of my head for Finnish and Swedish. Swedish has more users and uses two non-ASCII base letters that share the same ASCII base in NFD: å and ä, so Swedish is marginally more interesting technically. Both Swedish and Finnish collation data have the same special cases for å and ä, but Swedish input actually exercises both. Estonian would have both õ and ö, but I don't know how to create realistic fake data that exercises those off the top of my head.)

### DIFF
--- a/components/collator/benches/bench.rs
+++ b/components/collator/benches/bench.rs
@@ -124,6 +124,14 @@ pub fn collator_with_locale(criterion: &mut Criterion) {
             .rev()
             .collect::<Vec<&str>>(),
     );
+    let content_swedish: (&str, Vec<&str>) = (
+        "TestNames_Swedish",
+        include_str!("data/TestNames_Swedish.txt")
+            .lines()
+            .filter(|&s| !s.starts_with('#'))
+            .rev()
+            .collect::<Vec<&str>>(),
+    );
     let content_thai: (&str, Vec<&str>) = (
         "TestNames_Thai",
         include_str!("data/TestNames_Thai.txt")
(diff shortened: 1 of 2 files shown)


hsivonen:
Fixes #7559

components/collator/benches/bench.rs
```diff
@@ -124,6 +124,14 @@ pub fn collator_with_locale(criterion: &mut Criterion) {
             .rev()
             .collect::<Vec<&str>>(),
     );
+    let content_swedish: (&str, Vec<&str>) = (
+        "TestNames_Swedish",
+        include_str!("data/TestNames_Swedish.txt")
+            .lines()
+            .filter(|&s| !s.starts_with('#'))
+            .rev()
```
robertbastian, line 132: why the `.rev()`? to move the empty line to the back?
```suggestion
            .filter(|&s| !s.starts_with('#') && !s.is_empty())
```

hsivonen: This is copypaste from the existing harness.

PR #7581, opened 2026-02-04
Actually make CodePointTrieBuilder a builder type

robertbastian:
Hoping to make an impact on #1884 by not having to allocate big temporaries.

## Changelog

icu_codepointtrie_builder: (Breaking) Actually make `CodePointTrieBuilder` a builder type. This API has changed significantly, please look at the docs for the new API.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -806,12 +806,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "downcast-rs"
-version = "1.2.1"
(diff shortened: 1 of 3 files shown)


sffc:
If you want to fix the panic memory leak, great, if not, 🤷

robertbastian:
The panics in FFI code are existing, and we don't need to care much about it, because we don't catch panics.

PR #7583, opened 2026-02-04
Add MSRV entry to changelog

Manishearth:
GitHub's automerge feature has you confirm the commit message when you enable automerge.

Unfortunately, this means updates to the PR title (e.g. https://github.com/unicode-org/icu4x/pull/7576) will not be incorporated. This is not going to be an issue most of the time, we should just be wary of it.

However, now we have an incorrect commit message in our history. Preemptively updating the changelog to avoid mistakes later.


Manishearth:
CI failure is network intermittent

PR #7584, opened 2026-02-04
Use `umutablecptrie_setRange`

robertbastian:
## Changelog

icu_codepointtrie_builder: Optimize by using `umutablecptrie_setRange`


sffc (approved):
Is the code path tested?

robertbastian:
Yes, segmenter datagen uses it

PR #7586, opened 2026-02-05
Update PR template with AI disclosure

sffc:
wdyt?

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -9,8 +9,8 @@ information on contributing to ICU4X.
 Please fill in the following disclosure about your use of AI tools.
 -->
 
-Check one of the following boxes:
+AI tool disclosure:
 
-- [ ] All code in this pull request was written by me, a human.
-- [ ] This pull request was a collaboration between me, a human, and an AI agent.
-- [ ] This pull request was entirely written and orchestrated by aan AI agent.
+- [ ] I wrote this code myself, without help from an AI tool.
+- [ ] I wrote this code with help from an AI tool.
+- [ ] This pull request was entirely written and orchestrated by an AI tool.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -12,5 +12,5 @@ Please fill in the following disclosure about your use of AI tools.
 AI tool disclosure:
 
 - [ ] I wrote this code myself, without help from an AI tool.
-- [ ] I wrote this code with help from an AI tool.
-- [ ] This pull request was entirely written and orchestrated by an AI tool.
+- [ ] I wrote this code with help from an AI tool (Copilot, Claude, Gemini CLI, Codex, ...) as human-in-the-loop.
+- [ ] This pull request was entirely written by an AI tool as human-over-the-loop.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -6,7 +6,8 @@ Reminder: try to use [Conventional Comments](https://conventionalcomments.org/) 
 Please see https://github.com/unicode-org/icu4x/blob/main/CONTRIBUTING.md for general
 information on contributing to ICU4X.
 
-Please fill in the following disclosure about your use of AI tools.
+AI tools can be helpful to learn new things and accelerate your work. At the same time, we seek to give
+mentorship to humans, not AI agents. Please fill in the following disclosure about your use of AI tools.
 -->
 
 AI tool disclosure:
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,4 +5,12 @@ Reminder: try to use [Conventional Comments](https://conventionalcomments.org/) 
 
 Please see https://github.com/unicode-org/icu4x/blob/main/CONTRIBUTING.md for general
 information on contributing to ICU4X.
+
+Please fill in the following disclosure about your use of AI tools.
 -->
+
+Check one of the following boxes:
+
+- [ ] All code in this pull request was written by me, a human.
+- [ ] This pull request was a collaboration between me, a human, and an AI agent.
+- [ ] This pull request was entirely written and orchestrated by aan AI agent.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,13 +5,26 @@ Reminder: try to use [Conventional Comments](https://conventionalcomments.org/) 
 
 Please see https://github.com/unicode-org/icu4x/blob/main/CONTRIBUTING.md for general
 information on contributing to ICU4X.
-
-AI tools can be helpful to learn new things and accelerate your work. At the same time, we seek to give
-mentorship to humans, not AI agents. Please fill in the following disclosure about your use of AI tools.
 -->
 
-AI tool disclosure:
+### AI Tool Disclosure
 
-- [ ] I wrote this code myself, without help from an AI tool.
-- [ ] I wrote this code with help from an AI tool (Copilot, Claude, Gemini CLI, Codex, ...) as human-in-the-loop.
-- [ ] This pull request was entirely written by an AI tool as human-over-the-loop.
+<!--
+AI tools can be helpful to learn new things and accelerate your work. At the
+same time, we seek to give mentorship to humans, not AI agents. Please fill in
+the following statement.
+-->
+
+I used the following AI tools when writing this PR: _____ (N/A, Copilot, Claude, Gemini CLI, Codex, ...)
+
+### Changelog
+
+<!--
+If you added API or changed behavior in this pull request, please list it
+below in the given format, or say "N/A".
+-->
+
+- `icu_crate1`
+    - Description of change
+- `icu_crate2`
+    - Description of change
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,30 +1,21 @@
 <!--
-Thank you for your pull request to ICU4X!
+Thank you for your pull request to ICU4X! Please:
 
-Reminder: try to use [Conventional Comments](https://conventionalcomments.org/) to make comments clearer.
-
-Please see https://github.com/unicode-org/icu4x/blob/main/CONTRIBUTING.md for general
-information on contributing to ICU4X.
+1. Read CONTRIBUTING.md in full before opening a Pull Request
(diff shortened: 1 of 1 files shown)


Manishearth (commented):
Is the intention for this to be for newcomers? And known contributors can blank it out and be clear about AI in any way they feel makes the most sense?

I like this. We should remember to blank it when merging (I think?) but I think it's a good idea.

I'd rather not have to tick this box every time I make a PR but if your intent is for everyone to do it I think it's...fine.

sffc:
> Is the intention for this to be for newcomers? And known contributors can blank it out and be clear about AI in any way they feel makes the most sense?

Yes

> We should remember to blank it when merging (I think?)

I think it's okay for the commit log to contain a string describing the use of AI. It seems verbose for it to contain all three options with one of them being ticked in Markdown syntax. Suggestions?

Manishearth:
"if one of the AI check boxes is checked, reviewers should try to delete the other lines before merging" (as soft guidance for reviewers, not something we have to enforce)


I agree that having the full checklist there is overkill

sffc:
I changed the format and added a changelog section.

robertbastian (changes requested):
I don't care if contributors use tools to generate PRs. The PR author is responsible for all code they send for review, period. 

This is why I also don't like @sffc's attitude on his AI generated PRs that this is not his but Gemini's code and therefore he won't address feedback (like https://github.com/unicode-org/icu4x/pull/7534#issuecomment-3833918603).

This PR has the added problem of putting a lot of process boilerplate in PR descriptions that will end up in commit messages.

Manishearth:
@sffc As I said Friday I plan to write up a plan for the changelog thing and see how people feel. I'd like an opportunity to do that before we start doing the template.

Manishearth:
Though I guess there's no harm in having _some_ changelog template we can always change it later

sffc:
> I don't care if contributors use tools to generate PRs. The PR author is responsible for all code they send for review, period.

Given the rapid increase in third-party contributors, which I suspect has a causal relationship with the rapid increase in AI coding tools, I want more transparency in this area.

> This is why I also don't like @sffc's attitude on his AI generated PRs that this is not his but Gemini's code and therefore he won't address feedback (like [#7534 (comment)](https://github.com/unicode-org/icu4x/pull/7534#issuecomment-3833918603)).

My attitudes toward AI coding tools are rapidly evolving, and we did end up resolving that conflict.

> This PR has the added problem of putting a lot of process boilerplate in PR descriptions that will end up in commit messages.

I agreed and already fixed this by replacing the bulleted list by a single line.

robertbastian:
> Given the rapid increase in third-party contributors, which I suspect has a causal relationship with the rapid increase in AI coding tools, I want more transparency in this area.

And what does that transparency give us? There's also nothing stopping a contributor from saying no even if they used AI.

robertbastian (changes requested):
I think the correct approach to do this is to update `CONTRIBUTING.md`, which should be a TC decision. Then you can update the template to align with that.

Manishearth:
> The PR author is responsible for all code they send for review, period.


I find it useful to know when team members have used AI. I've said this before but when I review code by project members the fact that the project member made a conscious choice to write that code factors in to my generally high-by-default confidence in it, and my general tendency to approve things.

But when an AI makes a choice and you review it it's not the same thing. There may have been better choices to make that I would expect Shane or Robert to notice when actively writing it, but less so when prompting an AI to fix a problem.

To me, "this PR was written by Shane" and "this PR was carefully reviewed by Shane" mean different things, and will have chances for different mistakes.


A project member saying "This PR was primarily generated by AI" to me is roughly equivalent as a project member's intern making a PR with the project member present during the creation process and mentoring.

A project member saying "this commit containing a pretty silly refactor was primarily generated by AI" is something where I probably won't care, though.

> And what does that transparency give us? There's also nothing stopping a contributor from saying no even if they used AI.

If a contributor lies about it and it becomes obvious they are using an AI, well, then, we have an easy reason to not work with them. If a contributor is truthful about it we can have a conversation with them and ask them to not do so. It helps separate out the people who are trying to pull the wool over our eyes for GSOC points from people who are actually meaning to contribute but have not picked up on our stance on AI for newcomers.

sffc:
@Manishearth's post is well said.

> I think the correct approach to do this is to update CONTRIBUTING.md, which should be a TC decision. Then you can update the template to align with that.

Procedurally, I see basically two options here

1. Those of us commenting on this PR get alignment on the AI disclosure strategy, we amend CONTRIBUTING.md in the same PR, and land it after passing it by the TC for additional feedback. We then take what we came up with and present it to other Unicode TCs for their consideration.
2. I start by working with the TCG to establish a Unicode-wide AI policy. All the TC chairs give feedback, and then we update our ICU4X md files accordingly.
3. Do nothing.

I don't want to do nothing, because I don't like being in the dark on AI tool usage.

It's not clear to me whether you feel we should do nothing, or if you feel we should do something but just not the exact solution I proposed in this PR. I would prefer the bottom-up approach 1 rather than the top-down approach 2, and I believe you would too, but to do that I need concrete suggestions.

sffc:
> > Given the rapid increase in third-party contributors, which I suspect has a causal relationship with the rapid increase in AI coding tools, I want more transparency in this area.
> 
> And what does that transparency give us? There's also nothing stopping a contributor from saying no even if they used AI.

Just to add to what @Manishearth already said, contributors lying about their AI tool usage is a great way to filter. I'm very happy to mentor a contributor who uses AI but makes their own decisions, and I'm very happy to shoo away contributors who brainlessly copy everything into an LLM and lie about their AI tool usage. I think the incentives on all sides are to be honest and transparent.

Manishearth:
> I'm very happy to mentor a contributor who uses AI but makes their own decisions

One thing to highlight: if we know this is happening, we can _mentor contributors about AI use as well_. "I recommend telling your agent to format code and use `--all-features`", etc. I experimented a bit with Gemini on ICU4X recently and there are a bunch of ICU4X-specific tips that apply, as well as a lot of general "new" skills that can be learned from LLM usage in coding, which we may be able to mentor.

.github/pull_request_template.md
```diff
@@ -5,4 +5,26 @@ Reminder: try to use [Conventional Comments](https://conventionalcomments.org/)
...
+the following statement.
+-->
+
+I used the following AI tools when writing this PR: _____ (N/A, Copilot, Claude, Gemini CLI, Codex, ...)
+
+### Changelog
+
+<!--
+If you added API or changed behavior in this pull request, please list it
+below in the given format, or say "N/A".
```
robertbastian: why would we want to say "N/A" instead of just removing this section?

.github/pull_request_template.md
```diff
@@ -5,4 +5,26 @@ Reminder: try to use [Conventional Comments](https://conventionalcomments.org/)
...
+-->
+
+- `icu_crate1`
+    - Description of change
+- `icu_crate2`
+    - Description of change
```
robertbastian: most PRs only touch a single crate, lets just have one entry here so we don't have to remove the second one all the time
```suggestion
```

.github/pull_request_template.md
robertbastian: remove this while we're updating the template

.github/pull_request_template.md
```diff
@@ -5,4 +5,26 @@ Reminder: try to use [Conventional Comments](https://conventionalcomments.org/)
 
 Please see https://github.com/unicode-org/icu4x/blob/main/CONTRIBUTING.md for general
 information on contributing to ICU4X.
--->
\ No newline at end of file
+-->
+
```
robertbastian, line 8: if we're requiring a changelog entry, we should also require an issue

```
Fixes: #todo
```

.github/pull_request_template.md
```diff
@@ -5,4 +5,26 @@ Reminder: try to use [Conventional Comments](https://conventionalcomments.org/)
...
+### AI Tool Disclosure
+
+<!--
+AI tools can be helpful to learn new things and accelerate your work. At the
+same time, we seek to give mentorship to humans, not AI agents. Please fill in
+the following statement.
+-->
+
+I used the following AI tools when writing this PR: _____ (N/A, Copilot, Claude, Gemini CLI, Codex, ...)
```
robertbastian: I don't want to fill this out every time I open a PR

.github/pull_request_template.md
```diff
@@ -5,4 +5,26 @@ Reminder: try to use [Conventional Comments](https://conventionalcomments.org/)
 
 Please see https://github.com/unicode-org/icu4x/blob/main/CONTRIBUTING.md for general
 information on contributing to ICU4X.
```
robertbastian: if we're adding an AI disclosure, we should also add a "I have read `CONTRIBUTING.md`" requirement

.github/pull_request_template.md
Manishearth: I'd like to split out the changelog discussion from here in https://github.com/unicode-org/icu4x/issues/7154, that's where the details of this are, I don't really want that to be resolved in a random PR that does other things.

.github/pull_request_template.md
Manishearth: At least when I was talking about this I was expecting that this is something core contributors are allowed to blank (like the rest of the template).

Though if using a tool core contributors are still encouraged to disclose details in some fashion.

.github/pull_request_template.md
Manishearth: Again, we should not be talking about the changelog design here; that should be split out of this PR.

.github/pull_request_template.md
sffc: If regular contributors skip the form, it's _fine_; this is targeted mainly at infrequent contributors.

.github/pull_request_template.md
sffc: Okay

.github/pull_request_template.md
```diff
@@ -1,8 +1,21 @@
...
+
+### AI Tool Disclosure
+
+<!--
+AI tools can be helpful to learn new things and accelerate your work. At the
+same time, we seek to give mentorship to humans, not AI agents. Please fill in
+the following statements.
+-->
+
+I am a: ____ (Human or AI)
```
robertbastian, line 19: what

PR #7587, opened 2026-02-05
Add a Greek collation benchmark

### LINKED ISSUE #7558: Add a Greek collation benchmark

IshaanXCoder:
Add a Greek collation benchmark

Resolves #7558 
This PR adds a new collator benchmark for the Greek language to improve coverage.
Follows all the recommendations as mentioned in https://github.com/unicode-org/icu4x/issues/7558#issuecomment-3846085321


IshaanXCoder:
@hsivonen pls review this PR

hsivonen:
There appear to be a couple of lines here that fit what was sought for, but most lines in the test file don't fit the spec.

PR #7588, opened 2026-02-05
Add a Bengali collation benchmark

### LINKED ISSUE #7557: Add a Bengali collator benchmark

IshaanXCoder:
Add a Bengali collation benchmark

Resolves https://github.com/unicode-org/icu4x/issues/7557
This PR adds a new collator benchmark for the bengali language to improve coverage.
Follows all the recommendations as mentioned in https://github.com/unicode-org/icu4x/issues/7557#issuecomment-3846138401


IshaanXCoder:
@hsivonen pls review this pr

PR #7590, opened 2026-02-05
Add methods to load LSTM or dictionary data for existing segmenters

robertbastian:
Clients want to do custom dictinoary/lstm loading, without doing full custom data loading for segmenters. We can extend `Segmenter::new_for_non_complex_scripts` with methods `load_lstm` and `load_dictionary` so that the non-complex script data can be loaded from a different source, or behind a Cargo feature, without jumping through the hoops of defining a branching data provider or reloading data that has already been loaded.

## Changelog

icu_segmenter: Add methods to load LSTM or dictionary data for existing segmenters

 - New methods on `LineSegmenter` and `WordSegmenter`: `with_lstm_unstable()`, `with_lstm_with_buffer_provider()`, `with_dictionary_unstable()`, `with_dictionary_with_buffer_provider()`
 - New mthods on `LineSegmenterBorrowed` and `WordSegmenterBorrowed`: `with_lstm()` and `with_dictionary()`

### DIFF
--- a/components/segmenter/src/word.rs
+++ b/components/segmenter/src/word.rs
@@ -397,7 +397,7 @@ impl WordSegmenter {
     #[cfg(feature = "compiled_data")]
     #[cfg(feature = "lstm")]
     pub fn new_lstm(options: WordBreakInvariantOptions) -> WordSegmenterBorrowed<'static> {
-        Self::new_for_non_complex_scripts(options).with_lstm_unstable()
+        Self::new_for_non_complex_scripts(options).with_lstm()
     }
 
     #[cfg(feature = "lstm")]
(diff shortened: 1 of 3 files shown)


sffc:
Or they can just implement a custom DataProvider?

If this case is common enough I guess these methods are harmless.

components/segmenter/src/word.rs
```diff
@@ -680,6 +666,30 @@ impl<'data> WordSegmenterBorrowed<'data> {
 }
 
 impl WordSegmenterBorrowed<'static> {
+    /// Loads LSTM data for a [`WordSegmenter`] constructed with
+    /// [`WordSegmenter::new_for_non_complex_scripts`].
+    ///
+    /// ✨ *Enabled with the `compiled_data` and `lstm` Cargo features.*
+    #[cfg(feature = "lstm")]
+    #[cfg(feature = "compiled_data")]
+    pub fn with_lstm_unstable(mut self) -> Self {
```
sffc: Observation: `_unstable` is on the owned type and compiled data is on the Borrowed type.

components/segmenter/src/word.rs
```diff
@@ -592,6 +550,34 @@ impl WordSegmenter {
...
+    where
+        D: DataProvider<SegmenterLstmAutoV1> + ?Sized,
+    {
+        self.complex = self.complex.with_southeast_asian_lstms(provider)?;
+        Ok(self)
+    }
+
+    /// Loads dictionary data for a [`WordSegmenter`] constructed with
+    /// [`WordSegmenter::new_for_non_complex_scripts`].
+    pub fn with_dictionary_unstable<D>(mut self, provider: &D) -> Result<Self, DataError>
```
sffc, line 581: Question: Should we have the buffer versions, too?

Issue #6645, opened 2025-05-28
Make the demo page less formidable


macchiati:
I would suggest on the [demo page](https://icu4x.unicode.org/2_0/demo/) doing the following to make it a bit less daunting.
* Have an intro paragraph that explains what this does, eg 
    * This demo allows you to fill in method arguments for a variety of ICU4J methods, and see what the results are.
* Pre-fill all the input boxes with samples so that something meaningful/interesting happens when someone just opens the twisty and hits the > button
* For some of the more obscure methods (eg rataDie?), have a sentence that explains them. Eg for .segment, 
    * This demo inserts " . " at word-break positions in Thai.”
* Have long names for at least the more obscure datatypes
    * `op.x.f` means ???

sffc:
@macchiati sent additional feedback today:

> The demo is really great, but for someone coming to the page, it is hard to see what the import of the different functions are and the format that the input is supposed to take.
>
> Examples:
>
> - I click on ListFormatter.format, and type in 'ja', pick Wide, and 'a b c'. It just comes back with 'a b c'. I then realize that you want items delimited by commas. 
> - I click on Locale.normalize and type in iw, and it just comes back with iw.
> - I click on LocaleCanonicalizer.canonicalize, and it has Locale:Name. I type in a locale name (German), and it errors out. (I then realize that you are using "Name" instead of "ID" or "Code" (very confusing)). I then type in iw-BU and it works to get he-MM.
> - I click on WordSegmenter.segment, and it has preset examples. Great!
> - I click on ZonedDateTimeFormatterGregorian.formatIso, and fill in some locales (el and de) and a year/month, leaving everything else alone. No matter what locales I use, I get DateTimeFormatterLoadError.DataMarkerNotFound
> - I then click on ZonedTimeFormatter.format and type in 'ja' for the "[self.locale.name](http://self.locale.name/)" and "America/Los_Angeles" for the [zone.id.id](http://zone.id.id/). TimeZoneInvalidOffsetError
>
> My suggestion is that in the right-hand pane, when I open up a method, you 
>
> 1. Prepopulate the fields for every method with an interesting example, like you do for WordSegmenter.segment
> 1. Add three scrolling boxes in the right column, and populate them as follows.
>     1. The API description for the class
>     1. The API description for the method
>     1. Include some more sample interesting examples that people can just copy into the input boxes, eg he-MM above.
>
[...]

PR #7592, opened 2026-02-05
Demo page improvement

SATVIKsynopsis:
Addresses #6645.

I have improved the web demo’s usability by prepopulating parameters with sensible, type based default values, so methods are immediately runnable and input formats are clearer.

This uses the existing default value mechanism without altering core logic.

I verified this locally using the webpack dev server. 

This is an initial incremental step. then further metadata (like descriptions from diplomat) can build on this.

### DIFF
--- a/tools/web-demo/package-lock.json
+++ b/tools/web-demo/package-lock.json
@@ -18,8 +18,7 @@
 				"sass": "^1.77.8",
 				"sass-loader": "^16.0.0",
 				"webpack": "^5.94.0",
-				"webpack-cli": "^5.1.4",
-				"webpack-dev-server": "^5.2.3"
+				"webpack-cli": "^5.1.4"
 			}
 		},
 		"../../ffi/npm": {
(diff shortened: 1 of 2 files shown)


SATVIKsynopsis:
I reverted the manual edits to generated files under `tools/web-demo/gen/` to avoid touching diplomat output. Happy to follow up with a diplomat based solution if that's the preferred direction.

sffc:
This pull request is empty.

SATVIKsynopsis:
Thanks for confirming. I reverted the generated changes after realizing this belongs in diplomat. I will follow up with a diplomat side PR for this issue.

Manishearth:
I'm not sure if we should be guessing defaults here. I'd rather defaults be explicitly specified, with `diplomat::demo` attributes allowing one to specify per-parameter defaults on on each FFI method.

A pull request that guesses that all strings default to "en" would be rejected from Diplomat. That is not a sensible default.

SATVIKsynopsis:
Thanks for the clarification. I will close this PR since the change belongs in diplomat. I will follow up with a diplomat side PR to add explicit demo defaults and metadata as suggested.

PR #7595, opened 2026-02-05
Expose a compiled data version of titlecase_segment_with_only_case_data over FFI

Manishearth:
We have compiled data versions of other ones.

This helps reduce binary size for gboard since we can avoid pulling in GC data. GBoard does custom head adjustment.

## Changelog

icu_capi: Expose a compiled data version of `CaseMapper::titlecase_segment_with_only_case_data()`


sffc (approved):
The name is weird but seems to be a reasonable combination between FFI `lowercase_with_compiled_data` and Rust `titlecase_segment_with_only_case_data`

sffc (approved):

Issue #7510, opened 2026-01-27
Finalize name of ucd-tag/unihan-root CLI args


sffc:
These landed in https://github.com/unicode-org/icu4x/pull/7504.

There was an inconclusive email discussion (TODO: copy in the salient points). We should make sure we're all happy with the option names when shipping ICU4X 2.2.

My main question: are we okay with `--ucd-tag` controlling both the unihan source and the future UCD source? I think yes, because they will have different root tags for power users. We'll end up with:

1. `--ucd-root` sets a directory to UCD, except not unihan
2. `--unihan-root` sets a directory to unihan
3. `--ucd-tag` sets the URL substring on unicode.org to be used when downloading either UCD or Unihan.zip

@Manishearth @opnuub @robertbastian

Manishearth:
I think this is fine, yes.

sffc:
Also, I noticed that `--ucd-tag latest` resolves to https://www.unicode.org/Public/latest/ucd, but in the other options like `--cldr-tag` we special-case `latest` to mean "latest tested". What should we use to signify "latest tested" in `--ucd-tag`?

Manishearth:
`latest-tested`?

robertbastian:
> Also, I noticed that `--ucd-tag latest` resolves to https://www.unicode.org/Public/latest/ucd, but in the other options like `--cldr-tag` we special-case `latest` to mean "latest tested". What should we use to signify "latest tested" in `--ucd-tag`?

I think it's fine for "latest" to have our special meaning (as currently implemented), instead of unicode.org's special meaning.

sffc:
It seems reasonable for someone to want to pull from the unicode.org "latest" tag, though. We shouldn't shadow that tag name with some other meaning on our end.

robertbastian:
> We shouldn't shadow that tag name with some other meaning on our end.

It's not a tag name. It should be clear to everyone that it's an alias. Github also has a `latest` alias that we don't support.

sffc:
> Github also has a `latest` alias that we don't support.

Citation?

It's a shame if that's true, and I wouldn't want to repeat it

robertbastian:
https://github.com/unicode-org/icu4x/releases/latest/download/full.postcard

sffc:
Discussion with @robertbastian and @sffc:

- @robertbastian You should never use the `latest` alias; things will randomly break
- @sffc It's as valid a use case as using a nightly Rust or dev compiler
[...]

PR #7599, opened 2026-02-06
Add latest-tag value for icu4x-datagen tag arguments

sffc:
#7510

## Changelog

icu4x-datagen: Add `latest-tag` value for tag arguments

### DIFF
--- a/provider/icu4x-datagen/src/main.rs
+++ b/provider/icu4x-datagen/src/main.rs
@@ -173,7 +173,9 @@ struct Cli {
     icuexport_root: Option<PathBuf>,
 
     #[arg(long, value_name = "TAG", default_value = "17.0.0")]
-    #[arg(help = "Download versioned UCD from unicode.org.")]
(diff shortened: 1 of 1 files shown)


Manishearth (approved):
Makes sense.

robertbastian (dismissed):
Did you test this? There are two issues why I didn't implement this two years ago, neither of which seem to be addressed in this PR:
* The normal GitHub URL is `<repo>/releases/download/{tag}/foo.zip`, but the `latest` URL is `<repo>/releases/latest/download/foo.zip`. 
* CLDR-JSON contains the version in the file name, you can't just use `latest`. This points to the fundamental issue I have with this: CLDR-JSON doesn't use the same file name for each release because it's not designed to just swap in data from a different release

sffc:
I didn't try it for CLDR, no. Can I add it for ucd_tag?

robertbastian:
If you have to. I don't see a use case for `latest-tag`, but it's not in the way either.

Issue #6511, opened 2025-04-30
Investigate use cases for en-US-posix collation data adjust datagen ergonomics accordingly


hsivonen:
It appears that the purpose of the `en-US-posix` locale is to represent the `C`/`POSIX` locale in the BCP47 value space.

CLDR has a special collation tailoring for [`en_US_POSIX`](https://github.com/unicode-org/cldr/blob/main/common/collation/en_US_POSIX.xml).

Firefox wants the `C` locale to behave like `en-US`, so the tailoring is unwanted. I expect command-line utilities to want to use the `C` locale to opt into lexical sort on bytes without calling into a real collator, so the tailoring would be unnecessary data.

We should investigate use cases for the tailoring an adjust datagen accordingly. (To exclude it by default or to make it super-easy to exclude.)

robertbastian:
Chrome also seems to exclude it: https://source.chromium.org/chromium/chromium/src/+/main:third_party/icu/filters/common.json;l=188-190;drc=bbccc2f6efc1b825de5f2c903c48be685cd0cf22

PR #7600, opened 2026-02-06
Make the collator a lot faster

hsivonen:
#7528 needs to land first, but opening a PR to let reviewers take a look before that.

The key insight here is that by far the most comparisons stop at the first primary difference at the position of the first code unit difference, and that comparison is most often between simple collation unit types.


hsivonen:
This now statically assumes small tries in the collator, which around a 5% perf win give or take a % point. Our data pipeline has never supported generating fast tries. Putting on the discussion agenda for this point.

hsivonen:
> This now statically assumes small tries in the collator, which around a 5% perf win give or take a % point. Our data pipeline has never supported generating fast tries. Putting on the discussion agenda for this point.

We discussed this and concluded that statically assuming small-mode tries is not a semver break given what the data tooling has been like.

provider/source/src/collator/mod.rs
```diff
@@ -129,6 +129,7 @@ impl SourceDataProvider {
         Ok(self
             .icuexport()?
             .list(&format!("collation/{}", self.collation_root_han()))?
+            .filter(|name| !name.contains("POSIX")) // No known use cases
```
robertbastian, line 132: I'd like to know more about this. If it's actually not needed, then ICU shouldn't support and export it, and CLDR shouldn't have data for it.

I'd prefer removing it from CLDR, or from ICU export data. We should not remove things here, because then there's no way to include them, even if the data is there and the caller selects this.

Manishearth: I thought we were asked about it recently? Maybe I'm misremembering...

robertbastian: All I can remember was a recent discussion about posix in host_info: https://github.com/unicode-org/icu4x/issues/6574

There's an issue for collation data: https://github.com/unicode-org/icu4x/issues/6511

Manishearth: I think there was a comment from someone external. @sffc might remember? But yeah this is something CLDR can figure out

sffc:
WG discussion:

- @Manishearth It's hard to review the PR without normalizer landing.
- @hsivonen I found that after doing experiments with fast fata for the root, for chinese/japanese, data size for myanmar/khmer, ICU4C has small tries for all these, and how much data it adds to do fast tries, consider, if we assume small tries at compile time, we get like 5% boost across the board. So I think we should be assuming small tries at compile time, and design a better data structure. Given that we never shipped a data pipeline shipping fast mode tries for collation, can we say that it's not a semver break since we've never shipped tooling for otherwise? They would have had to patch more stuff than is reasonable.
- @Manishearth I find that compelling
- @sffc Does the datagen flag for trie small/fast not impact collator tries?
- @hsivonen It impacts normalizer and properties. Not on these.
- @Manishearth If a client mucked with the bytes in unsupported ways, we can break them.
- @sffc I agree. If a client couldn't get to this code path with a public icu4x-datagen invocation, then we can delete the code path.

components/collator/src/provider.rs
```diff
@@ -638,3 +638,20 @@ impl CollationSpecialPrimariesValidated<'_> {
...
+/// Chinese collations are listed as `und-Hani` with `und-Hant` and `und-Hans` resolving
+/// to `stroke` and `pinyin` despite not listing the collation keyword.
+#[cfg(all(feature = "compiled_data", feature = "unstable"))]
+pub fn list_locales() -> impl Iterator<Item = (DataLocale, tinystr::TinyAsciiStr<8>)> {
+    use icu_provider::baked::DataStore;
+    Baked::DATA_COLLATION_METADATA_V1.iter().map(|d| {
+        (
+            d.locale.clone(),
+            tinystr::TinyAsciiStr::<8>::try_from_str(d.marker_attributes.as_str())
+                .expect("Marker attribute invariants upheld"),
```
robertbastian: thought: return these as `CollationType`

PR #7604, opened 2026-02-06
fix: overflow  from_epoch_milliseconds_and_utc_offset … (#6512)

### LINKED ISSUE #6512: Handle possible integer overflow in from_epoch_milliseconds_and_utc_offset
from_epoch_milliseconds_and_utc_offset should be well-defined over all possible i64 millisecond timestamp values. We need to test and fix the conditions near the boundaries.

swar09:
Added tests confirming integer overflow when combining `i64::MAX/MIN` 
with non-zero UTC offsets. Function panics at `components/time/src/types.rs:378:40` during 
addition operation.

Fixes #6512

## Changelog

icu_time: Fix overflow in `from_epoch_milliseconds_and_utc_offset()`

### DIFF
--- a/provider/source/src/time_zones/mod.rs
+++ b/provider/source/src/time_zones/mod.rs
@@ -751,6 +751,7 @@ mod tests {
     use icu::time::zone::TimeZoneVariant;
 
     use super::*;
+
     #[test]
     fn basic_cldr_time_zones() {
         use icu::locale::langid;


sffc (dismissed):
Thanks for the contribution.

Some issues:

1. The function is on `ZonedDateTime` which is in the `icu_time` crate. Please add the test in that crate.
2. The issue asks for the overflow issue to be fixed. Please fix it, not just test for it.

swar09:
1. Added test  in `ZonedDateTime`  in the `icu_time` crate
2. Using `saturating_add()`  which clamp to `i64::MAX`  (max representable date)
3. Doesn't change function signature (`Self`) , any other method may change  signature to `Result<Self, Error>`  

@sffc Please review it .

sffc:
I agree with @robertbastian.

components/time/src/types.rs
```diff
@@ -370,23 +370,45 @@ impl ZonedDateTime<Iso, UtcOffset> {
...
+    ///         i64::MAX,
+    ///         max_offset
+    ///     );
+    ///
+    /// let min_offset = UtcOffset::try_from_seconds(-43200).unwrap(); // -12 hours
+    /// let zdt_min = ZonedDateTime::from_epoch_milliseconds_and_utc_offset(
+    ///         i64::MIN,
+    ///         min_offset
+    ///     );
+    /// ```
```
robertbastian: this should not be phrased in terms of "handled without overflow". while that's good phrasing for this PR, it's not useful documentation on the method (it's like "we don't have this particular bug"). instead, you should document the general behaviour of large values saturating on the day level

swar09:
@robertbastian  thanks for the feedback ! I have  updated the documentation to describe the general saturation behavior instead of focusing on overflow handling.

PR #7606, opened 2026-02-06
feat(datetime): Add YearStyle::NoEra

### LINKED ISSUE #6010: Add additional YearStyle variants
The ECMA-402 proposal for eraDisplay may propose additional options such as "implied era". For more information, see:

https://github.com/tc39/proposal-intl-eradisplay

mega123-art:
Adds YearStyle::NoEra option as per TC39 intl-eradisplay proposal.

Fixes #6010

## Changelog

icu_datetime: Support formatting without era field
  - New variant: `YearStyle::NoEra`

### DIFF
--- a/components/datetime/src/raw/neo.rs
+++ b/components/datetime/src/raw/neo.rs
@@ -198,10 +198,13 @@ impl DatePatternSelectionData {
         let variant = match (year_style, ambiguity) {
             (YearStyle::WithEra, _) => PackedSkeletonVariant::Variant1,
 
-            (YearStyle::Full, YearAmbiguity::EraAndCenturyRequired | YearAmbiguity::EraRequired) => {
-                PackedSkeletonVariant::Variant1
+            (
+                YearStyle::Full,
+                YearAmbiguity::EraAndCenturyRequired | YearAmbiguity::EraRequired,
+            ) => PackedSkeletonVariant::Variant1,
+            (YearStyle::Full, YearAmbiguity::CenturyRequired | YearAmbiguity::Unambiguous) => {
+                PackedSkeletonVariant::Variant0
             }
-            (YearStyle::Full, YearAmbiguity::CenturyRequired | YearAmbiguity::Unambiguous) => PackedSkeletonVariant::Variant0,
 
             (
                 YearStyle::Auto | YearStyle::NoEra,


sffc (changes requested):
`Always` is the same as `WithEra` and is not needed.

`Never` would be a useful addition, but it seems like not a good addition to the enum which is named `YearStyle`, not `EraDisplay` as the ECMA proposal.

Also please note the "implied" style in the ECMA proposal.

mega123-art:
> `Always` is the same as `WithEra` and is not needed.
> 
> `Never` would be a useful addition, but it seems like not a good addition to the enum which is named `YearStyle`, not `EraDisplay` as the ECMA proposal.
> 
> Also please note the "implied" style in the ECMA proposal.

Thanks for the review! After investigating the [TC39 proposal-intl-eradisplay](https://github.com/tc39/proposal-intl-eradisplay), I found:

**Current proposal status (Stage 2):**
- Only defines 3 values: `"never"`, `"always"`, `"auto"`
- The "implied" option you mentioned doesn't exist in the current spec yet

**Issues with my implementation:**
-  Agree: `Always` is redundant with `WithEra`
- Agree: `Never` belongs in an `EraDisplay` enum, not `YearStyle`
- The current [YearStyle](cci:2://file:///home/parth/icu4x/ffi/npm/lib/YearStyle.d.ts:8:0-24:1) mixes year precision (2-digit vs 4-digit) with era display

**Proposed path forward:**
Would it make more sense to:
1. Wait for TC39 proposal to reach Stage 3+ before implementing?
2. Create a separate `EraDisplay` option instead of extending [YearStyle](cci:2://file:///home/parth/icu4x/ffi/npm/lib/YearStyle.d.ts:8:0-24:1)?

Should I close this PR for now? Should I update issue #6010 with these findings?

sffc:
We would accept a PR adding something like `YearStyle::NoEra`, corresponding to the ECMA-402 proposal `eraDisplay: "never"`.

There is discussion of `eraDisplay: "implied"` in the ECMA-402 proposal repository.

mega123-art:
@sffc sir i tried to make those suggested changes

components/datetime/src/options/mod.rs
```diff
@@ -300,6 +300,50 @@ pub enum YearStyle {
...
+    /// );
+    /// assert_writeable_eq!(
+    ///     formatter.format(&Date::try_new_gregorian(77, 1, 1).unwrap()),
+    ///     "1/1/77"
+    /// );
+    /// // Modern years also show full year without era:
+    /// assert_writeable_eq!(
+    ///     formatter.format(&Date::try_new_gregorian(2025, 1, 1).unwrap()),
+    ///     "1/1/2025"
+    /// );
```
sffc: I think `YearStyle::NoEra` should be more ambiguous than `YearStyle::Auto`, so it should elide the century for nearby years:

```suggestion
    /// assert_writeable_eq!(
    ///     formatter.format(&Date::try_new_gregorian(1900, 1, 1).unwrap()),
    ///     "1/1/1900"
    /// );
    /// // Era and century both elided for nearby years:
    /// assert_writeable_eq!(
    ///     formatter.format(&Date::try_new_gregorian(2025, 1, 1).unwrap()),
    ///     "1/1/25"
    /// );
```

And please change the code to make the test pass.

components/datetime/src/options/mod.rs
sffc: Suggestion (optional): Split this test into tests on each variant, and replace the top-level docs here with a table summarizing the results.

components/datetime/src/raw/neo.rs
```diff
@@ -196,15 +196,16 @@ impl DatePatternSelectionData {
...
                 .unwrap_or(YearAmbiguity::EraAndCenturyRequired),
         ) {
-            (YearStyle::WithEra, _) | (_, YearAmbiguity::EraAndCenturyRequired) => {
-                PackedSkeletonVariant::Variant1
-            }
+            (YearStyle::WithEra, _) => PackedSkeletonVariant::Variant1,
+            (YearStyle::NoEra, YearAmbiguity::EraAndCenturyRequired)
+            | (YearStyle::NoEra, YearAmbiguity::EraRequired)
+            | (YearStyle::NoEra, YearAmbiguity::CenturyRequired) => PackedSkeletonVariant::Variant0,
+            (YearStyle::NoEra, _) => PackedSkeletonVariant::Standard,
```
sffc: Suggestion:

```suggestion
            (YearStyle::NoEra | YearStyle::Auto, YearAmbiguity::Unambiguous | YearAmbiguity::EraRequired) => PackedSkeletonVariant::Standard,
            (YearStyle::NoEra, _) =>  PackedSkeletonVariant::Variant0,
```

In general, please make this match statement exhaustive in the two dimensions. Avoid having a catch-all `_` match arm at the bottom, because it makes it less clear whether we covered all of the cases.

components/datetime/src/options/mod.rs
```diff
@@ -273,33 +194,137 @@ pub enum YearStyle {
...
+    /// let formatter = FixedCalendarDateTimeFormatter::<Gregorian, _>::try_new(
+    ///     locale!("en-US").into(),
+    ///     YMD::short().with_year_style(YearStyle::NoEra),
+    /// )
+    /// .unwrap();
+    ///
+    /// // Era elided even for dates before year 0:
+    /// assert_writeable_eq!(
+    ///     formatter.format(&Date::try_new_gregorian(-1000, 1, 1).unwrap()),
+    ///     "1/1/1001"
```
robertbastian, line 311: This seems like a footgun. Should `NoEra` use the extended year, so that it's -1000?

```suggestion
    ///     "1/1/-1000"
```

components/datetime/src/options/mod.rs
```diff
@@ -273,33 +194,137 @@ pub enum YearStyle {
...
+    /// );
+    /// assert_writeable_eq!(
```
robertbastian, line 313: this as well...

components/datetime/src/raw/neo.rs
mega123-art: I've refactored the match statement to be exhaustive and removed the _ arm. 
please check the latest commit

robertbastian (commented):
> We would accept a PR adding something like YearStyle::NoEra, corresponding to the ECMA-402 proposal eraDisplay: "never".

Can we take a step back please? Neither ECMA-402 nor Temporal have decided on this option.

components/datetime/src/raw/neo.rs
```diff
@@ -185,26 +185,35 @@ impl DatePatternSelectionData {
...
-                        .unwrap_or(YearAmbiguity::EraRequired)
-                })
-                .unwrap_or(YearAmbiguity::EraAndCenturyRequired),
-        ) {
-            (YearStyle::WithEra, _) | (_, YearAmbiguity::EraAndCenturyRequired) => {
+        let ambiguity = input
+            .year
+            .as_ref()
+            .and_then(|y| y.era())
+            .map(|e| e.ambiguity)
```
sffc: Observation: if `y.era()` returns `None`, you are changing the resolved behavior from `YearAmbiguity::EraRequired` to `YearAmbiguity::EraAndCenturyRequired`. This might resolve to a different pattern than before.

mega123-art: yes u were right, corrected in latest commit

components/datetime/src/options/mod.rs
sffc: @robertbastian It is to support this ECMA-402 behavior:

```javascript
new Intl.DateTimeFormat("en").format(new Date(-2000, 0, 1))
// '1/1/2001'
```

which is very footgunny, but _maybe_ okay in some contexts where the era is displayed separately, maybe a timeline where BC and AD are clearly marked.

components/datetime/src/raw/neo.rs
```diff
@@ -185,26 +185,38 @@ impl DatePatternSelectionData {
...
+                y.era()
+                    .map(|e| e.ambiguity)
+                    .unwrap_or(YearAmbiguity::EraRequired)
+            })
+            .unwrap_or(YearAmbiguity::EraAndCenturyRequired);
+
+        let variant = match (year_style, ambiguity) {
+            (YearStyle::WithEra, _) => PackedSkeletonVariant::Variant1,
+
+            (YearStyle::Full, YearAmbiguity::EraAndCenturyRequired) => {
```
sffc:
```suggestion
            (YearStyle::Full, YearAmbiguity::EraAndCenturyRequired | YearAmbiguity::EraRequired) => {
```

components/datetime/src/raw/neo.rs
```diff
@@ -185,26 +185,38 @@ impl DatePatternSelectionData {
...
                 PackedSkeletonVariant::Variant1
             }
-            (YearStyle::Full, _) | (_, YearAmbiguity::CenturyRequired) => {
-                PackedSkeletonVariant::Variant0
-            }
-            (YearStyle::Auto, YearAmbiguity::Unambiguous | YearAmbiguity::EraRequired) => {
-                PackedSkeletonVariant::Standard
+            (YearStyle::Full, _) => PackedSkeletonVariant::Variant0,
```
sffc:
```suggestion
            (YearStyle::Full, YearAmbiguity::CenturyRequired | YearAmbiguity::Unambiguous) => PackedSkeletonVariant::Variant0,
```

sffc:
> > We would accept a PR adding something like YearStyle::NoEra, corresponding to the ECMA-402 proposal eraDisplay: "never".
> 
> Can we take a step back please? Neither ECMA-402 nor Temporal have decided on this option.

ECMA-402 currently has only two choices, and they are "always" or "never" show the era. We support "always" but not "never".

mega123-art:
@sffc sir, since you’ve approved the changes, may we proceed with merging the PR?

PR #7608, opened 2026-02-07
Add tidy check for implicit features

### LINKED ISSUE #3004: Guard against implicit features
#2997 cleans up all implicit features (optional dependencies that generate features). We don't want those, as they're often unintended and unnecessarily increase the feature set.

psg-19:
Fixes #3004 

Summary :- the function `check_implicit_features` collect the paths of all `cargo.toml` files, which packages are in workspace and which dependencies are optional using `cargo metadata`. Then using the `toml` crate it parses the `cargo.toml` files and checks what features are explicitly defined by the developer and whether any feature references `dep:<name>` for each optional `dep`.
It detects implicit features when an optional dependency:
- does not have a same name feature in `[features]`
- and is not referenced via `dep:<name>` anywhere in `[features]`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -695,6 +695,11 @@ dependencies = [
 [[package]]
 name = "depcheck"
 version = "0.0.0"
+dependencies = [
+ "serde",
+ "serde_json",
+ "toml",
+]
 
 [[package]]
 name = "deranged"


tools/make/depcheck/src/main.rs
```diff
@@ -132,7 +133,141 @@ fn test_dep_list(
...
+    name: String,
+    #[serde(default)]
+    optional: bool,
+}
+
+// exception packages that can be allowed to have implicit features.
+const ALLOWED_IMPLICIT_FEATURES: &[(&str, &str)] = &[
+    // in this format - ("some-package", "some-dep"),
+    // ("icu_experimental", "log"), 
+];
```
robertbastian: there shouldn't be any exceptions, it's as easy to add the explicit feature as it is to add it here

robertbastian (commented):
this looks good, please fix the implicit feature in `icu_experimental` so that this passes

psg-19:
Hi just ran `cargo make bakeddata`, I don't know why its showing this much file changes although there no changes in most of those files at all.
Should I raise a new PR for this?
Can you guide me how to fix it ? 
@robertbastian

robertbastian:
> Hi just ran cargo make bakeddata

run it again

psg-19:
> > Hi just ran cargo make bakeddata
> 
> run it again

Did this but still it's the same

PR #7610, opened 2026-02-07
Add tcast crate with TransparentCast derive

### LINKED ISSUE #7607: Low-dependency ref-cast
Previous attempt: #5101
Also see: #6915

We have a decent amount of unsafe code involving ref-cast, which, if eliminated, would help make some crates be fully safe.

We could use the ref-cast crate, but it seems like a heavy dependency (involving `syn`) to replace just one `transmute` call.

#5101 was my first attempt at making `macro_rules` for this. My recollection is that one of the main issues with that approach was that it didn't eliminate the unsafe, it just shifted it around. Would moving that code to a separate crate be a viable option?

Another approach I was thinking was to check the fields by their type, as shown [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=36612361a70c869f0ed5078b065b794e).

What do we think about what to do here?

1. Keep using `transmute` and allow such call sites as in #6915
2. Use `ref-cast` and accept the extra cost
3. Use `ref-cast` when the feature is enabled and `transmute` when the feature is not enabled
4. Add our own crate doing something like #5101 or the above playground
5. Something else?

@Manishearth

sffc:
Fixes #7607
Replaces #5101
See #6915

I was trying to use `ref_cast`, but found that it had multiple deficiencies:

1. Doesn't support standard pointer wrappers like Box
2. Implements a trait by default, which can violate field invariants
    - Note: it supports a private conversion function via `RefCastCustom`
3. Not highly suited for use as a dev-dependency only

So, I made my own proc macro, `TransparentCast`.

The use cases are the same as `RefCast`, as listed in #7607, but it resolves the deficiencies listed above.

I am as sensitive to dependency bloat as everyone else, so I specifically designed this derive to be usable as a dev-dependency. See the example. However, this PR is not intended to be the place to debate whether we decide to use that strategy in icu4x.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3121,6 +3121,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "tcast"
+version = "0.1.0"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.108",
+]
+
+[[package]]
 name = "thiserror"
 version = "2.0.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,6 +70,7 @@ members = [
     "utils/litemap",
     "utils/resb",
     "utils/tinystr",
+    "utils/tcast",
     "utils/tzif",
     "utils/potential_utf",
     "utils/writeable",
@@ -299,6 +300,7 @@ rand = "0.9"
 rand_distr = "0.5"
 rand_pcg = "0.9"
 rayon = "1.3.0"
+ref_cast = "1.0"
 regex = "1.12.2"
 rkyv = "0.7"
 rmp-serde = "1.2.0"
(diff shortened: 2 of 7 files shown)


robertbastian (changes requested):
This does not pull its weight as a dependency. I don't see the point in obsessing over motivated uses of `unsafe`, and pushing them to other crates.

It also doesn't make sense as a dev-only dependency. That pattern increases complexity a lot and actually *removes testing from the unsafe code*.

Manishearth:
FWIW I do plan to look at this, but I haven't had the time.

Overall I think a crate like this is probably useful, but I'm skeptical of how useful it is to ICU4X. From an unsafe review point of view, a smattering of "simple" unsafe usage is far easier to audit than complicated macros and trait setups: "APIs for reducing unsafe" have the problem that they need to be audited for all potential usages ("must be safe no matter what the user throws at it") which is far more challenging than just checking for a `repr(transparent)` and explicit type annotations. "Number of loc unsafe" is not that meaningful a metric except for in the broadest order-of-magnitude sense.

This is counteracted by _popular_ unsafe code crates sometimes being widely audited and trustworthy though that needs to be weighed against their size.

So: I would support publishing this separately as a general non-ICU4X utility, but without having looked too closely, I do not think we should spend time on this or use this in ICU4X.

sffc:
Although I prefer this PR's approach, I don't intend to pursue this further, and I will instead focus on https://github.com/unicode-org/icu4x/pull/7788

Backed up as https://github.com/sffc/omnicu/tree/archive/tcast

PR #7611, opened 2026-02-08
added error codes for compile_fail

### LINKED ISSUE #5752: Change all compile_fail tests to use error number annotations
Apparently (#5747) we can use notation such as

    ```compile_fail,E0271
    // test that fails with code E0271
    ```

This isn't documented anywhere, but it is a feature for Rust developers according to @Manishearth. We should use it everywhere to make sure our `compile_fail` tests continue failing for the same reason.

CC @robertbastian

psg-19:
Fixes #5752 

Added error codes to `compile_fail`


sffc:
thank you!

PR #7612, opened 2026-02-08
Clarify docs on `extensions::Private`

### LINKED ISSUE #4149: Can we make private use subtags constructor always safe?
We have the function `icu::locid::extensions::private::Private::from_vec_unchecked`, but it's not clear to me which invariants need to be upheld for that function to be called "unchecked". Can we just name it `from_vec`?

https://unicode-org.github.io/icu4x/docs/icu_locid/extensions/private/struct.Private.html#method.from_vec_unchecked

@zbraniecki

psg-19:
Fixes #4149 

Summary :- 

- Added clear documentation stating the caller must provide sorted subtags, deduplicated subtags, canonical form ( lowercase ) subtags

- Updated all examples to show properly sorted input which is "bar" before "foo"

### DIFF
--- a/components/locale_core/src/extensions/private/mod.rs
+++ b/components/locale_core/src/extensions/private/mod.rs
@@ -114,8 +114,6 @@ impl Private {
 
     /// A constructor which takes a list of [`Subtag`]s.
     ///
-    /// The subtags should be in canonical form (lowercase).
-    ///
     /// ✨ *Enabled with the `alloc` Cargo feature.*
     ///
     /// # Examples


robertbastian:
I don't think private subtag lists actually need to be sorted (they're private, there shouldn't be any invariants on them). @zbraniecki?

zbraniecki:
Agree. sorting is not required, duplicates are allowed, canonical case is lowercase.

components/locale_core/src/extensions/private/mod.rs
```diff
@@ -112,7 +112,9 @@ impl Private {
         Err(ParseError::InvalidExtension)
     }
 
-    /// A constructor which takes a pre-sorted list of [`Subtag`].
+    /// A constructor which takes a list of [`Subtag`].
```
robertbastian:
```suggestion
    /// A constructor which takes a list of [`Subtag`]s.
```

components/locale_core/src/extensions/private/mod.rs
```diff
@@ -112,7 +112,9 @@ impl Private {
...
+    ///
+    /// The subtags should be in canonical form (lowercase).
```
robertbastian: the `Subtag` type cannot contain non-canonical values

```suggestion
```

PR #7616, opened 2026-02-09
Re-export ValidatedDurationFormatterOptions

### LINKED ISSUE #7615: Impossible to construct icu_experimental::duration::DurationFormatter outside of the crate
This is because the `icu_experimental::duration::validated_options` module is not public, and `ValidatedDurationFormatterOptions` is not re-exported.

Even something like this doesn't work because there's absolutely no way to express the types:

```rust
let formatter = DurationFormatter::try_new(locale!("es-419").into(), options.into().try_into())?;
```

teohhanhui:
Fixes #7615

### DIFF
--- a/components/experimental/src/duration/mod.rs
+++ b/components/experimental/src/duration/mod.rs
@@ -18,3 +18,4 @@ pub use duration::{Duration, DurationSign};
 pub use format::FormattedDuration;
 pub use formatter::DurationFormatter;
 pub use formatter::DurationFormatterPreferences;
+pub use validated_options::ValidatedDurationFormatterOptions;


teohhanhui:
> probably also want to export the unvalidated type?

It's already accessible as the `icu_experimental::duration::options` module is public:

https://docs.rs/icu_experimental/0.4.0/icu_experimental/duration/options/struct.DurationFormatterOptions.html

Or do you want me to add a re-export regardless?

robertbastian:
The whole design is weird. It should probably be 

```rust
DurationFormatterOptions::validate(self) -> Result<ValidatedDurationFormatterOptions>
```

instead of 

```rust
ValidatedDurationFormatterOptions::validate(DurationFormatterOptions) -> Result<Self>
```

and really, the constructors should take `DurationFormatterOptions` and perform the validation.

However, this is still unstable, and this PR unblocks you, so I'm happy to have this for now.

PR #7617, opened 2026-02-09
Add `chrono` and `jiff` integrations

robertbastian:
This adds `chrono` and `jiff` features to `icu_datetime`, which allows using those crate's types as inputs to `DateTimeFormatter`.

It also adds `chrono` and `jiff` features to `icu_time` and `icu_calendar` that allow converting to `icu` types.

## Changelog

This PR and https://github.com/unicode-org/icu4x/pull/7711

icu_calendar: Integrate with `chrono`, `jiff`, and `time`
 - New impls: `From<chrono::NaiveDate>`, `From<jiff::civil::Date>`, `From<time::Date>` for `Date<Gregorian>`

icu_datetime: Integrate with `chrono`, `jiff`, and `time`
 - New impls: Various scaffolding traits on `chrono::NaiveDate`, `jiff::civil::Date`, and `time::Date`, making them formattable.

### DIFF
--- a/components/datetime/Cargo.toml
+++ b/components/datetime/Cargo.toml
@@ -103,6 +103,8 @@ bench = false  # This option is required for Benchmark CI
 
 [package.metadata.cargo-all-features]
 skip_optional_dependencies = true
+# This has a lot of features, run a reduced test that is likely to catch 99% of bugs
+max_combination_size = 3
 
 [[bench]]
 name = "datetime"


components/datetime/src/chrono.rs
```diff
@@ -0,0 +1,330 @@
...
+
+impl GetField<Nanosecond> for chrono::NaiveTime {
+    fn get_field(&self) -> Nanosecond {
+        #[allow(clippy::unwrap_used)] // chrono returns valid fields
+        Nanosecond::try_from(self.nanosecond()).unwrap()
+    }
+}
+
+impl UnstableSealed for chrono::NaiveDate {}
+impl InFixedCalendar<Gregorian> for chrono::NaiveDate {}
```
sffc, line 45: Drive-by comment (not expressing an opinion on the PR)

I seems fine to implement this, but you should also implement `InFixedCalendar<Iso>`, and more importantly `ConvertCalendar` so that you can use this in `DateTimeFormatter` which is what most people should be doing.

robertbastian: `InFixedCalendar<Iso>` is useless, because `Iso` is not formattable. Also both `chrono` and `jiff` say that they implement the Gregorian calendar, not the ISO calendar.

I'm explicitly not implementing `ConvertCalendar`, as `chrono` and `jiff` cannot do that. If you want to use the non-fixed-calendar formatter, you have to convert to our `Date`.

components/time/Cargo.toml
```diff
@@ -32,6 +32,10 @@ zerovec = { workspace = true, features = ["derive", "yoke"] }
 databake = { workspace = true, optional = true, features = ["derive"] }
 serde = { workspace = true, features = ["derive"], optional = true }
 
+chrono = { version = "0.4", optional = true }
+jiff = { version = "0.2", optional = true }
```
BurntSushi: Please scope the feature name to the Jiff version. e.g., `jiff-0_2`. Jiff 1.0 should be out sometime in the first half of this year, and bumping it will be a breaking change. If you include the version in the feature name, then a `jiff` feature can be added for 1.0 in a semver compatible release (while keeping the `jiff-0_2` feature).

components/datetime/src/chrono.rs
sffc: We can implement `ConvertCalendar` where we return ICU4X Date as the conversion output. It seems like if we add these impls, we should add them for the best i18n formatter, not for the Gregorian-only formatter exclusively.

components/calendar/src/chrono.rs
```diff
@@ -0,0 +1,37 @@
+// This file is part of ICU4X. For terms of use, please see the file
+// called LICENSE at the top level of the ICU4X source tree
+// (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
+use crate::cal::gregorian::GregorianDateInner;
+use crate::calendar_arithmetic::ArithmeticDate;
+use crate::Date;
+use crate::Gregorian;
+
+impl From<chrono::NaiveDate> for Date<Gregorian> {
```
sffc, line 10: Issue: I'm not super happy using `Date<Gregorian>` instead of `Date<Iso>` as the canonical representation in icu4x.

The main value proposition of the icu4x calendar crate relative to chrono/jiff/time is our ability to represent different calendar systems.

`Date<Iso>` means "a date that has not yet been converted into a human-readable formattable calendar". If people want to format that as Gregorian, make them explicitly do the conversion.\* That is the intended design: the formatted calendar type should be explicit.

In other words, I want to prevent people from creating a `FixedCalendarDateTimeFormatter` that formats with Gregorian without ever having the word "gregorian" written in their code.

Another footgun, which is admittedly narrow, is that `DateTimeFormatter::format_same_calendar` will succeed for most locales by default but fail for the non-gregorian locales. Although I guess since you aren't implementing the `InSameCalendar` trait, users won't bit this footgun.

\* I think I remember approving a PR a few months ago that made that conversion free, i.e. not going through RD.

robertbastian: Both `chrono` and `jiff` understand themselves to be implementing the Gregorian calendar. The "ISO calendar" is a weird Temporal thing, it's not some magic interchange format.

Manishearth:
Slight worry that can probably be resolved with some research: Do we know the MSRV policy of these packages and their deps? So far we have managed to limit the number of larger deps in `icu_foo`, but we often get issues where we need to e.g. opt our larger tools out of the MSRV check.

Fortunately, it seems like these are small deptrees, so it's probably not a big deal. I suspect anything written by Andrew is sufficiently MSRV-conservative for us.

components/calendar/src/chrono.rs
BurntSushi: @sffc I agree. The Jiff docs could be improved on this point, and it's probably just down to a mixture of mild sloppiness and end user familiarity that it says "Gregorian calendar" everywhere. But AFAIK, Jiff does actually implement the ISO 8601 proleptic calendar because it has a year `0`. Similarly for `chrono`.

robertbastian:
`chrono` is 1.61

BurntSushi:
@Manishearth Speaking for Jiff, what policy I do have is documented here: https://github.com/BurntSushi/jiff?tab=readme-ov-file#minimum-rust-version-policy --- Which is basically the same as `regex`, but doesn't specifically commit to a specific time window.

I can't imagine a scenario in which I'll bump to anything newer than N-9 (i.e., 1 year old Rust). I should probably just add that to the policy so that folks can rely on it. OK, I put up a PR for that: https://github.com/BurntSushi/jiff/pull/512 (I did carve out a possible exception, although it seems unlikely to happen? It kind of depends on, e.g., how crates like `windows-sys` treat MSRV. In practice they're pretty conservative I think.) Based on https://github.com/unicode-org/icu4x/pull/7576, it seems like this should be good enough for y'all though.

In practice, I suspect I'll never bump to anything newer than what's in Debian stable.

Manishearth:
N-9 is much wider than our window, so that works!

components/calendar/src/chrono.rs
robertbastian: Our `Gregorian` calendar uses a year 0 in both its constructor as well as its internal representation. The whole BC/CE thing (which is the only difference between `Gregorian` and `Iso`) is a human-display difference. The reason why we want to integrate `jiff` with `icu` is for human display, so it makes sense to use the `Calendar` which is suitable for that, not the `Calendar` which implements a technical interchange format.

PR #7618, opened 2026-02-09
Bump icu_capi in the release

### LINKED ISSUE #7605: icu_properties 2.1.2 is not consistent with icu_capi 2.1.1
Running `cargo test` on a build containing the latest releases of `icu_properties` and `icu_capi` throws this error:

```
thread 'unstable::properties_enums::test::test_all_cases_covered' (5722) panicked at third_party/rust/icu_capi/v2/src/./properties_enums.rs:3034
Found Script value not supported in ffi
test unstable::properties_enums::test::test_all_cases_covered ... FAILED
```

(icu_capi has Chisoi but not Garay, icu_properties has Garay but not Chisoi)

Commenting out the Script test gives a similar issue for InSc.

I think we haven't made all of the patch releases that that issue needed.

Manishearth:
Fixes #7605

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ Several crates have had patch releases in the 2.1 stream:
         - Add conversion between `icu::properties::props::Script` and `icu::locale::subtags:Script` (unicode-org#7270)
     - (2.1.2) `icu_segmenter`
         - Add non-complex line and word break constructors (unicode-org#7268)
+    - (2.1.2) `icu_capi`
+        - Bring in alignment with 2.1.2 icu_properties changes.
 - Data model and providers
     - (2.1.2) `icu_provider_source`, `icu_locale_data`, `icu_datetime_data`, `icu_experimental_data`
         - Update to CLDR 48.1 (unicode-org#7396)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1224,7 +1224,7 @@ version = "2.1.1"
 
 [[package]]
 name = "icu_capi"
-version = "2.1.1"
+version = "2.1.2"
 dependencies = [
  "diplomat",
  "diplomat-runtime",
--- a/ffi/capi/Cargo.toml
+++ b/ffi/capi/Cargo.toml
@@ -15,7 +15,7 @@ include = [
     "LICENSE",
     "README.md"
 ]
-version.workspace = true
+version = "2.1.2"
 
 authors.workspace = true
 categories.workspace = true


robertbastian:
Please submit to main and cherry-pick to the release branch

Manishearth:
There's no "fix" to land on main. The fix is already on main, _and_ on the release branch, we just never published a release for this crate. We should have when we made this patch for `icu_properties`, but we didn't.

Manishearth:
I'm happy to just `cargo publish` my local commit, but I wanted a PR for CI and review tracking.

robertbastian:
The fact that `icu_capi` has been bumped to 2.1.2 should be on main. We should not have versions in release branches (and on crates.io) overtake main.

edit: versions and *changelog entries*

Manishearth:
Fair enough. I thought we weren't bumping component crates on main, but whatever.

robertbastian:
> I thought we weren't bumping component crates on main, but whatever.

We are, at least I am. I think it's particularly important that the changelog on main contains all the entries for patch releases.

Manishearth:
@robertbastian Yes, I was planning on forward-porting the changelog from this PR onto main.


I'm not sure I agree with this process overall. I think we should be ensuring that patch releases get CId on patch branches in some fashion, and the current process just doesn't cover that.

We seem to do this differently each time and I'm never clear on this. What we've written down in "release.md" talks about "fixing issues on main" but it's unclear how that applies for issues that are only issues on the release branch.

robertbastian:
> We seem to do this differently each time and I'm never clear on this. What we've written down in "release.md" talks about "fixing issues on main" but it's unclear how that applies for issues that are only issues on the release branch.

Well, fixing the issue includes bumping relevant versions. The reason why the CP should be main -> release is because our CI on main works slightly better, and it's easier to review PRs going into main as we're all familiar with the baseline there.

Manishearth:
Yes, but we should also get the signal that release branch CI passes, or fails for "CI is outdated" reasons.

I think the current set of cascading uplift issues would have been less likely had the initial uplifts been PRs to the release branch, where we would have noticed that more things need bumping.

A core issue is we need to be sure we uplift _everything_, and this works out differently for _new_ fixes vs uplifting an existing fix.

If I make a new fix and bump the versions/changelog, that is usually easy enough to cherry pick. But if I'm trying to do this for a PR that has already landed, we're not really CIing anything meaningful unless we make a PR to the release branch.

Release branch CI is less reliable, but it is still a signal.

Manishearth:
??

I merged a PR to main, then cherry picked it, and pushed to the release branch. I closed this PR.

PR #7620, opened 2026-02-10
data struct with codepoint builder for radicals

opnuub:
Issue: https://github.com/unicode-org/icu4x/issues/6941

Added:
- components/segmenter/src/provider/radicals.rs | data struct 

Modified: 
- components/segmenter/src/provider/mod.rs | added two lines to use radicals.rs also in provider
- provider/source/src/segmenter/unihan.rs | added data provider and test for new codepoint trie

### DIFF
--- a/components/segmenter/examples/experimental_segmenter.rs
+++ b/components/segmenter/examples/experimental_segmenter.rs
@@ -15,8 +15,8 @@ use adaboost::Predictor;
 use cnn::{CnnSegmenter, RawCnnData};
 use icu_segmenter::{options::WordBreakOptions, WordSegmenter, WordSegmenterBorrowed};
 use std::time::SystemTime;
-use icu::segmenter::provider::SegmenterUnihanIrgV1;   // ✅ needed to load IRG marker
-use icu_provider::prelude::*;                         // ✅ needed for DataRequest/DataResponse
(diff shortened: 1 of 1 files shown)


components/segmenter/src/provider/radicals.rs
```diff
@@ -0,0 +1,32 @@
...
+
+icu_provider::data_struct!(
+    UnihanIrgData<'_>,
+    #[cfg(feature = "datagen")]
+);
+
+icu_provider::data_marker!(
+    /// `SegmenterUnihanV1`
+    SegmenterUnihanV1,
+    "segmenter/unihan/v1",
```
robertbastian: is this the only data struct that can possibly be generated from unihan data? if not, it should have a more concrete name

opnuub:
> Is there an issue for this? What's the goal here?

Issue: #6941

components/segmenter/src/provider/radicals.rs
opnuub: I can name it SegmenterUnihanIrgV1 to be more concrete, but yes it is the only data struct generated from unihan data that's for the segmenter

sffc: Yes, SegmenterUnihanIrgV1 is a better name, in case we want to add more unihan data in the future.

PR #7621, opened 2026-02-10
Add BoundLocaleDataProvider and some impls for it

sffc:
#3260 

Please take a look at this design for a formatter that retains a piece of the data provider for runtime lookup of data marker attributes, intended for use cases such as currencies, units, and display names.

Reviewable commit-by-commit.

No AI was used when writing this, but I might use one to fix up the clippy and feature warnings in CI.

The docs are somewhat limited. Please focus on the docs tests for the end-to-end usage examples. I will add more docs before merging.

### DIFF
--- a/provider/blob/src/blob_data_provider.rs
+++ b/provider/blob/src/blob_data_provider.rs
@@ -3,7 +3,8 @@
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
 use crate::blob_schema::BlobBoundLocaleSchema;
-use crate::blob_schema::BlobSchema;
+use crate::blob_schema::RawBlob;
+use crate::blob_schema::ValidatedBlobSchema;
 #[cfg(feature = "alloc")]
 use alloc::boxed::Box;
 use icu_provider::Cart;
@@ -89,7 +90,7 @@ use yoke::*;
 /// ```
 #[derive(Clone)]
 pub struct BlobDataProvider {
-    pub(crate) data: Yoke<BlobSchema<'static>, Option<Cart>>,
+    pub(crate) data: Yoke<ValidatedBlobSchema<'static>, Option<Cart>>,
 }
 
 impl core::fmt::Debug for BlobDataProvider {
@@ -108,7 +109,7 @@ impl BlobDataProvider {
     pub fn try_new_from_blob(blob: Box<[u8]>) -> Result<Self, DataError> {
         Ok(Self {
             data: Cart::try_make_yoke(blob, |bytes| {
-                BlobSchema::deserialize_and_check(&mut postcard::Deserializer::from_bytes(bytes))
+                ValidatedBlobSchema::deserialize_and_check(&mut postcard::Deserializer::from_bytes(bytes))
             })?,
         })
     }
@@ -117,7 +118,7 @@ impl BlobDataProvider {
     /// [`try_new_from_blob`](BlobDataProvider::try_new_from_blob) and is allocation-free.
     pub fn try_new_from_static_blob(blob: &'static [u8]) -> Result<Self, DataError> {
         Ok(Self {
-            data: Yoke::new_owned(BlobSchema::deserialize_and_check(
+            data: Yoke::new_owned(ValidatedBlobSchema::deserialize_and_check(
                 &mut postcard::Deserializer::from_bytes(blob),
             )?),
         })
@@ -125,7 +126,7 @@ impl BlobDataProvider {
 
     #[doc(hidden)] // for testing purposes only: checks if it is using the Bigger format
     pub fn internal_is_using_bigger_format(&self) -> bool {
-        matches!(self.data.get(), BlobSchema::V003Bigger(..))
+        matches!(self.data.get(), ValidatedBlobSchema::V004Bigger(..))
     }
 }
 
@@ -135,15 +136,17 @@ impl DynamicDataProvider<BufferMarker> for BlobDataProvider {
         marker: DataMarkerInfo,
         req: DataRequest,
     ) -> Result<DataResponse<BufferMarker>, DataError> {
-        let payload: Yoke<(&[u8], Option<u64>), Option<Cart>> = self
+        let payload: Yoke<RawBlob, Option<Cart>> = self
             .data
             .try_map_project_cloned(|blob, _| blob.load(marker, req))?;
         let mut metadata = DataResponseMetadata::default();
-        metadata.buffer_format = Some(BufferFormat::Postcard1);
-        metadata.checksum = payload.get().1;
+        metadata.buffer_format = Some(payload.get().kind.into_buffer_format());
+        metadata.checksum = payload.get().checksum;
         Ok(DataResponse {
             metadata,
-            payload: DataPayload::from_yoked_buffer(payload.map_project(|(bytes, _), _| bytes)),
+            payload: DataPayload::from_yoked_buffer(
+                payload.map_project(|raw_blob, _| raw_blob.bytes),
+            ),
         })
     }
 }
@@ -204,17 +207,17 @@ pub struct BlobBoundLocaleDataProvider {
 }
 
 impl BindLocaleDataProvider<BufferMarker> for BlobDataProvider {
-    type BoundLocaleDataProvider<'data> = BlobBoundLocaleDataProvider;
+    type BoundLocaleDataProvider = BlobBoundLocaleDataProvider;
     fn bind_locale(
         &self,
         marker: DataMarkerInfo,
         req: DataRequest,
-    ) -> Result<BindLocaleResponse<Self::BoundLocaleDataProvider<'static>>, DataError> {
+    ) -> Result<BindLocaleResponse<Self::BoundLocaleDataProvider>, DataError> {
         let payload: Yoke<(BlobBoundLocaleSchema, Option<u64>), Option<Cart>> = self
             .data
             .try_map_project_cloned(|blob, _| blob.bind_locale(marker, req))?;
         let mut metadata = DataResponseMetadata::default();
-        metadata.buffer_format = Some(BufferFormat::Postcard1);
+        metadata.buffer_format = Some(payload.get().0.kind.into_buffer_format());
         metadata.checksum = payload.get().1;
         Ok(BindLocaleResponse {
             metadata,
@@ -233,7 +236,7 @@ impl BoundLocaleDataProvider<BufferMarker> for BlobBoundLocaleDataProvider {
         let blob = self.data.get();
         let payload = blob.load(req)?;
         let mut metadata = DataResponseMetadata::default();
-        metadata.buffer_format = Some(BufferFormat::Postcard1);
+        metadata.buffer_format = Some(blob.kind.into_buffer_format());
         // Note: the checksum is returned by `bind_locale()` instead of `load_bound()`
         Ok(BoundLocaleDataResponse { metadata, payload })
     }
--- a/provider/blob/src/blob_data_provider.rs
+++ b/provider/blob/src/blob_data_provider.rs
@@ -2,13 +2,19 @@
 // called LICENSE at the top level of the ICU4X source tree
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
+use crate::blob_schema::BlobBoundLocaleSchema;
 use crate::blob_schema::BlobSchema;
 #[cfg(feature = "alloc")]
 use alloc::boxed::Box;
-use icu_provider::Cart;
-use icu_provider::DynamicDryDataProvider;
 use icu_provider::buf::BufferFormat;
 use icu_provider::prelude::*;
+use icu_provider::unstable::BindLocaleDataProvider;
+use icu_provider::unstable::BindLocaleResponse;
+use icu_provider::unstable::BoundLocaleDataProvider;
+use icu_provider::unstable::BoundLocaleDataResponse;
+use icu_provider::unstable::DataAttributesRequest;
+use icu_provider::Cart;
+use icu_provider::DynamicDryDataProvider;
 use yoke::*;
 
 /// A data provider that reads from serialized blobs of data.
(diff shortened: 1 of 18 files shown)


sffc:
In case it helps, here's a summary of what you can do after this PR:

You can write code like

```rust
let formatter = HelloWorldAttributeFormatter::try_new_with_buffer_provider(
    &provider,
    locale!("en").into()
).unwrap();

assert_writeable_eq!(formatter.format("reverse").unwrap(), "Olleh Dlrow");
assert_writeable_eq!(formatter.format_for_prefix("rev").unwrap(), "Olleh Dlrow");
```

Please observe that the data marker attribute is given in the format function, not in the constructor.

To make this work, we save a small slice of the data provider structure inside of the HelloWorldAttributeFormatter, which has a generic parameter that varies based on the provider type, powered by some new traits.

This PR adds impls of the new traits in the BlobDataProvider and HelloWorldProvider universes. We will need to add them to baked data, and probably FsDataProvider.

provider/core/src/buf/serde.rs
```diff
@@ -59,6 +60,22 @@ where
...
+    ///
+    /// This requires enabling the deserialization Cargo feature
+    /// for the expected format(s):
+    ///
+    /// - `deserialize_json`
+    /// - `deserialize_postcard_1`
+    /// - `deserialize_bincode_1`
+    ///
+    /// ✨ *Enabled with the `serde` Cargo feature.*
+    pub fn new(inner: &'a P) -> Self {
```
robertbastian: the canonical way to wrap a provider in a `DeserializingBufferProvider` is to call `.as_deserializing()`. we use this in a lot of docs, and I don't think we should introduce an alternative API that does the same

provider/blob/src/blob_data_provider.rs
```diff
@@ -2,11 +2,16 @@
 // called LICENSE at the top level of the ICU4X source tree
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
+use crate::blob_schema::BlobBoundLocaleSchema;
 use crate::blob_schema::BlobSchema;
 #[cfg(feature = "alloc")]
 use alloc::boxed::Box;
 use icu_provider::buf::BufferFormat;
 use icu_provider::prelude::*;
+use icu_provider::unstable::BindLocale;
```
robertbastian: if you start depending on unstable APIs, do we need to use a ~ dep?

provider/core/src/hello_world.rs
```diff
@@ -291,12 +364,37 @@ pub struct HelloWorldFormatter {
...
+/// let fmt = HelloWorldAttributeFormatter::try_new_with_buffer_provider(
+///     &HelloWorldProvider.into_json_provider(),
+///     locale!("en").into(),
+/// )
+/// .expect("locale exists and has attributes");
+///
+/// assert_writeable_eq!(fmt.format("reverse").unwrap(), "Olleh Dlrow");
+/// ```
+#[derive(Debug)]
+pub struct HelloWorldAttributeFormatter<P: BoundLocaleDataProvider<HelloWorldV1>> {
```
robertbastian, line 392: is this a bad example, or does this mean every formatter using this pattern needs to be generic in a provider?

robertbastian (commented):
As far as I remember the display names design, it doesn't involve the formatter holding on to the provider. I don't think this approach gives nice APIs, as the formatters will need to be generic in the providers. This also requires a lot of new traits and types. Overall I'm very negative on this.

sffc:
The use case is when you need to load a list of many different data marker attributes.

If you load each one in sequence from a provider, you need to re-read the locale (and marker in the case of blob provider) for every lookup.

What this design does is it creates an intermediate "bound locale provider" that is halfway through a data load operation. You can then more efficiently perform attribute lookups.

sffc:
See https://github.com/unicode-org/icu4x/pull/7663 for benchmark results. @robertbastian

sffc:
My preference is to move forward with this design as the "primary" Display Names API. However, @robertbastian does not agree and prefers to make single-display-name the "primary" API. @Manishearth attempted to find compromise, saying that we can start with single-display-name and ask for feedback if multi-display-name is motivated based on user feedback.

I do not find this to be a desirable compromise, because we have what I see as compelling data that multi-display-name is significantly (2x) faster than single-display-name when loading a list of names, which is both the most common use case and also the one that amplifies any performance gains. I would prefer to not implement as the "primary" API one that _we know today_ is slow.

Another argument that was raised was that it is easy to refactor single-display-name into multi-display-name. I consider this half-true. On the one hand, the data markers are the same, and we only need to rewrite the traits and the implementation. On the other hand, the whole user-facing API is subject to change, with all its docs tests, FFI, etc., as well as the implementation. It believe it to be substantially less work to implement multi-display-name out of the barn than to implement single-display-name followed later by a refactor to multi-display-name.

In order to move forward, I intend to write a "minimal" single-display-name API in such a way that I spend most of my time on the things that are needed by both approaches, i.e., focusing on the data model. I will not add FFI or thorough test coverage. I will add only enough docs to make the types usable in a 2.2 experimental release.

This position is not intended to be an endorsement for the single-display-name API, and I have great fear that my implementing it this way will result in the temptation for clients to start using it and for ICU4X to graduate it. I am setting aside these fears in order to make engineering progress.

Manishearth:
>  @Manishearth attempted to find compromise, saying that we can start with single-display-name and ask for feedback if multi-display-name is motivated based on user feedback.

(This was not my suggestion. I shall write more later)

Manishearth:
My suggestion was that you move forward with [Option 4](https://github.com/unicode-org/icu4x/issues/3260#issuecomment-3881644475) (the iterator API), implemented internally the slow way. We document it as "this is somewhat slow, we hope to optimize it". We _could_ early optimize it if it can be done in a purely internal way.

We then link people to one of the issues here and encourage them to talk to us about their use cases.

This is a good use of the fanfare around experimental new features in release notes. We get to see what people care about, and this can inform API shape and where to spend time optimizing.


> Another argument that was raised was that it is easy to refactor single-display-name into multi-display-name. I consider this half-true. On the one hand, the data markers are the same, and we only need to rewrite the traits and the implementation. On the other hand, the whole user-facing API is subject to change, with all its docs tests, FFI, etc., as well as the implementation. It believe it to be substantially less work to implement multi-display-name out of the barn than to implement single-display-name followed later by a refactor to multi-display-name.



This is why I would prefer to do this all at once. But I like the sound of exposing the iterator API as an exploratory path for 2.2.

sffc:
Nota bene: For DisplayNames, I feel that we do not need to look far for feedback on use cases. Intl.DisplayNames is one of the newer ECMA-402 APIs. You can read about its motivation here:

https://github.com/tc39/proposal-intl-displaynames

It was designed as `new Intl.DisplayNames("en", { type: "region" }).of("UK")` because it was expected that you would want to load many display names for the same language.

Manishearth:
> I feel that we do not need to look far for feedback on use cases

Yes, but I think user feedback on API shape and performance is still useful. I don't think we _need_ it to proceed, but a solution that gives us a concrete thing to put in front of users without necessarily finalizing it is nice IMO.

sffc:
Following up from a conversation today with @robertbastian and @Manishearth, I wanted to show two more use cases for BoundLocaleDataProvider

### Unit formatting with conversion

Since we may not know the output unit when formatting with a usage context, the formatter needs to have access to multiple potential output units, somewhere between 1 and ~10.

A unit formatter could leverage BoundLocaleDataProvider to retain access to the units it might need.

Each formatting operation would involve hitting the ZeroTrie of attributes, which is not free, but I expect it to be small relative to the overall formatting operation.

Instead of storing a BoundLocaleDataProvider, an alternative solution would be to store a SmallVec of payloads. This is probably how we would do this in a world without BLDP, but BLDP still has advantages including stack size, no-alloc. (SmallVec's advantage is no generic on the formatter.)

### Getting display names as an iterator

It also occurred to me that a ZeroTrie also offers a fast iteration mechanism. After you've chosen your locale, a function with the following signature would be easy to implement:

```rust
impl RegionDisplayNames<B> {
    pub fn iter(&self) -> impl Iterator<(Region, &str)> + '_ { ... }
}
```

If loading _most_ or _all_ display names, this is even faster than the `.of` function in ECMA, since we don't need to look anything up in the trie; we just loop over it.

---

@robertbastian asked that if we implement this, that we also make a macro that adds a default "naive" implementation of the trait. This seems fine.

---

I also should follow up on how this gets implemented in the provider adapters. Some are easier than others.

The following are trivial: either, empty, filter, fixed.

The two interesting ones:

### Forking Providers

I think this works fine by wrapping both children in another forking provider.

1. BindLocaleDataProvider calls BindLocaleDataProvider on all children
2. It stores the resulting providers in its fields, returning the result as a BoundLocaleDataProvider
3. The BoundLocaleDataProvider simply delegates to one child or the other, as usual

### Fallback Providers

This is a bit more tricky and is one of the reasons I started with postcard instead of baked.  The most efficient solution for lookup speed would be for BoundLocaleFallbackProvider to have a smallvec of child providers, one for each locale in the fallback chain. In most cases we should end up with <=3 locales in the fallback chain that actually have data, so we should be able to usually fit this on the stack.

Manishearth:
The impact on other providers seems reasonable.

Aside from internal providers, who would have to care about this trait? People doing custom data loading impls, yes? (Which is a power user use case, so it's allowed to be complex)

sffc:
I plan to implement this for LocaleFallbackProvider because that will be an interesting signal for the benchmarks. I'm not entirely sure what to expect: a higher startup cost but lower runtime cost.

sffc:
I copied the benchmarks into this PR and added ones that use a `Box<dyn BoundLocaleDataProvider>`.

<details>
<pre>
locale_names/region/baked/all/1by1
                        time:   [16.277 ms 16.392 ms 16.537 ms]
                        change: [-8.4275% -7.2284% -5.9933%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) high mild
  3 (3.00%) high severe
locale_names/region/baked/10x10pct/1by1
                        time:   [175.21 µs 177.34 µs 179.89 µs]
                        change: [-12.197% -10.019% -7.5308%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 13 outliers among 100 measurements (13.00%)
  5 (5.00%) high mild
  8 (8.00%) high severe
locale_names/region/baked/all/old_map_struct
                        time:   [5.0773 ms 5.1166 ms 5.1610 ms]
                        change: [-5.8200% -4.2660% -2.8105%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) high mild
  5 (5.00%) high severe
locale_names/region/baked/10x10pct/old_map_struct
                        time:   [53.893 µs 54.137 µs 54.413 µs]
                        change: [-4.6229% -3.3484% -2.1997%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) low mild
  7 (7.00%) high mild
  3 (3.00%) high severe
locale_names/region/postcard/all/1by1
                        time:   [19.675 ms 20.003 ms 20.395 ms]
                        change: [-8.6210% -6.4702% -4.2571%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  1 (1.00%) low mild
  4 (4.00%) high mild
  6 (6.00%) high severe
locale_names/region/postcard/10x10pct/1by1
                        time:   [178.16 µs 181.21 µs 184.71 µs]
                        change: [-3.9104% -1.6390% +0.7084%] (p = 0.18 > 0.05)
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe
locale_names/region/postcard/all/boundlocale/typed
                        time:   [10.255 ms 10.348 ms 10.449 ms]
                        change: [-7.1759% -5.5889% -4.0667%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 15 outliers among 100 measurements (15.00%)
  3 (3.00%) low mild
  3 (3.00%) high mild
  9 (9.00%) high severe
locale_names/region/postcard/10x10pct/boundlocale/typed
                        time:   [79.610 µs 80.414 µs 81.480 µs]
                        change: [-5.2711% -3.8400% -2.4071%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  8 (8.00%) high mild
  2 (2.00%) high severe
locale_names/region/postcard/all/boundlocale/boxdyn
                        time:   [10.127 ms 10.181 ms 10.250 ms]
                        change: [-4.9644% -4.1112% -3.2546%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  5 (5.00%) high mild
  2 (2.00%) high severe
locale_names/region/postcard/10x10pct/boundlocale/boxdyn
                        time:   [81.931 µs 83.093 µs 84.454 µs]
Found 17 outliers among 100 measurements (17.00%)
  3 (3.00%) low mild
  6 (6.00%) high mild
  8 (8.00%) high severe


</pre>
</details>

In table form:

| Data Set | Provider Storage | Provider Trait Setup | Time |
| :--- | :--- | :--- | :--- |
| all | baked | 1by1 | 16.392 ms [16.277 ms, 16.537 ms] |
| all | baked | old_map_struct | 5.1166 ms [5.0773 ms, 5.1610 ms] |
| all | postcard | 1by1 | 20.003 ms [19.675 ms, 20.395 ms] |
| all | postcard | BoundLocale (Typed) | 10.348 ms [10.255 ms, 10.449 ms] |
| all | postcard | BoundLocale (Dyn Box) | 10.181 ms [10.127 ms, 10.250 ms] |
| 10x10pct | baked | 1by1 | 177.34 µs [175.21 µs, 179.89 µs] |
| 10x10pct | baked | old_map_struct | 54.137 µs [53.893 µs, 54.413 µs] |
| 10x10pct | postcard | 1by1 | 181.21 µs [178.16 µs, 184.71 µs] |
| 10x10pct | postcard | BoundLocale (Typed) | 80.414 µs [79.610 µs, 81.480 µs] |
| 10x10pct | postcard | BoundLocale (Dyn Box) | 83.093 µs [81.931 µs, 84.454 µs] |

So it seems adding the Dyn creates only a small percentage change to the performance.

Manishearth:
(comment moved to https://github.com/unicode-org/icu4x/issues/3913#issuecomment-4386340552)

sffc:
We have a few threads. I posted my currently preferred solution in https://github.com/unicode-org/icu4x/issues/3913#issuecomment-4385404708 and pushback on the data duplication solution in https://github.com/unicode-org/icu4x/issues/7824#issuecomment-4385122960 (which is where you first suggested it).

My intent was to focus this PR on issues relating specifically to the design of BoundLocaleDataProvider, not display names in general.

Manishearth:
Ah, sorry. Let me move this comment there.

sffc:
@robertbastian is okay with landing this behind an unstable feature in icu_provider. The displaynames crate will also have an unstable feature that enables the icu_provider unstable feature.

wdyt @Manishearth?

Manishearth:
I think that's fine.

sffc:
Updated design (I thought I posted it but I don't remember where): instead of an `unstable` feature on `icu_provider`, add a new crate, perhaps `icu_core_unstable`, name to be bikeshed, that is synced with a tilde dependency and has all of our internal unstable things.

EDIT: Here are the notes: https://github.com/unicode-org/icu4x/issues/3913#issuecomment-4421689920

sffc:
A change I made relative to my original design is that I want `FooDisplayName[Borrowed]` to borrow from the owned type without any cows. This means that the `BoundLocaleDataProvider` needs to return references to the struct fields, which can be done by either:

1. Owning the data struct object
2. Returning something that can be borrowed as the struct object

I implemented (2) by leveraging the existing `MaybeAsVarULE` trait. `BoundLocaleDataProvider` now returns `&EncodedStruct` instead of `Yoke<DataStruct>`. This means `BoundLocaleDataProvider` will only work for:

1. Structs that use the VarULE optimization, _and_
2. Providers that use the VarULE optimization

(1) is a reasonable restriction. (2) requires a new postcard blob schema, but it is backwards-compatible: it will store some structs as postcard and others as `VarULE`. Meanwhile, `AsDeserializingBufferProvider` needs to convert the `VarULE` to a concrete struct, which is doable with the correct bounds.

---

I plan to split this PR into smaller pieces. I already made #8224 and #8225. Another standalone piece will be supporting VarULE as a serialization format. I will put in the time to finish that PR only if there is clear agreement that such a PR will not be blocked from landing.

PR #7622, opened 2026-02-10
Implement `Into<Locale>` for preference structs.

jedel1043:
This will simplify https://github.com/boa-dev/boa/pull/4589 in the future.

### DIFF
--- a/components/locale_core/src/preferences/mod.rs
+++ b/components/locale_core/src/preferences/mod.rs
@@ -418,7 +418,7 @@ pub trait PreferenceKey: Sized {
 
 #[macro_export]
 #[doc(hidden)]
-// #[cfg(feature = "alloc")]
+#[cfg(feature = "alloc")]
 macro_rules! __impl_into_locale {
     ($name:ident, $($key:ident: $pref:ty),*) => {
         impl From<$name> for $crate::Locale {
@@ -439,12 +439,12 @@ macro_rules! __impl_into_locale {
     }
 }
 
-// #[macro_export]
-// #[doc(hidden)]
-// #[cfg(not(feature = "alloc"))]
-// macro_rules! __impl_into_locale {
-//     ($name:ident, $($key:ident: $pref:ty),*) => {}
-// }
+#[macro_export]
+#[doc(hidden)]
+#[cfg(not(feature = "alloc"))]
+macro_rules! __impl_into_locale {
+    ($name:ident, $($key:ident: $pref:ty),*) => {};
+}
 
 /// A macro to facilitate generation of preferences struct.
 ///
@@ -546,7 +546,7 @@ macro_rules! __define_preferences {
             }
         }
 
-        $crate::preferences::__impl_into_locale!($name, $($key: $pref,)*);
+        $crate::__impl_into_locale!($name, $($key: $pref),*);
 
         impl $name {
             /// Extends the preferences with the values from another set of preferences.
--- a/components/locale_core/src/preferences/mod.rs
+++ b/components/locale_core/src/preferences/mod.rs
@@ -416,6 +416,10 @@ pub trait PreferenceKey: Sized {
     }
 }
 
+// This utility macro needs to be split from the main `__define_preferences` macro
+// to gate the implementation behind the `alloc` feature; using `cfg(feature = "alloc")`
+// inside `__define_preferences` does not use the "alloc" feature from this
+// crate, it uses the "alloc" feature from the crate that expands (calls) the macro.
 #[macro_export]
 #[doc(hidden)]
 #[cfg(feature = "alloc")]
--- a/components/locale_core/src/preferences/mod.rs
+++ b/components/locale_core/src/preferences/mod.rs
@@ -416,6 +416,36 @@ pub trait PreferenceKey: Sized {
     }
 }
 
+#[macro_export]
+#[doc(hidden)]
+// #[cfg(feature = "alloc")]
+macro_rules! __impl_into_locale {
+    ($name:ident, $($key:ident: $pref:ty),*) => {
+        impl From<$name> for $crate::Locale {
+            fn from(other: $name) -> Self {
+                use $crate::preferences::PreferenceKey;
+                let mut result = Self::from(other.locale_preferences);
+                $(
+                    if let Some(value) = other.$key {
+                        if let Some(ue) = <$pref>::unicode_extension_key() {
+                            let val = value.unicode_extension_value().unwrap();
+                            result.extensions.unicode.keywords.set(ue, val);
+                        }
+                    }
+                )*
+                result
+            }
+        }
+    }
+}
+
+// #[macro_export]
+// #[doc(hidden)]
+// #[cfg(not(feature = "alloc"))]
+// macro_rules! __impl_into_locale {
+//     ($name:ident, $($key:ident: $pref:ty),*) => {}
+// }
+
 /// A macro to facilitate generation of preferences struct.
 ///
 ///
@@ -516,21 +546,7 @@ macro_rules! __define_preferences {
             }
         }
 
-        // impl From<$name> for $crate::Locale {
-        //     fn from(other: $name) -> Self {
-        //         use $crate::preferences::PreferenceKey;
-        //         let mut result = Self::from(other.locale_preferences);
-        //         $(
-        //             if let Some(value) = other.$key {
-        //                 if let Some(ue) = <$pref>::unicode_extension_key() {
-        //                     let val = value.unicode_extension_value().unwrap();
-        //                     result.extensions.unicode.keywords.set(ue, val);
-        //                 }
-        //             }
-        //         )*
-        //         result
-        //     }
-        // }
+        $crate::preferences::__impl_into_locale!($name, $($key: $pref,)*);
 
         impl $name {
             /// Extends the preferences with the values from another set of preferences.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,7 @@ Fully filled in up to 30c187f4b7
         - Add `LocalePreferences::from_locale_strict` (unicode-org#7377)
         - Fix `LanguageIdentifier::normalize_utf8` example (unicode-org#7372)
         - Use better types in `LocalePreferences` (unicode-org#7360)
+        - Implement `Into<Locale>` for preference structs (unicode-org#7622)
     - `icu_normalizer`
         - Move `harfbuzz-traits` implementations into component crates (unicode-org#7200)
     - `icu_pattern`


components/locale_core/src/preferences/mod.rs
```diff
@@ -416,6 +416,40 @@ pub trait PreferenceKey: Sized {
...
+macro_rules! __impl_into_locale {
+    ($name:ident, $($key:ident: $pref:ty),*) => {
+        impl From<$name> for $crate::Locale {
+            fn from(other: $name) -> Self {
+                use $crate::preferences::PreferenceKey;
+                let mut result = Self::from(other.locale_preferences);
+                $(
+                    if let Some(value) = other.$key {
+                        if let Some(ue) = <$pref>::unicode_extension_key() {
+                            let val = value.unicode_extension_value().unwrap();
```
sffc, line 435: Question/Issue: Will every `value` always have a unicode extension value?

sffc (commented):
Thanks! Would you mind adding some round-trip test cases?

robertbastian:
I don't think this makes much sense conceptually. A locale that is obtained from say `DateTimeFormatterPreferences` cannot be used for anything else, as the conversion into the preferences is lossy. Preferences that are not used by DTF won't round-trip.

sffc:
The use case is

```javascript
new Intl.DateTimeFormat("en-u-ca-chinese-co-pinyin").resolvedOptions().locale
// 'en-u-ca-chinese'
```

robertbastian:
Preferences are not designed to be the resolved options, they are not aware of any fallback that happens. Does `resolvedOptions()` just need to throw away irrelevant extensions, or also produce the result of data resolution?

It's very counterintuitive for `From` to be implemented in both directions, but not round-trip. I think this should be `into_locale()` or something. Another approach that might suffice for Temporal would be to implement `Writeable` on preferences to get the string representation.

sffc:
It appears to do filtering of relevant extension keywords, removing unsupported values, and resolving aliases.

```javascript
new Intl.DateTimeFormat("en-u-ca-ethiopic-amete-alem-co-pinyin-nu-latn1").resolvedOptions().locale
'en-u-ca-ethioaa'
```

https://tc39.es/ecma402/#sec-resolvelocale

sffc:
> It's very counterintuitive for `From` to be implemented in both directions, but not round-trip.

Well, preference-to-locale is the lossless direction. The direction we currently have is the lossy direction.

> I think this should be `into_locale()` or something.

I am always happy with concrete functions over `From` impls

> Another approach that might suffice for Temporal would be to implement `Writeable` on preferences to get the string representation.

We could do both, but I see no good reason to export Writeable without exporting the Locale conversion, since you can re-parse the writeable into a Locale and just be less efficient.

robertbastian:
> Well, preference-to-locale is the lossless direction. The direction we currently have is the lossy direction.

You can't really talk about lossless if you only have one direction. Right now, we have a Locale domain A and a preferences domain B (which is different from A). `From<Locale> for Preferences` is a function from domain A to domain B (which is not bijective, but that's fine). As soon as you add the B to A function though, not having bijectivity is very unexpected.


> We could do both, but I see no good reason to export Writeable without exporting the Locale conversion, since you can re-parse the writeable into a Locale and just be less efficient.

Yes, but using that `Locale` is also generally wrong, so I like making it more complicated.

jedel1043:
@robertbastian Hmm, what about only implementing a conversion into `extensions::unicode::Unicode` instead? TBH we don't really need the full `Locale` in the linked PR, we just need the unicode extensions part. Does that implementation make more sense?

robertbastian:
I would approve `fn as_keywords(&self) -> unicode::Keywords`.

sffc:
A potential issue with that signature is whether we should include -u-rg and -u-sd in the keywords. If we don't include them, then we aren't including all the keywords. If we do include them, then we aren't returning all of the region information.

Manishearth:
If it's lossy perhaps we should have `to_locale_lossy()` or `to_locale_for_resolved_options()`?

robertbastian:
> If we do include them, then we aren't returning all of the region information.

can you elaborate what you mean by this? we're returning all unicode keywords that are relevant to the preferences

sffc:
> > If we do include them, then we aren't returning all of the region information.
> 
> can you elaborate what you mean by this? we're returning all unicode keywords that are relevant to the preferences

The Preferences struct contains LocalePreferences plus zero or more keywords relevant to the formatter. It seems wrong to pull only some of the data out of LocalePreferences without returning all of the data from LocalePreferences.

robertbastian:
- @sffc I support the proposal. `as_keywords()` is weird because it indlues some things from LocalePreferences but not all the things.
- @robertbastian Agreed, I walk back that suggestion
- @zbraniecki What's wrong with the PR?
- @robertbastian It's not clear why they need it. They need it for resolvedOptions?
- @zbraniecki It's the wrong way to do it
- @manishearth Maybe they're doing it the wrong way on purpose, boa has the freedom to be "mostly correct" especially temporarily.
- @sffc It's the right way to do resolved locale, can you elaborate?
- @zbraniecki It doesn't handle fallback
- @robertbastian
- @zbanieecki e.g. might ask for zulu-buddhist and get en-u-ca-gregory
- @robertbastian it's a different type and doesn't represent anything useful
- @hsivonen think it's a bit weird .... First of all preferences may be possibly hard to understand from otusdie icu4x. E.g. "what's collator options, what's locale, ..."
- @hsivonen my understanding is that CollatorOptions is what the app dev should set, and CollatorPreferences is "could come from the user but it can also be a negotiation", kind of fuzzy
- @hsivonen once we have a Preferences object, if there was a use case for serializing it.. then the serialization would look like a locale. In that sense, if we do have these objects it seems a bit weird that we're not allowed to go from prefs to locales and have that be the serializable thing
- @Manishearth I mostly agree with @hsivonen; it is useful to look at these in terms of locales, for debugging and understandability. The BCP-47 extensions are the canonical way of discussing these across projects. Having `as_bcp_47()` is useful
- @robertbastian These types were designed as input types to ICU4X. They are narrow so that ICU4X can take the right preferences from a more complicated locale object, maybe merge it with another preference object. They aren't meant for Boa to do more.
- @robertbastian Debugging is useful; we could debug it to a BCP-47 string, but we could just debug it as fields.
- @sffc My impression was that preference bags were not only input types but also a mechanism for merging preferences coming from many sources.
- @zbraniecki I can see an argument to serialize it into a locale. I haven't thought about this. If you take a preference bag composed from many sources, and you pass it to a constructor...
- @robertbastian If you take a preference bag from one constructor and convert it to a locale and back to a preference bag for a different constructor, it is lossy
- @zbraniecki If the use case is debug and serialize, we can impl Debug and Serialize the preference bag.
- @hsivonen Then it's hostile that you need to go through a string instead of type to type. UTS 35 defines a way to represent a bunch of keywords
- @sffc Should we add impl Writeable and impl Display, too?
- @Manishearth I don't want impl Serialize... it is expensive and not what people will usually want
- @zbraniecki Should we make `impl Display` be "LocalePreferences { en-US-u-ca-buddhist }"
- @sffc I thought we agreed that serialization was a valid use case. Yes?
- @Manishearth Yes, but we should serialize as fields. I want users to have the *option*. I think BCP47 is a bad default option.
- @robertbastian I would say serialization is an antipattern. You don't want to store these in a database. You want to store the user's whole locale in the database.
- @Manishearth You might want to persist things, like a user language profile
- @sffc We have a serialization format for these, it is BCP-47, we should not invent new serialization formats
- @Manishearth/@robertbastian (talking over each other, hard to understand)
- @zbraniecki Logging systems will log the input locale preferences. 
- @robertbastian They can log Debug.- @zbraniecki That's very verbose. What I proposed in line 37 is just easier to read
- @zbraniecki That's very verbose. What I proposed in line 37 is just easier to read
- @hsivonen Each one of these is a subset of BCP-47, and BCP-47 is the serialization format. If you go from BCP-47 to Preferences, and you convert back, you get back a locale that drops the unreachable keywords. Is that really so bad that we need to be hostile that we make clients jump through hoops like @zbraniecki's string format above? If we don't want it automatic, then we can make an explicit function.
- @sffc +1
- @robertbastian We're trying to build a library usable by non-i18n experts. If we implement `impl Into`, it is too easy to do the wrong thing.
- @robertbastian You can implement this conversion in userland
- @sffc No you can't because it is non-exhaustive so you can't keep up with the latest preferences. It is ICU4X that knows which preferes are there. It is better to implement this in a library than in userland.
- @Manishearth I think we should have a method, not an impl. I think it's useful
- @sffc Should it be `to_locale` or should it be scarier?
- @Manishearth `to_locale`, `into_locale`, whatever it should be called
- @robertbastian Can we call it `to_locale_lossy`?
- @hsivonen The earlier step is what was lossy
- @sffc `recover_locale`?
- @robertbastian It's not recovering the locale that was in the beginning
- @robertbastian I still don't understand the motivation for this
- @zbraniecki I want to see what Boa wants to do with this

Conclusion:

- Add a method. Maybe `to_locale` or maybe something else like `to_locale_lossy`
- Method name to be bikeshed in the issue

LGTM: @hsivonen @sffc @Manishearth 

Abstain: @robertbastian @zbraniecki

robertbastian:
@jedel1043 could you explain your use case for this a bit more?

jedel1043:
@robertbastian Boa migrated from custom structs to ICU4X's preferences with a trait that abstracts the concept of a "ServicePreference" https://github.com/jedel1043/boa/blob/203e8989ebd9f237b4891112ba742f9b46e87149/core/engine/src/builtins/intl/mod.rs#L234

However, we needed to have something to convert from preferences to `Locale`'s Unicode extensions such that we could set the extensions on the final locale per ECMA402's requirements.

We kinda implemented the conversion by setting the unicode extensions manually, but it would be better to have that in ICU4X so that it doesn't get out of date if new prefs get introduced.

Manishearth:
It seems like `set_unicode_extension()` on these preference types might be useful.


> such that we could set the extensions on the final locale per ECMA402's requirements.


When does this come up?

jedel1043:
@Manishearth on [`ResolveLocale`](https://tc39.es/ecma402/#sec-resolvelocale), emphasis on:

> ... returns a representation of the match that also includes corresponding data from localeData and a resolved value for each element of relevantExtensionKeys (defaulting to data from the matched locale, superseded by data from the requested Unicode locale extension sequence if present and then by data from options if present). If the matched element from requestedLocales contains a Unicode locale extension sequence, it is copied onto the language tag in the [[Locale]] field of the returned Record, omitting any keyword Unicode locale nonterminal whose key value is not contained within relevantExtensionKeys or type value is superseded by a different value from options.

sffc:
> It seems like `set_unicode_extension()` on these preference types might be useful.

We already eliminated this choice; see discussion. I think the option we agree on is a function that returns a `Locale`, modulo the bikeshed, and modulo a compelling use case from @jedel1043.

> on ResolveLocale

So the issue here is that Preferences resolves keywords, but it doesn't yet know if the value is actually supported. The ResolveLocale spec requires actually checking inside the data. You can see this in e.g.

```javascript
new Intl.NumberFormat("en-u-nu-arab").resolvedOptions().locale
'en-u-nu-arab'
new Intl.NumberFormat("en-u-nu-zzzz").resolvedOptions().locale
'en'
```

zbraniecki:
@jedel1043 ok, thanks so much for this explanation. If I understand your case correctly that corroborates the concern that @robertbastian and I raised - `Preferences` is an *input* object, `resolvedLocales` is an *outcome* object.
They cannot be interchanged.

We need to construct `ResolvedOptions` from what data was resolved into, so the pipeline is:

```
Preferences -> DataResolution -> ResolvedOptions
```

sffc:
Since this API doesn't solve @jedel1043's use case, I'm closing this PR; but if anyone can articulate a use case where this type of API is needed, see the above discussion for how to name it, and we can land it then.

PR #7623, opened 2026-02-10
Improve performance of property name lookup

### LINKED ISSUE #5599: Make property name lookup `get_loose` more efficient
Address some of the follow-ups in https://github.com/unicode-org/icu4x/pull/5576.

CC @robertbastian

psg-19:
Fixes #5599 

Summary :- 
- made `step()` function to return `bool`.
- removed `whitespace characters`  and `hyphen` check from the trie-skipping loop
- made the `original case` and `other case` recurse only if `step` returned `true` instead of using `or_else`
- added `other_case != ascii && cursor.step(other_case)` check which avoids redundant work for digits and symbols because digits and  symbols don't have another case, so `to_ascii_lowercase()` returns the same byte.

## Changelog

icu_properties: Improve performance of property name lookup

### DIFF
--- a/provider/source/src/properties/enum_codepointtrie.rs
+++ b/provider/source/src/properties/enum_codepointtrie.rs
@@ -84,7 +84,7 @@ impl super::uprops_serde::enumerated::EnumeratedPropertyMap {
         Ok(code_point_trie)
     }
 
-    pub(crate) fn names_to_values(&self) -> Result<BTreeMap<&str, u16>, DataError> {
+    pub(crate) fn names_to_values(&self) -> BTreeMap<&str, u16> {
         let mut map = BTreeMap::new();
 
         for range in &self.ranges {
@@ -103,16 +103,7 @@ impl super::uprops_serde::enumerated::EnumeratedPropertyMap {
             }
         }
 
-        for name in map.keys() {
-            if name.contains('-') || name.bytes().any(|b| b.is_ascii_whitespace()) {
-                return Err(
-                    DataError::custom("Property name contains '-' or whitespace")
-                        .with_display_context(name),
-                );
-            }
-        }
-
-        Ok(map)
+        map
     }
 
     pub(crate) fn values_to_names_long(&self) -> BTreeMap<u16, &str> {
@@ -256,7 +247,16 @@ macro_rules! expand {
                         core::str::from_utf8(<$prop as EnumeratedProperty>::NAME).unwrap(),
                         core::str::from_utf8(<$prop as EnumeratedProperty>::SHORT_NAME).unwrap()
                     )?;
-                    let trie = data.names_to_values()?
+                    let map = data.names_to_values();
+                    for name in map.keys() {
+                        if name.contains('-') || name.bytes().any(|b| b.is_ascii_whitespace()) {
+                            return Err(
+                                DataError::custom("Property name contains '-' or whitespace")
+                                    .with_display_context(name),
+                            );
+                        }
+                    }
+                    let trie = map
                         .into_iter()
                         .map(|(k, v)| (k, v as usize))
                         .collect::<ZeroTrieSimpleAscii<_>>()


utils/zerotrie/src/cursor.rs
```diff
@@ -210,8 +210,8 @@ impl ZeroTrieSimpleAsciiCursor<'_> {
     /// assert_eq!(cursor.take_value(), None);
     /// ```
     #[inline]
-    pub fn step(&mut self, byte: u8) {
-        reader::step_parameterized::<ZeroTrieSimpleAscii<[u8]>>(&mut self.trie.store, byte);
+    pub fn step(&mut self, byte: u8) -> bool {
```
sffc: Please document what the return value `bool` indicates.

components/properties/src/names.rs
```diff
@@ -304,17 +304,24 @@ fn get_strict_u16(payload: &PropertyValueNameToEnumMap<'_>, name: &[u8]) -> Opti
...
+        // Skip underscore, hyphen in trie, only recurse if step succeeds
+        for skip in [b'_', b'-'] {
             let mut skip_cursor = cursor.clone();
-            skip_cursor.step(skip);
-            if let Some(r) = recurse(skip_cursor, rest) {
-                return Some(r);
+            if skip_cursor.step(skip) {
+                if let Some(r) = recurse(skip_cursor, rest) {
+                    return Some(r);
+                }
```
sffc: Observation: We unconditionally skip `b'_' | b'-'` in the name, so can we unconditionally skip them here, too, instead of recursing?

components/properties/src/names.rs
```diff
@@ -304,17 +304,24 @@ fn get_strict_u16(payload: &PropertyValueNameToEnumMap<'_>, name: &[u8]) -> Opti
 
 /// Avoid monomorphizing multiple copies of this function
 fn get_loose_u16(payload: &PropertyValueNameToEnumMap<'_>, name: &[u8]) -> Option<u16> {
+    // limit input length to prevent stack overflow
+    const MAX_NAME_LENGTH: usize = 100;
+    if name.len() > MAX_NAME_LENGTH {
+        return None;
+    }
+
```
sffc: Nit: Although I previously expressed concern about stack overflow, I don't think we should add a guard like this unless we can actually reproduce it. The recursion appears to be limited by the length of the longest value name in the trie, so it's hard to see this actually overflowing.

components/properties/src/names.rs
```diff
@@ -304,17 +304,24 @@ fn get_strict_u16(payload: &PropertyValueNameToEnumMap<'_>, name: &[u8]) -> Opti
...
     fn recurse(mut cursor: ZeroTrieSimpleAsciiCursor, mut rest: &[u8]) -> Option<usize> {
         if cursor.is_empty() {
             return None;
         }
 
-        // Skip whitespace, underscore, hyphen in trie.
-        for skip in [b'\t', b'\n', b'\x0C', b'\r', b' ', 0x0B, b'_', b'-'] {
+        // Skip underscore, hyphen in trie, only recurse if step succeeds
+        for skip in [b'_', b'-'] {
```
robertbastian: does this assume that the trie doesn't contain `b'\t', b'\n', b'\x0C', b'\r', b' ', 0x0B`? If so, validate this invariant in `SourceDataProvider`.

robertbastian: the tries don't contain `-` either, so you only have to skip `_`.

you might want to validate these assumptions in `SourceDataProvider`.

components/properties/src/names.rs
```diff
@@ -304,17 +304,24 @@ fn get_strict_u16(payload: &PropertyValueNameToEnumMap<'_>, name: &[u8]) -> Opti
...
             let mut skip_cursor = cursor.clone();
-            skip_cursor.step(skip);
-            if let Some(r) = recurse(skip_cursor, rest) {
-                return Some(r);
+            if skip_cursor.step(skip) {
+                if let Some(r) = recurse(skip_cursor, rest) {
+                    return Some(r);
+                }
             }
```
robertbastian: I'm not convinced this change is necessary. `recurse` immediately returns if the cursor is empty

utils/zerotrie/src/cursor.rs
robertbastian: huh, this is semver-breaking, why does our semver CI not notice this?

components/properties/src/names.rs
psg-19: Yes recurse returns immediately if the cursor is empty, that guard only saves one function call since it hits return none after entering the recursion.  I'll remove the `step()` check here and keep the original pattern.

components/properties/src/names.rs
psg-19: we cannot unconditionally skip them because a property might be a prefix of another.

utils/zerotrie/src/cursor.rs
sffc: does Rust consider it semver-breaking to change a return type from `()` to something else? I guess you could have been calling the function as the last expression in a block that needs to return `()`?

components/properties/src/names.rs
psg-19: should I add a check in `icu4x/provider/source/src/properties/enum_codepointtrie.rs` for this ?

robertbastian: yes

provider/source/src/properties/enum_codepointtrie.rs
```diff
@@ -104,10 +104,7 @@ impl super::uprops_serde::enumerated::EnumeratedPropertyMap {
         }
 
         for name in map.keys() {
-            assert!(
-                !name.contains('-'),
-                "Property name {name:?} contains '-'"
-            );
+            assert!(!name.contains('-'), "Property name {name:?} contains '-'");
```
robertbastian: issue: this should return a `DataError`, not panic. you might have to do this later where you can return `DataError`s

robertbastian: issue: also check for the other characters that you removed from the check in names.rs

components/properties/src/names.rs
robertbastian: also in the two places below

psg-19: if so then I guess there's no need to return a boolean in the `step()` function, I'll remove that too

provider/source/src/properties/enum_codepointtrie.rs
```diff
@@ -103,7 +103,16 @@ impl super::uprops_serde::enumerated::EnumeratedPropertyMap {
             }
         }
 
-        map
+        for name in map.keys() {
```
robertbastian: nit: I'd prefer if this lived in line 255. the `names_to_values` collecting here should be infallible, this requirement starts being relevant when you create the trie from this `BTreeMap`

psg-19:
> I preferred the version where `step` returns a bool and we skip the extra call to `recurse` but 🤷

as suggested by @robertbastian, I reverted that change 😅

PR #7626, opened 2026-02-10
Actually reject date strings with invalid calendar annotations

robertbastian:
The current code is buggy because it only performs this check if the annotation parses into a `CalendarAlgorithm`. That is stupid.

## Changelog

icu_calendar: Reject date strings with invalid calendar annotations


Manishearth (approved):
Should we have special handling here for the islamic subsetting?

robertbastian:
what do you mean?

Manishearth:
@robertbastian using calendar `islamic-tbla` while specifying `[u-ca=islamic]` or something. AIUI internally that's `CalendarAlgorithm::Hijri(None)`

robertbastian:
I don't think we should do that

PR #7627, opened 2026-02-10
Use `Date::from_raw` in `Date::try_from_str`

robertbastian:
This saves a RD round-trip for Gregorian calendars.

### DIFF
--- a/components/time/src/ixdtf.rs
+++ b/components/time/src/ixdtf.rs
@@ -755,12 +755,13 @@ impl Time {
             .transpose()?
             .unwrap_or_default();
 
-        Ok(Self::try_new(
-            time_record.hour,
-            time_record.minute,
-            time_record.second,
-            nanosecond,
-        )?)
+        // values are in range by construction
+        Ok(Time {
+            hour: crate::types::Hour(time_record.hour),
+            minute: crate::types::Minute(time_record.minute),
+            second: crate::types::Second(time_record.second),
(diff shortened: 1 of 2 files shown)


Manishearth:
In particular, in the process of fixing that I discovered that tests you have written are not straightforward to "still make sense" in the new world (specifically, `test_iso_overflow`), so if you care deeply about that test, it might be better for you to do this.

robertbastian:
this should be independent of any work on the range, as this constructor doesn't need to change

robertbastian:
`test_iso_overflow` just needs to be adapted to the tighter RD and constructor ranges, there is not conceptual change

this is not the place to discuss this

components/calendar/src/ixdtf.rs
```diff
@@ -112,21 +113,24 @@ impl<A: AsCalendar> Date<A> {
...
+        );
+        let inner = GregorianDateInner(ArithmeticDate::new_unchecked(
+            date_record.year,
+            date_record.month,
+            date_record.day,
+        ));
 
-        // `rd` is in range, see `_RD_RANGE_REGRESSION_CHECK`
-        Ok(Date::from_rata_die(rd, calendar))
+        Ok(Date::from_raw(inner, Gregorian).to_calendar(calendar))
```
Manishearth, line 129: thought: I'd kind of prefer to reduce `Date::from_raw()` calls but this is fine.

Issue #7035, opened 2025-10-04
Make DateTime Ord


michalmoc:
Since `DateTime` is just `Date` and `Time`, both of which are already Ord, it seems to me the `DateTime` should be Ord too, sorting first by date and then by time. Or are there any particular reasons it cannot be?

sffc:
It's worth considering. What do you need it for? We have wanted to keep `DateTime` a minimal type focused on localization, and it's not clear what use case an `Ord` impl would serve, except perhaps storing it in a BTreeSet.

A reason to err on the side of caution is that impl Ord for DateTime is a bit misleading since it does not contain a time zone. And, if we add impl Ord to DateTime, then it raises questions about whether we should put it on ZonedDateTime as well, and how that impl would behave.

michalmoc:
It's a matter of common use. It's not rare to want to compare which event was earlier, or to sort a list of events by date(time) or as you've said, to store it in BTree. I know it can be done on user-side with something like `(lhs.date, lhs.time) < (rhs.date, rhs.time)`, but it's annoying and why would you write a crate which is annoying?

If you consider non-timezoned times misleading, then the Ord should be removed from Time, for sake of consistence. But again, I believe even low-level crates should contain some bare minimum of quality of life, especially when it is free like in this case.

And yes, ZonedDateTime probably should be Ord too, with semantic of "convert both Times to UTC (which may shift Date) and then compare". A bit more tricky to write, but it seems to me like something in scope of the project? Like user should be able to compare dates not wondering which country they come from?

sffc:
Thanks for clarifying.

Would you mind elaborating on your usecase for using ICU4X DateTime for this purpose as opposed to the equivalents in `time`, `chrono`, `jiff`, `temporal_rs`, etc?

> And yes, ZonedDateTime probably should be Ord too, with semantic of "convert both Times to UTC (which may shift Date) and then compare". A bit more tricky to write

Right, that seems like the most reasonable expectation for how two ZonedDateTimes would compare. However, it's not as easy as it sounds. ICU4X supports TimeZones with a location but not an offset, such as `2025-10-07T00:00[Europe/Berlin]`. Such ZonedDateTimes can't be converted to UTC without a timezone database to look up the correct offset, but ICU4X by design does not carry a timezone database dependency, since it is not needed for time zone localization. (Most/all of the other crates listed above carry a timezone database.)

michalmoc:
[...]

PR #7630, opened 2026-02-10
Make some `ZonedDateTime<Iso, UtcOffset>` functions generic in calendar

robertbastian:
I don't know why we only had these for the `Iso` calendar, there's nothing stopping us from having them for all calendars, because they only need `.to_rata_die()`.

## Changelog
icu_time: 
* Make some `ZonedDateTime<Iso, UtcOffset>` functions generic in calendar

### DIFF
--- a/components/time/src/zone/mod.rs
+++ b/components/time/src/zone/mod.rs
@@ -287,7 +287,7 @@ impl<'a> zerovec::maps::ZeroMapKV<'a> for TimeZone {
 ///
 /// // ... or by adding a local time
 /// let time_zone_at_time = time_zone.at_date_time(DateTime {
-///     date: Date::try_new_gregorian(2023, 12, 2).unwrap(),
+///     date: Date::try_new_coptic(1996, 12, 2).unwrap(),
 ///     time: Time::start_of_day(),
 /// });
 /// ```


Manishearth:
I think Shane made the choice here, @sffc was this intentional?

components/time/src/types.rs
```diff
@@ -306,20 +306,26 @@ where
 {
 }
 
-const UNIX_EPOCH: RataDie = calendrical_calculations::gregorian::fixed_from_gregorian(1970, 1, 1);
-
-impl Ord for ZonedDateTime<Iso, UtcOffset> {
+impl<A: AsCalendar> Ord for ZonedDateTime<A, UtcOffset> {
```
sffc, line 315: Issue: We have both resisted adding the generic Ord impl to ZonedDateType. Why the change of heart?

components/time/src/types.rs
```diff
@@ -395,10 +401,12 @@ impl ZonedDateTime<Iso, UtcOffset> {
...
         }
     }
+}
 
-    pub(crate) fn to_epoch_milliseconds_utc(self) -> i64 {
+impl<A: AsCalendar> ZonedDateTime<A, UtcOffset> {
+    pub(crate) fn epoch_millis(&self) -> i64 {
         let ZonedDateTime { date, time, zone } = self;
-        let days = date.to_rata_die() - UNIX_EPOCH;
+        let days = date.to_rata_die() - UNIX_EPOCH_RD;
```
sffc: Observation: this is probably not a lot more efficient in Iso than in other calendars that have fast RD conversions, so there isn't much technical cost with this.

sffc (changes requested):
We have always treated ZonedDateTime (and DateTime) as a thin container object. We had a very narrow exception which was that `ZonedDateTime<Iso>` had these semantics about having an ordering and being convertible to a timestamp, since we lack another timestamp type. For a while the ZoneNameTimestamp constructors took positional arguments, but it made a number of other things cleaner if it took a single argument. That's the only reason why that conversion function has that particular signature.

I thought this position was something where not only we agreed but that you were a champion.

sffc:
Here's the PR where you wrote and landed docs that said to use ZonedDateTime as a formatter input type and not to expect it to implement Ord: https://github.com/unicode-org/icu4x/pull/7275

Also https://github.com/unicode-org/icu4x/issues/7035#issuecomment-3380725855, with a slightly different tone but the same conclusion.

components/datetime/src/provider/time_zones.rs
```diff
@@ -415,7 +415,7 @@ pub(crate) mod legacy {
             "2022-03-12 13:00Z",
             "2023-03-08 16:00Z",
         ] {
-            let t = ZoneNameTimestamp::from_zoned_date_time_iso(
+            let t = ZoneNameTimestamp::from_zoned_date_time(
```
sffc: As far as the ZoneNameTimestamp APIs go, the ZoneNameTimestamp is a power-user type. We don't need to make it be nice and pretty and generic. It really should be constructible not from a ZonedDateTime but rather from a UNIX Epoch timestamp integer, like

```rust
impl ZoneNameTimestamp {
    pub fn from_epoch_millis(epoch_millis: i64) -> Self
}
```

But, we instead landed the more typesafe thing, despite the imperfection of assigning semantic meaning to a type that is otherwise supposed to be just a bag of fields.

components/time/src/types.rs
robertbastian: yes, and there are savings of not having to do `.to_calendar(Iso)` first, which is two more RD conversions

components/datetime/src/provider/time_zones.rs
robertbastian: If you have a `DateTime<Buddhist>` and want to make that zoned, you'll have to convert the `Date` to `Iso` just to create a `ZoneNameTimestamp` for that date. This is pointless, because it will go Buddhist -> RD -> Iso -> RD instead of just Buddhist -> RD.

robertbastian:
> Here's the PR where you wrote and landed docs that said to use ZonedDateTime as a formatter input type and not to expect it to implement Ord: https://github.com/unicode-org/icu4x/pull/7275

Added in that PR:

> /// [^1]: [`ZonedDateTime<UtcOffset>`] is an exception to this, as the whole time zone
///       identity is part of the type, and there are no local time discontinuities.
///       Therefore it actually identifies an absolute time and implements [`PartialOrd`].

This should not just apply to `ZonedDateTime<Iso, UtcOffset>`, but `ZonedDateTime<C, UtcOffset>`

robertbastian:
All this PR does is replace functions that take `DateTime<Iso>` or `ZonedDateTime<Iso, Z>` by functions that take `DateTime<C>` or `ZonedDateTime<C, Z>`. There are no "new" APIs or implementations, it just removes needless ISO conversions from call sites when the API just needs an RD internally.

components/time/src/types.rs
robertbastian: as you can see the implementation is already there, just less generic

components/datetime/src/fieldsets.rs
```diff
@@ -947,7 +947,7 @@ macro_rules! impl_zone_marker {
         /// // Time zone info for America/Chicago in the summer
         /// let zone = TimeZone(subtag!("uschi"))
         ///     .with_offset("-05".parse().ok())
-        ///     .at_date_time_iso(DateTime{ date: Date::try_new_iso(2022, 8, 29).unwrap(), time: Time::start_of_day() });
+        ///     .at_date_time(DateTime{ date: Date::try_new_gregorian(2022, 8, 29).unwrap(), time: Time::start_of_day() });
```
sffc: Nit: don't use this PR to propagate the use of `Gregorian`


```suggestion
        ///     .at_date_time(DateTime{ date: Date::try_new_iso(2022, 8, 29).unwrap(), time: Time::start_of_day() });
```

components/datetime/src/fieldsets.rs
```diff
@@ -1318,7 +1318,7 @@ pub mod zone {
         /// // Time zone info for Europe/Istanbul in the winter
         /// let zone = TimeZone(subtag!("trist"))
         ///     .with_offset("+03".parse().ok())
-        ///     .at_date_time_iso(DateTime{ date: Date::try_new_iso(2022, 1, 29).unwrap(), time: Time::start_of_day() });
+        ///     .at_date_time(DateTime{ date: Date::try_new_gregorian(2022, 1, 29).unwrap(), time: Time::start_of_day() });
```
sffc:
```suggestion
        ///     .at_date_time(DateTime{ date: Date::try_new_iso(2022, 1, 29).unwrap(), time: Time::start_of_day() });
```

I'm not going to annotate the rest of these.

components/datetime/src/provider/time_zones.rs
sffc: What do you feel about adding `ZoneNameTimestamp::from_epoch_milliseconds` and `ZonedDateTime::to_epoch_milliseconds()` that talk to each other via i64

sffc:
> > Here's the PR where you wrote and landed docs that said to use ZonedDateTime as a formatter input type and not to expect it to implement Ord: #7275
> 
> Added in that PR:
> 
> > /// [^1]: [`ZonedDateTime<UtcOffset>`] is an exception to this, as the whole time zone
> > ///       identity is part of the type, and there are no local time discontinuities.
> > ///       Therefore it actually identifies an absolute time and implements [`PartialOrd`].
> 
> This should not just apply to `ZonedDateTime<Iso, UtcOffset>`, but `ZonedDateTime<C, UtcOffset>`

ok, you are making the calendar generic on this one but not the zone. Fair enough.

components/datetime/src/fieldsets.rs
robertbastian: I'm using Gregorian as an example of a formattable calendar. `Iso` is not

sffc: You aren't formatting the calendar, you are constructing a timestamp.

robertbastian: Yes, but these are examples. `TimeZoneInfo`s are almost always used in `ZonedDateTime`, where you already have a `Date<C>`.

components/datetime/src/provider/time_zones.rs
robertbastian: > the imperfection of assigning semantic meaning to a type that is otherwise supposed to be just a bag of fields.

lol you added this stuff and I didn't even get to review

> What do you feel about adding `ZoneNameTimestamp::from_epoch_milliseconds` and `ZonedDateTime::to_epoch_milliseconds()` that talk to each other via i64

negatively, because then the user has `i64`s floating around, and I don't want to define a `Timestamp` newtype in addition to this. also the status quo is fine, because `TimeZoneInfo::at_date_time` abstracts this away

components/datetime/src/fieldsets.rs
sffc: If you are reusing a `Date<Gregorian>` that you are also formatting, ok, but don't use `Date<Gregorian>` for the purposes of constructing a ZoneNameTimestamp.

I will block the PR on this. We should not use `Date<Gregorian>` for anything other than explicitly formatting of Gregorian dates.

robertbastian: that's so petty

components/datetime/src/provider/time_zones.rs
sffc: > lol you added this stuff

You have landed multiple changes to ZoneNameTimestamp, including the one that added this particular constructor (https://github.com/unicode-org/icu4x/pull/6746)

> and I didn't even get to review

Huh? You were on the reviewer list for https://github.com/unicode-org/icu4x/pull/6513, which replaced my first attempt https://github.com/unicode-org/icu4x/pull/6322 based on your feedback.

components/datetime/src/fieldsets.rs
sffc: What makes you think it's petty? Iso vs Gregorian is a foundational distinction. Iso is for machine dates, Gregorian is for human dates. If this were in test code, I wouldn't block, but this is in docs code, and docs should not use the Gregorian type for things it is not meant for.

components/datetime/src/provider/time_zones.rs
sffc: What if we left the current constructor as-is (perhaps deprecated) and replaced it with a constructor with the signature

```rust
impl ZoneNameTimestamp {
    pub fn from_timestamp_parts(rata_die: RataDie, time: Time, offset: UtcOffset)
}
```

components/datetime/src/fieldsets.rs
robertbastian: > Iso is for machine dates, Gregorian is for human dates.

That's your personal interpretation. Rata die are for machine dates, ISO is an abomination of a `Calendar`, because it's identical to Gregorian in all but the era ('default' is a stupid era name), and it's the only calendar that is not in CLDR and is not formattable.

sffc: I wish you would not call it my "personal interpretation". It is an interpretation deeply encoded into Temporal's design.

robertbastian: Well we're not Temporal. We use `RataDie` for machine dates.

sffc: The design of calendars in icu_calendar was based on Temporal. RataDie is a calendrical_calculations type which you added as a re-export in icu_calendar at a later time. It was never agreed that RataDie would "replace" the ISO calendar in icu_calendar for machine dates.

robertbastian: RataDie is a calendrical_calculations type that we started using when we based our implementations on calendrical_calculations. Using it as an interchange type is a performance improvement over using Iso.

robertbastian: I've changed some examples to use different calendars than Gregorian

components/datetime/src/fieldsets.rs
```diff
@@ -1698,7 +1700,7 @@ pub mod zone {
         /// // Time zone info for Asia/Tokyo
         /// let zone = TimeZone::from_iana_id("Asia/Tokyo")
         ///     .with_offset(UtcOffset::try_from_seconds(9 * 3600).ok())
-        ///     .at_date_time_iso(DateTime{ date: Date::try_new_iso(2022, 1, 29).unwrap(), time: Time::start_of_day() });
+        ///     .at_date_time(DateTime{ date: Date::try_new_japanese_with_calendar("reiwa", 2, 1, 29, Japanese::new()).unwrap(), time: Time::start_of_day() });
```
sffc: ok, I admit it's cute that you can use reiwa when setting your zone name timestamp.

We still should use ISO over Gregorian in our examples, though, since most often your zone name timestamp is set from a timestamp, and ISO is for machine dates.

robertbastian: Gregorian is as much a machine date as ISO, they're the same calendar.

the point here is to show that this is in winter, so it's fair to use an actual calendar. if we just wanted to set a machine date, I'd set a UNIX timestamp.

sffc (changes requested):
No, this makes docs objectively worse than before. Please just leave the docs alone and I will hapily approve this.

robertbastian:
please keep that discussion in https://github.com/unicode-org/icu4x/pull/7720

components/time/src/zone/mod.rs
```diff
@@ -285,8 +286,8 @@ impl<'a> zerovec::maps::ZeroMapKV<'a> for TimeZone {
 /// );
 ///
 /// // ... or by adding a local time
-/// let time_zone_at_time = time_zone.at_date_time_iso(DateTime {
-///     date: Date::try_new_iso(2023, 12, 2).unwrap(),
+/// let time_zone_at_time = time_zone.at_date_time(DateTime {
+///     date: Date::try_new_gregorian(2023, 12, 2).unwrap(),
```
sffc: Nit: please change this to either `iso`, for showing how to set it to a machine date, or a non-Gregorian calendar like `hebrew`.

robertbastian: no

sffc: I won't approve a PR that adds documentation that uses to `try_new_gregorian` outside the context of demonstrating how to use a FixedCalendar formatter.

robertbastian: It's weird that you'd approve Hebrew though

sffc: I offered Hebrew as an alternative because it is more clear that it is a human calendar and not a machine date.

sffc: I offered Hebrew as an alternative because it is more clear that it is a human calendar and not a machine date.

robertbastian: But isn't Gregorian a "human calendar" by your definition

sffc: Hebrew = definitely human date
Iso = definitely computer date
Gregorian = human date, but more likely to be confused for a computer date, so let's just avoid it except in the situations where we really need it (FixedCalendar formatter)

robertbastian: a date can be a "human date" and a "computer date" at the same time. all our dates are "computer dates" because they are on a computer.

ISO is a fake calendar, so let's just avoid it.

your whole reasoning is so stupid, but you win because I'm going on vacation and want this merged. please consider whether this tantrum is worth discussing this PR *for a whole month*, and how that fits into our recent discussion about team velocity and PR reviews.

sffc: You could have fixed this when I first left this comment on Feb 11 and it would have been merged a month sooner. Let's not escalate language by calling each other's reasoning "stupid" or "tantrum".

PR #7631, opened 2026-02-10
Reduce checks in `ZonedDateTime` parsing

robertbastian:
Currently the date and time are reparsed to populate the time zone timestamp.

Also fixes the `PartialOrd` impl for `ZonedDateTime<Iso, UtcOffset>`, which previously only used millisecond accuracy, whereas `Time` has nanosecond.

Also simplifies `ZoneNameTimestamp` construction, which only needs second-level accuracy.

## Changelog

icu_time: Reduce unnecessary checks during `ZonedDateTime` parsing.

### DIFF
--- a/components/time/src/types.rs
+++ b/components/time/src/types.rs
@@ -179,6 +179,10 @@ impl Time {
             subsecond: nanosecond.try_into()?,
         })
     }
+
+    pub(crate) const fn seconds_since_midnight(self) -> u32 {
+        (self.hour.0 as u32 * 60 + self.minute.0 as u32) * 60 + self.second.0 as u32
+    }
 }
 
 /// A date and time for a given calendar.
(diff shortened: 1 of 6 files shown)


components/time/src/types.rs
```diff
@@ -310,8 +314,47 @@ const UNIX_EPOCH: RataDie = calendrical_calculations::gregorian::fixed_from_greg
...
+            return core::cmp::Ordering::Less;
+        }
+        if srd - 3 >= ord {
+            return core::cmp::Ordering::Greater;
+        }
+
+        let mut ss = self.time.seconds_since_midnight() as i32 - self.zone.to_seconds();
+        let mut os = other.time.seconds_since_midnight() as i32 - other.zone.to_seconds();
+
+        // the seconds can wrap into the day
```
sffc, line 333: Suggestion (optional): This might be more readable in a small helper type,

```rust
pub(crate) struct RataDieWithTime /* or Moment? */ {
    pub rata_die: RataDie,
    pub nanoseconds: u64
}

impl RataDieWithTime {
    pub(crate) fn from_rata_die_time_and_offset(rata_die: RataDie, time: Time, offset: UtcOffset) -> Self {
        // merge the three fields into positive nanoseconds less than 86400e9
    }
}
```

sffc (approved):

components/time/src/types.rs
robertbastian: > merge the three fields into positive nanoseconds less than 86400e9

that'll overflow which is why I'm not doing that

sffc: I mean, merge the 3 fields into the 2 fields of the struct as defined in my suggestion, such that the nanoseconds field is positive and less than 86400e9

PR #7634, opened 2026-02-11
Update date range comments

robertbastian:
#7076


components/calendar/src/calendar_arithmetic.rs
```diff
@@ -417,6 +420,7 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
         Ok(Self::new_unchecked(year, constrained_month, day))
     }
 
+    // Used by calendar-specific constructors (non-lunisolar), checks `VALID_YEAR_RANGE`
```
Manishearth, line 423: ty for documenting the internal stuff too, this will be helpful

PR #7635, opened 2026-02-11
fix: windows CI by adding cross-platform Makefile support (#3109)

### LINKED ISSUE #3109: Make test-cargo CI run on Windows
When run on Windows, I get errors such as

```
cp -p "D:\a\icu4x\icu4x\docs\tutorials\cargo_tests/target"/release/demo_buffer bin/demo_buffer
./bin/icu4x-datagen \
	--format blob \
	--keys-for-bin bin/demo_buffer \
	--locales my en-ZA \
	--cldr-tag latest \
	--overwrite \
	--out buffer_data.postcard
Error: bin/demo_buffer

Caused by:
    The system cannot find the file specified. (os error 2)
```

Need to debug. Probably easy to do locally; just get `cargo make test-cargo-tutorial-local` to succeed.

swar09:
Fix Windows CI by adding cross-platform Makefile support

- Add .exe extension detection for Windows binaries in Makefiles
- Update CI workflow to run on  Windows
- Install harfbuzz dependencies for Windows using vcpkg
- Fix file path separators and binary extensions in `tools/make/tests.toml`
- Quote file paths properly to handle spaces in Windows paths

Fixes #3109


swar09:
Screenshot 2026-02-11 191834

swar09:
image

sffc:
Good work!

I noticed that "Install harfbuzz (Windows)" takes 7m 10s. Can we use a faster installation process for harfbuzz and/or cache the installation so that we don't need to wait for it on every PR moving forward?

.github/workflows/build-test.yml
```diff
@@ -158,7 +158,10 @@ jobs:
 
   # ci-job-test-cargo
   test-cargo:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, windows-latest ]
```
robertbastian, line 163: I don't think the cost of an additional runner is worth it. the linux job is barely worth it, because we never touch the code that this tests

robertbastian (commented):
it's nice to get this working on windows, but I don't think we should spend CI resources on it

.github/workflows/build-test.yml
sffc: Yeah I'm on the fence. wdyt about running it only on Windows? Then at least the job is covering something that most of the other jobs don't cover.

Manishearth: I'm also fine with having Windows makefile support for harfbuzz being best-effort: we don't CI it but we try to maintain it when people ask us to fix things.

swar09:
Thanks for the feedback! I see the concern about CI . 

Would you prefer if I  
1. Keep only the Makefile changes for cross-platform compatibility
2. Revert the CI workflow changes

.github/workflows/build-test.yml
sffc: Having it in .github/workflows means that it remains reproducible, which is nice.

Wdyt about making it a trigger-only job (requires pressing a button in the Actions tab) that we run periodically such as when preparing a release (except we don't want to make releases harder)

sffc:
> Thanks for the feedback! I see the concern about CI .
> 
> Would you prefer if I
> 
> 1. Keep only the Makefile changes for cross-platform compatibility
> 2. Revert the CI workflow changes

I don't think the maintainers have decided yet how we want to handle the CI, but if you want to split the Makefile changes into their own PR, I think that would be landable.

.github/workflows/build-test.yml
robertbastian: > Wdyt about making it a trigger-only job (requires pressing a button in the Actions tab) that we run periodically such as when preparing a release (except we don't want to make releases harder)

then you might as well run it locally. all that the CI would test is that harfbuzz is installed correctly – this is not something we need to test because it's outside of ICU4X.

sffc:
@swar09 Please resolve conflicts with the main branch. Thanks.

examples/cargo/buffer/Makefile
```diff
@@ -31,7 +31,7 @@ bin/tutorial_buffer.wasm:
 	rustup target add wasm32-wasip1 --toolchain ${PINNED_CI_NIGHTLY}
 	cargo +${PINNED_CI_NIGHTLY} build --target wasm32-wasip1 --release
 	mkdir -p bin
-	cp -p "$(CARGO_TARGET_DIR)"/wasm32-wasip1/release/tutorial_buffer.wasm bin/
+	cp -p "$(CARGO_TARGET_DIR)/wasm32-wasip1/release/tutorial_buffer.wasm" bin/
```
robertbastian, line 34: this is actually the only change that's required for the `Makefile`s to work on windows?

sffc: The other changes were split into https://github.com/unicode-org/icu4x/pull/7650

.github/workflows/build-test.yml
sffc: I think I still have a slight preference to enforce Windows client builds in CI (not just tests).

I don't care much about harfbuzz in particular as a CI thing; I would be happy if we built the cargo example crates without the harfbuzz one.

robertbastian: then lets run `test-c` on windows (we already run `test-dart`). keep in mind that windows CI takes 2x longer than ubuntu

robertbastian:
WG discussion:

- it'd be nice to have the examples work on all platforms
- it's probably overkill to run this CI on all PRs, because we barely touch the code
- this could also apply to other CI, like FFI tests, but that's out of scope of this dicussion

Conclusion:
- Change the test-cargo job to only run on changes in examples/cargo
- Run the test-cargo job unconditionally on cronjob CI 
- Also add macOS to the matrix

LGTM: @robertbastian @sffc @manishearth

PR #7637, opened 2026-02-11
added location field to DataError

### LINKED ISSUE #4048: Tracking caller on `DataError` construction

psg-19:
Fixes #4048 

Summary :- 
- Added a location field to `DataError` that stores the caller file, line and column
- Added `#[track_caller]` to all error constructors and `from impls` so `Location::caller()` captures the real call site
- Updated `display` to include the location in the `error` message

### DIFF
--- a/provider/core/src/buf/serde.rs
+++ b/provider/core/src/buf/serde.rs
@@ -232,6 +232,7 @@ where
 
 #[cfg(feature = "deserialize_json")]
 impl From<serde_json::error::Error> for DataError {
+    #[track_caller]
     fn from(e: serde_json::error::Error) -> Self {
         DataErrorKind::Deserialize
             .with_str_context("serde_json")
@@ -241,6 +242,7 @@ impl From<serde_json::error::Error> for DataError {
 
 #[cfg(feature = "deserialize_bincode_1")]
 impl From<bincode::Error> for DataError {
+    #[track_caller]
     fn from(e: bincode::Error) -> Self {
         DataErrorKind::Deserialize
             .with_str_context("bincode")
@@ -250,6 +252,7 @@ impl From<bincode::Error> for DataError {
 
 #[cfg(feature = "deserialize_postcard_1")]
 impl From<postcard::Error> for DataError {
+    #[track_caller]
     fn from(e: postcard::Error) -> Self {
         DataErrorKind::Deserialize
             .with_str_context("postcard")
--- a/provider/core/src/error.rs
+++ b/provider/core/src/error.rs
@@ -87,11 +87,14 @@ pub struct DataError {
 
     /// Whether this error was created in silent mode to not log.
     pub silent: bool,
+
+    /// The source location where the error was created.
+    pub location: core::panic::Location<'static>,
 }
 
 impl fmt::Display for DataError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "ICU4X data error")?;
+        write!(f, "ICU4X data error at {}", self.location)?;
         if self.kind != DataErrorKind::Custom {
             write!(f, ": {}", self.kind)?;
         }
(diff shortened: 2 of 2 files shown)


sffc:
hmm

```
error: the `Err`-variant returned from this function is very large
   --> components/datetime/src/neo.rs:501:10
    |
501 |     ) -> Result<Self, DateTimeFormatterLoadError>
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
   ::: components/datetime/src/error.rs:16:5
    |
 16 |     Names(PatternLoadError),
    |     ----------------------- the largest variant contains at least 128 bytes
    |
```

core::panic::Location is 24 bytes, and it's all Copy. Seems _fine_? But DataError is already 80 bytes, so this seems to push things that contain a DataError over the edge.

DataError doesn't really contain anything super big, just a lot of pointers, and enum discriminants that probably use 8 bytes due to alignment.

Manishearth:
It is a public field so adding it is a commitment, and we may not be able to feature-gate it.

sffc:
I do think the lint is telling us something useful (we don't want error types to be so big that they start causing stack frames to not fit in registers or l1 caches, etc), but the problem is not Location, it is the inefficient layout of DataError, which is not something we can easily fix since the type is public and used widely.

So I think my preferred short-term fix is to suppress this clippy lint on the errors it is impacting, and open an issue to fix in 3.0 to reduce the stack size of DataError.

Manishearth:
Given that the type is `non_exhaustive`, can we add this as a private field that gets CFG'd into `()`? Then only the people with the logging feature enabled pay the cost.

sffc:
> Given that the type is `non_exhaustive`, can we add this as a private field that gets CFG'd into `()`? Then only the people with the logging feature enabled pay the cost.

I don't know what problem that solves?

We could cfg the whole field, but we will still get the clippy warnings for the error being too big since we run clippy with `--all-features`

Manishearth:
> I don't know what problem that solves?

Users who have not enabled logging 

I don't consider the fact that clippy is annoyed to be a "problem", I consider the underlying size bloat caused adding 24 more bytes to an error type to be a bit of a problem, and I would prefer to avoid those costs where possible.

We can add an `attach_location(&mut self)` method to DataError that would allow people to attach the current `Location`, and encapsulate the Location in the Debug impl.

sffc:
I don't really consider increasing DataError from 80 to 104 bytes to be a major issue, since the Location field is probably _more_ important than all the other context information fields it already has. But, if we really want to avoid adding the field, we can just print out the location as we do in `.with_debug_context` and not persist it in the DataError.

Manishearth:
Yeah, though one benefit of persisting a location is that it works well with things like forking providers, where errors are returned through the normal course of business.

robertbastian:
> We can add an `attach_location(&mut self)` method to DataError that would allow people to attach the current `Location`, and encapsulate the Location in the Debug impl.

the whole point here is this being automatic. people can already attach arbitrarily detailed messages to the errors, but they don't.

> I don't really consider increasing DataError from 80 to 104 bytes to be a major issue

On my machine it's currently 96 bytes. `PatternLoadError`, the other variant in `DateTimeFormatterLoadError` is 104 bytes, so it's not like we're currently well clear of the (arbitrary) 128 bytes boundary anyway.

Manishearth:
> the whole point here is this being automatic. people can already attach arbitrarily detailed messages to the errors, but they don't.



Yes, we can continue to be automatic, but if people wish to reset the location we have a path for that. I was mostly saying that we can make this private without losing any of the benefits of a public field.

> On my machine it's currently 96 bytes. PatternLoadError, the other variant in DateTimeFormatterLoadError is 104 bytes, so it's not like we're currently well clear of the (arbitrary) 128 bytes boundary anyway.

Sure, it's arbitrary, and we should improve this in 3.0, but I want to try and not make the problem _worse_.

robertbastian:
WG discussion:

- current size is 80 bytes without the export feature, 96 with the export feature (extra debug path)
- this adds 24 bytes, so 104/120 bytes
- 120 gets close to two 64-bit registers, especially with the Result discriminant, but we don't care _that_ much if the export feature is enabled
- could log the location with the logging feature without persisting it
- could incubate storing this under the export feature

Conclusion:
- log the location during DataError construction
- file an issue for 3.0 to decrease the size of DataError, at which point we'll have more space available to store it

LGTM: @robertbastian @sffc @manishearth

@psg please remove the field then, and log the location if the `cfg(feature = "log")` is enabled.

PR #7638, opened 2026-02-11
Move `host_info` back to the workspace

robertbastian:
The crate doesn't even currently compile because there's no CI for it. I'd also like to use it in examples within the workspace, so we need to make it work on CI.

I think this fixes the problem that made us remove it from the workspace, by adding a build script that detects whether `gio` is available.


utils/host_info/build.rs
```diff
@@ -0,0 +1,10 @@
+// This file is part of ICU4X. For terms of use, please see the file
+// called LICENSE at the top level of the ICU4X source tree
+// (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
+fn main() {
+    println!("cargo:rustc-check-cfg=cfg(gio)");
+    if std::fs::exists("/usr/include/gio/gio.h").ok() == Some(true) {
+        println!("cargo:rustc-cfg=gio");
```
sffc: Nit: these are global so please put them in the icu4x namespace as we do elsewhere

```suggestion
        println!("cargo:rustc-cfg=icu4x_has_gio");
```

utils/host_info/src/backends/linux.rs
```diff
@@ -88,7 +87,7 @@ impl RawHostInfoBackend for LinuxHostInfoBackend {
     }
 }
 
-#[cfg(feature = "gnome")]
+#[cfg(gio)]
```
sffc: It's possible to have the gio header file but not have linked the gio dependency which is gated on `target_os = "linux"`

```suggestion
#[cfg(all(icu4x_has_gio, feature = "gio"))]
```

sffc: Or `#[cfg(all(icu4x_has_gio, target_os = "linux""))]`

robertbastian: the `linux.rs` module is, unsurprisingly, `cfg(target_os = "linux")`

robertbastian:
Ok so this *might* break the `gio` call, I have no way of validating this on my machine.

However CI passes, the crate is fixed for non-Gnome Linux users, and it lets us use the crate in the workspace.

PR #7640, opened 2026-02-11
Implement new specced reference year behavior

Manishearth:
Temporal specced reference years in more detail, using our code and feedback. Temporal now rejects MonthDay combinations that produce reference years in the far past, though as the "specced valid" range expands Temporal will likely

Constraining these dates turns leap months into regular months.

https://tc39.es/proposal-intl-era-monthcode/#chinese-dangi-iso-reference-years

I'm returning MonthNotInYear for now since the constrain behavior is identical to the month not being in the year. We could do RangeError, I think that's a valid thing to explore in a followup.

## Changelog: N/A

Covered by graduation changelog

### DIFF
--- a/components/calendar/src/cal/east_asian_traditional.rs
+++ b/components/calendar/src/cal/east_asian_traditional.rs
@@ -32,9 +32,12 @@ enum EastAsianCalendarKind {
     Korean,
 }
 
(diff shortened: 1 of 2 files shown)


Manishearth:
The test needs to be updated to handle this

Manishearth:
Okay, the reference_year test is basically incorrect in the face of constraining and needs a lot more work, I think.

Manishearth:
I think I've fixed up the test

PR #7641, opened 2026-02-11
Correct Hijri reference years

Manishearth:
The existing code incorrectly produced past years for nonexistant Hijri MonthDays, which ends up in stuff constraining to the wrong month.

The reference_year function must be able to handle any number thrown at it: if you throw `day = 35` , it should resolve to the reference year for `day = 30`, or if that's not a thing, `day = 29`, so that the outer constraining logic can constrain and produce the spec-correct year.

### DIFF
--- a/components/calendar/tests/reference_year.rs
+++ b/components/calendar/tests/reference_year.rs
@@ -45,6 +45,8 @@ where
         for is_leap in [false, true] {
             let mut valid_day_number = 1;
             let month_validity = valid_md_condition(month_number, is_leap, valid_day_number);
+            // The last valid monthday produced with Overflow::Reject in this month
+            let mut last_valid_date = None;
             for day_number in 1..=32 {
                 let md_validity = valid_md_condition(month_number, is_leap, day_number);
                 if md_validity == ValidityState::Valid {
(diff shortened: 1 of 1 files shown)


components/calendar/src/cal/hijri.rs
```diff
@@ -321,22 +321,17 @@ impl Rules for TabularAlgorithm {
...
             (8, _) => 1392,
             (9, _) => 1392,
-            (10, 30..) => 1389,
             (10, _) => 1392,
-            (11, ..=26) if self.epoch == TabularAlgorithmEpoch::Thursday => 1392,
-            (11, ..=25) if self.epoch == TabularAlgorithmEpoch::Friday => 1392,
+            (11, ..26) if self.epoch == TabularAlgorithmEpoch::Friday => 1392,
+            (11, ..27) if self.epoch == TabularAlgorithmEpoch::Thursday => 1392,
             (11, _) => 1391,
             (12, 30..) => 1390,
```
sffc, line 336: Question: Year 1392 contained all of these 30-day months? Was this just completely wrong before? Did the reference year correctness test not catch this?!

Manishearth: No, the removed entries correspond to cases where that 30 day month is impossible. 1392 will exhibit constrain/reject behavior for d=30.

This was wrong before, it was returning a different year for no reason other than the generation code assuming that all Islamic calendars have all monthday combinations in the tested period of 5 years.

This is incorrect, since M02-29 and M02-30 should resolve to the same referenceYear if a calendar constrains M02-30 to M02-29.

robertbastian (commented):
this PR is making behavioural changes that it claims are fixes, but it doesn't add any tests that would let me review that the changes are sensible

Manishearth:
Sure, I can tweak the reference year test to ensure that MDs that get their days constrained end up with the same reference year.

Manishearth:
FWIW my process for test262 failures is to get a PR out quickly since backporting is arduous (more so as icu_calendar has drifted significantly from 2.1), and then continue to improve it while getting a confirmation about the broad approach. Perhaps I should open as a draft then, but in most cases it's typically landable IMO even if not perfect.

Anyway, added a test.

Manishearth:
Merging for spec correctness, or at least better spec correctness. Happy to do followups if requested.

Issue #7603, opened 2026-02-06
Expose casemap-relevant static locales over FFI


Manishearth:
Previously: https://github.com/unicode-org/icu4x/issues/6388

Currently, our casemapping APIs take  language identifiers, but casemapping only cares about a small handful of locales (root, Turkish, Lithuanian, Greek, Dutch, Armenian). We internally match on a `LanguageIdentifier` and produce this enum.

This is low cost in Rust: it's a quick `match` statement. If you wish to avoid the codesize of locale parsing, you can use the `locale!` macro and hardcode these locales.

This is not low cost in C++, since you _must_ parse a locale. I'd previously filed https://github.com/unicode-org/icu4x/issues/6388 as a way for us to make it easier for people trying to cache locales, but also as a way to reduce this cost. People were opposed to that.

I think a solution is to have `fn tr() -> &'static Locale {...}` over FFI, for each casemap relevant locale.

Open questions:

 - Should we only do casemap-relevant locales, or all "basic" locales?
 - Should this live on `Locale`, `CaseMapper`, or some third type? If we're doing all "basic" locales `LocaleConsts` might be a good type to add.
 - Currently `Locale::unknown()` returns a `Box<Locale>`, which is wasteful. Should we also add an API that returns a static?

Also note that JS and Dart [currently don't support](https://github.com/rust-diplomat/diplomat/issues/1044) returning `&'static Opaque`. This support can be added.

This is important for a client but not a client blocker, since I am able to patch in the necessary APIs for now.

sffc:
`LocaleConsts` seems fine. I wouldn't mind `CaseMapper::create_tr()` for the 5 locales you need, but you're not constructing these; you are using `lowercase_with_compiled_data` which takes a `&Locale` argument. So I think it makes the most sense to add the things you need on `LocaleConsts`.

I don't have a strong opinion between returning `&'static Locale` or `Box<Locale>`. If you return a box, then it acts more like a normal FFI object. Clients can save the boxes into local statics, which I assume is easy. If that is not easy, then `&'static Locale` seems more motivated.

I think it's fine if it is C++ only for now. I want all features to be in all languages, but I'm fine if micro optimizations don't land in all the platforms right away, as long as there is a path for them.

Manishearth:
Would it make sense to start with a small LocaleConsts with ~5 locales, and then expand it as requested? Perhaps with an eye towards eventually picking

[...]

PR #7642, opened 2026-02-12
Add FFI type CaseMapLocaleConsts

Manishearth:
Fixes https://github.com/unicode-org/icu4x/issues/7603

## Changelog

icu_capi: Add FFI type CaseMapLocaleConsts for cheap casemapping from known locale tailorings.


ffi/capi/src/locale_core.rs
```diff
@@ -192,6 +193,55 @@ pub mod ffi {
...
+    #[diplomat::attr(not(cpp), disable)]
+    #[diplomat::opaque]
+    pub struct LocaleStatics;
+
+    impl LocaleStatics {
+        /// Construct a [`Locale`] "und".
+        pub fn und() -> &'static Locale {
+            Box::new(Locale(icu_locale_core::Locale::UNKNOWN))
+        }
+        /// Construct a [`Locale`] "tr".
```
sffc: Nit: Please put these in alphabetical order please.

ffi/capi/src/locale_core.rs
```diff
@@ -192,6 +194,61 @@ pub mod ffi {
...
+        /// Construct a [`Locale`] "und".
+        pub fn und() -> &'static Locale {
+            static UND: icu_locale_core::Locale = icu_locale_core::Locale::UNKNOWN;
+            Locale::transparent_convert(&UND)
+        }
+
+        /// Construct a [`Locale`] "az".
+        pub fn az() -> &'static Locale {
+            static AZ: icu_locale_core::Locale = locale!("az");
+            Locale::transparent_convert(&AZ)
```
robertbastian: you shouldn't need to use `transparent_convert`:

```rust
static AZ: Locale = Locale(locale!("az"));
&AZ
```

or, more concise

```suggestion
const { &Locale(locale!("az")) }
```

Manishearth:
> These &'statics solve a problem for casemap, they should live on the casemapper. If another type also needs locale_nl() in the future, we can also add it on that type – a method that returns a pointer is very cheap.

This is a design I am open to, @sffc dispreferred it. Please see the discussion on https://github.com/unicode-org/icu4x/pull/7642 (though that specific design may have been dispreferred when we discussed this in person)

The idea is that this type can add more locales over time if needed.

robertbastian:
This is me joining the discussion

Manishearth:
Yes, please continue the discussion there, because some of this has already been discussed a bit and I would not like to have to rehash things.

Manishearth:
Repurposed this type for the agreement on https://github.com/unicode-org/icu4x/issues/7603

Manishearth:
One thing to note: I moved `und` over to `Locale` because _that_ one makes sense and we should at least have a `static` version for C++. I called it `unknown_static()`.

ffi/capi/src/casemap.rs
```diff
@@ -494,6 +495,55 @@ pub mod ffi {
...
+    /// In most cases, you should be taking a locale from the user, but in some
+    /// limited cases you may be parsing from a limited set of locales and not
+    /// wish to pull in full-fledged parsing code.
+    ///
+    /// For now, this only supports locales that have special meaning to `icu_casemap`,
+    /// since it only cares about a small set of locales and locale parsing takes
+    /// up a relatively large amount of binary size in the context of casemapping.
+    ///
+    /// Please file an issue on ICU4X if you have a use case for more locales here.
+    #[diplomat::attr(not(cpp), disable)]
```
sffc, line 509: Is there a diplomat feature flag for "supports statics"?

ffi/capi/src/locale_core.rs
```diff
@@ -47,6 +47,16 @@ pub mod ffi {
             Box::new(Locale(icu_locale_core::Locale::UNKNOWN))
         }
 
+        /// Construct an unknown [`Locale`] "und" as a C++ const/static,
+        /// without allocating.
+        ///
+        /// (Can potentially be exposed to other backends if they gain static support)
+        #[diplomat::attr(not(cpp), disable)]
+        pub fn unknown_static() -> &'static Locale {
```
sffc: Nit: rustlink to https://unicode-org.github.io/icu4x/rustdoc/icu/locale/struct.Locale.html#associatedconstant.UNKNOWN

ffi/capi/src/casemap.rs
Manishearth: No because everyone should just support it. I usually add feature flags for things where it actually makes sense for a backend to have opinions on, maybe because the language doesn't support the needed feature (like named ctors) or because the backend has made a choice not to.
Otherwise per backend toggles are better. Then when the backend adds support you make a call on whether you want those things enabled.

It's a backlog item for me to add this support.

ffi/capi/src/casemap.rs
```diff
@@ -494,6 +495,55 @@ pub mod ffi {
...
     }
+
+    /// A type that provides access to `'static` Locales, to avoid
+    /// needing to parse user locales when the code knows what locale it wants.
+    ///
+    /// In most cases, you should be taking a locale from the user, but in some
+    /// limited cases you may be parsing from a limited set of locales and not
+    /// wish to pull in full-fledged parsing code.
+    ///
+    /// For now, this only supports locales that have special meaning to `icu_casemap`,
```
robertbastian:
```suggestion
    /// For now, this only supports locales that have special meaning to `CaseMapper`,
```

ffi/capi/src/casemap.rs
```diff
@@ -494,6 +495,55 @@ pub mod ffi {
...
+    /// since it only cares about a small set of locales and locale parsing takes
+    /// up a relatively large amount of binary size in the context of casemapping.
+    ///
+    /// Please file an issue on ICU4X if you have a use case for more locales here.
```
robertbastian: link to Github

ffi/capi/src/casemap.rs
```diff
@@ -494,6 +495,55 @@ pub mod ffi {
...
+    #[diplomat::attr(not(cpp), disable)]
+    #[diplomat::opaque]
+    pub struct CaseMapLocaleConsts;
+
+    impl CaseMapLocaleConsts {
```
robertbastian: "Const" is not really a term that scales across all languages, which is something we need here. both C++ and Dart (and I think JS) have their own notions of "const".

also the type is called "CaseMapper", not "CaseMap"

```suggestion
    impl CaseMapperLocales {
```

ffi/capi/src/locale_core.rs
robertbastian: this is first and foremost a reference. it's a reference with a static *lifetime*, but those don't exist in other languages. it's not a static as found in most languages.

```suggestion
        pub fn unknown_ref() -> &'static Locale {
```

ffi/capi/src/casemap.rs
```diff
@@ -494,6 +495,55 @@ pub mod ffi {
...
+        /// Construct a [`Locale`] "az".
```
robertbastian: the point here is that no "construction" takes places


```suggestion
        /// Returns a borrowed "az" [`Locale`].
```

ffi/capi/src/locale_core.rs
```diff
@@ -47,6 +47,16 @@ pub mod ffi {
             Box::new(Locale(icu_locale_core::Locale::UNKNOWN))
         }
 
+        /// Construct an unknown [`Locale`] "und" as a C++ const/static,
+        /// without allocating.
```
robertbastian:
```suggestion
        /// Returns a borrowed "und" (unknown) [`Locale`].
```

ffi/capi/src/casemap.rs
Manishearth: (a) this was what we decided on in the meeting

(b) this was intended to be a C++ only API. But I can call it CaseMapperRefs.

ffi/capi/src/casemap.rs
Manishearth: This isn't needed anymore since it's casemap specific

ffi/capi/src/casemap.rs
robertbastian: yes and I only realised this now, we can reopen the whole discussion if you want

ffi/capi/src/locale_core.rs
```diff
@@ -47,6 +47,16 @@ pub mod ffi {
             Box::new(Locale(icu_locale_core::Locale::UNKNOWN))
         }
 
+        /// Returns a borrowed unknown ("und") [`Locale`], without allocating
+        ///
+        /// (Can potentially be exposed to other backends if they gain static support)
```
robertbastian:
```suggestion
        //
        // (Can potentially be exposed to other backends if they gain static support)
```

ffi/capi/src/casemap.rs
```diff
@@ -494,6 +495,53 @@ pub mod ffi {
                 .write_to(write);
         }
     }
+
+    /// A type that provides access to `'static` Locales, to avoid
```
robertbastian: 'static doesn't mean anything to a C++ user
```suggestion
    /// A type that provides access to preconstructed Locales, to avoid
```

ffi/capi/src/casemap.rs
```diff
@@ -494,6 +495,53 @@ pub mod ffi {
...
+    /// needing to parse user locales when the code knows what locale it wants.
+    ///
+    /// In most cases, you should be taking a locale from the user, but in some
+    /// limited cases you may be parsing from a limited set of locales and not
+    /// wish to pull in full-fledged parsing code.
+    ///
+    /// This type is for locales that have special meaning to `CaseMap`,
```
robertbastian: the type is called `CaseMapper`

ffi/capi/src/locale_core.rs
```diff
@@ -47,6 +47,16 @@ pub mod ffi {
             Box::new(Locale(icu_locale_core::Locale::UNKNOWN))
         }
 
+        /// Returns a borrowed unknown ("und") [`Locale`], without allocating
```
robertbastian:
```suggestion
        /// Returns a borrowed unknown ("und") [`Locale`], without allocating.
```

ffi/capi/src/casemap.rs
```diff
@@ -494,6 +495,53 @@ pub mod ffi {
...
+    /// since it only cares about a small set of locales and locale parsing takes
+    /// up a relatively large amount of binary size in the context of casemapping.
+    #[diplomat::attr(not(cpp), disable)]
+    #[diplomat::opaque]
+    pub struct CaseMapLocaleRefs;
+
+    impl CaseMapLocaleRefs {
+        /// Returns a borrowed "az" [`Locale`].
```
robertbastian:
```suggestion
        /// Returns a borrowed "az" [`Locale`], without allocating.
```

ffi/capi/src/casemap.rs
```diff
@@ -494,6 +495,53 @@ pub mod ffi {
...
+    /// wish to pull in full-fledged parsing code.
+    ///
+    /// This type is for locales that have special meaning to `CaseMap`,
+    /// since it only cares about a small set of locales and locale parsing takes
+    /// up a relatively large amount of binary size in the context of casemapping.
+    #[diplomat::attr(not(cpp), disable)]
+    #[diplomat::opaque]
+    pub struct CaseMapLocaleRefs;
+
+    impl CaseMapLocaleRefs {
```
robertbastian:
```suggestion
    impl CaseMapperLocaleRefs {
```

ffi/capi/src/casemap.rs
sffc: `CaseMap` because the component is called `casemap`. These are the locales for the casemap component, which happens to export a type called `CaseMapper`.

As an analogy, these would be called `DateTimeLocaleConsts` rather than `DateTimeFormatterLocaleConsts`.

`CaseMapLocales` is fine with me as I previously stated in the meeting. If @robertbastian is onboard with that name then that seems like a better name.

Manishearth: Right, in fact some of these locales are only relevant to TitlecaseMapper.

PR #7646, opened 2026-02-12
data struct with codepoint builder for radicals

opnuub:
Issue: #6941 

Added:

components/segmenter/src/provider/radicals.rs | data struct
provider/registry/src/lib.rs to add marker
Modified:

components/segmenter/src/provider/mod.rs | added two lines to use radicals.rs also in provider
provider/source/src/segmenter/unihan.rs | added data provider and test for new codepoint trie
experimental_segmenter, test/adaboost, python_test_output to use the trie

rest are auto-generated by cargo make testdata and bakeddata

### DIFF
--- a/components/segmenter/tests/adaboost/main.rs
+++ b/components/segmenter/tests/adaboost/main.rs
@@ -4,59 +4,39 @@
 
 #![allow(dead_code)]
 
+use icu::segmenter::provider::UnihanIrgData;
+use icu_segmenter::provider::Baked;
 use std::collections::HashMap;
+
+fn load_irg_from_baked() -> &'static UnihanIrgData<'static> {
+    Baked::SINGLETON_SEGMENTER_UNIHAN_RADICAL_V1
+}
 
 static MODEL_FOR_TEST: &str = include_str!("model.json");
 static MODEL_FOR_TEST_THAI: &str = include_str!("model_thai.json");
 
-static CODEPOINTS: &[u16] = &[
-    20008, 20022, 20031, 20057, 20101, 20108, 20128, 20154, 20799, 20837, 20843, 20866, 20886,
-    20907, 20960, 20981, 20992, 21147, 21241, 21269, 21274, 21304, 21313, 21340, 21353, 21378,
-    21430, 21448, 21475, 22231, 22303, 22763, 22786, 22794, 22805, 22823, 22899, 23376, 23424,
-    23544, 23567, 23586, 23608, 23662, 23665, 24027, 24037, 24049, 24062, 24178, 24186, 24191,
-    24308, 24318, 24331, 24339, 24400, 24417, 24435, 24515, 25096, 25142, 25163, 25903, 25908,
-    25991, 26007, 26020, 26041, 26080, 26085, 26352, 26376, 26408, 27424, 27490, 27513, 27571,
-    27595, 27604, 27611, 27663, 27668, 27700, 28779, 29226, 29238, 29243, 29247, 29255, 29273,
-    29275, 29356, 29572, 29577, 29916, 29926, 29976, 29983, 29992, 30000, 30091, 30098, 30326,
-    30333, 30382, 30399, 30446, 30683, 30690, 30707, 31034, 31160, 31166, 31348, 31435, 31481,
-    31859, 31992, 32566, 32593, 32650, 32701, 32769, 32780, 32786, 32819, 32895, 32905, 33251,
-    33258, 33267, 33276, 33292, 33307, 33311, 33390, 33394, 33400, 34381, 34411, 34880, 34892,
-    34915, 35198, 35211, 35282, 35328, 35895, 35910, 35925, 35960, 35997, 36196, 36208, 36275,
-    36523, 36554, 36763, 36784, 36789, 37009, 37193, 37318, 37324, 37329, 38263, 38272, 38428,
-    38582, 38585, 38632, 38737, 38750, 38754, 38761, 38859, 38893, 38899, 38913, 39080, 39131,
-    39135, 39318, 39321, 39340, 39592, 39640, 39647, 39717, 39727, 39730, 39740, 39770, 40165,
-    40565, 40575, 40613, 40635, 40643, 40653, 40657, 40697, 40701, 40718, 40723, 40736, 40763,
-    40778, 40786, 40845, 40860, 40864,
-];
-
-pub(crate) fn get_radical(ch: char) -> u8 {
-    let id = ch as u32;
-
-    if !(19968..=40869).contains(&id) {
-        return 0;
-    }
-
-    let idx = CODEPOINTS.partition_point(|&b| (b as u32) <= id);
-    (idx as u8) + 1
+pub(crate) fn get_radical(irg: &UnihanIrgData<'_>, ch: char) -> u8 {
+    irg.trie.get(ch)
 }
 
-pub(crate) struct Predictor {
+pub(crate) struct Predictor<'a> {
     pub(crate) model: HashMap<String, HashMap<String, i16>>,
+    irg: &'a UnihanIrgData<'a>,
 }
 
-impl Predictor {
-    pub(crate) fn from_json(json: &str) -> Self {
+impl<'a> Predictor<'a> {
+    pub(crate) fn from_json(json: &str, irg: &'a UnihanIrgData<'a>) -> Self {
         let model: HashMap<String, HashMap<String, i16>> =
             serde_json::from_str(json).unwrap_or_default();
-        Self { model }
+        Self { model, irg }
     }
 
-    pub(crate) fn for_test() -> Self {
-        Self::from_json(MODEL_FOR_TEST)
+    pub(crate) fn for_test(irg: &'a UnihanIrgData<'a>) -> Self {
+        Self::from_json(MODEL_FOR_TEST, irg)
     }
 
-    pub(crate) fn for_test_thai() -> Self {
-        Self::from_json(MODEL_FOR_TEST_THAI)
+    pub(crate) fn for_test_thai(irg: &'a UnihanIrgData<'a>) -> Self {
+        Self::from_json(MODEL_FOR_TEST_THAI, irg)
     }
 
     pub(crate) fn predict(&self, sentence: &str) -> Vec<i16> {
@@ -73,7 +53,7 @@ impl Predictor {
 
             let mut score: i16 = 4;
 
-            let rad4 = get_radical(c);
+            let rad4 = get_radical(self.irg, c);
             if rad4 != 0 {
                 if let Some(map) = self.model.get("RSRID") {
                     let key = format!("{}:{}", c_prev, rad4);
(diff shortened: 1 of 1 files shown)


components/segmenter/tests/adaboost/main.rs
```diff
@@ -4,8 +4,15 @@
...
+use icu_segmenter::provider::Baked;
 use std::collections::HashMap;
 
+fn load_irg_from_baked() -> DataPayload<SegmenterUnihanIrgV1> {
+    Baked.load(DataRequest::default()).unwrap().payload
+}
+
 static MODEL_FOR_TEST: &str = include_str!("model.json");
 
 static CODEPOINTS: &[u16] = &[
```
sffc: Can you delete `CODEPOINTS` now?

The `dead_code` lint should detect this, but you have it turned off

components/segmenter/tests/adaboost/python_test_output.txt
```diff
@@ -24,7 +24,7 @@
 2404
 -1604
 2824
-440
+320
```
sffc, line 27: Why the diff?

provider/source/src/segmenter/unihan.rs
```diff
@@ -4,9 +4,63 @@
...
+                    crate::TrieType::Fast => codepointtrie::TrieType::Fast,
+                    crate::TrieType::Small => codepointtrie::TrieType::Small,
+                },
+            );
+
+            for (ch, irg_val) in irg_map.iter() {
+                builder.set_value(*ch as u32, irg_val.value);
+            }
+
+            let trie = builder.build();
```
sffc, line 44: FYI @robertbastian, we're adding a new `CodePointTrieBuilder` call site

components/segmenter/Cargo.toml
```diff
@@ -52,6 +52,7 @@ default = ["compiled_data", "auto"]
 serde = ["dep:serde", "potential_utf/serde", "zerovec/serde", "icu_collections/serde", "icu_provider/serde"]
 datagen = ["serde", "dep:databake", "potential_utf/databake", "zerovec/databake", "icu_collections/databake", "icu_provider/export"]
 lstm = ["dep:core_maths"]
+radicals = []
```
sffc: Please name the feature `unstable` for now.

components/segmenter/tests/adaboost/main.rs
opnuub: will we want to run a test to compare naive binary search vs codepoint builder? like on speed, memory usage etc

components/segmenter/tests/adaboost/python_test_output.txt
opnuub: naive binary search approach gives different values

components/segmenter/Cargo.toml
opnuub: will it be radicals = ["unstable"]?

provider/data/segmenter/fingerprints.csv
```diff
@@ -24,3 +24,4 @@ segmenter/lstm/auto/v1, und/Burmese_codepoints_exclusive_model4_heavy, 91365B, 9
 segmenter/lstm/auto/v1, und/Khmer_codepoints_exclusive_model4_heavy, 74665B, 74368B, cc6fe9f66fed196d
 segmenter/lstm/auto/v1, und/Lao_codepoints_exclusive_model4_heavy, 72160B, 71863B, 3f52a4025c7d618f
 segmenter/lstm/auto/v1, und/Thai_codepoints_exclusive_model4_heavy, 72327B, 72030B, 4486b38238d7c651
+segmenter/unihan/irg/v1, <singleton>, 27212B, 27152B, 26ca4955491a2686
```
sffc: Observation: this is 27152B? Please sanity check this. It seems bigger than you need.

If you can get good performance using your own smaller structure, feel free to do that.

components/segmenter/Cargo.toml
sffc: No, just a single feature named `unstable` which contains your radicals stuff. Eventually we may want to graduate the things into a `radicals` feature, but we don't need to decide that right now

components/segmenter/tests/adaboost/python_test_output.txt
sffc: This is supposed to test whether Rust aligns with Python. If you get different results, then is there now a difference in behavior? We shouldn't land things that make behavior change.

components/segmenter/tests/adaboost/main.rs
sffc: Yeah, that's a good idea. Now that you have your code implemented both ways, it should be easy to get a benchmark.

provider/data/segmenter/fingerprints.csv
opnuub: would it be a valid size since the trie stoes about 100k chars of radical ids?

components/segmenter/tests/adaboost/main.rs
opnuub: trie is 8x slower than dict, binary search is 4x slower than dict. memory usage wise binary search is much smaller at 27 bytes

provider/data/segmenter/fingerprints.csv
opnuub: after sanity checking it is indeed 27152 bytes

sffc:
As noted in chat: the 27 kB struct appears to be the correct size; there is just a whole lot more data than the experimental implementation. Some comes from adding support for historical characters, and some comes from supporting more exceptions in the base plane.

My suggestion is to settle on a specific mapping table, and use the same table identically for both training and inference.

components/segmenter/examples/experimental_segmenter.rs
```diff
@@ -13,13 +13,21 @@ mod cnn;
...
+use icu_provider::prelude::*;
+use icu_segmenter::provider::Baked;
 use icu_segmenter::{options::WordBreakOptions, WordSegmenter, WordSegmenterBorrowed};
 use std::time::SystemTime;
 
 const REPETITIONS: usize = 1000;
 
 fn main_adaboost(args: &[String]) {
-    let segmenter = Predictor::for_test();
+    let response: DataResponse<SegmenterUnihanIrgV1> = Baked.load(DataRequest::default()).unwrap();
```
robertbastian: this is a singleton, use the constant

components/segmenter/Cargo.toml
robertbastian: this is not resolved

components/segmenter/src/provider/mod.rs
```diff
@@ -120,6 +125,13 @@ icu_provider::data_marker!(
     RuleBreakData<'static>,
     is_singleton = true
 );
+icu_provider::data_marker!(
+    /// `SegmenterUnihanIrgV1`
+    SegmenterUnihanIrgV1,
```
robertbastian: issue: I had to look up what IRG means, and it's Ideographic Research Group. This name, even if it were to be spelled out, doesn't tell me anything about what this data is. please use a better name

components/segmenter/tests/adaboost/main.rs
```diff
@@ -4,8 +4,15 @@
 
 #![allow(dead_code)]
 
+use icu::segmenter::provider::{SegmenterUnihanIrgV1, UnihanIrgData};
+use icu_provider::prelude::*;
+use icu_segmenter::provider::Baked;
 use std::collections::HashMap;
 
+fn load_irg_from_baked() -> DataPayload<SegmenterUnihanIrgV1> {
+    Baked.load(DataRequest::default()).unwrap().payload
```
robertbastian: again, use the singleton

provider/source/src/segmenter/unihan.rs
```diff
@@ -4,9 +4,63 @@
...
+            DataError::custom("Unihan data generation requires use_wasm or use_icu4c")
+                .with_req(SegmenterUnihanIrgV1::INFO, req),
+        );
+
+        #[cfg(any(feature = "use_wasm", feature = "use_icu4c"))]
+        {
+            self.check_req::<SegmenterUnihanIrgV1>(req)?;
+
+            let unihan_cache = self.unihan()?;
+            let irg_map = unihan_cache.irg_sources()?;
```
robertbastian: all the juice of this implementation is in `irg_sources`, which creates an intermediate `LiteMap` instead of just adding the data to the `CodePointTrieBuilder`. there's not much point in caching this data in `unihan_cache`, because it's only used in a single `DataProvider` implementation. please move this code here and remove the caching and intermediate `LiteMap`.

components/segmenter/src/provider/mod.rs
```diff
@@ -120,6 +125,13 @@ icu_provider::data_marker!(
     RuleBreakData<'static>,
     is_singleton = true
 );
+icu_provider::data_marker!(
+    /// `SegmenterUnihanIrgV1`
```
robertbastian: proper docs please

components/segmenter/src/provider/mod.rs
```diff
@@ -120,6 +125,13 @@ icu_provider::data_marker!(
...
+    SegmenterUnihanIrgV1,
+    "segmenter/unihan/irg/v1",
+    UnihanIrgData<'static>,
```
robertbastian, line 132: you don't need to use an intermediate data struct

```suggestion
    CodePointTrie<'static, u8>,
```

components/segmenter/src/provider/radicals.rs
```diff
@@ -0,0 +1,32 @@
...
+    #[cfg_attr(feature = "serde", serde(borrow))]
+    pub trie: CodePointTrie<'data, u8>,
+}
+
+icu_provider::data_struct!(
+    UnihanIrgData<'_>,
+    #[cfg(feature = "datagen")]
+);
+
+icu_provider::data_marker!(
```
robertbastian, line 26: issue: you're defining this data marker twice

components/segmenter/src/provider/mod.rs
sffc: maybe call it `SegmenterUnihanRadicalV1`

provider/source/src/segmenter/unihan.rs
sffc: I asked @opnuub to use the caching strategy to be more consistent with how we load all our other structured data, and because we might want to use the IRG data in another component later.

components/segmenter/src/provider/mod.rs
sffc: There is more data in IRG the file, though. We are using the radicals only, but there is more data.

provider/source/src/segmenter/unihan.rs
robertbastian: if we ever use the *IRG file* in in multiple places, we can start caching it. however the current data that is being cached is not the IRG file, it's the processed radicals data (filtered and in a `LiteMap`).

components/segmenter/src/provider/mod.rs
robertbastian: and what would require all data from the IRG file to be in the same data marker?

sffc: The definition of the data marker is "the IRG data needed for radaboost segmentation"... if anything this should be merged in with the other radaboost data marker.

robertbastian: that shouldn't be the definition of a data marker. we don't have a `CldrSupplementalV1` data marker either.

PR #35, opened 2020-04-16
Initialize the minimal set of components.

zbraniecki:
Fixes #16.

This PR introduces three no-code components with crates in version `0.0.1`.

I'd like to land that, and then add a PR for #31 on top of it.

PR #7651, opened 2026-02-13
feat(decimal): implement fraction grouping support compliant with TR35

### LINKED ISSUE #7403: Support grouping fraction digits
https://unicode.org/reports/tr35/tr35-numbers.html#Number_Format_Patterns:~:text=The%20grouping%20separator%20may%20also%20occur%20in%20the%20fractional%20part%2C%20such%20as%20in%20%E2%80%9C%23%2C%23%230.%23%23%23%2C%23%E2%80%9D.%20This%20is%20most%20commonly%20done%20where%20the%20grouping%20separator%20character%20is%20a%20thin%2C%20non%2Dbreaking%20space%20(U%2B202F)%2C%20such%20as%20%E2%80%9C1.618%E2%80%AF033%E2%80%AF988%E2%80%AF75%E2%80%9D.%20See%20physics.nist.gov/cuu/Units/checklist.html.

mega123-art:
This PR implements support for grouping fractional digits in `icu_decimal`, following UTS #35 (LDML) and NIST/SI guidelines. fixes #7403 
### Changes:
- Added `fraction_grouping_size: Option<u8>` to `DecimalFormatterOptions`.
- Implemented `check_fraction` in `grouper.rs` with support for `min_grouping` (Min2 rule).
- Updated `format.rs` to insert grouping separators in the fractional part when applicable.
- Added comprehensive tests for various grouping strategies and edge cases.
- Ensured compliance with NIST checklist (e.g., suppressing grouping for 4-digit fractions when using Min2).


components/decimal/src/options.rs
```diff
@@ -13,12 +13,18 @@ pub struct DecimalFormatterOptions {
     ///
     /// Default is [`GroupingStrategy::Auto`]
     pub grouping_strategy: Option<GroupingStrategy>,
+
+    /// The size of groups in the fractional part of the number.
+    ///
+    /// If `None`, grouping separators will never be shown in the fractional part.
+    pub fraction_grouping_size: Option<u8>,
```
sffc: This should probably be `Option<GroupingStrategy>` since the grouping strategy options can equivalently be applied to the fraction part, correct?

components/decimal/src/decimal_formatter.rs
```diff
@@ -226,9 +227,21 @@ impl Writeable for FormattedUnsignedDecimal<'_> {
...
+                    if grouper::check_fraction(
+                        lower_magnitude,
+                        m,
+                        self.options.grouping_strategy.unwrap_or_default(),
+                        self.options.fraction_grouping_size,
+                        self.symbols.grouping_sizes.min_grouping,
+                    ) && range.peek().is_some()
+                    {
+                        w.with_part(parts::GROUP, |w| {
+                            w.write_str(self.symbols.grouping_separator())
```
sffc, line 241: Should fractional grouping use the same symbol as integer grouping? Can you show where exactly in TR 35 this behavior is specified?

components/decimal/src/options.rs
mega123-art: That makes sense. I've updated the field to `fraction_grouping: Option<GroupingStrategy>`. This allows the fractional part to respect the same `Min2` logic as the integer part, which is core to the NIST/SI requirement of only grouping when there are at least 5 digits.

components/decimal/src/decimal_formatter.rs
mega123-art: Yes, they use the same symbol. TR 35 Part 3 specifies this in two places:

1. In Section 3, Table **'Number Pattern Character Definitions'**, the `, `(grouping separator) is defined as: 'May occur in both the integer part and the fractional part.'
2. In Section 3.10 'Number Patterns', it notes: 'The grouping separator may also occur in the fractional part, such as in “#,##0.###,#”. This is most commonly done where the grouping separator character is a thin, non-breaking space (U+202F)...'

components/decimal/src/grouper.rs
```diff
@@ -51,6 +51,53 @@ pub fn check(
...
+    // we do not group.
+    // e.g. fraction=3, min2=2. Total digits must be 5+.
+    if -lower_magnitude < fraction + min_grouping {
+        return false;
+    }
+
+    // Magnitude is negative: -1, -2, -3...
+    // We want to group at -fraction, -2*fraction, etc.
+    // e.g. fraction=3. Group at -3, -6...
+    (-magnitude) % fraction == 0
```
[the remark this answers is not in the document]
robertbastian: not relevant in this PR

components/decimal/src/grouper.rs
```diff
@@ -60,31 +60,28 @@ pub fn check_fraction(
...
+    if grouping_size == 0 {
+        return false;
+    }
     let min_grouping = {
         use GroupingStrategy::*;
         match strategy {
-            Never => return false,
+            Never | Auto => return false,
             // Note: Auto and Always are the same for DecimalFormatter.
             // When currencies are implemented, this will change.
```
robertbastian, line 76: outdated comment

components/decimal/src/decimal_formatter.rs
```diff
@@ -226,9 +227,21 @@ impl Writeable for FormattedUnsignedDecimal<'_> {
...
+                while let Some(m) = range.next() {
                     #[expect(clippy::indexing_slicing)] // digit_at in 0..=9
                     w.write_char(self.digits[self.value.digit_at(m) as usize])?;
+                    if grouper::check_fraction(
+                        lower_magnitude,
+                        m,
+                        self.options.grouping_strategy.unwrap_or_default(),
+                        self.options.fraction_grouping_size,
+                        self.symbols.grouping_sizes.min_grouping,
+                    ) && range.peek().is_some()
```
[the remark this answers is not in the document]
robertbastian: irrelevant, `Default::default` is effectively const and will be inlined

components/decimal/src/grouper.rs
```diff
@@ -51,6 +51,50 @@ pub fn check(
...
+///
+/// `lower_magnitude` is the magnitude of the lowest-power digit, used for resolving minimum
+/// grouping digits.
+/// `magnitude` is the magnitude of the digit, which is negative for fraction digits.
+pub fn check_fraction(
+    lower_magnitude: i16,
+    magnitude: i16,
+    strategy: GroupingStrategy,
+    min_grouping: u8,
+    grouping_size: u8,
```
robertbastian: nit: `check` accepts a `GroupingSizes`, `check_fraction` individual fields. please align

components/decimal/src/options.rs
```diff
@@ -13,12 +13,18 @@ pub struct DecimalFormatterOptions {
     ///
     /// Default is [`GroupingStrategy::Auto`]
     pub grouping_strategy: Option<GroupingStrategy>,
+
+    /// The grouping strategy to use for the fractional part of the number.
+    ///
+    /// If `None`, grouping separators will never be shown in the fractional part.
+    pub fraction_grouping: Option<GroupingStrategy>,
```
robertbastian: question: do we actually want to expose this as a separate option? what's the use case for using a different grouping strategy for fractions than for integers?

components/decimal/src/options.rs
```diff
@@ -13,12 +13,18 @@ pub struct DecimalFormatterOptions {
...
 }
 
 impl From<GroupingStrategy> for DecimalFormatterOptions {
     fn from(grouping_strategy: GroupingStrategy) -> Self {
         Self {
             grouping_strategy: Some(grouping_strategy),
+            fraction_grouping: None,
```
robertbastian: issue: should this not set the `fraction_grouping`?

components/decimal/src/provider.rs
```diff
@@ -306,7 +306,7 @@ impl<'a> core::ops::Deref for DecimalSymbols<'a> {
 
 impl DecimalSymbols<'static> {
     /// Create a new en-US format for use in testing
-    #[cfg(feature = "datagen")]
+    #[cfg(any(feature = "datagen", test))]
```
robertbastian: nit: we don't generally write our code to compile with `cargo test` without `--all-features`, so we don't add these `cfg`s

components/decimal/src/decimal_formatter.rs
```diff
@@ -226,9 +227,21 @@ impl Writeable for FormattedUnsignedDecimal<'_> {
...
-                for m in range.by_ref() {
+                while let Some(m) = range.next() {
                     #[expect(clippy::indexing_slicing)] // digit_at in 0..=9
                     w.write_char(self.digits[self.value.digit_at(m) as usize])?;
+                    if grouper::check_fraction(
+                        lower_magnitude,
+                        m,
+                        self.options.fraction_grouping.unwrap_or_default(),
+                        self.symbols.grouping_sizes.min_grouping,
+                        self.symbols.grouping_sizes.primary,
```
robertbastian: question: should this actually use the integer grouping values, or do we need to parse fractional grouping values from the patterns?

components/decimal/src/grouper.rs
```diff
@@ -203,3 +247,129 @@ fn test_grouper() {
         }
     }
 }
+
+#[test]
+fn test_fraction_grouper() {
```
robertbastian, line 256: issue: this unit-tests the grouper, but I'm missing a test of `DecimalFormatter` with the new options (which I think can produce bogus values)

components/decimal/src/options.rs
mega123-art: The main use case is scientific and technical formatting (e.g., NIST SP 811), where requirements for fractional grouping are often stricter or different than integer grouping.

For example, a user might want standard locale grouping for integers (1,000,000), but need to suppress fractional grouping (0.12345) or enforce the Min2 rule strictly for fractions (0.1234 not grouped, 0.12345 grouped), while keeping the integer part standard. separating the options provides the granularity needed to cover these mixed scenarios.

components/decimal/src/decimal_formatter.rs
mega123-art: TR 35 is unfortunately a bit vague, but typically fractional grouping is implied to mirror integer grouping size (usually 3). The spec says:

"The grouping separator may also occur in the fractional part, such as in “#,##0.###,#”. This is most commonly done where the grouping separator character is a thin, non-breaking space..."

In practice (and in CLDR data), fractional grouping sizes aren't explicitly defined separately from integer primary/secondary sizes in the standard patterns. So reusing the primary integer grouping size is the standard interpretation. If we want to support explicit fractional grouping sizes (e.g., grouping by 2s in fractions but 3s in integers), we'd likely need new data model support, but using the primary size covers the vast majority of cases (like the NIST 'group by 3' rule).

components/decimal/src/decimal_formatter.rs
```diff
@@ -308,3 +320,28 @@ pub fn test_es_mx() {
...
+    let d1 = "12345.1".parse().unwrap();
+    assert_writeable_eq!(fmt.format(&d1), "12,345.1");
+
+    // 4 digits fraction: Min2 suppression applies (size 3 + min_grouping 2 = 5 needed)
+    let d4 = "12345.1234".parse().unwrap();
+    assert_writeable_eq!(fmt.format(&d4), "12,345.1234");
+
+    // 5 digits fraction: grouping appears
+    let d5 = "12345.12345".parse().unwrap();
+    assert_writeable_eq!(fmt.format(&d5), "12,345.123,45");
```
robertbastian: this is not a desired output

components/decimal/src/options.rs
robertbastian: but fraction grouping should only be used when the CLDR pattern explicitly opts in to it. it only makes sense when the grouping separator is a space, we don't want commas or periods in the fractional part

robertbastian: for the NIST use case we would want a special constructor, it should not be achieved through options on localised patterns data

components/decimal/src/decimal_formatter.rs
mega123-art: Agreed, grouping with commas in the fraction part is non-standard for general usage. This test case was intended to verify the programmatic control over the formatter (i.e., that fraction_grouping works), but I can update it to use a mock locale with space separators so the output is valid scientific notation (e.g., 12 345.123 45)

components/decimal/src/options.rs
mega123-art: I understand. The concern is that mixing locale data (commas) with fractional grouping options can easily produce invalid output.

However, `DecimalFormatter` needs some mechanism to enable fractional grouping when the data supports it (e.g. space separators). If we remove fraction_grouping from `DecimalFormatterOptions`, how should we expose this capability for locales that do use space separators?

Is the suggestion to make fraction_grouping strictly data-driven (i.e. if the pattern has grouping in the fraction part, we do it; otherwise we don't), without a user option to override?

Or should we hide fraction_grouping from the general options and only expose it via a specific constructor/setter for advanced use cases?

robertbastian: > Is the suggestion to make fraction_grouping strictly data-driven (i.e. if the pattern has grouping in the fraction part, we do it; otherwise we don't), without a user option to override?

Yes.

> Or should we hide fraction_grouping from the general options and only expose it via a specific constructor/setter for advanced use cases?

I think we might need a constructor where you can set all the data-driven values (grouping separator, primary and secondary grouping lengths, fractional grouping, digits, signs), which can then be used to construct a `DecimalFormatter` for the NIST style.

components/decimal/src/decimal_formatter.rs
robertbastian: the API should not allow constructing formatters that produce bogus output like this. this is worse than not having this functionality

components/decimal/src/provider.rs
```diff
@@ -207,6 +208,25 @@ pub struct GroupingSizes {
     /// The minimum number of digits required before the first group. For example, if `primary=3`
     /// and `min_grouping=2`, grouping separators will be present on 10,000 and above.
     pub min_grouping: u8,
+
+    /// The size of groups in the fractional part of the number.
+    ///
+    /// If 0, grouping separators will never be shown in the fractional part.
+    pub fraction: u8,
```
robertbastian, line 214: issue: this breaks deserialization of existing data.  the way to solve this is to introduce a new data marker (`V2`) to use in `try_new_unstable`, and some code in `try_new_with_buffer_provider` that deserializes either the old or the new marker

components/decimal/src/provider.rs
```diff
@@ -193,6 +193,7 @@ pub const MARKERS: &[DataMarkerInfo] = &[DecimalSymbolsV1::INFO, DecimalDigitsV1
 #[cfg_attr(feature = "serde", derive(serde::Deserialize))]
 #[cfg_attr(feature = "datagen", derive(serde::Serialize, databake::Bake))]
 #[cfg_attr(feature = "datagen", databake(path = icu_decimal::provider))]
+#[non_exhaustive]
```
robertbastian: code in `provider` is inherently unstable, so it doesn't need to be marked non-exhaustive

components/decimal/Cargo.toml
```diff
@@ -58,5 +58,10 @@ unstable = ["alloc", "dep:icu_pattern", "dep:icu_plurals"]
 name = "fixed_decimal_format"
 harness = false
 
+[package.metadata.cargo-semver-checks.lints]
+workspace = true
+constructible_struct_adds_field = "allow"
```
robertbastian: don't add these. we just ignore errors for `provider` modules

ffi/npm/test/data-providers.mjs
```diff
@@ -9,128 +9,34 @@ import { Decimal, Locale, DecimalFormatter, DecimalGroupingStrategy, DataProvide
...
-    0xf0, 0x9e, 0xa5, 0x90, 0x04, 0xf0, 0x9e, 0xa5, 0x91, 0x04, 0xf0, 0x9e,
-    0xa5, 0x92, 0x04, 0xf0, 0x9e, 0xa5, 0x93, 0x04, 0xf0, 0x9e, 0xa5, 0x94,
-    0x04, 0xf0, 0x9e, 0xa5, 0x95, 0x04, 0xf0, 0x9e, 0xa5, 0x96, 0x04, 0xf0,
-    0x9e, 0xa5, 0x97, 0x04, 0xf0, 0x9e, 0xa5, 0x98, 0x04, 0xf0, 0x9e, 0xa5,
-    0x99, 0x0e, 0x01, 0x01, 0x02, 0x02, 0x03, 0x04, 0x2d, 0x2b, 0x2e, 0x2c,
-    0x62, 0x65, 0x6e, 0x67, 0x03, 0x02, 0x01, 0x0e, 0x01, 0x01, 0x02, 0x02,
-    0x03, 0x04, 0x2d, 0x2b, 0x2e, 0x2c, 0x6c, 0x61, 0x74, 0x6e, 0x03, 0x02,
-    0x01, 0x0e, 0x01, 0x01, 0x02, 0x02, 0x03, 0x04, 0x2d, 0x2b, 0x2e, 0x2c,
-    0x6c, 0x61, 0x74, 0x6e, 0x03, 0x03, 0x01
+    0x03, 0x08, 0x31, 0x43, 0x45, 0x84, 0x99, 0x47, 0xd1, 0xf9, 0x35, 0x02,
```
robertbastian, line 12: this should not change, the old bytes need to keep working

components/decimal/Cargo.toml
mega123-art: Yaa actually I did these bec semver check wasn't passing

mega123-art:
@robertbastian Thanks for the review. I've reverted the `#[non_exhaustive]` changes and Cargo.toml edits as requested.

Regarding the data compatibility:

1. Should I create a separate `GroupingSizesV1` struct for the old format, or is there a way to handle this with just the new GroupingSizes?
2. Since `try_new_with_buffer_provider` is currently macro-generated, should I replace the macro usage with a manual implementation to handle the fallback logic independently?
3. Is there an existing example in the codebase of a V1 -> V2 migration with buffer fallback that I can follow?

components/decimal/src/decimal_formatter.rs
sffc: Those sections of the spec are talking about the pattern syntax. I'm asking about the localized grouping symbol. For example, are there any locales that use, say, a comma (,) for thousands but a NNBSP (U+202F) for fractional grouping? If the spec is not clear about this, we should bring it up with CLDR.

components/decimal/src/options.rs
sffc: > what's the use case for using a different grouping strategy for fractions than for integers?

It seems you'd want it on for integers and off for fractions, or on for integers and locale-default for fractions.

components/decimal/src/provider.rs
sffc: The use case is so super niche that I don't really want to disrupt data in 2.x, especially for a low-level component. We already know that zero locales have fractional grouping data. I prefer to copy the grouping sizes from the integer portion and leave an issue open to investigate this further and add data in 3.0.

components/decimal/src/decimal_formatter.rs
robertbastian: there can only be a single grouping separator per locale: https://github.com/unicode-org/cldr/blob/main/common/main/root.xml#L3028

robertbastian:
I don't know, this is a tricky migration that is easier to do myself than explain to someone through Github. How about you merge the internal changes, leaving the datagen changes alone. You haven't implemented them anyway, you just set it to 0.

sffc:
Yes, I prefer implementing this without changing the data (which it seems like we can do) and keep any potential data changes to a standalone PR.

PR #7652, opened 2026-02-13
Add regression tests for locale default hour cycle (Fixes #594)

### LINKED ISSUE #594: Choose a default hour cycle for a locale when no preferences are provided
This is a follow-up from https://github.com/unicode-org/icu4x/pull/587#discussion_r605070056

> [When no preference is provided the preference] should default to the locale default, which is region-based (h12 for US, h23 for GB, etc). This is in CLDR, but we need to load it as well as think about the best architecture for where that data loading code should reside.

afrdbaig7:
Fixes #594

This PR adds regression tests to make sure datetime formatting falls back
to the correct default hour cycle based on CLDR region data when no
explicit -u-hc- preference is provided.
I added both runtime and datagen tests so that the behavior is covered
at formatting time as well as at the data level.

Runtime tests (simple_test.rs):
- Check default behavior:
  - en → h12
  - fr, ja, en-GB → h23
- Verify that explicit -u-hc- overrides still take priority
- Add a midnight edge case to ensure formatting is correct

Datagen test (semantic_skeletons/tests.rs):
- Validate that preferred_hour_cycle() returns the expected
  CoarseHourCycle:
  - en → H11H12
  - fr, ja, en-ZA → H2
Note: The runtime tests require `feature = "experimental"` for
`components::Bag` access.
These tests should help prevent regressions and make the expected
locale-based hour cycle behavior more clearly covered.
## Changelog
N/A (Add regression tests for `components::Bag::hour_cycle` resolution across locales.)

### DIFF
--- a/components/datetime/tests/resolved_components.rs
+++ b/components/datetime/tests/resolved_components.rs
@@ -106,3 +106,42 @@ fn test_date_and_time() {
         "en-u-hc-h23".parse::<Locale>().unwrap(),
     );
 }
+
+/// More thorough test of `components::Bag::hour_cycle` across multiple locales,
+/// including region-based defaults and explicit overrides.
+/// Fixes <https://github.com/unicode-org/icu4x/issues/594>.
+#[test]
+fn test_hour_cycle_resolved_components() {
+    let skeleton = CompositeDateTimeFieldSet::Time(TimeFieldSet::T(fieldsets::T::short()));
+
+    let check = |locale_str: &str, expected: HourCycle| {
+        let locale: Locale = locale_str.parse().unwrap();
+        let dtf = FixedCalendarDateTimeFormatter::<Gregorian, _>::try_new(locale.into(), skeleton)
+            .unwrap();
+        let datetime = DateTime {
+            date: Date::try_new_gregorian(2024, 1, 15).unwrap(),
+            time: Time::try_new(21, 22, 0, 0).unwrap(),
+        };
+        let bag = components::Bag::from(&dtf.format(&datetime).pattern());
+        assert_eq!(
+            bag.hour_cycle,
+            Some(expected),
+            "{locale_str}: expected {expected:?}, got {:?}",
+            bag.hour_cycle,
+        );
+    };
+
+    // h12 locales
+    check("en", HourCycle::H12);
+    check("ko", HourCycle::H12);
+
+    // h23 locales
+    check("fr", HourCycle::H23);
+    check("ja", HourCycle::H23);
+    check("de", HourCycle::H23);
+    check("en-GB", HourCycle::H23); // same language, different region
+
+    // explicit -u-hc- overrides
+    check("fr-u-hc-h12", HourCycle::H12);
+    check("en-u-hc-h23", HourCycle::H23);
+}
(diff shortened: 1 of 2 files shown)


sffc (approved):
I think we have tests for this already, but adding additional tests specifically for the `components::Bag` behavior as integration tests is harmless and directly fixes the issue.

afrdbaig7:
> I think we have tests for this already, but adding additional tests specifically for the `components::Bag` behavior as integration tests is harmless and directly fixes the issue.

Thanks for the review
Yes, the main addition here is the components::Bag integration test in simple_test.rs. It structurally checks the resolved hour_cycle for different locales (for example, en → H12 and fr/ja/en-GB → H23). It also covers explicit -u-hc- overrides to make sure they still take precedence, along with a midnight edge case to ensure formatting behaves correctly.
Additionally, I added a datagen-level test in semantic_skeletons/tests.rs that verifies preferred_hour_cycle() directly against the CLDR source data. This acts as an extra safety check so that both the data layer and the runtime behavior stay aligned.

sffc:
Please fix the merge conflict

sffc (approved):

sffc:
Please fix clippy. Thanks

components/datetime/tests/simple_test.rs
```diff
@@ -220,6 +220,89 @@ fn test_5387() {
...
+
+    for (locale_str, expected_hour_cycle) in cases {
+        let locale: Locale = locale_str.parse().unwrap();
+        let prefs = DateTimeFormatterPreferences::from(&locale);
+        let formatter =
+            FixedCalendarDateTimeFormatter::<Gregorian, _>::try_new(prefs, fieldsets::T::short())
+                .unwrap();
+        let formatted = formatter.format(&datetime);
+        let resolved_pattern = formatted.pattern();
+        let bag = components::Bag::from(&resolved_pattern);
```
sffc, line 272: Observation: the main thing you are testing in this PR is the behavior of `components::Bag::hour_cycle`.

There is one small test that does this, `test_length_time`

We also have `test_hour_cycle_selection` which tests the hour cycle resolution through `DateTimeFormatterPreferences` but not all the way through to `components::Bag`

So I lean toward believing that this test is a value-add, but we have this behavior mostly already tested elsewhere.

I would be more confident if you moved the test into `resolved_components.rs` and described it as being a more thorough test of `components::Bag::hour_cycle`.

sffc (approved):

sffc (commented):
(@afrdbaig7 please address my comment from above)

afrdbaig7:
> (@afrdbaig7 please address my comment from above)
Thanks for the reminder will push an update shortly.

sffc (commented):
Thanks, please fix the fmt error and then we can merge

afrdbaig7:
> Thanks, please fix the fmt error and then we can merge

Ok i will fix it and push the update shortly

afrdbaig7:
@sffc  Thanks for your help  I’ve addressed your comment.
I’d like to continue contributing, so I’ll start exploring other open issues as well.
Also, is there a preferred community channel (Discord/Slack) where contributors usually coordinate?

sffc (approved):

PR #7656, opened 2026-02-16
Use `RetainBaseLanguages` deduplication and `DataLocaleFamily::single` in testdata, add `en-US-posix`

robertbastian:
`make_testdata` should use `DataLocaleFamily::single`, as the downloaded test data is for single locales only, so this is not a good test for `::with_descendants`.

I've also changed the deduplication to `::RetainBaseLanguages`, which removes 1313 files from the repo, without significantly impacting the readability of generated data (it's only removes "trivial" fallbacks to the base language).

Also added the `en-US-posix` locale, for which collator has data, but which exposed some issues with markers that don't call `self.check_req`.


Manishearth:
Seems fine to me, did not look super closely.

PR #7658, opened 2026-02-17
Fix resb on big endian platforms

miladfarca:
Node.js/V8 are now using Temporal with a dependency on Rust. On big endian platforms ICU4C provides data in native BE order causing ZoneInfo64 parsing to fail with "Failed to load timezone info" and is causing test failures on Node.js after enabling Temporal:
https://github.com/nodejs/node/issues/61808

Following the instruction in this page:
https://chromium.googlesource.com/chromium/src/+/28fe000/third_party/rust/chromium_crates_io/patches/
I have uploaded a patch to Chromium to fix this issue on that repository:
http://crrev.com/c/7583895
And I have also applied the same patch here to ICU4X upstream.

## Changelog

resb: Support big-endian platforms

### DIFF
--- a/utils/resb/src/binary/header.rs
+++ b/utils/resb/src/binary/header.rs
@@ -114,7 +114,7 @@ impl TryFrom<&[u8]> for BinReprInfo {
         let (reserved_word, value) = read_u16(value)?;
 
         let (endianness, value) = (Endianness::try_from(value[0])?, &value[1..]);
-        if (endianness == Endianness::Little) == cfg!(target_endian = "little") {
+        if (endianness == Endianness::Little) != cfg!(target_endian = "little") {
             return Err(BinaryDeserializerError::unsupported_format(
                 "bundle endianness does not match platform endianness",
             ));


utils/resb/src/binary/header.rs
```diff
@@ -117,11 +117,6 @@ impl TryFrom<&[u8]> for BinReprInfo {
         // contents of the repr info, we explicitly depend on little endian data
         // in order to ensure compatibility with `zerovec`.
         let (endianness, value) = (Endianness::try_from(value[0])?, &value[1..]);
-        if endianness != Endianness::Little {
```
robertbastian, line 120: issue: don't remove this check, make it compare to the platform

utils/resb/src/binary/header.rs
robertbastian: remove the comment, it doesn't make sense

utils/resb/src/binary/header.rs
Manishearth: You can use `cfg!(target_endian = "little")` to get a boolean

utils/resb/src/binary/deserializer.rs
```diff
@@ -138,7 +138,7 @@ impl<'de> ResourceTreeDeserializer<'de> {
                 ))
             }
         };
-        let descriptor = u32::from_le_bytes(descriptor);
+        let descriptor = u32::from_ne_bytes(descriptor);
```
Manishearth, line 146: We should document `from_words` as requiring a native-endian file, noting that zoneinfo64.res comes in both formats, and then every time we do this we should say "See documentation on from_words".

miladfarca: I've added a comment under `utils/resb/src/binary/deserializer.rs`.

utils/resb/src/binary/header.rs
miladfarca: Thanks, I've created a platform specific check.

utils/resb/src/binary/header.rs
miladfarca: done.

utils/resb/src/binary/header.rs
```diff
@@ -113,13 +113,11 @@ impl TryFrom<&[u8]> for BinReprInfo {
         let (size, value) = read_u16(value)?;
         let (reserved_word, value) = read_u16(value)?;
 
-        // While the consumer is responsible for verifying acceptability of most
-        // contents of the repr info, we explicitly depend on little endian data
-        // in order to ensure compatibility with `zerovec`.
         let (endianness, value) = (Endianness::try_from(value[0])?, &value[1..]);
-        if endianness != Endianness::Little {
+        let native_le = cfg!(target_endian = "little");
+        if endianness != if native_le { Endianness::Little } else { Endianness::Big } {
```
robertbastian:
```suggestion
        if (endianness == Endianness::Little) == cfg!(target_endian = "little") {
```

miladfarca: I think the logic is inverted here, it should be 
```
if (endianness == Endianness::Little) != cfg!(target_endian = "little") {
```
I'll patch it.

robertbastian:
@miladfarca please sign the CLA

miladfarca:
@robertbastian I'll need to check if we (IBM) are allowed to sign this CLA as we are not in your current list:
https://api.unicode.org/cla/corporate

miladfarca:
@robertbastian I have now signed the CLA, sorry for the delay.

PR #7659, opened 2026-02-17
Refactor web-demo to support multiple demos(#2360)

### LINKED ISSUE #2360: Wasm Demo Frontend Framework
The [wasm demo](https://github.com/unicode-org/icu4x/tree/main/ffi/diplomat/wasm/wasm-demo) is currently written in vanilla TypeScript. This makes it difficult to add new demos to the webpage, since all the HTML and TS pieces have to be added manually, and rules for what to update when an input is updated are encoded by hand.

A frontend framework like Svelte would simplify this process, and make adding new demos as simple as "use these inputs, have them call these Diplomat bindings, and show the output here."

Currently, the demo is generated for every PR, so this would ideally enable authors to easily add a demo for their new feature that the reviewers could interact with in real time.

afrdbaig7:
Fixes #2360

This PR restructures tools/web-demo to support multiple independent demos instead of a single hardcoded one. The goal was to make it easier to extend while keeping the setup simple and lightweight.

Changes:
Router (src/router.mjs) – Added a small hash-based router to handle navigation, loading state, and proper teardown before mounting a new demo. Includes a guard to avoid race conditions on fast navigation.
Demo registry (src/demos/registry.mjs) – Central place to register demos with slug, label, and lazy load() via dynamic import.
Wasm loader (src/utils/loader.mjs) – Singleton wrapper so the icu module is initialized only once.
Playground demo – Moved existing demo to src/demos/playground/ and adapted it to use setup/teardown.
Template demo – Added a minimal starter demo for contributors.
Updated webpack entry to src/index.mjs and adjusted HTML to include #nav and #app.
To add a new demo: copy the template folder, implement setup/teardown, and register it in the demo registry.


Manishearth:
This seems to be a lot of changes at once and it's pretty hard to review. It also doesn't appear to fix the filed issue.

Refactoring this into multiple files is welcome, but please do it piecewise as commits, without intermixed behavior changes.

sffc:
I looked over the code and I'm not sure what it achieves. It seems like a refactoring.

CC @mega123-art who worked on this part of the code in https://github.com/unicode-org/icu4x/pull/7567

afrdbaig7:
> This seems to be a lot of changes at once and it's pretty hard to review. It also doesn't appear to fix the filed issue.
> 
> Refactoring this into multiple files is welcome, but please do it piecewise as commits, without intermixed behavior changes.

Thanks for the review. I see your point. I’ll split this up into smaller, incremental commits first keeping it to pure refactors, and then layering the behavior changes on top so it’s easier to review.
I’ll make sure the final structure clearly and directly addresses the original issue.

afrdbaig7:
> I looked over the code and I'm not sure what it achieves. It seems like a refactoring.
> 
> CC @mega123-art who worked on this part of the code in #7567

Thanks for the feedback. I didn’t realize #7567 had already introduced hash-based deep linking inside the existing demo. My intention with this PR was a bit different from #2360 I was trying to make the structure support multiple separate demo pages (like a playground, formatter demo, segmenter demo, etc.) with lazy loading and proper setup/teardown, rather than deep-linking within one demo.
 I should’ve reviewed #7567 more carefully before opening this since there’s definitely some overlap in the routing logic. I’m completely fine with closing this and reworking it so it builds on top of the current deep-link router instead of replacing it, if that makes more sense. Just let me know how you’d prefer I move forward

Manishearth:
@afrdbaig7 Are you using an LLM to primarily drive code writing and commenting?

sffc:
Yes I think we should build on top of #7567.

afrdbaig7:
> @afrdbaig7 Are you using an LLM to primarily drive code writing and commenting?

I’ve used tools here and there to double check  but I review everything myself before submiting it.

afrdbaig7:
> Yes I think we should build on top of #7567.

Ok I'll close this and rework on top of #7567. Thanks for the guidance.

PR #7660, opened 2026-02-17
Special-case hebrew months in names data struct

robertbastian:
This moves the "membrane" between `icu_calendar` and `icu_datetime` to not having to expose the non-standard "formatting month code".

#7149

### DIFF
--- a/components/datetime/src/pattern/names.rs
+++ b/components/datetime/src/pattern/names.rs
@@ -3773,25 +3773,37 @@ impl RawDateTimeNamesBorrowed<'_> {
         let month_index = usize::from(month.number() - 1);
         let name = match month_names {
             MonthNames::Linear(linear) => {
-                if month.is_formatting_leap() {
+                if month.is_leap() {
                     None
                 } else {
                     linear.get(month_index)
                 }
             }
+            MonthNames::LeapLinear(hebrew) if hebrew.get(12).unwrap_or_default().is_empty() => {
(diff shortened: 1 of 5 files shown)


components/datetime/src/pattern/names.rs
```diff
@@ -3773,25 +3773,37 @@ impl RawDateTimeNamesBorrowed<'_> {
...
         let name = match month_names {
             MonthNames::Linear(linear) => {
-                if month.is_formatting_leap() {
+                if month.is_leap() {
                     None
                 } else {
                     linear.get(month_index)
                 }
             }
+            MonthNames::LeapLinear(hebrew) if hebrew.get(12).unwrap_or_default().is_empty() => {
```
sffc, line 3782: "if M01L is empty" is a heuristic that works.

The heuristic I had half-baked in the call was:

1. If Month is leap, select the leap display name.
2. Else, if ordinal number > month number and the leap display name is nonempty, select the leap display name.
3. Else, select the normal display name.

But I'm not convinced that either of these heuristics are "good".

robertbastian: that doesn't work, it'd use the leap name for every chinese month after the leap month

robertbastian: > But I'm not convinced that either of these heuristics are "good".

They work. I'd like to introduce a `Hebrew` variant instead, but that would break old-code-new-data.

PR #7662, opened 2026-02-18
Add LocaleNamesRegionLongV1 data struct with data

sffc:
#3260


sffc:
Adding the data without using the VarULE optimization increases the build time of icu_experimental from 18s to 30s (on my fast computer). Building it takes 34s. Enabling the VarULE optimization brings build time down to 19s and data build time down to 14s.

CAUTION: These numbers are preliminary and could be way off. I tried to time the build of only `icu_experimental` but it might not be a fully clean build.

```
# NOT OPTIMIZED:
2026-02-18T01:14:31.027Z INFO  [icu_provider_export::export_impl] Generated marker LocaleNamesRegionLongV1 (34.943s, 'en-GB/KW' in 932.101µs, flushed in 22.194s)

# WITH VARULE OPTIMIZATION:
2026-02-18T01:26:35.962Z INFO  [icu_provider_export::export_impl] Generated marker LocaleNamesRegionLongV1 (13.983s, 'scn/IL' in 2.334ms, flushed in 1.689s)

# BASELINE:
sffc@sffc0:components/experimental$ time cargo build --all-features
   Compiling icu_experimental v0.5.0-dev (components/experimental)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 18.30s

real    0m18.372s
user    0m18.315s
sys     0m0.857s

# WITH DATA, NOT OPTIMIZED:
sffc@sffc0:components/experimental$ time cargo build --all-features
   Compiling icu_experimental v0.5.0-dev (components/experimental)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 29.99s

real    0m30.058s
user    0m29.625s
sys     0m1.263s

# WITH DATA, VARULE OPTIMIZATION:
sffc@sffc0:components/experimental$ time cargo build --all-features
   Compiling icu_experimental v0.5.0-dev (components/experimental)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 19.18s

real    0m19.249s
user    0m19.046s
sys     0m0.908s
```

sffc:
I added a benchmark. To load all names for all locales: 6.1783 ms without optimization, 6.3271 ms with VarULE. So, basically no difference. The time is all in the lookup, not the data dereferencing.

I also ran it for postcard data. Only about 5% slower, which again suggests that the bottleneck is the zerotrie lookup, which both code paths share.

sffc:
Data size:

ZeroMap:

```
region/display/names/v1, <lookup>, 1123B, 224 identifiers
region/display/names/v1, <total>, 1162567B, 1143806B, 208 unique payloads
```

Attributes:

```
locale/names/region/long/v1, <lookup>, 232109B, 43462 identifiers
locale/names/region/long/v1, <total>, 1274068B, 607866B, 28966 unique payloads
```

I haven't applied my dense zerotrie optimization, but it would reduce the data size of only the lookup table. Even if it cuts it down by half, we are still only the same size as the ZeroMap structure. 🤔 

EDIT 1: The ZeroMap structure contains both long and short names. The Attributes structure contains only long names. Still, I expect a bigger difference.

EDIT 2: The second column, which shows "postcard structs size", is substantially smaller for attributes than for zeromap. Since we are using the VarULE optimization, the "baked structs size" should be closer to the postcard size than usual, since we don't store the 64-bit pointers to each individual string in static memory.

components/experimental/src/displaynames/provider.rs
```diff
@@ -153,3 +153,38 @@ pub struct VariantDisplayNames<'data> {
...
+            name: VarZeroCow::new_borrowed(name),
+        }
+    }
+}
+
+icu_provider::data_struct!(
+    SingleDisplayName<'_>,
+    varule: str,
+    #[cfg(feature = "datagen")]
+    encode_as_varule: |v: &SingleDisplayName<'_>| &*v.name
```
Manishearth, line 180: nit (optional): potentially incldue a comment saying what about this data makes it VarULE-optimization-suitable so future readers of the code are not confused.

Manishearth:
(I did not review the large benchmark since Shane said he planned to remove it)

components/experimental/src/displaynames/provider.rs
sffc: I'll do this when I add the rest of these.

sffc:
Since the commits are interesting for future study on the VarULE optimization, I archived the branch at https://github.com/sffc/omnicu/tree/archive/displaynames-single-data

components/experimental/src/displaynames/provider.rs
```diff
@@ -153,3 +153,38 @@ pub struct VariantDisplayNames<'data> {
...
+);
+
+icu_provider::data_marker!(
+    /// Data marker for collation tailorings.
+    LocaleNamesRegionLongV1,
+    "locale/names/region/long/v1",
+    SingleDisplayName<'static>,
+    #[cfg(feature = "datagen")]
+    attributes_domain = "locale_names_region",
```
robertbastian, line 189: we usually use hyphens for our identifiers, not underscores

components/experimental/src/displaynames/provider.rs
```diff
@@ -153,3 +153,38 @@ pub struct VariantDisplayNames<'data> {
...
+
+icu_provider::data_struct!(
+    SingleDisplayName<'_>,
+    varule: str,
+    #[cfg(feature = "datagen")]
+    encode_as_varule: |v: &SingleDisplayName<'_>| &*v.name
+);
+
+icu_provider::data_marker!(
+    /// Data marker for collation tailorings.
```
robertbastian, line 184: .

components/experimental/src/displaynames/provider.rs
```diff
@@ -153,3 +153,38 @@ pub struct VariantDisplayNames<'data> {
...
+    LocaleNamesRegionLongV1,
+    "locale/names/region/long/v1",
+    SingleDisplayName<'static>,
```
robertbastian, line 187: why the extra type?

```suggestion
    VarZeroCow<'static, str>
```

components/experimental/src/displaynames/provider.rs
sffc: Discussed elsewhere: I used underscores because of `numbering_system` and `datetime_year_length`. If you feel strongly we can make a PR to change to hyphens across all of the attribute domains and discuss it there.

components/experimental/src/displaynames/provider.rs
sffc: I considered it, but it didn't work on my first try so I wrapped it. I don't remember exactly why

sffc: https://github.com/unicode-org/icu4x/pull/7683

components/experimental/src/displaynames/provider.rs
sffc: #7683

PR #7663, opened 2026-02-18
Benchmark for BoundLocaleDataProvider

sffc:
Issue: #3260 

Merge of #7662 and #7621 with a benchmark.

```
$ ICU4X_REGION_POSTCARD_PATH=../../_debug/regionnames.postcard cargo bench --all-features region

...

locale_names/region/postcard/all/1by1
                        time:   [6.8448 ms 6.9295 ms 7.0177 ms]
Found 6 outliers among 100 measurements (6.00%)
  6 (6.00%) high mild

locale_names/region/postcard/all/boundlocale
                        time:   [3.6443 ms 3.6895 ms 3.7372 ms]
Found 34 outliers among 100 measurements (34.00%)
  15 (15.00%) low severe
  1 (1.00%) low mild
  3 (3.00%) high mild
  15 (15.00%) high severe
```

I built regionnames.postcard like this: `cargo run -p icu4x-datagen --all-features -- -m LocaleNamesRegionLongV1 CaseMapV1 PluralsCardinalV1 TimezoneNamesLocationsOverrideV1 --format blob -o _debug/regionnames.postcard`

I included the target marker as well as a small handful of others to simulate walking through the marker trie.


sffc:
I added a bench on the old ZeroMap struct:

```
locale_names/region/baked/all/1by1
                        time:   [5.9787 ms 6.0411 ms 6.1111 ms]
Found 5 outliers among 100 measurements (5.00%)
  5 (5.00%) high mild

locale_names/region/baked/all/old_map_struct
                        time:   [2.3787 ms 2.3947 ms 2.4130 ms]
Found 18 outliers among 100 measurements (18.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
  16 (16.00%) high severe

locale_names/region/postcard/all/1by1
                        time:   [6.9537 ms 7.0637 ms 7.1818 ms]
Found 14 outliers among 100 measurements (14.00%)
  13 (13.00%) high mild
  1 (1.00%) high severe

locale_names/region/postcard/all/boundlocale
                        time:   [3.3650 ms 3.3756 ms 3.3897 ms]
Found 11 outliers among 100 measurements (11.00%)
  2 (2.00%) high mild
  9 (9.00%) high severe
```

robertbastian:
Can you add a benchmark for loading like 10 names? I don't think loading 300+ names at once is the most realistic use case.

sffc:
> Can you add a benchmark for loading like 10 names? I don't think loading 300+ names at once is the most realistic use case.

It is a realistic use case, because pickers have a lot of regions/languages, but I added one to load 10% of the names.

```
locale_names/region/baked/10x10pct/1by1
                        time:   [72.051 µs 72.475 µs 72.879 µs]
Found 17 outliers among 100 measurements (17.00%)
  5 (5.00%) low severe
  4 (4.00%) high mild
  8 (8.00%) high severe

locale_names/region/baked/10x10pct/old_map_struct
                        time:   [24.617 µs 24.832 µs 25.026 µs]

locale_names/region/postcard/10x10pct/1by1
                        time:   [68.274 µs 69.017 µs 69.688 µs]
Found 23 outliers among 100 measurements (23.00%)
  1 (1.00%) high mild
  22 (22.00%) high severe

locale_names/region/postcard/10x10pct/boundlocale
                        time:   [34.649 µs 34.948 µs 35.321 µs]
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
```

sffc:
I want to point out that the benchmark performance is more dependent on the shape of the data build than the exact workload. I built a postcard file containing display names data for all regions and all locales. If I were to use a postcard file containing display names for only one locale, then the diff is going to be smaller.

sffc:
From a discussion today:

@robertbastian and @Manishearth both want to see "real life end-to-end benchmarks", but at the same time acknowledge that there isn't a cohesive way to write such benchmarks. For example, I could allocate a string in these benchmarks, like we will need to do for V8, but then how about UTF-16? How about writing into the DOM? @robertbastian said he thinks the cost here will drop to 1% of end-to-end; I personally expect more like 10-20% for that particular workload. But, there are cases like in Rust that _don't_ need the extra allocations or all this extra work and therefore will see a greater marginal difference, more in line with my benchmarks in this thread.

Therefore, I think a reasonable conclusion is that end-to-end performance will improve by _50% or less_ with the BoundLocaleDataProvider design. The full 50% comes in the best case of a low-level Rust program, and a smaller percent comes in the worst-case of lots of FFI loading into a DOM.

---

@Manishearth stated that these numbers, being in the microseconds, are all so slow that performance-sensitive applications that care about fps performance will want to offload this work to another thread anyway, which leads to the question, do we really care about the difference between 24us and 69us?

I must admit that I find this line of reasoning to be beside the point. The 24us is just a benchmark, 10x10pct. An actual workload may end up with 10% or 10x of that raw figure. I also _really_ don't want to have to claim to our users that "DisplayNames is slow and the maintainers decided there's nothing we can do about it". I want us to claim "DisplayNames may or may not be slow depending on the workload; the maintainers put in the time to make it as fast as possible."

Manishearth:
> I also really don't want to have to claim to our users that "DisplayNames is slow and the maintainers decided there's nothing we can do about it". I want us to claim "DisplayNames may or may not be slow depending on the workload; the maintainers put in the time to make it as fast as possible."

I don't want us to design an API and data model for people it won't be useful to. I would rather have people come to us with their performance needs, rather than us guess their performance needs and get it wrong. I would consider it a failure if we designed an API motivated by the FPS use case, and the performance isn't good enough for them so they design their own caching wrapper.

Performance is not a property of an API, it is a property of a usage pattern. There are many, many usage patterns here, especially once you start looking at caching, a thing which is likely to be important for these users.

"DisplayNames is fast for _these_ use cases, and probably fast _enough_ for _these other_ use cases but if it isn't please talk to us" is a very respectable position I think.

Part of the trickiness here is that when it comes to caching, there is always the option to just preload using the 1x1 API. We don't implement a "multi" provider that does this because this is something someone can do on their own. That's not a choice to keep DisplayNames slow, that's a choice to keep caching on the other side of the API boundary.

sffc:
It seems the effort I put into making these benchmarks did _absolutely nothing_ to change anyone's mind. It convinced me that the single display name design is very slow and should be avoided, and it somehow convinced @robertbastian and @Manishearth that it's not that slow (yes, it's more nuanced than that, I'm not interested in continuing this argument because I don't think it leads anywhere).

I'm closing this PR but I will restore the benchmarks into #7621

Manishearth:
I think the benchmarks changed my mind pretty significantly, actually? I was previously very unsure, and now I am in favor of multi-style optimizations. There are other competing concerns and I will try to explain my position on #7621, but my current conclusion is I want to keep Bound in our back pocket for when we have users with this need, so we can use that to inform our decisions, instead of predesigning something with Bound.

I think perhaps my pushback against _specific_ performance arguments yesterday has been interpreted as dislike of the overall approach.

PR #7664, opened 2026-02-18
Add more constructors to UnvalidatedTinyAsciiStr

sffc:
Split from #7663

## Changelog

tinystr: Add more constructors to UnvalidatedTinyAsciiStr
 - New methods: `UnvalidatedTinyAsciiStr::try_from_utf8()`
 - New associated const: `TinyAsciiStr::EMPTY`

### DIFF
--- a/utils/tinystr/src/unvalidated.rs
+++ b/utils/tinystr/src/unvalidated.rs
@@ -27,9 +27,6 @@ impl<const N: usize> fmt::Debug for UnvalidatedTinyAsciiStr<N> {
 }
 
 impl<const N: usize> UnvalidatedTinyAsciiStr<N> {
-    /// The empty string as a default value.
-    pub const DEFAULT: Self = Self([0u8; N]);
-
     #[inline]
     /// Converts into a [`TinyAsciiStr`]. Fails if the bytes are not valid ASCII.
     pub fn try_into_tinystr(self) -> Result<TinyAsciiStr<N>, ParseError> {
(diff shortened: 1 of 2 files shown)


utils/tinystr/src/unvalidated.rs
```diff
@@ -27,14 +27,31 @@ impl<const N: usize> fmt::Debug for UnvalidatedTinyAsciiStr<N> {
 }
 
 impl<const N: usize> UnvalidatedTinyAsciiStr<N> {
+    /// The empty string as a default value.
+    pub const DEFAULT: Self = Self([0u8; N]);
```
Manishearth: EMPTY? but fine either way

utils/tinystr/src/unvalidated.rs
```diff
@@ -27,14 +27,31 @@ impl<const N: usize> fmt::Debug for UnvalidatedTinyAsciiStr<N> {
...
+
     #[inline]
     /// Converts into a [`TinyAsciiStr`]. Fails if the bytes are not valid ASCII.
     pub fn try_into_tinystr(self) -> Result<TinyAsciiStr<N>, ParseError> {
         TinyAsciiStr::try_from_raw(self.0)
     }
 
     #[inline]
-    /// Unsafely converts into a [`TinyAsciiStr`].
+    /// Creates one of these from a byte slice, returning an error if the bytes are too long.
```
Manishearth: nit: probably worth explicitly documenting that this doesn't check TinyAsciiStr invariants

utils/tinystr/src/unvalidated.rs
robertbastian: add this on `TinyAsciiStr`. we already have a const conversion into unvalidated

robertbastian: +1

utils/tinystr/src/unvalidated.rs
Manishearth: (yes, this type is "unvalidated", but "unvalidated" types are still intended to _mostly be well-formed_, so it is not obvious that a constructor that performs _some_ validation is not performing _all_ the validation)

robertbastian (approved):

PR #7666, opened 2026-02-18
Add `MonthNames::LeapPattern` and make `LeapLinear` be serde-only

robertbastian:
Currently we generate all month names in datagen, we can halve our data size if we just store the leap pattern.

This stores both the `leap` pattern, as well as the `standardAfterLeap` and `combined` calendars, which might be needed for future calendars.

It also models Hebrew names using the `leap` and `standardAfterLeap` patterns.

## Changelog

icu_datetime: Reduce data size by storing leap patterns where possible


provider/source/src/datetime/names.rs
```diff
@@ -352,96 +336,70 @@ fn months_convert(
...
-            } else {
-                let mut index: usize = k
+        // CLDR's numbering for hebrew has Adar I as 6, Adar as 7, and Adar II as 7-yeartype-leap
+        let leap_pattern = SinglePlaceholderPattern::try_from_str(
+            &months["6"].replace(&months["5"], "{0}"),
+            Default::default(),
+        )
+        .unwrap();
+        let standard_after_leap_pattern = SinglePlaceholderPattern::try_from_str(
+            &months["7-yeartype-leap"].replace(&months["7"], "{0}"),
```
Manishearth: issue: this is fragile and not guaranteed to work.

provider/source/src/cldr_serde/ca.rs
```diff
@@ -104,6 +104,10 @@ pub(crate) struct MonthSymbols(pub(crate) HashMap<String, String>);
 #[derive(Debug, PartialEq, Clone, Deserialize)]
 pub(crate) struct MonthPatternSymbols {
     pub(crate) leap: PatternString<SinglePlaceholder>,
+    #[serde(default, rename = "standardAfterLeap")]
+    pub(crate) standard_after_leap: PatternString<SinglePlaceholder>,
```
Manishearth: issue: no CLDR data actually sets this. We should probably just panic when we see this because we may need code changes.

Manishearth (commented):
I don't like trying to guess at patterns. If we want to compress Hebrew's case, we should compress it as the actual data, not by inventing a pattern.

I think it's fine to explore using string patterns for Chinese. We did make a deliberate decision here but that was also before we realized we'd need patterns anyway for strings.

Also, as noted elsewhere, we should not be calling this "standard after leap", CLDR's implementation here works as "Month 7, yeartype leap"

provider/source/src/datetime/names.rs
sffc: You could keep LeapLinear for cases when this find/replace fails.

components/datetime/src/provider/names.rs
```diff
@@ -308,6 +312,13 @@ pub enum MonthNames<'data> {
         )]
         Cow<'data, SinglePlaceholderPattern>,
     ),
+
+    /// This represents the formatting to apply to calendars with leap months.
+    /// The last three elements are patterns:
+    /// * N-3: `SinglePlaceholderPattern` for leap months
+    /// * N-2: `SinglePlaceholderPattern` for standard months after leap months
+    /// * N-1: `DoublePlaceholderPattern` for combined leap months
+    LeapPattern(VarZeroVec<'data, str>),
```
sffc, line 320: Suggestion: the biggest win is in Chinese/Dangi. It would be simpler to only model those, with the leap month pattern and N-1, without these more esoteric patterns that don't come from CLDR and might not always work. But, I also see the argument that since we're adding this extra data struct, we might as well figure out how to make it support Hebrew.

Thought: We could do like we do with modern Japanese era indices, where the era index is included in the YearInfo struct. We could define a Hebrew month name index and put it in the MonthInfo struct.

provider/source/src/datetime/names.rs
robertbastian: why do you think this is fragile? it cannot "fail"

Manishearth: This entirely depends on the formatting data containing a substring.

I guess if it doesn't it still works because then it doesn't have a pattern. Still feels like a misuse.

This trick will work for Hebrew because it only has one such month.

Manishearth: I think _if_ we are relying on that property I would like far more docs on that.

robertbastian: > This trick will work for Hebrew because it only has one such month.

This code is for Hebrew only, so if it works for Hebrew it's correct

provider/source/src/cldr_serde/ca.rs
robertbastian: the code changes for this are in this PR. if a calendar starts using this, it'll work

components/datetime/src/provider/names.rs
robertbastian: > Thought: We could do like we do with modern Japanese era indices, where the era index is included in the YearInfo struct. We could define a Hebrew month name index and put it in the MonthInfo struct.

and what would that index be? how would you model the combination of month number and month leap status? you'd almost need to define a language (like for plural cases). in any case, this would add considerable data size that is unnecessary if we can model this in a fixed way

sffc: It would be an index 0-11 and 12-13 (Adar I/II). You could name the field on MonthInfo `Option<u8>` and then use the Linear month names structure. Hebrew knows internally what the indices mean, but they are opaque. Just like the Japanese era indices are opaque. You don't need LeapStatus on the membrane.

I have not discussed this before and don't know if it is a good or a bad idea.

robertbastian: that would be a Hebrew-specific mapping. we can do that, but that's basically what this is without being Hebrew-specific

robertbastian:
very constructive as always

provider/source/src/cldr_serde/ca.rs
Manishearth: I don't want it to. I think CLDR's standardAfterLeap is a mistake (it makes sense for _one_ calendar and that calendar doesn't use it) and if they introduce that I want to be more deliberate about how we map it.

provider/source/src/datetime/names.rs
Manishearth: Right, that's fine. Just add docs please.

components/datetime/src/provider/names.rs
sffc: What you have here is a whole lot of machinery that exists because of Hebrew peculiarities. A design like mine shifts those peculiarities away from the membrane.

sffc:
> very constructive as always

We are having a constructive discussion? I wanted to reflect that in the review status instead of someone accidentally landing this with an open design question that impacts data stability.

components/calendar/src/types.rs
```diff
@@ -500,12 +493,13 @@ pub enum LeapStatus {
     Normal,
     /// A leap month.
     Leap,
-    /// A month that is not itself considered a leap month, but might
-    /// have special formatting because it occurs after a leap month.
+    /// A standard month that has a corresponding leap month.
```
Manishearth: "in the same year"

components/datetime/src/pattern/names.rs
```diff
@@ -3785,7 +3785,13 @@ impl RawDateTimeNamesBorrowed<'_> {
             #[cfg(feature = "serde")]
             MonthNames::LeapLinear(leap_linear) => {
                 let num_months = leap_linear.len() / 2;
-                if month.leap_status() != LeapStatus::Normal {
+                if month.leap_status() == LeapStatus::Leap {
+                    leap_linear.get(month_index + num_months)
+                } else if month.leap_status() == LeapStatus::LeapBase
+                    && month.number() < month.ordinal
```
Manishearth, line 3791: praise: clever hack.


I think this lets us land the LeapBase part of this PR without any data model changes, and would prefer that sequencing since LeapBase is a 2.2 blocker and the data model is not.

This also gives us better impetus to improve the data model for 2.x.

components/datetime/src/provider/names.rs
```diff
@@ -308,6 +312,12 @@ pub enum MonthNames<'data> {
         )]
         Cow<'data, SinglePlaceholderPattern>,
     ),
+
+    /// This represents the formatting to apply to calendars with leap months.
```
Manishearth, line 316: observation: this could be designed as containing a struct instead to be better documented

but that would likely have larger data size. VZV is fine.

components/datetime/src/pattern/names.rs
robertbastian: https://github.com/unicode-org/icu4x/pull/7701?

components/datetime/src/provider/names.rs
robertbastian: data size would probably be the same, but stack size would be larger

components/datetime/src/pattern/names.rs
Manishearth: Thanks!

components/datetime/src/provider/names.rs
Manishearth: Hmm true.

components/pattern/src/frontend/mod.rs
```diff
@@ -154,6 +153,12 @@ impl<B: PatternBackend> Pattern<B> {
         // Safety: Pattern is repr(transparent) over B::Store
         unsafe { &*(store as *const B::Store as *const Self) }
     }
+
+    #[doc(hidden)] // B::Store is doc(hidden)
+    pub fn from_ref_store(store: &B::Store) -> Result<&Self, Error> {
+        B::validate_store(store)?;
+        Ok(Self::from_ref_store_unchecked(store))
```
sffc, line 160: Suggestion(optional): instead of adding the new function, you can use `impl VarULE`

provider/source/src/datetime/names.rs
```diff
@@ -352,96 +336,71 @@ fn months_convert(
...
-                    continue;
-                }
-                let replaced = leap
-                    .0
-                    .interpolate([&symbols[i]])
-                    .write_to_string()
-                    .into_owned();
-                symbols[nonleap + i] = replaced.into();
-            }
-            Ok(MonthNames::LeapLinear((&symbols).into()))
```
sffc, line 501: Issue (follow-up OK): please keep generating `LeapLinear` for serde clients

components/datetime/src/provider/names.rs
```diff
@@ -320,6 +320,79 @@ pub enum MonthNames<'data> {
     LeapPattern(VarZeroVec<'data, str>),
 }
 
+#[cfg(feature = "serde")]
+impl serde::Serialize for MonthNames<'_> {
```
Manishearth, line 325: nit: include a comment as to why we do all this

Manishearth (approved):

provider/source/src/datetime/names.rs
sffc: > Issue (follow-up OK): please keep generating `LeapLinear` for serde clients

Did the issue get filed, or is it fixed already?

sffc: ^ The answer to this question that sat unanswered for 14 days appears to be that there's some sniffing for Hebrew in `impl Serialize` in the icu_datetime crate in order to output LeapLinear.

components/datetime/src/provider/names.rs
```diff
@@ -308,6 +312,88 @@ pub enum MonthNames<'data> {
...
+                use alloc::vec::Vec;
+                use zerovec::vecs::VarZeroVecOwned;
+
+                let leap_pattern = l.get(l.len() - 2).unwrap_or_default();
+                let leap_base_pattern = l.get(l.len() - 1).unwrap_or_default();
+
+                let normal_names = l.iter().take(l.len() - 2);
+
+                let r = if leap_pattern.starts_with('\0') {
+                    // The leap pattern is not actually a pattern (no placeholder) - this means it's Hebrew
```
sffc, line 354: issue: does this heuristic actually work all the time?

provider/source/src/datetime/names.rs
robertbastian: I didn't think it was necessary to respond to this question, the last commit here, which you appear to have reviewed, is called "revert serde" and implements this request

components/datetime/src/provider/names.rs
sffc: I thought the leap pattern for Hebrew was like "{0} II" in some locales.

Manishearth: CLDR doesn't use standardAfterLeap patterns currently

PR #7673, opened 2026-02-19
Enable clippy lints for transmutes

robertbastian:
And clean up a lot of transmutes.


utils/yoke/derive/src/lib.rs
```diff
@@ -318,29 +318,15 @@ fn yokeable_derive_impl(input: &DeriveInput) -> TokenStream2 {
...
-                        ::core::mem::size_of::<Self::Output>(),
-                        ::core::mem::size_of::<Self>(),
-                    );
-                    let this = ::core::mem::ManuallyDrop::new(this);
-                    // `Deref` coercion and then `&` to `*const` coercion.
-                    let ptr: *const Self = (&this as &Self::Output as *const Self::Output).cast();
-
-                    unsafe { ::core::ptr::read(ptr) }
+                unsafe fn make(from: Self::Output) -> Self {
+                    ::core::mem::transmute::<Self::Output, Self>(from)
```
Manishearth, line 322: worry: I'm not convinced this will always work. Probably fine, can revert if it's a problem.

utils/tinystr/src/int_ops.rs
```diff
@@ -141,23 +126,23 @@ impl Aligned4 {
         (upper_alpha & mask) == 0
     }
 
-    pub const fn to_ascii_lowercase(&self) -> Self {
+    pub const fn to_ascii_lowercase(&self) -> [AsciiByte; 4] {
         let word = self.0;
         let result = word | (((word + 0x3f3f_3f3f) & !(word + 0x2525_2525) & 0x8080_8080) >> 2);
-        Self(result)
+        unsafe { AsciiByte::to_ascii_byte_array(&result.to_ne_bytes()) }
```
Manishearth, line 132: nit: safety comments, please. I legitimately don't understand what's going on here, safety wise.

Sufficient to prove that the above does not produce anything out of range, don't need to prove that it actually lowercases.

Manishearth: (yes, I know the previous code was worse)

utils/tinystr/src/int_ops.rs
```diff
@@ -283,33 +253,33 @@ impl Aligned8 {
...
-    pub const fn to_ascii_lowercase(&self) -> Self {
+    pub const fn to_ascii_lowercase(&self) -> [AsciiByte; 8] {
         let word = self.0;
         let result = word
             | (((word + 0x3f3f_3f3f_3f3f_3f3f)
                 & !(word + 0x2525_2525_2525_2525)
                 & 0x8080_8080_8080_8080)
                 >> 2);
-        Self(result)
+        unsafe { AsciiByte::to_ascii_byte_array(&result.to_ne_bytes()) }
```
Manishearth, line 263: nit: safety comments here too

utils/yoke/derive/src/lib.rs
robertbastian: wdym "always"?

Manishearth: This is the output of a derive macro, and previously Rust was not smart enough to figure out that this transmute was safe in all cases. Rust is smarter now, but we don't know if it is smart enough for all possible users of the macro.

robertbastian: well it works for all applications of this pattern in `yoke` (not through the macro), and for all test cases of the macro...

Manishearth: yes, I understand, that is why this is a "worry" not an "issue"

Manishearth: We don't know what changed in rustc to fix this, whether it was intentional or not.

Yoke touches some tricky edge cases of the compiler, it always has, and I'm generally wary of changes that make it more likely to trip up the compiler.

Like I said, it's fine, we can revert if it's a problem, I'm just worried

utils/tinystr/src/int_ops.rs
robertbastian: I don't want to spend the time to understand and prove the bit twiddling, this is preexisting and will stay like this (I don't like it either).

utils/yoke/derive/src/lib.rs
robertbastian: ok

utils/tinystr/src/int_ops.rs
Manishearth: I have a slight preference towards keeping the preexisting code, then, because this appears to move invariants around which is why I wanted to review it.

Manishearth:
Can we undo the `transmute_ptr_to_ptr` changes? Basically, I think `transmute::<&Foo, &Bar>` is clearer.

utils/tinystr/src/int_ops.rs
robertbastian: well I'm not going to do the work to partially revert this PR because of your slight preference. the previous unsafe code was relying on non-local invariants that were not stated, now at least the safety can be figured out by just looking at this function

robertbastian:
> Can we undo the `transmute_ptr_to_ptr` changes? Basically, I think `transmute::<&Foo, &Bar>` is clearer.

this PR is take it or leave it. pointer casts are better because you're forced to be explicit about types, and you can't mess up by dropping a `&`.

Manishearth:
I guess my worry with pointer casts is that it's easy to accidentally cast the wrong pointer.


I find the pointer casts to be far better than `transmute(x)`, but `transmute::<&T, &U>` is clearer than both. YOu can't mess up by dropping an `&` there because the sizes won't match. Pointer casts don't check the sizes, so you _can_ mess up by having an extraneous `&` somewhere, which happens all the time in Rust.

> this PR is take it or leave it

This just makes me inclined to leave it. I like most of these lints, but this PR merges fixes from various rather different lints into one hard-to-separate commit, and also includes a mix of automated/boring fixes vs tricky fixes that change internal API/etc.

Normally I don't care that much, but this is unsafe code. +131 -192 is nbd on a regular PR, but for unsafe code it's a fair amount. I spent the time reviewing that, found some issues, and highlighted them. I don't think it's a big change to revert, either, this is something you could probably find-and-replace with the right regex.

Manishearth:
If you don't want to do that I may do that myself, that's also fine.

robertbastian:
> I find the pointer casts to be far better than `transmute(x)`, but `transmute::<&T, &U>` is clearer than both.

the lint that enforces explicit types for transmutes doesn't actually seem to work, I've enabled it with the hope that it starts working in the future.

I also think some other lint doesn't like reference transmuting, I think if the reference is a slice, then `transmute_undefined_repr` fires

Manishearth:
> the lint that enforces explicit types for transmutes doesn't actually seem to work, I've enabled it with the hope that it starts working in the future.



Okay, that's a good enough reason to use the explicit pointer casts for now.

PR #7676, opened 2026-02-20
Correctly check the RD range for Date::try_from_fields and Date::try_add*

Manishearth:
Progress on https://github.com/unicode-org/icu4x/issues/7076

This does a bunch of things:

 - Adds `GENEROUS_YEAR_RANGE`, which is a year range that is slightly wider than `VALID_RD_RANGE` for all calendars, "good enough" for early checks.
 - Uses `GENEROUS_YEAR_RANGE` as a precondition in `try_from_fields`, and also checks `VALID_RD_RANGE` as a postcondition.
 - Tests for `try_from_fields`.
 - Adds (internal) `YearOverflowError` used for internal functions which can only error when out of year range
 - Changes (internal) `DateFieldsResolver::year_info_from_era` to `extended_year_from_era_year`. This splits its responsibilities more cleanly: it doesn't have to bother with generating a YearInfo (which is risky), it just adjusts the era year.
 - Adds a (internal) `DateFieldsResolver::year_info_from_extended_checked` which checks the generous range and uses it in  `from_fields` and arithmetic code.
 - Adds `DateAddError` based on consensus in https://github.com/unicode-org/icu4x/issues/7010.
 - Updates the fuzzer to remove guards around addition

If there are issues in DateAddError or the test, I would prefer to work on them separately, since this PR is getting pretty large.

## Changelog

For this PR, https://github.com/unicode-org/icu4x/pull/7062, and  https://github.com/unicode-org/icu4x/pull/7629

icu_calendar: Introduce year/date ranges to all APIs, documented on the APIs themselves. `Date` now has a fundamental range (ISO years between ±999,999), and most constructors enforce a stricter range of ±9999 years for input years.

### DIFF
--- a/components/calendar/src/calendar_arithmetic.rs
+++ b/components/calendar/src/calendar_arithmetic.rs
@@ -58,7 +58,7 @@ pub(crate) const GENEROUS_YEAR_RANGE: RangeInclusive<i32> = -1_040_000..=1_040_0
 /// must ensure they do not have overflows or panics for year (era year or extended year) within this range.
 ///
 /// This value is computed by going past the max value producible by adding a duration,
-/// which is preemptively capped at `GENEROUS_MAX_YEARS`, `GENEROUS_MAX_MONTHS, `GENEROUS_MAX_DAYS`
+/// which is preemptively capped at `GENEROUS_MAX_YEARS`, `GENEROUS_MAX_MONTHS`, `GENEROUS_MAX_DAYS`
 /// in `added()`.
 pub(crate) const SAFE_YEAR_RANGE: RangeInclusive<i32> = -8_000_000..=8_000_000;
 


components/calendar/src/calendar_arithmetic.rs
```diff
@@ -815,6 +815,27 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
         cal: &C,
         options: DateAddOptions,
     ) -> Result<Self, DateAddError> {
+        // We preemptively protect ourselves from overly-large values
+        //
+        // This is mostly needed for avoiding overflows on Duration arithmetic,
```
Manishearth, line 847: Open to different approaches here. It feels a bit strange to have all of this nice infrastructure range-checking years and we still need a coarse preemptive check because the duration math overflows.

One thing we could do is have `surpasses_inner` returning a YearOverflowError, and `surpasses()` returning a bool where the error is turned into `true`.

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -506,17 +510,18 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
...
-        let year_info = cal.year_info_from_era(
+        let extended_year = cal.extended_year_from_era_year(
             era.as_bytes(),
             range_check(year, "era_year", VALID_YEAR_RANGE)?,
         )?;
         // check the extended year in terms of the year
-        let offset = year - year_info.to_extended_year();
+        let offset = year - extended_year;
         range_check(
             year, // == year_info.to_extended_year() + offset
```
robertbastian: comment not correct anymore

components/calendar/src/error.rs
```diff
@@ -461,6 +461,40 @@ mod unstable {
...
+    /// including in SemVer minor releases. Do not use this type unless you are prepared for things to occasionally break.
+    ///
+    /// Graduation tracking issue: [issue #7161](https://github.com/unicode-org/icu4x/issues/7161).
+    /// </div>
+    ///
+    /// ✨ *Enabled with the `unstable` Cargo feature.*
+    #[derive(Debug, Copy, Clone, PartialEq, Display)]
+    #[non_exhaustive]
+    pub enum DateAddError {
+        /// The day is invalid for the given month.
```
robertbastian, line 480: add that this can only happen with `Overflow::Reject`

components/calendar/src/error.rs
```diff
@@ -461,6 +461,40 @@ mod unstable {
...
+        #[displaydoc("Invalid day for month, max is {max}")]
+        InvalidDay {
+            /// The maximum allowed value (the minimum is 1).
+            max: u8,
+        },
+        /// The specified month does not exist in this calendar.
```
robertbastian: how can we ever arrive at a month that's not in the calendar?

components/calendar/src/error.rs
```diff
@@ -461,6 +461,40 @@ mod unstable {
...
+        #[displaydoc("The specified month does not exist in this calendar")]
+        MonthNotInCalendar,
+        /// The specified month exists in this calendar, but not in the specified year.
```
robertbastian, line 510: add that this can only happen with `Overflow::Reject`

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -22,15 +22,32 @@ use core::ops::RangeInclusive;
...
+/// date types. Because this range exceeds the [`VALID_YEAR_RANGE`], only
+/// the valid year range is checked in most constructors.
+///
+/// This is the range used by `Date::from_rata_die`, `Date::try_from_fields`,
+/// and Date arithmetic operations.
 pub const VALID_RD_RANGE: RangeInclusive<RataDie> =
     calendrical_calculations::gregorian::fixed_from_gregorian(-999999, 1, 1)
         ..=calendrical_calculations::gregorian::fixed_from_gregorian(999999, 12, 31);
 
+/// We *must* ensure dates are within `VALID_RD_RANGE` some point before constructing them.
```
robertbastian, line 36: nit: as the first sentence, state what this is: a year range that exceeds to RD range for all calendars

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -22,15 +22,32 @@ use core::ops::RangeInclusive;
...
+///
+/// However, we may need to perform a fair amount of calendar arithmetic before
+/// getting to the point where we know if we are in range, and the calendar arithmetic
+/// is fragile (chance of math issues, slowness, debug assertions) at high ranges.
+///
+/// So we try to early-check year values where possible. We use a "generous" year range
+/// which is known to be wider than the valid year range for any era in any currently
```
```suggestion
/// which is known to be wider than the valid RD range for any era in any currently
```

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -22,15 +22,32 @@ use core::ops::RangeInclusive;
...
+/// supported calendar.
+///
+/// `VALID_RD_RANGE` maps to 1031332 BH..=1030050 AH in the Islamic calendars, which have
+/// the shortest years. We pick a slightly wider
```
robertbastian: half sentence

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -22,15 +22,32 @@ use core::ops::RangeInclusive;
...
+///
+/// The tests in `extrema.rs` ensure that all in-range dates can be produced here,
```
robertbastian: "produced here"? nothing is being produced here

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -324,23 +354,18 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
...
         // to return structural type errors (`NotEnoughFields`) before checking for range errors.
         // This isn't behavior we *must* have, but it is not much additional work to maintain
         // so we make an attempt.
         let year = match (fields.era, fields.era_year) {
             (None, None) => match fields.extended_year {
                 Some(extended_year) => {
-                    if !VALID_YEAR_RANGE.contains(&extended_year) {
+                    if !GENEROUS_YEAR_RANGE.contains(&extended_year) {
                         return Err(DateFromFieldsError::Overflow);
                     }
```
robertbastian: `year_info_from_extended_checked` already checks this

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -370,13 +395,14 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
...
-                if !VALID_YEAR_RANGE.contains(&era_year) {
+                if !GENEROUS_YEAR_RANGE.contains(&era_year) {
                     return Err(DateFromFieldsError::Overflow);
                 }
-                let year = calendar.year_info_from_era(era, era_year)?;
-                if !VALID_YEAR_RANGE.contains(&year.to_extended_year()) {
+                let extended_year = calendar.extended_year_from_era_year(era, era_year)?;
+                if !GENEROUS_YEAR_RANGE.contains(&extended_year) {
                     return Err(DateFromFieldsError::Overflow);
                 }
```
robertbastian: `year_info_from_extended_checked` already checks this

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -237,14 +269,17 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
...
         calendar: &C,
     ) -> Result<Self, DateError> {
-        let year = if let Some(era) = era {
-            let year = calendar
-                .year_info_from_era(era.as_bytes(), range_check(year, "year", VALID_YEAR_RANGE)?)?;
-            range_check(year.to_extended_year(), "era_year", VALID_YEAR_RANGE)?;
+        let extended_year = if let Some(era) = era {
+            let year = calendar.extended_year_from_era_year(
+                era.as_bytes(),
+                range_check(year, "year", VALID_YEAR_RANGE)?,
```
```suggestion
                range_check(year, "era_year", VALID_YEAR_RANGE)?,
```

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -237,14 +269,17 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
...
+            )?;
+            range_check(year, "era_year", VALID_YEAR_RANGE)?;
```
```suggestion
            range_check(year, "extended_year", VALID_YEAR_RANGE)?;
```

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -237,14 +269,17 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
...
             year
         } else {
-            calendar.year_info_from_extended(range_check(year, "extended_year", VALID_YEAR_RANGE)?)
+            range_check(year, "extended_year", VALID_YEAR_RANGE)?
         };
+        let year = calendar.year_info_from_extended(extended_year);
```
```suggestion
            year
        } else {
            year
        };
        let year = calendar.year_info_from_extended_checked(extended_year)?;
```

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -274,12 +309,7 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
         }
         let year = calendar.year_info_from_extended(year);
 
-        let month = calendar
-            .ordinal_from_month(year, month, Default::default())
-            .map_err(|e| match e {
-                MonthError::NotInCalendar => LunisolarDateError::MonthNotInCalendar,
-                MonthError::NotInYear => LunisolarDateError::MonthNotInYear,
-            })?;
+        let month = calendar.ordinal_from_month(year, month, Default::default())?;
```
robertbastian, line 366: nit: I don't like defining a `From` impl, that is then hidden in a `?` invocation, for a single call site. `MonthError` is private, so at least it's not a public impl, but the previous code was more readable imho

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -370,13 +395,14 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
                 },
             },
             (Some(era), Some(era_year)) => {
-                if !VALID_YEAR_RANGE.contains(&era_year) {
+                if !GENEROUS_YEAR_RANGE.contains(&era_year) {
                     return Err(DateFromFieldsError::Overflow);
                 }
```
robertbastian, line 448: do we actually need to check the era year *and* the extended year? I guess the era -> extended conversion could overflow? maybe define this precondition on `extended_year_from_era_year`

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -561,8 +608,11 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
         debug_assert!(u8::try_from(resolved_day).is_ok());
         let resolved_day = resolved_day as u8;
         // 1. Return the Record { [[Year]]: _resolvedYear_, [[Month]]: _resolvedMonth_, [[Day]]: _resolvedDay_ }.
-        // TODO: does not obey precondition
-        Self::new_unchecked(resolved_year, resolved_month, resolved_day)
+        Ok(Self::new_unchecked(
```
robertbastian: issue: the todo comment should be replaced by an explanation why the precondition is satisfied. afaik this is *not* actually currently guaranteed to be in `VALID_RD_RANGE`, only in `GENEROUS_YEAR_RANGE`

components/calendar/src/calendar_arithmetic.rs
robertbastian: I don't know this code enough to comment on that. However, if you're establishing an invariant here, propagate it in documentation to the relevant functions (I guess `add_year_to`, `month_from_ordinal`?)

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -750,27 +809,42 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
...
+        )?;
         // 1. Let _m0_ be MonthCodeToOrdinal(_calendar_, _y0_, ! ConstrainMonthCode(_calendar_, _y0_, _parts_.[[MonthCode]], _overflow_)).
         let base_month = cal.month_from_ordinal(self.year(), self.month());
-        let m0 = cal
-            .ordinal_from_month(
-                y0,
-                base_month,
-                DateFromFieldsOptions::from_add_options(options),
-            )
-            .map_err(|e| {
```
robertbastian, line 765: prefer this as well

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -782,24 +856,28 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
...
         // 1. Return _result_.
-        Ok(Self::new_balanced(
+        let balanced = Self::new_balanced(
             end_of_month.year(),
             i64::from(end_of_month.month()),
             duration.add_weeks_and_days_to(regulated_day),
             cal,
-        ))
+        )?;
+        // We early checked for a generous range of years, now we must check
```
robertbastian: ah, here we go. move this check into `new_balanced` as that calls `ArithmeticDate::new_unchecked` and violates its invariant

components/calendar/src/error.rs
Manishearth: It comes from MonthError. I'll just fix the mapping; technically it should never occur, I don't think it's important enough to try to programmatically prevent.

components/calendar/src/calendar_arithmetic.rs
Manishearth: Yes.

I'll add that.

components/calendar/src/calendar_arithmetic.rs
Manishearth: Yeah.

Unfortunately if we need all VALID_RD_RANGE dates to be fully functional under arithmetic we must have a wider range for `new_balanced`. I'll update the invariant on `new_unchecked`.

robertbastian: no don't update that invariant. that is the single place where the RD invariant exists

components/calendar/src/calendar_arithmetic.rs
Manishearth: Why? In this case the errors map cleanly. If the errors stop mapping cleanly we can introduce a new match the way we did before.

components/calendar/src/calendar_arithmetic.rs
robertbastian: well, not `new_unchecked` but the `ArithmeticDate` type, so `new_unchecked` needs to obey it

components/calendar/src/calendar_arithmetic.rs
robertbastian: this is less readable if you're trying to figure out how an error was produced. rust-analyzer cannot tell you which `?`s implicitly use the `From` implementation

components/calendar/src/calendar_arithmetic.rs
Manishearth: We cannot, `new_balanced` must be able to produce slightly-out-of-range years.

I can't see a way to make the code work without it. I documented this new invariant on `new_balanced`, and I'm about to push a commit that documents it more. I also updated the documentation on `new_unchecked` with the new invariant.

When I wrote the tests my core desired property was that for arithmetic should work as long as the input _and_ output were in RD range. The arithmetic algorithm needs to be able to temporarily "peek" out of range to do so, see the `end_of_month` calculation here (and basically all of the `surpasses` stuff used in `until`). So `new_balanced` needs to work for near-edge dates on either side of the range boundary, which means we cannot have it follow the strict RD range.

If we want to give up that property we need to more carefully figure out what the desired behavior is here. I find that property to be rather sensible and nice to have.

We could add `new_unchecked_with_assertion` that debug asserts in all of the other call sites that it is within RD range. But I don't really think it's necessary.

components/calendar/src/calendar_arithmetic.rs
Manishearth: I tweaked the code to just check extedned year directly.

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -22,15 +22,35 @@ use core::ops::RangeInclusive;
 pub const VALID_YEAR_RANGE: RangeInclusive<i32> = -9999..=9999;
```
robertbastian: let's call this `CONSTRUCTOR_YEAR_RANGE`, "valid" isn't really accurate anymore

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -219,7 +257,10 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
         C::YearInfo::unpack_day(self.0)
     }
 
-    // Precondition: the date is in the VALID_RD_RANGE
+    /// Precondition: the date is in `GENEROUS_YEAR_RANGE`.
+    ///
+    /// If giving this date to *users*, it must also be in `VALID_RD_RANGE`. `GENEROUS_YEAR_RANGE`
+    /// is fine to use for intermediate dates.
```
robertbastian: please don't do this. "If giving this date to \*users\*" is not something that can be locally checked. make `new_balanced` return `(i32, u8, u8)`, it's never used arithmetically anyway

also note that the *type invariant* is still stated on `ArithmeticDate` as "VALID_RD_RANGE contains the date"

Manishearth: Yeah, that works.

components/calendar/src/calendar_arithmetic.rs
Manishearth: That's just a general problem with Rust errors: that's an argument against using `?` period. These errors are going to go through `?` a half dozen more times before hitting the user. Changing that here won't do anything.

There are Rust error libraries that will attach context to `?` for you but they're heavier so we don't use them.

robertbastian: > That's just a general problem with Rust errors: that's an argument against using `?` period.

yep. doesn't mean it's not an argument here.

> These errors are going to go through ? a half dozen more times before hitting the user.

that's not our problem. they don't go through another `?` before being returned from our public API, and appearing in our unit tests where it's often needed to figure out where they're from

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -19,18 +19,38 @@ use core::hash::{Hash, Hasher};
 use core::ops::RangeInclusive;
 
 /// This is checked by constructors. Internally we don't care about this invariant.
```
```suggestion
/// This is checked by convenience constructors and `from_codes`. Internally we don't care about this invariant.
```

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -19,18 +19,38 @@ use core::hash::{Hash, Hasher};
...
+pub const CONSTRUCTOR_YEAR_RANGE: RangeInclusive<i32> = -9999..=9999;
 
 /// This is a fundamental invariant of `ArithmeticDate` and by extension all our
-/// date types. Because this range slightly exceeds the [`VALID_YEAR_RANGE`], only
-/// the valid year range is checked in constructors. Only the `Date::from_rata_die`
-/// constructor actually uses this, but for clamping instead of erroring.
-// This range is the tightest possible range that includes all valid years for all
-// calendars, this is asserted in [`test_validity_ranges`].
+/// date types. Because this range exceeds the [`CONSTRUCTOR_YEAR_RANGE`], only
+/// the valid year range is checked in most constructors.
```
```suggestion
/// date types. Constructors that don't check [`CONSTRUCTOR_YEAR_RANGE`] check this range.
```

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -81,6 +101,29 @@ impl<C: DateFieldsResolver> Hash for ArithmeticDate<C> {
     }
 }
 
+/// Same data as ArithmeticDate, but may be out of [`VALID_RD_RANGE`]
+#[derive(Debug)]
+pub(crate) struct UncheckedArithmeticDate<C: DateFieldsResolver> {
```
robertbastian: let's keep this private, I don't see a use case for this outside this module

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -139,15 +182,33 @@ pub(crate) trait DateFieldsResolver: Calendar {
 
     /// Converts the era and era year to a [`Self::YearInfo`]. If the calendar does not have eras,
     /// this should always return an Err result.
-    fn year_info_from_era(
+    ///
+    /// This function does not guard against overflow, please range-check input years
+    /// against [`CONSTRUCTOR_YEAR_RANGE`] or [`GENEROUS_YEAR_RANGE`] before passing to it.
```
robertbastian: this should be phrased in a way that works both for callers and implementers


```suggestion
    /// Precondition: `era_year` is in [`GENEROUS_YEAR_RANGE`]
```

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -139,15 +182,33 @@ pub(crate) trait DateFieldsResolver: Calendar {
...
         &self,
         era: &[u8],
         era_year: i32,
-    ) -> Result<Self::YearInfo, UnknownEraError>;
+    ) -> Result<i32, UnknownEraError>;
 
     /// Converts an extended year to a [`Self::YearInfo`].
     fn year_info_from_extended(&self, extended_year: i32) -> Self::YearInfo;
 
+    /// `year_info_from_extended` will debug assert if given a too-large year
```
robertbastian: can we actually be explicit about what the condition is? "too-large year value" is not

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -139,15 +182,33 @@ pub(crate) trait DateFieldsResolver: Calendar {
...
+    /// This function does not guard against overflow, please range-check input years
+    /// against [`CONSTRUCTOR_YEAR_RANGE`] or [`GENEROUS_YEAR_RANGE`] before passing to it.
+    fn extended_year_from_era_year(
         &self,
         era: &[u8],
         era_year: i32,
-    ) -> Result<Self::YearInfo, UnknownEraError>;
+    ) -> Result<i32, UnknownEraError>;
 
     /// Converts an extended year to a [`Self::YearInfo`].
```
robertbastian, line 204: where is this still used? what is the precondition?

components/calendar/src/calendar_arithmetic.rs
Manishearth: Wait, hold on, I already fixed this instance of this pattern because the errors do not map cleanly.. I thought you were talking about the other instance of this pattern.

But still, I disagree that we should avoid `?`. It's annoying in unit tests, but it's also annoying to wade through extra match statements.

robertbastian: well it's not an extra match statement, the `MonthError` to `LunisolarDateError` conversion only happens in once, let's just keep it there and not move it to another file

Manishearth: Okay, so you are talking about the other instance of this pattern.

It's an extra match statement within complicated code. Moving the match statement elsewhere is better. All of the calendar arithmetic/construction code is already messy.

I could introduce a `.map_err(MonthError::into)?` if you really want but I still think that is kind of silly. The Rust community has pretty universally settled on the style of using ? wherever it works and I don't plan on changing how I use ? unless you argue for and land a style rule for it.

components/calendar/src/cal/east_asian_traditional.rs
```diff
@@ -588,18 +588,18 @@ impl<R: Rules> DateFieldsResolver for EastAsianTraditional<R> {
...
-    ) -> Result<Self::YearInfo, UnknownEraError> {
+    ) -> Result<i32, UnknownEraError> {
         // This calendar has no era codes
         Err(UnknownEraError)
     }
 
     #[inline]
     fn year_info_from_extended(&self, extended_year: i32) -> Self::YearInfo {
-        debug_assert!(crate::calendar_arithmetic::VALID_YEAR_RANGE.contains(&extended_year));
+        debug_assert!(crate::calendar_arithmetic::GENEROUS_YEAR_RANGE.contains(&extended_year));
```
sffc: Thought: This check doesn't really achieve anything since GENEROUS_YEAR_RANGE is bigger than the RD range. I guess it's harmless

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -18,19 +18,39 @@ use core::fmt::Debug;
...
+/// which is known to be wider than the valid RD range for any era in any currently
+/// supported calendar.
+///
+/// `VALID_RD_RANGE` maps to 1031332 BH..=1030050 AH in the Islamic calendars, which have
+/// the shortest years. We pick a slightly wider range so that temporary intermediate
+/// values are not affected by this.
+///
+/// The tests in `extrema.rs` ensure that all in-range dates can be produced by the APIs that,
+/// use this check, and that these year numbers map to out-of-range values for every era.
+pub const GENEROUS_YEAR_RANGE: RangeInclusive<i32> = -1_040_000..=1_040_000;
```
sffc:
```suggestion
pub(crate) const GENEROUS_YEAR_RANGE: RangeInclusive<i32> = -1_040_000..=1_040_000;
```

components/calendar/src/tests/extrema.rs
```diff
@@ -51,32 +55,304 @@ super::test_all_cals!(
...
+            VALID_RD_RANGE.start().add(100),
+            VALID_RD_RANGE.start().add(10000),
+            RataDie::new(-1000),
+            RataDie::new(0),
+            RataDie::new(1000),
+            VALID_RD_RANGE.end().add(10000),
+            VALID_RD_RANGE.end().add(100),
+            VALID_RD_RANGE.end().add(5),
+            VALID_RD_RANGE.end().add(1),
+            *VALID_RD_RANGE.end(),
```
sffc: Issue: These are all overshooting the end of the range, so `from_rata_die` saturates them to all the same value (the max RD).

components/calendar/src/tests/extrema.rs
```diff
@@ -51,32 +55,304 @@ super::test_all_cals!(
...
+        };
+        let reject = DateFromFieldsOptions {
+            overflow: Some(Overflow::Reject),
+            ..Default::default()
+        };
+
+        // First we want to test that large values all get range checked
+        for era in [first_era, last_era, None] {
+            // We want to ensure that the "early" generous year range check
+            // AND the
```
sffc: Sentence fragment

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -229,22 +305,24 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
...
     ) -> Result<Self, DateError> {
-        let year = if let Some(era) = era {
-            let year = calendar
-                .year_info_from_era(era.as_bytes(), range_check(year, "year", VALID_YEAR_RANGE)?)?;
-            range_check(year.to_extended_year(), "era_year", VALID_YEAR_RANGE)?;
-            year
+        let extended_year = if let Some(era) = era {
+            calendar.extended_year_from_era_year(
+                era.as_bytes(),
+                range_check(year, "era_year", CONSTRUCTOR_YEAR_RANGE)?,
```
sffc, line 330: Issue: I think this one should check GENEROUS_YEAR_RANGE because it seems really weird for era year and extended year ranges to be double-checked. It results in a smaller effective range. This is especially noticeable in a calendar where the era and extended years are not the same, like Ethiopian with input era "aa"; only ~15000 dates can be constructed instead of ~20000 dates when the era and extended year are the same.

Alternatively, don't check the extended year range if the input was an era year.

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -139,15 +182,46 @@ pub(crate) trait DateFieldsResolver: Calendar {
 
     /// Converts the era and era year to a [`Self::YearInfo`]. If the calendar does not have eras,
     /// this should always return an Err result.
-    fn year_info_from_era(
+    ///
+    /// Precondition: `era_year` is in [`GENEROUS_YEAR_RANGE`].
+    fn extended_year_from_era_year(
```
sffc: Suggestion, here and elsehwere there is a precondition: name it `unchecked`

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -437,50 +497,54 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
...
             return Err(DateFromFieldsError::InvalidDay { max: max_day });
         };
-        // date is in the valid year range, and therefore in the valid RD range
+        let rd = C::to_rata_die_inner(year, month, day);
+
+        // We early checked for a generous range of years, now we must check
+        // to ensure we are actually in range for our core invariant.
+        if !VALID_RD_RANGE.contains(&rd) {
+            return Err(DateFromFieldsError::Overflow);
+        }
```
sffc, line 518: Issue: I don't love converting to RD just to check the RD range. I would rather have a calendar-specific check. This function is already generic in the calendar. (ok for follow-up)

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -437,50 +497,54 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
...
-        let year_info = cal.year_info_from_extended(range_check(year, "year", VALID_YEAR_RANGE)?);
+        let year_info =
+            cal.year_info_from_extended(range_check(year, "year", CONSTRUCTOR_YEAR_RANGE)?);
         // check the extended year in terms of the year
         let offset = year - year_info.to_extended_year();
         range_check(
             year, // == year_info.to_extended_year() + offset
             "year",
-            (VALID_YEAR_RANGE.start() + offset)..=(VALID_YEAR_RANGE.end() + offset),
+            (CONSTRUCTOR_YEAR_RANGE.start() + offset)..=(CONSTRUCTOR_YEAR_RANGE.end() + offset),
```
sffc, line 537: Issue (here and elsewhere, similar to above): I really don't like checking the year against `CONSTRUCTOR_YEAR_RANGE` multiple times

components/calendar/src/error.rs
```diff
@@ -478,18 +478,82 @@ mod unstable {
...
+        ///
+        /// ```
+        /// use icu::calendar::cal::Hebrew;
+        /// use icu::calendar::types::{DateDuration, Month};
+        /// use icu::calendar::Date;
+        /// use icu::calendar::error::DateAddError;
+        /// use icu::calendar::options::{DateAddOptions, Overflow};
+        ///
+        /// // Hebrew year 5784 is a leap year, 5785 is not.
+        /// // Adar I (the leap month) is month 5 in a leap year.
```
robertbastian, line 524: isn't it month 6?

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -503,7 +577,8 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
         //   1. Set _monthsInYear_ to CalendarMonthsInYear(_calendar_, _resolvedYear_).
         //   1. Set _resolvedMonth_ to _resolvedMonth_ + _monthsInYear_.
         while resolved_month <= 0 {
-            resolved_year = cal.year_info_from_extended(resolved_year.to_extended_year() - 1);
+            resolved_year =
+                cal.year_info_from_extended_checked(resolved_year.to_extended_year() - 1)?;
```
sffc: Issue: I thought we were going to make sure the input to this function (`new_balanced`) was such that we never cause calendar-specific integer overflow. A precondition should be that `year` is in the generous range, `ordinal_month` does not exceed the generous range converted to months, and `day` does not exceed the generous range converted to days.

components/calendar/src/calendar_arithmetic.rs
sffc: I think the "coarse preemptive check" should be the only check, until the final RD check at the end.

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -750,56 +840,82 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
...
+    ) -> Result<Self, DateAddError> {
+        // We preemptively protect ourselves from overly-large values
+        //
+        // This is mostly needed for avoiding overflows on Duration arithmetic,
+        // this could also be handled by updating DateDuration::add_foo_to()
+        // to return an error. However, that would lead to more error handling cluttering
+        // surpasses().
+        const GENEROUS_MAX_YEARS: u32 =
+            (*GENEROUS_YEAR_RANGE.end() - *GENEROUS_YEAR_RANGE.start()) as u32;
+        const GENEROUS_MAX_MONTHS: u32 = GENEROUS_MAX_YEARS * 13;
```
sffc, line 853: Issue: I think this needs to be doubled, because you can have a min date and add 2x generous months to get the max date.

sffc (changes requested):

components/calendar/src/cal/east_asian_traditional.rs
Manishearth: Yeah, this was added deliberately because previously it was quite haphazard, it's nice to have a check right before we call the tricky calendrical APIs.

components/calendar/src/calendar_arithmetic.rs
Manishearth: This is existing documented behavior (on from_codes). Separate discussion to change if desired.

components/calendar/src/calendar_arithmetic.rs
Manishearth: I don't really wish to maintain calendar-specific range checking code, it's too easy to get wrong.

components/calendar/src/calendar_arithmetic.rs
Manishearth: Preexisting behavior, often (but not always?) documented. @robertbastian introduced this I believe.

components/calendar/src/calendar_arithmetic.rs
Manishearth: This is already handled by GENEROUS_MAX_YEARS.

Manishearth:
Regarding the coarse presumtpive checks and the other checks:

My main worry is that YearInfo calculations should never get a year that is too large for them to handle, and should be able to preemptively assert that they get a good year. I don't want _calendar-specific_ code to need to get into the business of worrying too much about range. 

My first commit here introduced `SAFE_YEAR_RANGE`, which was "the year range that is safe for all internal arithmetic. I got rid of it, though, since it would need to be much larger to work in the arithmetic space. I don't like expanding the safe year range too much (we have previously seen fuzzer issues from years in the millions. This might have since been fixed.)

I also want to limit overly-slow APIs. https://github.com/unicode-org/icu4x/issues/7077 is the issue for fixing that, and I'm _hoping_ to work on it, but it's not a release blocker.

<s>I'm happy to make a new PR that reintroduces that change. It ought to simplify the code significantly. But I'd like to land this first.</s>

I made the change. It's a separate commit. If people later disagree about this architecture, please open a PR with a revert of that commit.

sffc (approved):

components/calendar/src/calendar_arithmetic.rs
sffc: Fine for follow-up. `to_rata_die` is slow and we otherwise don't need it. Maintaining a const calendar-specific min and max date where we use `Ord` to check if in-range is fine and better.

components/calendar/src/calendar_arithmetic.rs
robertbastian: I guess with the small constructor range it's fine to only check it once

components/calendar/src/calendar_arithmetic.rs
robertbastian: I don't like that you're completely ignoring me here. this change is not related to this PR, you're just changing code that I explicitly wrote this way to the style you prefer

Manishearth: (I wasn't. I was engaging with this topic on the other thread below)

components/calendar/src/calendar_arithmetic.rs
Manishearth: https://github.com/unicode-org/icu4x/issues/7686

components/calendar/src/calendar_arithmetic.rs
Manishearth: https://github.com/unicode-org/icu4x/pull/7687

PR #7679, opened 2026-02-23
Add more collator data and filtering to testdata; add transliterator attributes domain

robertbastian:
Currently some of the testdata is only not generated because some files are not downloaded to repo source data. However, this is not what clients can/should do, all of this should be controllable through options on `SourceDataProvider`, in particular marker attribute filters.

Running `make_testdata.rs` with `SourceDataProvider::new()` instead of `SourceDataProvider::new_testing()` should yield the same results, `new_testing` is convenient for avoiding the network, but should not behave differently.

### DIFF
--- a/components/decimal/src/provider.rs
+++ b/components/decimal/src/provider.rs
@@ -163,7 +163,7 @@ icu_provider::data_marker!(
     "decimal/digits/v1",
     [char; 10],
     #[cfg(feature = "datagen")]
-    attributes_domain = "numbering-system"
+    attributes_domain = "numbering_system"
 );
 
 #[cfg(feature = "datagen")]
--- a/provider/source/src/tests/make_testdata.rs
+++ b/provider/source/src/tests/make_testdata.rs
@@ -85,7 +85,7 @@ fn make_testdata() {
             "001" | "419" | "FR" | "JP" | "CG" | "HK" | "ZA"
         )
     })
-    .with_marker_attributes_filter("numbering-system", |attrs| {
+    .with_marker_attributes_filter("numbering_system", |attrs| {
         matches!(attrs.as_str(), "arab" | "beng" | "cakm" | "latn" | "thai")
     })
     .with_marker_attributes_filter("transliterator", |attrs| {


provider/source/src/tests/make_testdata.rs
```diff
@@ -85,6 +85,24 @@ fn make_testdata() {
             "001" | "419" | "FR" | "JP" | "CG" | "HK" | "ZA"
         )
     })
+    .with_marker_attributes_filter("numbering_system", |attrs| {
+        matches!(attrs.as_str(), "arab" | "beng" | "cakm" | "latn" | "thai")
```
Manishearth, line 89: TIL we filter this for testdata.

observation: we do not filter this for bakeddata, which is the correct behavior

components/decimal/src/provider.rs
```diff
@@ -157,13 +157,13 @@ icu_provider::data_marker!(
...
 icu_provider::data_marker!(
-    /// The digits for a given numbering system. This data ought to be stored in the `und` locale with an auxiliary key
+    /// The digits for a given numbering system. This data ought to be stored in the `und` locale with a marker attribute
     /// set to the numbering system code.
     DecimalDigitsV1,
     "decimal/digits/v1",
     [char; 10],
     #[cfg(feature = "datagen")]
-    attributes_domain = "numbering_system"
+    attributes_domain = "numbering-system"
```
sffc: Question: why did you change the casing? And in make_testdata.rs, you use `_`. I wasn't sure whether we should use `_` or `-` so I started using `_` on my newly added attribute domains since it seems that what we are using elsewhere.

provider/source/src/tests/make_testdata.rs
sffc: We have filters because it reduces the number of json files.

components/decimal/src/provider.rs
robertbastian: make_testdata was an oversight.

we use `-` in most ICU4X identifiers, I don't think we use `_` anywhere. `-` is easier to type and more pleasing to read

provider/source/src/tests/make_testdata.rs
robertbastian: we currently generate the numbering systems for all locales in the data. because we only have select locales in the testdata, this behaves differently between `new_testing` and `new`. should we actually filter the numbering systems by locales somehow? the code is: https://github.com/unicode-org/icu4x/blob/8b31bf168b494d1c4f2e65b65a8c10ba5e4aa9d0/provider/source/src/decimal/mod.rs#L102-L120

components/decimal/src/provider.rs
sffc: It's not just here, it is other places, too:

https://github.com/search?q=repo%3Aunicode-org%2Ficu4x%20attributes_domain&type=code

I don't want us to become inconsistent. I would rather make a separate PR to change all the instances at the same time, rather than switching just this one. (Please don't change all the others in this PR)

provider/source/src/tests/make_testdata.rs
sffc: I don't understand the question?

sffc (approved):

PR #7680, opened 2026-02-23
Update APIs used in time zone docs

robertbastian:
Instead of constructing `TimeZoneInfo` from BCP-47 subtags and offset strings, we should make our examples construct them from IANA IDs and offset seconds. Those are the actual pieces that clients will have from their TZ arithmetic code.


sffc (approved):
`TimeZone::from_iana_id` is good. I don't have a strong opinion between `"+09".parse()` and `try_from_seconds(9 * 3600)`. I have another open PR to change these, and I will need to rebase it.

robertbastian:
Well it's more likely that the offset, which is an arithmetic value, is available as a number than as a string in a particular format.

PR #7683, opened 2026-02-24
Feedback on LocaleNamesRegionLongV1

sffc:
Unsure if this is an improvement. I am opening this based on feedback from @robertbastian on #7662.

(this PR was generated with help from an AI)


components/experimental/src/displaynames/provider.rs
```diff
@@ -154,37 +154,11 @@ pub struct VariantDisplayNames<'data> {
...
-icu_provider::data_struct!(
-    SingleDisplayName<'_>,
-    varule: str,
-    #[cfg(feature = "datagen")]
-    encode_as_varule: |v: &SingleDisplayName<'_>| &*v.name
-);
-
 icu_provider::data_marker!(
-    /// Data marker for collation tailorings.
+    /// Data marker for display names.
```
robertbastian:
```suggestion
    /// Data marker for region display names.
```

sffc:
Since you both approved this I assume it means you believe that this is better than using the wrapper struct. Ok.

Manishearth:
> Since you both approved this I assume it means you believe that this is better than using the wrapper struct. Ok.

Slight preference to having fewer types to think about. Very slight, I don't think we should be thinking about this most of the time.

PR #7684, opened 2026-02-24
Speed up features CI

robertbastian:
Many of our crates have accumulated a lot of features. Restricting the feature combinations to 3 gives significant speed-ups:

```
icu_calendar: 76.41 -> 29.77
icu_capi: 349.21 -> 276.34
icu_collator: 16.82 -> 12.61
icu_collections: 11.11 -> 11.47
icu_datetime: 330.39 -> 105.51
icu_decimal: 25.40 -> 11.74
icu_experimental: 57.89 -> 38.23
icu_normalizer: 68.97 -> 12.49
icu_pattern: 14.57 -> 7.42
icu_plurals: 13.70 -> 10.25
icu_properties: 42.88 -> 19.42
icu_provider_export: 14.28 -> 10.48
icu_provider_source: 46.93 -> 34.05
icu_provider: 64.23 -> 26.58
icu_segmenter: 12.32 -> 8.43
icu_time: 28.41 -> 15.26
icu4x-datagen: 115.50 -> 113.88
litemap: 10.01 -> 8.02
tinystr: 11.41 -> 7.50
zerotrie: 60.98 -> 12.23
zerovec: 62.24 -> 36.91
```

### DIFF
--- a/components/calendar/Cargo.toml
+++ b/components/calendar/Cargo.toml
@@ -20,7 +20,7 @@ rust-version.workspace = true
 all-features = true
 
 [package.metadata.cargo-all-features]
-max_combination_size = 2
+max_combination_size = 3
 
 [dependencies]
 calendrical_calculations = { workspace = true }
--- a/components/collator/Cargo.toml
+++ b/components/collator/Cargo.toml
@@ -20,7 +20,7 @@ rust-version.workspace = true
 all-features = true
 
 [package.metadata.cargo-all-features]
-max_combination_size = 2
+max_combination_size = 3
 
 [dependencies]
 icu_collections = { workspace = true }
--- a/components/collections/Cargo.toml
+++ b/components/collections/Cargo.toml
@@ -20,7 +20,7 @@ version.workspace = true
 all-features = true
 
 [package.metadata.cargo-all-features]
-max_combination_size = 2
+max_combination_size = 3
 
 [dependencies]
 displaydoc = { workspace = true }
--- a/components/datetime/Cargo.toml
+++ b/components/datetime/Cargo.toml
@@ -103,7 +103,7 @@ bench = false  # This option is required for Benchmark CI
 
 [package.metadata.cargo-all-features]
 skip_optional_dependencies = true
-max_combination_size = 2
+max_combination_size = 3
 
 [[bench]]
 name = "datetime"
--- a/components/decimal/Cargo.toml
+++ b/components/decimal/Cargo.toml
@@ -20,7 +20,7 @@ version.workspace = true
 all-features = true
 
 [package.metadata.cargo-all-features]
-max_combination_size = 2
+max_combination_size = 3
 
 [dependencies]
 fixed_decimal = { workspace = true }
--- a/components/experimental/Cargo.toml
+++ b/components/experimental/Cargo.toml
@@ -23,7 +23,7 @@ independent = true
 all-features = true
 
 [package.metadata.cargo-all-features]
-max_combination_size = 2
+max_combination_size = 3
 
 [dependencies]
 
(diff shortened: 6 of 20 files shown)


robertbastian:
`icu_capi` takes so long, it should be it's own job. however, it's not currently possible to do that with `cargo-all-features`.

Manishearth:
The max combination size of 3 was chosen deliberately, since we have std and serde as common features, and I wanted to make sure all feature combos were tested with and without std/serde.

BUT I think 2 will also mostly work. We have had a couple cases where a very specific feature combo doesn't work. Unsure if they would have always been caught by this max combination size.


> icu_capi takes so long, it should be it's own job. however, it's not currently possible to do that with cargo-all-features.


CAF lets you chunk by feature sets instead of by packages, which would likely lead to better chunking where icu_capi gets fewer other packages in its chunk. This may have bad build cache locality though. Unsure.


Another way to do this would be to exclude icu_capi from CAF and have a separate task that runs CAF just on icu_capi, editing the exclude rule out of Cargo.toml to do so.

Manishearth:
With this PR it seems like features-1 takes 10min and the others take 5, 4, and 2 min. Hm; that's probably due to icu_capi.

Manishearth (approved):
Want confirmation from Shane that 2 is fine here. We can maybe use 3 otherwise? I'm pretty convinced 3 will be fine.

robertbastian:
> The max combination size of 3 was chosen deliberately, since we have std and serde as common features, and I wanted to make sure all feature combos were tested with and without std/serde.

We don't have std anymore usually.

> Another way to do this would be to exclude icu_capi from CAF and have a separate task that runs CAF just on icu_capi, editing the exclude rule out of Cargo.toml to do so.

I tried this, but it's not possible to exclude a crate from CAF.

> With this PR it seems like features-1 takes 10min and the others take 5, 4, and 2 min. Hm; that's probably due to icu_capi.

Yeah adding a fourth job doesn't do much if it's so unbalanced.

Manishearth:
> I tried this, but it's not possible to exclude a crate from CAF.

Yes it is, set `skip_package = true` in the config. I added this feature, ICU4X uses it already.

> Yeah adding a fourth job doesn't do much if it's so unbalanced.

Should we still do it then?

-------

Regarding "CAF lets you chunk by feature sets instead of by packages": Try using `--split-by-feature`: https://github.com/frewsxcv/cargo-all-features/pull/65. That will do a balanced chunk, with the tradeoff that some packages may get split across jobs.

robertbastian:
> Yes it is, set `skip_package = true` in the config. I added this feature, ICU4X uses it already.

That'd be global, but we need this per invocation. If we set `skip_package = true`, then we cannot run CAF on `icu_capi` at all.

robertbastian:
cool this makes features-2 be icu_capi only

Manishearth:
> That'd be global, but we need this per invocation. If we set `skip_package = true`, then we cannot run CAF on `icu_capi` at all.

yes, that is why I said "And have a separate task that runs CAF just on icu_capi, editing the exclude rule out of Cargo.toml to do so." when I suggested this strat. It's not a great strat. It's an option.

Manishearth:
7, 7, 4, 4, not bad.

robertbastian:
sorry I missed that, editing Cargo.toml files in CI is just not something I want to do

Manishearth:
Same. Either way, the chunking strategy flag appears to work.

Manishearth (approved):
LGTM with no concerns for max size 3

Manishearth:
(Though I assume you're doing that to compare numbers and haven't decided which you like yet)

robertbastian:
11, 10, 5, 5, I guess that's fine still

PR #7685, opened 2026-02-24
Remove `UtcOffset` range invariant

robertbastian:
We currently have an invariant on `UtcOffset` that it's +-18:00:00. I don't know where this comes from but
* we don't use it anywhere. for formatting, we can format any offset, and for the `TimeZonePeriodsV1` data marker's custom encoding, we only support values in TZDB anyway
* it's not sufficient for Temporal IXDTF. there's currently a test that asserts that the string with the maximum valid offset produces a failure – that doesn't make sense

## Changelog
`icu_calendar`: Remove `UtcOffset` range invariant

### DIFF
--- a/components/time/src/time_crate.rs
+++ b/components/time/src/time_crate.rs
@@ -30,7 +30,7 @@ impl From<time::PrimitiveDateTime> for DateTime<Gregorian> {
 
 impl From<time::UtcOffset> for UtcOffset {
     fn from(other: time::UtcOffset) -> Self {
-        UtcOffset::from_seconds_unchecked(other.whole_seconds())
+        UtcOffset::from_seconds(other.whole_seconds())
     }
 }
 


robertbastian:
We don't serialize these. Formatting handles this alright: https://github.com/unicode-org/icu4x/blob/500a9e43d4231567e0197b6091ca0ac6db3ffeba/components/datetime/src/format/time_zone.rs#L355

sffc (changes requested):
I think the limit exists because the invariant allows us to pack it, which we do (or have done) in various places.

For a while we had infallible `from_eighths_of_hour`. Not sure what happened to it.

If Temporal requires +/- 24, then let's do that, so long as it doesn't break anything. I don't think I support removing the bound entirely. Restricting the private fields of structs is almost always a good thing.

robertbastian:
> I think the limit exists because the invariant allows us to pack it, which we do (or have done) in various places.

I listed the use cases of this invariant in the OP. For packing, the 18h invariant does not help, it requires the offset to be x:00, x:10, x:15, x:20, x:30, x:40, x:45, x:50, or 00:44:30, where, yes, x < 18, but it's fallible anyway and we rely on the fact that TZDB doesn't contain x > 18, we don't need this as an invariant on the `UtcOffset` type.

> If Temporal requires +/- 24, then let's do that, so long as it doesn't break anything. I don't think I support removing the bound entirely. Restricting the private fields of structs is almost always a good thing.

Temporal requires +/-24, but other systems will have other requirements. The `chrono` and `jiff` conversion code currently uses `from_seconds_unchecked`, which is not ideal, but the conversion shouldn't be fallible because we can actually format arbitrary offsets.

sffc:
I can't get `Option<(i16, u8)>` to fit in 4 bytes, but I can get `Option<(i16, u8, bool)>` to fit. 🤷 

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=fb7bb35d9ac2c407087c7ec0e2dca841

robertbastian:
Using an `i32` is nice because everyone uses `i32` and this way there's no fallibility when converting:

* https://docs.rs/chrono/latest/chrono/struct.FixedOffset.html#method.local_minus_utc
* https://docs.rs/jiff/latest/jiff/tz/struct.Offset.html#method.seconds

We can still have a niche optimisation in the future by saturating the `i32::MIN` bit pattern to `i32::MIN + 1` or something.

sffc:
What if we made it infallible but saturated at a reasonable value, like `23:59:59` or `99:59:59`

sffc (commented):
> Using an `i32` is nice because everyone uses `i32` and this way there's no fallibility when converting:
> 
> * https://docs.rs/chrono/latest/chrono/struct.FixedOffset.html#method.local_minus_utc
> * https://docs.rs/jiff/latest/jiff/tz/struct.Offset.html#method.seconds

They return `i32` but their constructors do the range checking, too (at least chrono does)

https://docs.rs/chrono/latest/src/chrono/offset/fixed.rs.html#62-68

robertbastian:
Yes, they do range checking because they need the invariant for doing arithmetic. We don't need the invariant for our formatting input type, we can format anything.

robertbastian:
This is meant to be a cleanup because the current jiff, chrono, and time integrations break the documented invariant. But if you'd rather have that, I will close this.

PR #7687, opened 2026-02-24
Undo my change to lunisolar error matching

Manishearth:
In https://github.com/unicode-org/icu4x/pull/7676 I moved lunisolar errors over to a `?` operator.

Robert disagrees: https://github.com/unicode-org/icu4x/pull/7676#discussion_r2841888168

I landed that PR in the interest of unblocking future work, but I don't mind undoing it. I still have a slight preference to not have error matches in this (rather large) file, but I don't care enough in this case since it's a simple function.

### DIFF
--- a/components/calendar/src/calendar_arithmetic.rs
+++ b/components/calendar/src/calendar_arithmetic.rs
@@ -363,7 +363,11 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
         }
         let year = calendar.year_info_from_extended(year);
 
-        let month = calendar.ordinal_from_month(year, month, Default::default())?;
+        let month = match calendar.ordinal_from_month(year, month, Default::default()) {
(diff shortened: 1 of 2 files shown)


Manishearth:
Also, for context as to why I changed this: I had a strong motivation to make this change in the code I _was_ touching, since the arithmetic code is quite messy and I didn't want it to be messier with error matches. While doing that, I noticed the other instance of this pattern, and proactively cleaned it up. I then undid my other change since the match became no-longer-obvious.

I think "this is the only place this error is generated so we should consolidate it" is a decent argument.

In general I am opposed to using local `match` over `?`: 99% of the time I do not want error handling code to be visible. But it's fine here.

Manishearth:
Also filed https://github.com/rust-lang/rust-analyzer/issues/21700 about `?` in rust-analyzer.

PR #7688, opened 2026-02-24
Reduce year range checks

### LINKED ISSUE #7686: Constructor year range checking should be for *input* range (only)
https://github.com/unicode-org/icu4x/pull/7676 made progress on #7076 by making every ICU4X API do _some_ range checking.

In the process, Shane, [noted](https://github.com/unicode-org/icu4x/pull/7676#discussion_r2843401735) that we year-range-check twice in most constructors: for the given year and the extended year.

This was introduced in https://github.com/unicode-org/icu4x/pull/7219 and documented in #7634 . 

Now that our constructor year ranges are  much smaller, we can probably just range check the input number and not double range checking. From the pov of the user it's always nice to get errors talking about values they have given us (not internal values)

We should fix this, unless people see a strong reason to dobule range check.

(Assigning to Robert since he wrote most of the docs here over the course of multiple PRs, and the tricky thing here is fixing up all the docs)

robertbastian:
Fixes #7686

## Changelog: N/A

(covered by other range check PR)

### DIFF
--- a/components/calendar/src/calendar_arithmetic.rs
+++ b/components/calendar/src/calendar_arithmetic.rs
@@ -7,7 +7,7 @@ use calendrical_calculations::rata_die::RataDie;
 use crate::duration::{DateDuration, DateDurationUnit};
 use crate::error::{
     range_check, DateAddError, DateFromFieldsError, EcmaReferenceYearError, LunisolarDateError,
-    MonthCodeParseError, MonthError, UnknownEraError, YearOverflowError,
+    MonthCodeParseError, MonthError, UnknownEraError,
 };
 use crate::options::{DateAddOptions, DateDifferenceOptions};
 use crate::options::{DateFromFieldsOptions, MissingFieldsStrategy, Overflow};
@@ -121,13 +121,13 @@ struct UncheckedArithmeticDate<C: DateFieldsResolver> {
 }
 
 impl<C: DateFieldsResolver> UncheckedArithmeticDate<C> {
-    fn into_checked(self) -> Result<ArithmeticDate<C>, YearOverflowError> {
+    fn into_checked(self) -> Option<ArithmeticDate<C>> {
         let rd = C::to_rata_die_inner(self.year, self.ordinal_month, self.day);
         if !VALID_RD_RANGE.contains(&rd) {
-            return Err(YearOverflowError);
+            return None;
         }
         // Invariant checked above
-        Ok(ArithmeticDate::new_unchecked(
+        Some(ArithmeticDate::new_unchecked(
             self.year,
             self.ordinal_month,
             self.day,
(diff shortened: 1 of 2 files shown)


Manishearth:
fmt

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -865,9 +840,11 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
 
         // 1. Let _parts_ be CalendarISOToDate(_calendar_, _isoDate_).
         // 1. Let _y0_ be _parts_.[[Year]] + _duration_.[[Years]].
-        let y0 = cal.year_info_from_extended_checked(
-            duration.add_years_to(self.year().to_extended_year()),
-        )?;
+        let extended_year = duration.add_years_to(self.year().to_extended_year());
+        if !GENEROUS_YEAR_RANGE.contains(&extended_year) {
+            return Err(DateAddError::Overflow);
+        }
```
sffc, line 846: Issue: I don't like the double-checking here. @Manishearth removed it from the first PR which made me happy.

Manishearth: (I hadn't removed this case because in my PR I still had the `_checked` function everywhere protecting YearInfo constructions. I still _kind_ of prefer redundant checks but not much. Not opposed to further cleanup here.)

robertbastian: this is the inlined `year_info_from_extended_checked`, I'm not adding this check, but I'm also not removing it because we need it in order to be allowed to call `year_info_from_extended`

Manishearth: The whole add code starts with a range check that should obviate the need for a check here, in theory.

Personally I prefer double checking because it's an invariant and I disprefer invariants being upheld by complex math, but I won't block a chance to remove the check.

robertbastian: it checks the duration that's being added to exit early if it's entirely impossible that that duration will keep the date in bounds, but if the date itself is close to the boundary, even a small duration can make the year exceed the bounds

Manishearth: The model changed as a result of discussion: see https://github.com/unicode-org/icu4x/pull/7676#issuecomment-3947896410. I switched back to my original plan.

That is, there is an internal `SAFE_YEAR_RANGE`which is our _internal_ absolute invariant that we assert on and design math for. During arithmetic, we are allowed to escape the `VALID_RD_RANGE` as long as we are still in the `SAFE_YEAR_RANGE`.

The API should never produce an `ArithmeticDate` that is outside of the RD range, but it may produce YMD values, including YearInfos, in that range.

So the "add duration at the boundary" doesn't actually have problems.

robertbastian: hmm that seems counterproductive, because then you'll do a lot of potentially very expensive arithmetic even though you could have figured out very easily that you're already out of range. if the year is out of range here, the smaller units cannot make it come back into range

Manishearth: Yes, that was one of the reasons why I prefer the double checking.

PR #7689, opened 2026-02-24
Add more docs on Date's range errors

Manishearth:
Fixes https://github.com/unicode-org/icu4x/issues/7076 (I think everything else there is resolved)

Decided to add a nice section on `Date` that the other docs can all point to. I chose to publicize our range but clearly document it as subject to change within limits.

We _may_ wish to publicly expose `VALID_RD_RANGE`.

## Changelog: N/A

(covered on other PR)

### DIFF
--- a/components/calendar/src/date.rs
+++ b/components/calendar/src/date.rs
@@ -103,6 +103,25 @@ impl<C> Deref for Ref<'_, C> {
 /// 3. From a RFC 9557 string via [`Self::try_from_str()`]
 /// 4. From a [`RataDie`] via [`Self::from_rata_die()`]
 ///
+/// # Date ranges
+///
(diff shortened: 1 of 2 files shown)


Manishearth:
I want to write more tests, but the current tests are sufficient imo.

We have a few "out of range" tests that we generate by nudging a duration out of range. It's an okay tactic, could do more interesting things there.

I went through ctors when doing this. They all seemed documented.

Manishearth:
> could do more interesting things there.

(This is mostly blocked on blazing fast until. I don't know if we'll have that soon. I'll file a followup though)

Manishearth:
https://github.com/unicode-org/icu4x/issues/7695

PR #7697, opened 2026-02-25
Add RegionDisplayName that directly implements Writeable

sffc:
#3260 / #3913

## Changelog

icu_experimental/displaynames: Add RegionDisplayName for loading a single display name
- New type: `RegionDisplayName`

### DIFF
--- a/components/experimental/src/displaynames/displaynames.rs
+++ b/components/experimental/src/displaynames/displaynames.rs
@@ -613,8 +613,10 @@ fn test_region_load_long() {
     use icu_locale_core::subtags::region;
     use writeable::assert_writeable_eq;
 
-    let mut prefs = DisplayNamesPreferences::default();
-    prefs.locale_preferences = (&locale!("en-001")).into();
+    let prefs = DisplayNamesPreferences {
+        locale_preferences: (&locale!("en-001")).into(),
+        ..Default::default()
+    };
     let data = RegionDisplayName::try_new(prefs, region!("AE")).unwrap();
     assert_writeable_eq!(data, "United Arab Emirates");
 }


sffc:
Is this roughly what you envision for the one-at-a-time display names API?

components/experimental/src/displaynames/displaynames.rs
```diff
@@ -88,11 +110,58 @@ impl RegionDisplayNames {
         .or_else(|| data.names.get(&region.to_tinystr().to_unvalidated()))
         // TODO: Respect options.fallback
     }
+
+    #[doc = icu_provider::gen_buffer_unstable_docs!(UNSTABLE, Self::load_long)]
+    pub fn load_long_unstable<D: DataProvider<LocaleNamesRegionLongV1> + ?Sized>(
```
robertbastian: why does this live on `RegionDisplayNameS`? please replace the old API instead of keeping both around to untangle later

components/experimental/src/displaynames/displaynames.rs
```diff
@@ -534,3 +603,14 @@ fn test_language_display() {
...
+
+#[test]
+fn test_region_load_long() {
+    use icu_locale_core::locale;
+    use icu_locale_core::subtags::region;
+
+    let mut prefs = DisplayNamesPreferences::default();
+    prefs.locale_preferences = (&locale!("en-001")).into();
+    let data = RegionDisplayNames::load_long(prefs, region!("AE")).unwrap();
+    assert_eq!(data.to_string(), "United Arab Emirates");
```
robertbastian:
```suggestion
    assert_writeable_eq!(data, "United Arab Emirates");
```

components/experimental/src/displaynames/displaynames.rs
sffc: done

components/experimental/src/displaynames/displaynames.rs
sffc: done

sffc:
Discussion: Most/all formatter outputs that implement Writeable borrow from somewhere. With this design, the RegionDisplayName owns its payload. If we add other ways to load RegionDisplayName that borrow instead, we'll need another type.

components/experimental/src/displaynames/displaynames.rs
```diff
@@ -46,6 +46,78 @@ pub struct RegionDisplayNames {
...
+                    &locale,
+                ),
+                ..Default::default()
+            })?
+            .payload;
+        Ok(Self { payload })
+    }
+}
+
+impl writeable::Writeable for RegionDisplayName {
```
Manishearth, line 105: heh, that's interesting

robertbastian:
> Perhaps we should add an `as_str()` API as well, and the Writeable impl is there for consistency

I would rather not, this boxes us in representation-wise. In the future, we might have things like "Korea (North)", "Korea (South)" that are assembled from parts. Also, `LocaleDisplayName` won't be able to do `as_str()` at all.

components/experimental/src/displaynames/displaynames.rs
```diff
@@ -52,6 +52,55 @@ pub struct RegionDisplayName {
...
+        ///
+        /// let mut prefs = DisplayNamesPreferences::default();
+        /// prefs.locale_preferences = (&locale!("en-001")).into();
+        /// let display_name = RegionDisplayName::try_new(prefs, region!("AE"))
+        ///     .expect("Data should load successfully");
+        ///
+        /// assert_eq!(display_name.to_string(), "United Arab Emirates");
+        /// ```
+        functions: [
+            try_new,
```
robertbastian, line 77: please add all constructors

Manishearth: Not sure which other constructors belong here?

Manishearth (approved):
I think we should have this in our experimental release.

PR #7700, opened 2026-02-25
Use `YearNames::FixedEras` for Japanese

robertbastian:
I've marked this API breaking because it breaks new-data-old-code. We either wait until 3.0 or introduce a new data marker, that might be worth it for a 26kB size reduction.


Manishearth:
I like the size reduction. I think we have had _some_ discussion about new-data-old-code before: where this kind of breakage is fine, but maybe worth waiting multiple releases for.

This is a thing whenever we add variants to provider types, which is not infrequent, and we usually don't seem to worry too much about new-data-old-code then.

Manishearth:
https://github.com/unicode-org/icu4x/blob/main/documents/process/data_versioning.md#objectives



> Older code should be able to read newer data files.
>  - For example, an app that is a few years old should be able to consume an ICU4X data file built with the latest features and CLDR version.
>  - Old code should continue to work with its existing functionality, but it doesn't need to automatically support newer features.

...

> IV. Aspects of Data File Versioning
>
> There are several aspects of a data file that could undergo version updates:
> ...
> 2. Additional variants added to existing data structs (example: PropertyCodePointSetV1)
> ...
> Case 2: Depending on the minimum version specified in datagen, the newest backwards-compatible variant can be generated in the output data.

Manishearth:
I think the Case 2 stuff is relevant, but the policy refers to "minimum supported version from datagen", which isn't something we really do?

Probably makes sense for us to have a best-effort "compatible-with" flag for datagen (that may even error if asked for things that are fundamentally not hackable)

provider/source/data/debug/datetime/names/year/japanese/v1/3/en.json
```diff
@@ -1,22 +1,11 @@
...
-      "Taishō"
-    ]
+  "FixedEras": [
+    "BC",
+    "Meiji",
+    "",
+    "Taishō",
+    "Shōwa",
+    "Heisei",
+    "Reiwa"
```
sffc: Shouldn't it be BC, AD, Meiji, Taisho, ... ? Where is AD

sffc (commented):
As far as old-code-new-data is concerned, perhaps we could consider exporting the era map by default in postcard?

We should update/clarify the guidelines on old-code-new-data and new-code-old-data. I think we could say that we only support this for buffer provider data.

For a change like this that makes data a bit smaller / more efficient but doesn't unlock any new features, I'm inclined to just keep postcard using the era map until 3.0. I don't see the necessity of a flag.

provider/source/data/debug/datetime/names/year/japanese/v1/3/en.json
robertbastian: good catch

Manishearth:
> As far as old-code-new-data is concerned, perhaps we could consider exporting the era map by default in postcard?



Should we be adding a `compatible-with` flag to datagen like the policy assumes we have?

> We should update/clarify the guidelines on old-code-new-data and new-code-old-data. I think we could say that we only support this for buffer provider data.


I agree.

sffc:
I think we should keep icu4x-datagen generating postcard data compatible with the start of the major version until we have a good reason to change it.

Defensive Clarification: this comment is intended as "this is the easy thing to do because it needs fewer flags and branches", not "shane is decreeing something harder to do"

Manishearth:
That's a reasonable stance. I'm worried about test coverage though.

My impression is that we have _already_ been adding new variants to data quite often, and using it. We're considering doing that in https://github.com/unicode-org/icu4x/pull/7666. I _think_ we have done this type of thing before; adding a variant to a data struct enum and using it where we previously didn't. Maybe I'm wrong?

Given my impression of us doing this semi-regularly, I'm a little reluctant to do it in a way that reduces test coverage every time we do it. This might be resolved via universally all-providers'ing the tests, but there are other options available.


Like I said, I could be wrong about this being a semi-regular occurrence. If it isn't, a policy where we say "postcard must be compatible with the first version the marker is introduced with" AND "changing data in a way that breaks new-data-old-code requires thorough compat tests" would be fine.

"thorough compat tests" in the case of e.g. a datetime marker would basically mean _most_ datetime formatting tests are run with the old data. There's a lot of subtle interactions in datetime formatting especially around anything indexed.

## Preferred outcome


I think my _preferred_ outcome here is:

 - We add a `--compatible-with` flag. This flag generates compat data for relevant markers when asked.
   -  I'm not sure what it should do when run with markers that do not exist in the compatible version: in theory it should error, in practice this would mean we need to tag every marker with a version. That's tedious, and silently succeeding is not too bad here.
 - We declare `--compatible-with` as _best effort_. This means we do not extensively test the output (ideally we test it a little).

This makes it easy to switch to new variants when we want; we just need to pass through a flag during exporting, and if we break the old data format in a subtle way we can patch it.

I recognize that this is a nontrivial amount of work, but I also think the level of testing I'm hoping for otherwise is a nontrivial amount of work, so I think if we want to make this type of change then we are locked in to a nontrivial amount of work either way. Maybe people disagree with me about the amount of testing we should have.


## Appendix: Provider agnostic testing

I think there are a couple ways of doing this. One way is to have [a `test_all_providers!()` macro](https://github.com/boa-dev/temporal/blob/614f3d0a49a2838f1be958d2d6b1c563efc93343/src/builtins/core/zoned_date_time/tests.rs#L57). In regular builds it simply tests baked providers, but in a special CI build it could detect an env var and also test a postcard build. (I'm assuming we don't ever plan to check in a massive postcard file: we can generate it on demand in CI)


There are three downsides:

 - Doesn't work for doctests
 - A lot of churn for our tests: we have to switch every compiled API over to its provider version, which is hard to automate
 - This isn't just "testing with different data", this is _testing different APIs_. Baked data APIs are our primary APIs and I like that most of our tests use them. As ICU4X devs we experience our default ergonomics when we write tests and often notice API issues that way. Using the provider APIs will make it worse.

Another option is to have CI generate compat data _for baked_, and run tests with it, with the ICU4X_DATA_DIR override. I think that's probably easier

components/datetime/src/provider/names.rs
```diff
@@ -247,6 +247,52 @@ pub enum YearNames<'data> {
...
+            VariableEras(&'a YearNamesMap<'a>),
+            Cyclic(&'a VarZeroVec<'a, str>),
+        }
+
+        let x: YearNamesMap;
+
+        match self {
+            // Japanese eras are now generated as `FixedEras`, but we want to keep serializing
+            // them as VariableEras. They are the only calendar with 7 eras.
+            Self::FixedEras(e) if e.len() == 7 => {
```
sffc, line 274: `e.len() == 7` seems like a not good heuristic, hmm

robertbastian: why not? no other calendar is going to reach 7 eras anytime soon, and it's all we have

Manishearth: I think this is fine and we should perhaps file 3.0 blocking issues for both of these new-data-old-code hacks

Manishearth (approved):
I'm still worried about the testing. The approach here seems fine, though.

Maybe we should have a followup about testing new-data-old-code? I have posted some ideas above.

robertbastian:
@sffc are you planning to review this any further?

robertbastian:
> Please explain the choice to put the backwards compatibility stuff in `impl Serialize` instead of directly in datagen as it was before.
>
> Is it because we want impl Serialize and impl Bake to do different things?

The whole point of this PR is to use the smaller variant. Then you asked to not use it for serde.

> Do you have a reason why impl Serialize is the custom impl instead of impl Bake?

Yes, because it's easier to clean up when we can remove this backward compatibility logic.

> Is this a style that you anticipate we use regularly moving forward to solve similar types of issues?

I'm solving this issue, nothing more.

Manishearth:
Skipping review since I'm also unclear on what the discussion ended up at because I thought this was the plan.

robertbastian:
you previously approved and I only fixed a merge conflict

robertbastian:
@sffc you approved the same pattern in #7666. It's very tiring how you drag your feet on every single one of my PRs, I responded to your questions within 20 minutes, and now it's been 15 hours without a reaction from you.

sffc:
> @sffc you approved the same pattern in #7666.

I did not realize that I approved the same pattern in #7666. I have left some more comments on that PR.

Manishearth:
> you previously approved and I only fixed a merge conflict

Yes, and since then it seemed like the situation got more complex, so I wanted to wait until it was figured out. r+ either way

Manishearth:
To be clear, when I said "I thought this was the plan" was me referring to how you had done it.

PR #7702, opened 2026-02-25
zerovec: niche-optimize Result for UleError and VarZeroVecFormatError

### LINKED ISSUE #2801: Measure impact of changing ZeroVecError to start at 1
@kupiakos found that making error enums start at 1 can help code size:

https://twitter.com/kupiakos/status/1587506293996036097

We have a lot of cases of `Result<(), ZeroVecError>` throughout the code base. We should try to apply this optimization to `ZeroVecError` and measure the impact.

afrdbaig7:
This PR adds niche optimization for Result<(), UleError> and Result<(), VarZeroVecFormatError> as asked in #2801. In a few places in the codebase, we return Result<(), E>, and when the error type E has an unused bit pattern (a niche), Rust can optimize the representation so that Ok(()) fits in the niche. This makes Result<(), E> the same size as E> instead of needing an extra spot for a discriminant. To achieve this, I added #[repr(u8)] to both UleError and VarZeroVecFormatError and specified their first variants (InvalidLength and Metadata, respectively) to have a discriminant of 1, leaving 0 for Ok(()). I also added compile-time const assertions to make sure that size_of::<Result<(), E>>() == size_of::<E>(), which helps prevent regressions in the future. The documentation comments were updated to explain the reasoning behind these representations and discriminant values. I tested this by running cargo check -p zerovec --all-features, cargo test -p zerovec --all-features (all tests passed), cargo fmt, and cargo clippy with no new warnings.
 This closes #2801.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -148,6 +148,7 @@ Fully filled in up to 30c187f4b7
         - Add `ZeroAsciiDenseSparse2dTrie` for more efficient storage of data keys with many attributes (unicode-org#7264, unicode-org#7304, unicode-org#7305)
     - `zerovec`: `0.11.5 -> ???`
         - `schemars` support (unicode-org#7209)
+        - Niche-optimize `Result<(), UleError>` and `Result<(), VarZeroVecFormatError>` (#2801)
     - `zerovec`: No change (`0.11.2`)
     - `zoneinfo64`: `0.2.1 -> ???`
         - Internal cleanups


afrdbaig7:
> No, this makes the enums bigger in size.
> 
> Alyssa's tweet was about fieldless enums. The issue is incorrect.

@Manishearth thanks for clarifying this for me; I see now how data in UleError can be aligned differently resulting in a total increase in size of 24 to 32 bytes with the addition of #[repr(u8)] forcing the discriminant into its own alignment slot. I initially thought that #2801's intent was to affect the layout of any enum but it actually only affects fieldless error enums and allows for niche optimizations that can lead to size benefits.

PR #7703, opened 2026-02-26
Update Diplomat to new opaque_mut annotations

Manishearth:
https://github.com/rust-diplomat/diplomat/pull/1065 made Diplomat start enforcing `opaque_mut`. Diplomat doesn't use this to check anything yet, but it's step 1 of the process in making FFI sound.

### DIFF
--- a/ffi/capi/src/timezone.rs
+++ b/ffi/capi/src/timezone.rs
@@ -283,7 +283,7 @@ pub mod ffi {
         #[diplomat::rust_link(icu::time::zone::TimeZoneVariant, Enum, compact)]
         #[allow(deprecated)]
         pub fn infer_variant(
-            &mut self,
+            &self,
             _offset_calculator: &crate::unstable::variant_offset::ffi::VariantOffsetsCalculator,
         ) -> Option<()> {
             Some(())
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -746,7 +746,7 @@ dependencies = [
 [[package]]
 name = "diplomat"
 version = "0.14.0"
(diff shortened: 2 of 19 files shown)


ffi/capi/src/timezone.rs
```diff
@@ -283,7 +283,7 @@ pub mod ffi {
         #[diplomat::rust_link(icu::time::zone::TimeZoneVariant, Enum, compact)]
         #[allow(deprecated)]
         pub fn infer_variant(
-            &mut self,
+            &self,
```
Manishearth, line 286: this won't change FFI in any language except C/C++ and in those this is strictly more permissive, so it should compile barring type introspection things

Manishearth:
Diplomat does enforce these now. It just doesn't _use_ them yet, that's the next step.

I'm not yet sure if I'm going to have the enforcement in the same Diplomat release, since these are disruptive changes. But this is a thing we've been wanting to do for a long time now.

Manishearth:
> You're adding these to the segmenter types, which we conceptually don't want to be mutable because we want the iterators.

Yes, and I filed a followup issue to remove them.

This PR is just "update to latest Diplomat so that whenever someone needs a diplomat update they don't have to think about it". I don't like it when Diplomat changes get all hatched up and then you get a larger update PR that takes a bunch of work.

robertbastian:
> Diplomat does enforce these now. It just doesn't _use_ them yet, that's the next step.

Please don't strawman. You know what I mean.

Manishearth:
> Please don't strawman. You know what I mean.

No, I legitimately did not. Either you had misunderstood the Diplomat change, so the first half of my comment clarifies that.

Or you had understood it and was against Diplomat making the change without the other half (or against me doing the partial update). I addressed that in the second half. It's a disruptive change and doing it in parts will be helpful to users. I don't yet know if I plan to do the "full enforcement" now.

PR #7705, opened 2026-02-26
Add FFI for date arithmetic

Manishearth:
Aside from the `fmt + fixes` commit, this was _entirely_ agent-generated, one prompt per commit.

I have quickly checked the diff before moving on in each step. I plan to do a more thorough review but I haven't yet.

(I understand if people do not wish to review this until I have had a look. I wanted to put this out there as an example of pure agent work which I think @sffc will be interested in at least)


ffi/capi/src/date.rs
```diff
@@ -30,6 +33,82 @@ pub mod ffi {
...
+        #[diplomat::rust_link(icu::calendar::types::DateDuration::from_str, FnInStruct, hidden)]
+        #[diplomat::attr(all(supports = fallible_constructors, supports = named_constructors), named_constructor)]
+        pub fn from_string(v: &DiplomatStr) -> Result<DateDuration, DateDurationParseError> {
+            Ok(icu_calendar::types::DateDuration::try_from_utf8(v)?.into())
+        }
+
+        /// Returns a new [`DateDuration`] representing a number of years.
+        #[diplomat::rust_link(icu::calendar::types::DateDuration::for_years, FnInStruct)]
+        #[diplomat::attr(all(supports = fallible_constructors, supports = named_constructors), named_constructor)]
+        pub fn for_years(years: i32) -> DateDuration {
```
Manishearth, line 80: question: is it actually worth exposing these over FFI? feels wasteful to do an FFI call for this.

sffc: I think I have a slight preference to keep it, because the DateDuration fields are just pretty awful and error-prone to manipulate manually. I almost think we should keep it opaque over FFI but I don't want the extra cost of allocating it. I don't care much about making a function call that will probably be inlined in some languages anyway.

ffi/capi/src/date.rs
```diff
@@ -185,6 +266,39 @@ pub mod ffi {
...
             self.0.days_in_year()
         }
+
+        /// Returns a new [`IsoDate`] with the given duration added to it.
+        ///
+        /// 🚧 This API is unstable and may experience breaking changes outside major releases.
+        #[diplomat::rust_link(icu::calendar::Date::try_added_with_options, FnInStruct)]
+        #[diplomat::rust_link(icu::calendar::Date::try_add_with_options, FnInStruct, hidden)]
+        #[cfg(feature = "unstable")]
+        pub fn try_added_with_options(
```
sffc: Question: `add` or `added`?

Do we anticipate wanting to add a `&mut` version of this ever? If not, then maybe just call it `add`? But maybe `added` is clearer and just better.

ffi/capi/src/date.rs
```diff
@@ -185,6 +266,39 @@ pub mod ffi {
...
+        #[diplomat::rust_link(icu::calendar::Date::try_until_with_options, FnInStruct)]
+        #[cfg(feature = "unstable")]
+        pub fn try_until_with_options(
+            &self,
+            other: &IsoDate,
+            options: DateDifferenceOptions,
+        ) -> DateDuration {
+            self.0
+                .try_until_with_options(&other.0, options.into())
+                .unwrap() // Infallible for Iso
```
sffc: Issue: don't use unwrap. If it is infallible then you can assign it to `Ok(duration) = ...`

ffi/capi/src/date.rs
```diff
@@ -511,6 +625,38 @@ pub mod ffi {
...
+        /// Calculating the duration between `other - self`
+        ///
+        /// 🚧 This API is unstable and may experience breaking changes outside major releases.
+        #[diplomat::rust_link(icu::calendar::Date::try_until_with_options, FnInStruct)]
+        #[cfg(feature = "unstable")]
+        pub fn try_until_with_options(
+            &self,
+            other: &Date,
+            options: DateDifferenceOptions,
+        ) -> Result<DateDuration, CalendarDateDifferenceError> {
```
sffc: Question/Discussion: Should this be just DateTimeMismatchedCalendarError (also in Rust), which we already have in errors.rs

ffi/capi/src/date.rs
```diff
@@ -30,6 +33,84 @@ pub mod ffi {
...
+    #[non_exhaustive]
+    pub enum DateDurationUnit {
+        Years,
+        Months,
+        Weeks,
+        Days,
+    }
+
+    #[diplomat::rust_link(icu::calendar::types::DateDuration, Struct)]
+    #[cfg(feature = "unstable")]
```
robertbastian, line 49: should all unstable APIs be documented as such? it's very easy to get a linker or runtime load error if they are missing

ffi/capi/src/date.rs
robertbastian: yes, the error here is `core::convert::Infallible`

we might want to add `Date::until_with_options` for calendars that are infallible.

ffi/capi/src/date.rs
robertbastian: `DateTimeMismatchedCalendarError` is formatting specific, it's a struct with fields `this_kind` and `date_kind`. `AnyCalendarDifferenceError` is a singleton enum (non-exhaustive for some reason)

ffi/capi/src/date.rs
```diff
@@ -560,3 +706,47 @@ impl<'a> From<ffi::DateFields<'a>> for icu_calendar::types::DateFields<'a> {
         fields
     }
 }
+
+#[cfg(feature = "unstable")]
+impl From<icu_calendar::types::DateDuration> for ffi::DateDuration {
+    fn from(other: icu_calendar::types::DateDuration) -> Self {
+        Self {
+            is_negative: other.is_negative,
```
robertbastian, line 723: weren't these fields meant to be private?

ffi/capi/src/errors.rs
```diff
@@ -98,6 +98,45 @@ pub mod ffi {
...
+        Overflow = 0x03,
+    }
+
+    #[derive(Debug, PartialEq, Eq)]
+    #[repr(C)]
+    #[diplomat::rust_link(icu::calendar::cal::AnyCalendarDifferenceError, Enum, compact)]
+    #[cfg(all(feature = "unstable", feature = "calendar"))]
+    #[non_exhaustive]
+    #[diplomat::attr(auto, error)]
+    pub enum CalendarDateDifferenceError {
```
robertbastian: I don't see the value in making this an enum instead of a ZST struct. For `AnyCalendar` there won't be additional variants, if another calendar needs a different error, they can use a different type

ffi/capi/src/errors.rs
```diff
@@ -250,6 +289,56 @@ impl From<icu_calendar::error::DateFromFieldsError> for CalendarDateFromFieldsEr
...
+                Self::MismatchedCalendars
+            }
+            _ => Self::Unknown,
+        }
+    }
+}
+
+#[cfg(feature = "calendar")]
+#[cfg(all(feature = "unstable", feature = "calendar"))]
+impl From<core::convert::Infallible> for CalendarDateDifferenceError {
```
robertbastian: issue: the existence of this impl means we can spawn `CalendarDateDifferenceError` in return types from nothing. please remove

ffi/capi/src/date.rs
Manishearth: I don't want to expose mutation over FFI where we don't have to. So `add` is good.

ffi/capi/src/date.rs
Manishearth: We should call this `until` for Iso. Good catch.

ffi/capi/src/date.rs
Manishearth: No, we discussed and decided they were public.

Manishearth: (and we decided against the alternative: a large number of fallible mutation methods)

ffi/capi/src/errors.rs
Manishearth: This is an enum in ICU4X as well. We should fix both at the same time? I'm going to note this on the graduation issue (but not fix this here). Good catch.

ffi/capi/src/date.rs
Manishearth: https://github.com/unicode-org/icu4x/issues/7059#issuecomment-3785809131

ffi/capi/src/errors.rs
Manishearth: https://github.com/unicode-org/icu4x/pull/7713

sffc (approved):

PR #7709, opened 2026-02-26
Simplify FFI time zone timestamp code

robertbastian:
This encapsulates the `ZoneNameTimestamp` calculation/guessing in `ffi::TimeZoneInfo::as_rust_at_time` instead of having it in generated code across multiple formatters. It also uses `TimeZoneInfo::at_rd_time` instead of `TimeZoneInfo::at_date_time_iso`, as the RD is already calculated, and, especially if there is a calendar conversion, this saves doing multiple conversions and passing around `iso_date_obj: impl FnOnce() -> Date<Iso>`.

### DIFF
--- a/components/datetime/src/format/input.rs
+++ b/components/datetime/src/format/input.rs
@@ -9,8 +9,9 @@ use crate::scaffold::*;
 use icu_calendar::types::{DayOfYear, RataDie};
 use icu_calendar::{AsCalendar, Calendar};
 use icu_time::scaffold::IntoOption;
+use icu_time::zone::models::{AtTime, Base};
 use icu_time::zone::ZoneNameTimestamp;
-use icu_time::{Hour, Minute, Nanosecond, Second};
+use icu_time::{Hour, Minute, Nanosecond, Second, TimeZoneInfo};
 
 use icu_calendar::Date;
 use icu_time::{zone::UtcOffset, Time, TimeZone};
@@ -86,6 +87,19 @@ impl DateTimeInputUnchecked {
     /// Sets the time zone UTC offset.
     pub fn set_time_zone_utc_offset(&mut self, utc_offset: UtcOffset) {
         self.zone_offset = Some(utc_offset);
+    }
+
+    /// Sets all fields from a [`TimeZoneInfo<Base>`] input.
+    pub fn set_time_zone_info_base_fields(&mut self, info: TimeZoneInfo<Base>) {
+        self.zone_id = Some(info.id());
+        self.zone_offset = self.zone_offset.or_else(|| info.offset());
+    }
+
+    /// Sets all fields from a [`TimeZoneInfo<AtTime>`] input.
+    pub fn set_time_zone_info_at_time_fields(&mut self, info: TimeZoneInfo<AtTime>) {
+        self.zone_id = Some(info.id());
+        self.zone_offset = self.zone_offset.or_else(|| info.offset());
+        self.zone_name_timestamp = Some(info.zone_name_timestamp());
     }
 
     /// Sets the time zone ID.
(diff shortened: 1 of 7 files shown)


components/datetime/src/format/input.rs
```diff
@@ -88,6 +91,19 @@ impl DateTimeInputUnchecked {
...
+        self.zone_offset = self.zone_offset.or_else(|| info.offset());
+    }
+
+    /// Sets all fields from a [`TimeZoneInfo<AtTime>`] input.
+    pub fn set_time_zone_info_at_time_fields(&mut self, info: TimeZoneInfo<AtTime>) {
+        self.zone_id = Some(info.id());
+        self.zone_offset = self.zone_offset.or_else(|| info.offset());
+        self.zone_name_timestamp = Some(info.zone_name_timestamp());
+    }
+
```
sffc, line 104: Suggestion: do this here on the DateTimeInputUnchecked instead of adding more FFI code

```suggestion
    pub fn infer_time_zone_name_timestamp(&mut self) {
        match (self.rata_die, self.hour, self.minute, self.zone_offset) {
            (Some(rata_die), Some(hour), Some(minute), Some(offset)) => {
                // calculate and set the zone name timestamp
            }
            _ => {}
        }
    }
```

Then in FFI you replace the complicated `input.set_time_zone_name_timestamp` with simply `input.infer_time_zone_name_timestamp()`

robertbastian: I don't want to do this because
* In Rust you should use the correct types (`TimeZoneInfo<Base>` vs `TimeZoneInfo<AtTime>`), we should only do this hack for our FFI `TimeZoneInfo`
* this is sensitive to the order in which fields are added to `DateTimeInputUnchecked`

sffc: 1. It seems harmless to add `set_time_zone_info_base_fields`, but we (I) had designed this type to split out the time zone fields into separate setters.
2. It is called "unchecked" for a reason.

For more background on my reasoning when I wrote this type: at first I had all the fields be public. But it was cumbersome to manually extract each Date field at every call site. So then I made these convenience setters for Date and Time, and I made all the fields private, but I left the individual field setters for time zone, because that is how this type was envisioned to work.

components/datetime/src/format/input.rs
```diff
@@ -73,6 +75,7 @@ impl DateTimeInputUnchecked {
         self.day_of_month = Some(date_in_calendar.day_of_month());
         self.weekday = Some(date_in_calendar.weekday());
         self.day_of_year = Some(date_in_calendar.day_of_year());
+        self.rata_die = Some(date_in_calendar.to_rata_die());
```
sffc: Issue: Please remove this from `set_date_fields_unchecked`, which calls it unconditionally when basically no one needs it, and create a `set_rata_die` standalone fn.

robertbastian: this is https://github.com/unicode-org/icu4x/pull/7708, which is why this PR is draft and you shouldn't be reviewing it.

components/datetime/src/format/input.rs
robertbastian: The inference just should not live on `DateTimeInputUnchecked`. It also needs to take into account the explicit timestamp that might be set on the FFI TZI.

ffi/capi/src/timezone.rs
```diff
@@ -300,6 +300,26 @@ pub mod ffi {
...
+        &self,
+        date: Option<icu_calendar::Date<C>>,
+        time: Option<icu_time::Time>,
+    ) -> icu_time::TimeZoneInfo<icu_time::zone::models::AtTime> {
+        let base = self.id.with_offset(self.offset);
+        if let Some(zone_name_timestamp) = self.zone_name_timestamp {
+            base.with_zone_name_timestamp(zone_name_timestamp)
+        } else if let Some(date) = date {
+            base.at_date_time_iso(icu_time::DateTime {
+                date: date.to_calendar(icu_calendar::Iso),
```
sffc, line 314: I assume you have plans to convert this to use Cal -> RataDie instead of Cal -> Iso -> RataDie

sffc (approved):

PR #7711, opened 2026-02-26
Add `time` integration

robertbastian:
Similar to https://github.com/unicode-org/icu4x/pull/7617.

`time` is actually the largest of the three popular Rust datetime crates, having more crates.io downloads (total and recent) than either `jiff` or `chrono`.

## Changelog: N/A

Merged with #7617


components/datetime/src/time_crate.rs
```diff
@@ -0,0 +1,349 @@
...
+impl InFixedCalendar<()> for time::Weekday {}
+impl GetField<()> for time::Weekday {
+    fn get_field(&self) {}
+}
+
+impl GetField<Weekday> for time::Weekday {
+    fn get_field(&self) -> Weekday {
+        (*self).into()
+    }
+}
```
sffc: Question: Does this mean a `time::Weekday` can be passed as input to a formatter created with `fieldsets::E`? If so, that's great, everything working as designed. We don't have a weekday type in icu4x core.

robertbastian: > We don't have a weekday type in icu4x core.

We do, I'm adding conversions for it in this very PR

robertbastian: it's in this code snippet even

robertbastian:
can I get a review please

PR #7713, opened 2026-02-26
Rename AnyCalendarDifferenceError to MismatchedCalendarsError (and make it a unit struct)

Manishearth:
Progress on https://github.com/unicode-org/icu4x/issues/3964

@robertbastian noticed that this probably should be a unit struct.

I agree. We have already committed[^1] `until` to not returning any computation-relevant errors for non-AnyCalendar, since we set the error to Infallible. This means AnyCalendar should also not be returning any computation-relevant errors. The only possible error for AC on top of regular until errors (i.e. nothing) is "mismatched calendar". I can't see any reason to expand that.

If people don't want to make this a unit struct, we should still land the first commit.

 [^1]: Unstable API, can be changed, but I think everyone likes this design for `until`.

This will conflict with https://github.com/unicode-org/icu4x/pull/7705, but that's fine.

## Changelog: N/A

covered in graduation docs

### DIFF
--- a/components/calendar/src/error.rs
+++ b/components/calendar/src/error.rs
@@ -601,7 +601,7 @@ mod unstable {
     ///
     /// N
     #[derive(Clone, Copy, PartialEq, Debug, Display)]
-    #[displaydoc("Attempted to diff two Date<AnyCalendar>`s with different calendars")]
+    #[displaydoc("Attempted to diff two `Date<AnyCalendar>`s with different calendars")]
     #[allow(
         clippy::exhaustive_structs,
         reason = "This is the only possible error with paired AnyCalendar operations"


components/calendar/src/error.rs
```diff
@@ -561,6 +561,40 @@ mod unstable {
     }
 
     impl core::error::Error for DateAddError {}
+
+    /// Error returned when comparing two [`Date`](crate::Date)s with [`AnyCalendar`](crate::AnyCalendar).
```
robertbastian: issue, preexisting: it's not comparing (PartialOrd), more like subtracting

components/calendar/src/error.rs
```diff
@@ -561,6 +561,40 @@ mod unstable {
...
+    /// let d2 = Date::try_new_persian(1562, 1, 1).unwrap().to_any();
+    ///
+    /// assert_eq!(
+    ///     d1.try_until_with_options(&d2, Default::default())
+    ///         .unwrap_err(),
+    ///     MismatchedCalendarsError,
+    /// );
+    ///
+    /// // To compare the dates, convert them to the same calendar,
+    /// // such as ISO.
```
robertbastian: issue, preexisting: it doesn't make sense to suggest converting to a third calendar, because then the difference will be in terms of that calendar. instead, show how the choice of calendar matters:

```rust
let diff1 = d1.to_calendar(*d2.calendar())
    .try_until_with_options(&d2, Default::default())
    .unwrap();
let diff2 = d1
    .try_until_with_options(&d2.to_calendar(*d1.calendar())
    .unwrap(); 

// diff1 and diff2 are probably different
```

components/calendar/src/error.rs
```diff
@@ -561,6 +561,40 @@ mod unstable {
...
+    ///
+    /// d1.to_iso()
+    ///     .try_until_with_options(&d2.to_iso(), Default::default())
+    ///     .unwrap();
+    /// ```
+    #[derive(Clone, Copy, PartialEq, Debug, Display)]
+    #[displaydoc("Attempted to compare two AnyCalendars with different months")]
```
robertbastian:
```suggestion
    #[displaydoc("Attempted to diff two Date<AnyCalendar>s with different calendars")]
```

components/calendar/src/error.rs
```diff
@@ -561,6 +561,40 @@ mod unstable {
...
+    #[allow(
+        clippy::exhaustive_structs,
+        reason = "This is the only possible error with paired AnyCalendar operations"
+    )]
+    pub struct MismatchedCalendarsError;
```
robertbastian: https://unicode-org.github.io/icu4x/rustdoc/icu/datetime/struct.MismatchedCalendarError.html

https://unicode-org.github.io/icu4x/rustdoc/icu/calendar/enum.ParseError.html#variant.MismatchedCalendar

```suggestion
    pub struct MismatchedCalendarError;
```

Manishearth: Should `icu::datetime` just reuse this error?

Manishearth: Nope, it cannot.

components/calendar/src/error.rs
```diff
@@ -561,6 +561,54 @@ mod unstable {
...
+    ///     .try_until_with_options(&d2.to_calendar(d1.calendar().clone()),
+    ///                             options)
+    ///     .unwrap();
+    ///
+    /// assert_ne!(diff1, diff2);
+    /// ```
+    ///
+    /// N
+    #[derive(Clone, Copy, PartialEq, Debug, Display)]
+    #[displaydoc("Attempted to diff two Date<AnyCalendar>`s with different calendars")]
```
robertbastian:
```suggestion
    #[displaydoc("Attempted to diff two `Date<AnyCalendar>`s with different calendars")]
```

Manishearth:
Merging now as it impacts my FFI PR. @sffc should also sign off, or we should revisit when I properly do a graduation review.

(This and the other issue will be noted)

PR #7714, opened 2026-02-26
Take self by ref in try_added_with_options

Manishearth:
See discussion in https://github.com/unicode-org/icu4x/issues/3964#issuecomment-3963697368

## Changelog: N/A

covered in graduation changelog

### DIFF
--- a/components/calendar/src/date.rs
+++ b/components/calendar/src/date.rs
@@ -397,10 +397,7 @@ impl<A: AsCalendar> Date<A> {
             .calendar
             .as_calendar()
             .add(&self.inner, duration, options)?;
-        Ok(Self {
-            inner,
-            calendar: self.calendar.clone(),
-        })
+        Ok(Self::from_raw(inner, self.calendar.clone()))
     }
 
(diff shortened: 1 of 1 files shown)


Manishearth:
(I could also call `.clone()` and then call `try_add`, wasn't sure which is cleaner)

components/calendar/src/date.rs
```diff
@@ -382,12 +386,21 @@ impl<A: AsCalendar> Date<A> {
...
+    ) -> Result<Self, DateAddError>
+    where
+        A: Clone,
+    {
+        let inner = self
+            .calendar
+            .as_calendar()
+            .add(&self.inner, duration, options)?;
+        Ok(Self {
+            inner,
```
robertbastian: use Date::from_raw to be more explicit about invariants

robertbastian (approved):

Manishearth:
Do want to make sure @sffc is okay with this as well but either way this will be looked at as a part of the graduation.

Merging now since it impacts my FFI PR.

PR #7717, opened 2026-02-26
Add snapshots for date arithmetic coverage test

### LINKED ISSUE #7215: Add snapshot test of Date add/until

sffc:
Fixes #7215 

It is fast enough that I am running it on the whole set. I removed all days in 3-26 which also significantly sped up the test. I can still change it to run on a deterministic subset if you like. The test takes less than 3s on my machine.

### DIFF
--- a/components/calendar/src/tests/arithmetic.rs
+++ b/components/calendar/src/tests/arithmetic.rs
@@ -3,7 +3,7 @@
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
 use crate::options::{DateAddOptions, DateDifferenceOptions, Overflow};
-use crate::types::{DateDuration, DateDurationUnit, Month, MonthCode};
+use crate::types::{DateDuration, DateDurationUnit, Month};
 use crate::{AsCalendar, Calendar, Date};
 use core::fmt;
 


components/calendar/src/tests/arithmetic.rs
```diff
@@ -3,11 +3,71 @@
...
+    pub start: ArithmeticDateForDebug,
+    pub end: ArithmeticDateForDebug,
+    pub duration: DateDuration,
+}
+
+impl fmt::Display for TestOutput {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{} {} {:2}y {:2}m {:2}w {:2}d = {} ({})",
```
Manishearth: praise: I like the succinctness of this over just using Debug outputs.

Manishearth:
I'm not sure we need to do a deterministic subset, but a valid reason to do so would be to reduce the size of the snapshots.

Manishearth (approved):

sffc:
ok I did work to make the test better: it now does actual proper assertions for 90%+ of cases. It only adds snapshot tests when something potentially odd is going on with regard to constraining behavior. This reduces the snapshots to be about 500 KB per calendar.

Manishearth (approved):
filtering, yay

Manishearth:
@sffc I think we can mark this as closing https://github.com/unicode-org/icu4x/issues/7078

sffc:
I think it fixes #7215. It looks like #7078 wants us to test reject behavior, too. It could be added to this test.

Issue #7710, opened 2026-02-26
`NoCalendarFormatter` fails with fieldset `E`


robertbastian:
This works

```rust
    NoCalendarFormatter::try_new(icu::locale::locale!("fr").into(), fieldsets::T::long()).unwrap();
```

but this doesn't

```rust
    NoCalendarFormatter::try_new(icu::locale::locale!("fr").into(), fieldsets::E::long()).unwrap();
```

sffc:
The line for NoCalendarFormatter is drawn as the Date/Time/Zone boundary, and `E` is considered a date field. 🤷

sffc:
CLDR also considers weekdays to be calendar-dependent fields; it stores weekday names under `<calendar type="gregorian">` and friends.

However, ICU4X has only a single `DatetimeNamesWeekdayV1`, and we seem to generate it from Gregorian only.

Suffice it to say that we're not consistent here.

The workaround is to use a fixed calendar formatter, and any external `Weekday` type needs to implement the trait that allows it to be formatted with that fixed calendar.

sffc:
I haven't done a full investigation for what would need to change in the scaffold traits to allow this, but one starting point is that we use the date skeleton data key for storing the weekday patterns, because again CLDR says they are calendar-dependent. We would need to split out the weekday skeleton patterns or move them to a different calendar-independent data key.

robertbastian:
`FixedCalendarDateTimeFormatter::<Gregorian, _>::try_new(locale!("fr").into(), fieldsets::E::long())` doesn't compile.

sffc:
> `FixedCalendarDateTimeFormatter::<Gregorian, _>::try_new(locale!("fr").into(), fieldsets::E::long())` doesn't compile.

I'm not sure why not. I landed an example demonstrating how it compiles.

robertbastian:
> I landed an example demonstrating how it compiles.

You landed a fix for the issue, not an example demonstrating that there's no issue.

sffc:
I didn't fix anything having to do with the call site `FixedCalendarDateTimeFormatter::<Gregorian, _>::try_new(locale!("fr").into(), fieldsets::E::long())`. I landed trait impls to allow `Weekday` to be allowed as input to that formatter.

PR #7719, opened 2026-02-27
Add datetime formatting impls on Weekday with test

sffc:
See #7710

## Changelog

icu_datetime: Allow formatting `Weekday`
 - New trait impls: `InSameCalendar` and `ConvertCalendar` on `UtcOffset` and `Weekday`

### DIFF
--- a/components/datetime/src/scaffold/calendar.rs
+++ b/components/datetime/src/scaffold/calendar.rs
@@ -601,6 +601,11 @@ impl CalMarkers<ErasedPackedPatterns> for FullDataCalMarkers {
 }
 
 /// A type that can be converted into a specific calendar system.
+///
+/// You often want to set `Converted` to an ICU4X built-in type.
+///
+/// If the type is not calendar-specific, such as a time or time zone, set `Converted`
+/// to the same type and return it in the implementation.
 // This trait is implementable
 pub trait ConvertCalendar {
     /// The converted type. This can be the same as the receiver type.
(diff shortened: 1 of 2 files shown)


components/datetime/src/scaffold/calendar.rs
```diff
@@ -649,6 +650,14 @@ impl<C: IntoAnyCalendar, A: AsCalendar<Calendar = C>, Z: Copy> ConvertCalendar
...
 }
 
+impl ConvertCalendar for UtcOffset {
+    type Converted<'a> = UtcOffset;
+    #[inline]
+    fn to_calendar<'a>(&self, _: &'a AnyCalendar) -> Self::Converted<'a> {
+        *self
+    }
+}
+
```
sffc, line 665: This one seemed to be missing. We implement this trait on Time and TimeZoneInfo already.

sffc:
the CI errors are both network

components/datetime/src/scaffold/calendar.rs
```diff
@@ -727,6 +744,13 @@ impl<O: TimeZoneModel> InSameCalendar for TimeZoneInfo<O> {
     }
 }
 
+impl InSameCalendar for Weekday {
```
robertbastian, line 752: can you document the set of traits that an input type needs to implement somewhere?

sffc: done

PR #7728, opened 2026-03-02
Don't expose CLDR's Hebrew month numbering

### LINKED ISSUE #7727: Bad data for narrow Hebrew months
CLDR uses its month numbering for narrow Hebrew months: 

https://github.com/unicode-org/cldr-json/blob/10a18ab6a5580e71b43dcc76c69bdc3e4cfdb724/cldr-json/cldr-cal-hebrew-full/main/en/ca-hebrew.json#L28-L43

This results in 

https://github.com/unicode-org/icu4x/blob/3a81a1fe582645f609ef2b1805cf93e3aa7774b3/provider/source/data/debug/datetime/names/month/hebrew/v1/4/en.json

Which is incorrect according to ICU4X's Hebrew month model. This is an issue that we need to work around in ICU4X, as we don't use the same month numbering as CLDR.

robertbastian:
Fixes #7727 

## Changelog

icu_datetime: Don't expose CLDR's Hebrew month numbering during formatting


provider/source/src/datetime/names.rs
```diff
@@ -357,7 +357,64 @@ fn months_convert(
     let (month_count, has_leap) = calendar_months(calendar);
     let mut symbols = vec![Cow::Borrowed(""); month_count];
 
-    if calendar == DatagenCalendar::Hebrew {
+    if calendar == DatagenCalendar::Hebrew
+        && length == Length::Narrow
+        && months.0["10"].starts_with(&months.0["1"])
```
Manishearth, line 362: observation: this is the key check in case CLDR decides to [fix this](https://unicode-org.atlassian.net/browse/CLDR-19282): if CLDR fixes this, then month 10 (M09) will no longer start with some variant of the digit 1.

PR #7731, opened 2026-03-03
Handle permanent DST

robertbastian:
In case we want to model `America/Vancouver`'s upcoming changes as permanent DST within the existing `America_Pacific` metazone, this adds the datagen code to show disambiguations ("Pacific Time (Vancouver)") for permanent-DST zones.

See also #6755 and #7162.

## Changelog

icu_time: Handle permanent DST


robertbastian:
> Why the data change in iedub?

Because we now look at the explicit CLDR-defined stdOffset, where previously we tried to infer it only from earlier or later periods, which didn't work for that iedub period.

> Is America/Vancouver really the first to try doing this? I thought some of those regions in North Africa were observing the permanent DST

Nowhere in North Africa does permanent DST. I think Yukon is the only jurisdiction that has announced "permanent DST", but CLDR modeled this as "Yukon Time" instead of "Pacific Daylight Time". I want to keep the door open to model America/Vancouver as year-round "Pacific Daylight Time".

PR #7732, opened 2026-03-04
Move DateDurationUnit to options, document weeks behavior

Manishearth:
This is an option, it should be in options.

"weeks" is interesting since it is not possible to get weeks out of `until` _unless_ you specifically ask for them. I've documented this on DateDifferenceOptions. We could also document it on DateDurationUnit.

### DIFF
--- a/ffi/capi/bindings/cpp/icu4x/DateDurationUnit.d.hpp
+++ b/ffi/capi/bindings/cpp/icu4x/DateDurationUnit.d.hpp
@@ -27,7 +27,7 @@ namespace capi {
 
 namespace icu4x {
 /**
- * See the [Rust documentation for `DateDurationUnit`](https://docs.rs/icu/2.1.1/icu/calendar/types/enum.DateDurationUnit.html) for more information.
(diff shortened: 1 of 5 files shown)


components/calendar/src/options.rs
```diff
@@ -206,6 +211,7 @@ mod unstable {
         /// assert_eq!(
         ///     d1.try_until_with_options(&d2, options_weeks).unwrap(),
         ///     DateDuration {
+        ///         // This is the only time there is a `weeks` value.
         ///         weeks: 58,
```
robertbastian, line 215: Should we get rid of the week fields, and have getters/setters that use the days field?

Manishearth: I think that's an option. I feel like for the complex algorithms here it is better to try and just match the spec where possible.

robertbastian: I think our complex algorithms all go through `DateDuration::add_weeks_and_days_to`, so they already treat weeks and days as just a combined days value.

Manishearth: Until has a separate specced Weeks branch.

PR #7734, opened 2026-03-04
Validate `DateInner` compatibility

robertbastian:
The until operation only works if both calendars are the same. This is not just a concern for `AnyCalendar`, any calendar can be parameterised, and some of our calendars actually are (~although none in a way that affects the until method~). Given that `Calendar::DifferenceError` is an associated type, we need to allow calendar implementations to actually figure out whether the calendars are comparable. In `AnyCalendar` we currently do this by checking that both dates and one calendar have the same parameters, but we should actually check both calendars and both dates.

Update: this also needs to be done for `PartialEq` and `PartialOrd`

Update 2: being explicit about calendar identity actually revealed an issue: we allow interactions between different `Hijri<TabularAlgorithm>` calendars – this is wrong

## Changelog

icu_calendar: Improve validation in scenarios where you may try comparing dates from different calendars. `PartialEq` and `PartialOrd` on two different `HijriTabular` calendars will now produce `None`.

### DIFF
--- a/components/calendar/src/cal/abstract_gregorian.rs
+++ b/components/calendar/src/cal/abstract_gregorian.rs
@@ -163,7 +163,7 @@ impl<Y: GregorianYears> Calendar for AbstractGregorian<Y> {
         date1.until(date2, &AbstractGregorian(IsoEra), options)
     }
 
-    fn eq_calendars(&self, other: &Self) -> Result<(), Self::IdentityError> {
+    fn check_identity(&self, other: &Self) -> Result<(), Self::IdentityError> {
         // The GregorianYears generic only affects constructors and year_info,
         // it does not affect date identity.
         let _ignored = other;
@@ -317,12 +317,12 @@ macro_rules! impl_with_abstract_gregorian {
                     .until(&date1.0, &date2.0, options)
             }
 
-            fn eq_calendars(&self, other: &Self) -> Result<(), Self::IdentityError> {
+            fn check_identity(&self, other: &Self) -> Result<(), Self::IdentityError> {
                 let $self_ident = self;
                 let c1 = crate::cal::abstract_gregorian::AbstractGregorian($eras_expr);
                 let $self_ident = other;
                 let c2 = crate::cal::abstract_gregorian::AbstractGregorian($eras_expr);
-                c1.eq_calendars(&c2)
+                c1.check_identity(&c2)
             }
 
             fn year_info(&self, date: &Self::DateInner) -> Self::Year {
--- a/components/calendar/src/cal/abstract_gregorian.rs
+++ b/components/calendar/src/cal/abstract_gregorian.rs
@@ -30,10 +30,6 @@ pub(crate) trait GregorianYears: Clone + core::fmt::Debug {
         None
     }
 
-    type IdentityError: core::fmt::Debug;
-
-    fn check_identity(&self, other: &Self) -> Result<(), Self::IdentityError>;
-
     fn debug_name(&self) -> &'static str;
 }
 
@@ -88,7 +84,7 @@ impl<Y: GregorianYears> crate::cal::scaffold::UnstableSealed for AbstractGregori
 impl<Y: GregorianYears> Calendar for AbstractGregorian<Y> {
     type DateInner = ArithmeticDate<AbstractGregorian<IsoEra>>;
     type Year = EraYear;
-    type IdentityError = Y::IdentityError;
+    type IdentityError = core::convert::Infallible;
 
     fn from_codes(
         &self,
@@ -170,7 +166,10 @@ impl<Y: GregorianYears> Calendar for AbstractGregorian<Y> {
     }
 
     fn eq_calendars(&self, other: &Self) -> Result<(), Self::IdentityError> {
-        self.0.check_identity(&other.0)
+        // The GregorianYears generic only affects constructors and year_info,
+        // it does not affect date identity.
+        let _ignored = other;
+        Ok(())
     }
 
     fn eq_dates(
--- a/components/calendar/src/cal/abstract_gregorian.rs
+++ b/components/calendar/src/cal/abstract_gregorian.rs
@@ -156,6 +156,7 @@ impl<Y: GregorianYears> Calendar for AbstractGregorian<Y> {
     #[cfg(feature = "unstable")]
     fn until(
         &self,
+        _: &Self,
         date1: &Self::DateInner,
         date2: &Self::DateInner,
         options: DateDifferenceOptions,
@@ -301,13 +302,14 @@ macro_rules! impl_with_abstract_gregorian {
             #[cfg(feature = "unstable")]
             fn until(
                 &self,
+                _: &Self,
                 date1: &Self::DateInner,
                 date2: &Self::DateInner,
                 options: crate::options::DateDifferenceOptions,
             ) -> Result<crate::types::DateDuration, Self::DifferenceError> {
                 let $self_ident = self;
                 crate::cal::abstract_gregorian::AbstractGregorian($eras_expr)
-                    .until(&date1.0, &date2.0, options)
+                    .until(&crate::cal::abstract_gregorian::AbstractGregorian($eras_expr), &date1.0, &date2.0, options)
             }
 
             fn year_info(&self, date: &Self::DateInner) -> Self::Year {
--- a/components/calendar/src/cal/abstract_gregorian.rs
+++ b/components/calendar/src/cal/abstract_gregorian.rs
@@ -317,7 +317,10 @@ macro_rules! impl_with_abstract_gregorian {
                     .until(&date1.0, &date2.0, options)
             }
 
-            fn check_date_compatibility(&self, other: &Self) -> Result<(), Self::DateCompatibilityError> {
+            fn check_date_compatibility(
+                &self,
+                other: &Self,
+            ) -> Result<(), Self::DateCompatibilityError> {
                 let $self_ident = self;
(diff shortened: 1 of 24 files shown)


components/calendar/src/cal/abstract_gregorian.rs
```diff
@@ -156,13 +160,35 @@ impl<Y: GregorianYears> Calendar for AbstractGregorian<Y> {
...
         date1: &Self::DateInner,
         date2: &Self::DateInner,
         options: DateDifferenceOptions,
-    ) -> Result<types::DateDuration, Self::DifferenceError> {
+    ) -> Result<types::DateDuration, Self::IdentityError> {
+        self.0.check_identity(&other.0)?;
         Ok(date1.until(date2, &AbstractGregorian(IsoEra), options))
     }
 
+    fn eq_dates(
```
Manishearth: issue: I think Eq and Ord should just use RD so that they don't need any of this other scaffolding.

that said, this scaffolding could be used to avoid needing to pay the cost of RD conversion on most calendars.

Manishearth (commented):
Hmm. IdentityError is an interesting concept. We already have commtited to DifferenceError not getting new variants because in most cases it is infallible, so explicitly calling it IdentityError is good in my book.

I'm not as clear on if I want to have more Calendar methods to power the Eq/Ord impls. I ... guess it's fine? I want to be careful to not add too many new things to Calendar, since these methods are stable to call.

@sffc, thoughts?

components/calendar/src/cal/abstract_gregorian.rs
robertbastian: comparing RDs is significantly more expensive than comparing the YMD.

Manishearth: Oh, I thought I deleted this comment. I had thought about it more and realized that I wanted Eq to include the calendar.

Using the Option for the PartialOrd is good. Though it would be nice to be able to BTreeMap these with Ord, but there are other ways to do that.

robertbastian: hm? e.g. `Date<Hebrew>` is `Ord`, before this PR and after. Just `Date<AnyCalendar>` is not.

components/calendar/src/date.rs
```diff
@@ -597,18 +599,22 @@ where
...
+            .cmp_dates(other.calendar(), &self.inner, &other.inner)
+            .ok()
     }
 }
 
 impl<C, A> Ord for Date<A>
 where
-    C: Calendar,
-    C::DateInner: Ord,
+    C: Calendar<IdentityError = core::convert::Infallible>,
```
robertbastian: this is a semver breakage, especially because this is removing `Ord` from `Date<HijriTabular>`.

components/calendar/src/cal/abstract_gregorian.rs
Manishearth: Right, I meant with AC. But like I said, there are other ways to do that.

components/calendar/src/date.rs
Manishearth: Hmmm. I think it makes sense to do so, but we should carefully consider the breakage likelihood.

My impression is that not many people use the individual calendar types, and `Ord` is not something they would use them with.

Manishearth (commented):
My two worries here are:

 - The minor breaking change. OTOH I'm not sure that _not_ doing the breaking change is the right call either: the "default buggy" Ord behavior seems bad. I _really_ don't want assertions like that in our code
 - More Calendar APIs: this adds more methods to `Calendar`, which are insta-stable to call. I'm not a huge fan of that dynamic, I would perhaps prefer `doc(hidden) pub fn unstable_eq_..` or something

sffc (commented):
I'm trying to figure out what bug this is trying to fix. I see no tests that demonstrate why all this extra code and machinery is motivated.

My best guess: you can have two `Hijri<Tabular>`, one constructed with Thursday rules, and the other with Friday rules. Since they share the same type identity, the PartialOrd impls are turned on, which delegate to ArithmeticDate impls, which don't know that you need to add or subtract a day.

If this is the bug, are there more concise ways to fix it?

robertbastian:
The original motivation for this PR was to make assumptions over calendar equality explicit, in a way that doesn't affect runtime code. Finding the tabular bug was the result of that work.

This PR cannot be tested, because it removes functionality that should not have existed.

robertbastian:
> * The minor breaking change. OTOH I'm not sure that _not_ doing the breaking change is the right call either: the "default buggy" Ord behavior seems bad. I _really_ don't want assertions like that in our code

One thing we could do is fall back to RD equality if calendar identity fails. This way we can have `Ord` on `Date<HijriTabular>` and `Date<AnyCalendar>`. I haven't thought this all the way through though, this might not be a valid `Ord`.

Manishearth:
> One thing we could do is fall back to RD equality if calendar identity fails. This way we can have Ord on Date<HijriTabular> and Date<AnyCalendar>. I haven't thought this all the way through though, this might not be a valid Ord.

Yeah, this was the thing I suggested in a comment a few days ago, but I myself am not super sure of that. RD calculations are more expensive than straight up comparison; feels like a potential footgun especially if you're using `Ord` for BTreeMaps.

But I'm not opposed.

robertbastian:
We can decide this in another PR. For now I've left a comment pointing out the incorrect comparison.

Manishearth:
I think my preference for this PR is to actually include the very minor breaking change on `impl Ord for Date<Hijri<Tabular>>`, and have a followup for reexpanding.

robertbastian:
I don't want to bog down this PR by including a breaking change.

Manishearth:
Fair point, I also don't want to bog down the PR with something inefficient, but it seems like the current impl retains the old fast behavior of comparing YearInfos. I'll rereview.

components/calendar/src/cal/hijri.rs
```diff
@@ -371,6 +392,15 @@ impl Rules for TabularAlgorithm {
             extended_year,
         }
     }
+
+    type IdentityError = crate::error::MismatchedCalendarError;
```
Manishearth: I think this is the only actual behavior change, and I think it's the right call.

Manishearth (approved):
I like this.

components/calendar/src/date.rs
sffc, line 614: Suggestion: add `C: Ord` (make sure this works for all built-in calendars and add it to the "icu_calendar considerable changes" section of the changelog) instead of the complicated `check_identity` thing. (Also `C: PartialOrd`, `C: Eq`, etc., above.)

sffc (changes requested):
I think I understand the problem; I am not convinced that this is the right solution

sffc:
On the type `Date<C>`, I feel like the bound should be

```rust
impl<A> Ord for Date<A>
where
    A: AsCalendar,
    A::Calendar: Ord,
    A::Calendar::DateInner: Ord,
```

and then you just compare the two fields (calendar first).

This is another case where I slightly regret adding these impls, but I think it's _fine_

components/calendar/src/date.rs
robertbastian: I am 100% convinced that `C: Ord` should not be required for `Date<C>: Ord`. You can always compare dates (worst case falling back to RD comparison), but defining `Ord` on calendars is super arbitrary.

robertbastian:
> and then you just compare the two fields (calendar first).

That doesn't make any sense. This would lead to a lexicographic order over (calendar, year, month, day), where the calendar ordering is arbitrary.

sffc (approved):
`Identity` is a terrible name. We should bikeshed the name before 2.2

components/calendar/src/calendar.rs
```diff
@@ -159,7 +162,13 @@ pub trait Calendar: crate::cal::scaffold::UnstableSealed {
...
         date2: &Self::DateInner,
         options: DateDifferenceOptions,
-    ) -> Result<types::DateDuration, Self::DifferenceError>;
+    ) -> types::DateDuration;
+
+    /// Returns whether [`Self::DateInner`] represents the same date in both calendars.
+    ///
+    /// This is checked by [`Date::try_until_with_options`](crate::Date::try_until_with_options),
+    /// `impl PartialEq for Date<C>`, `impl PartialOrd for Date<C>`, and `impl Ord for Date<C>`.
+    fn check_identity(&self, other: &Self) -> Result<(), Self::IdentityError>;
```
robertbastian:
```suggestion
    fn check_date_compatibility(&self, other: &Self) -> Result<(), Self::IdentityError>;
```

Manishearth (approved):

Manishearth (approved):

components/calendar/src/cal/abstract_gregorian.rs
```diff
@@ -159,8 +159,15 @@ impl<Y: GregorianYears> Calendar for AbstractGregorian<Y> {
...
         options: DateDifferenceOptions,
-    ) -> Result<types::DateDuration, Self::DifferenceError> {
-        Ok(date1.until(date2, &AbstractGregorian(IsoEra), options))
+    ) -> types::DateDuration {
+        date1.until(date2, &AbstractGregorian(IsoEra), options)
+    }
+
+    fn check_date_compatibility(&self, other: &Self) -> Result<(), Self::DateCompatibilityError> {
+        // The GregorianYears generic only affects constructors and year_info,
+        // it does not affect date identity.
```
sffc, line 168:
```suggestion
        // The GregorianYears generic only affects constructors and year_info.
        // The DateInners always use ISO yeears.
```

sffc (commented):
(old pending comment)

PR #7735, opened 2026-03-04
Move try_added_with_options over to taking by-self again

Manishearth:
https://github.com/unicode-org/icu4x/issues/3964

I am convinced that we should take by `self` here, the implicit clone is the worst of all the options. We should consider making AnyCalendar Copy.

### DIFF
--- a/ffi/capi/src/date.rs
+++ b/ffi/capi/src/date.rs
@@ -641,6 +641,11 @@ pub mod ffi {
             duration: DateDuration,
             options: DateAddOptions,
         ) -> Result<Box<Date>, CalendarDateAddError> {
+            // This is a cheap clone: DateInner is Copy, and AnyCalendar
+            // while not Copy does not contain any types with destructors.
+            //
+            // It *may* add calendars with non-cheap clones in the future, but that will
+            // be weighted against the performance of this FFI code.
             Ok(Box::new(Date(self.0.clone().try_added_with_options(
                 duration.into(),
                 options.into(),


ffi/capi/src/date.rs
robertbastian, line 639: observation: this implicit clone is effectively a `Copy`.

ffi/capi/src/date.rs
```diff
@@ -641,7 +641,7 @@ pub mod ffi {
             duration: DateDuration,
             options: DateAddOptions,
         ) -> Result<Box<Date>, CalendarDateAddError> {
-            Ok(Box::new(Date(self.0.try_added_with_options(
+            Ok(Box::new(Date(self.0.clone().try_added_with_options(
```
sffc, line 649: Nit: Add a comment explaining why this is a cheap clone

ffi/capi/src/date.rs
```diff
@@ -641,7 +641,12 @@ pub mod ffi {
             duration: DateDuration,
             options: DateAddOptions,
         ) -> Result<Box<Date>, CalendarDateAddError> {
-            Ok(Box::new(Date(self.0.try_added_with_options(
+            // This is a cheap clone: DateInner is Copy, and AnyCalendar
```
robertbastian, line 644: well now you should annotate all calendar clones in this file. I don't think this comment is necessary though

Manishearth: This is a nit on a nit fix, I don't wish to spend any time on this whatsoever.

PR #7736, opened 2026-03-04
Add changelog policy and CI enforcement

### LINKED ISSUE #7153: Have a better API changelog

### LINKED ISSUE #7154: Improved changelog procedure

Manishearth:
As discussed in https://github.com/unicode-org/icu4x/issues/7154, this adds a new changelog policy

Fixes #7153, fixes #7154. This fixes both issues, my understanding is that the new policy produces a CHANGELOG.md that satisfies all currently stated requirements of an "API changelog".

This PR contains tooling but of course I cannot test this yet since ICU4X doesn't have PR bodies with this info yet. The tooling was agent-generated, and seems to mostly work with my experimentation.

I have a slight preference for landing this _now_ (with enforcement) even though it won't be used: we can use the 2.2 release to test out tooling around this.

I can commit to playing around with the script more and improving it: I would prefer not to block landing this on the quality of the script. But I can also remove it from this PR if that is what others want.

### DIFF
--- a/documents/process/changelog.md
+++ b/documents/process/changelog.md
@@ -31,7 +31,7 @@ For API changes, you MUST list all newly added or modified public items. This in
 * New methods (excluding those on new types).
 * New traits.
 * New trait items (excluding those on new traits).
-* New trait implementations (excluding those on new types).
+* New trait implementations (excluding those on new types), including standard library traits such as `PartialOrd` and `Copy`.
 * New re-exports and type aliases.
 * New type parameters on existing items.
 * New struct fields or enum variants.


documents/process/changelog.md
```diff
@@ -0,0 +1,109 @@
...
+* New trait items (excluding those on new traits).
+* New trait implementations (excluding those on new types, or ones with new traits; also excluding common ones like `Debug`, `Clone`, and `PartialEq`).
+* New re-exports and type aliases.
+* New type parameters on existing items.
+* New struct fields or enum variants.
+* New Cargo features.
+
+Changes to *stable* APIs are mandatory. Changes to *unstable* APIs are optional but encouraged if they are significant.
+
+Additions to FFI that mirror the primary API need not be mentioned in the changelog unless the API getting an FFI version was already publicly available in a prior release. FFI-specific changes should of course be mentioned in the changelog.
```
Manishearth, line 42: note: this is a yet-undiscussed addition that I think will be uncontroversial: we don't tend to have an FFI changelog entry for each new API unless that addition is "interesting" in some way.

documents/process/changelog.md
```diff
@@ -0,0 +1,109 @@
...
+
+A good changelog entry should be concise but informative.
+
+For API changes, you MUST list all newly added or modified public items. This includes:
+
+* New types (structs, enums, unions).
+* New methods (excluding those on new types).
+* New traits.
+* New trait items (excluding those on new traits).
+* New trait implementations (excluding those on new types, or ones with new traits; also excluding common ones like `Debug`, `Clone`, and `PartialEq`).
```
sffc: Nit: I thought we agreed to include new derived trait impls on existing types. I think the three specific ones you listed here are highly unlikely to be controversial, but should we call out that `Copy`, `PartialOrd`, ... should definitely be included?

documents/process/changelog.md
sffc: seems fine

documents/process/changelog.md
```diff
@@ -0,0 +1,109 @@
...
+If a large number of APIs are added (e.g., during a Unicode update), it is acceptable to link to the PR or another document for the full list.
+
+## Changelog Collation
+
+The collation process occurs during the release process and is typically driven by the release driver. It can also be done mid-cycle resulting in an `# Unreleased` section in the changelog.
+
+
+### Gathering Entries
+
+The release driver gathers all changelog entries from the `# Changelog` sections of all PRs merged since the last release. This can be done by scraping PR descriptions via the GitHub API.
```
sffc:
```suggestion
The release driver gathers all changelog entries from the `## Changelog` sections of all PRs merged since the last release. This can be done by scraping PR descriptions via the GitHub API.
```

documents/process/changelog.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Version changes
+
+Most ICU4X components are versioned with the "ICU4X version", but ICU4X utils and some crates like `icu_pattern` are not.
+
+The changelog entry for these crates should mention a version diff, e.g. `icu_pattern: 0.4.1 -> 0.4.2`, or "No change" if there was no change made.
+
+### Handling Deferrals
+
+For any PRs with `TBD` in the `# Changelog` entry, the collator is responsible for:
```
sffc:
```suggestion
For any PRs with `TBD` in the `## Changelog` entry, the collator is responsible for:
```

documents/process/changelog.md
Manishearth: Ah, I misremembered. I'll fix.

documents/process/changelog.md
```diff
@@ -0,0 +1,109 @@
...
+
+A good changelog entry should be concise but informative.
+
+For API changes, you MUST list all newly added or modified public items. This includes:
+
+* New types (structs, enums, unions).
+* New methods (excluding those on new types).
+* New traits.
+* New trait items (excluding those on new traits).
+* New trait implementations (excluding those on new types).
```
sffc: Just to explicitly call this out, since newer programmers might not realize that `#[derive(X)]` is the same thing as a "new trait implementation"

```suggestion
* New trait implementations (excluding those on new types), including standard library traits such as `PartialOrd` and `Copy`.
```

PR #7742, opened 2026-03-06
In icu_time, limit use of the icu crate to doctests

musicinmybrain:
This is a really minor tweak, but one that’s useful to me as downstream maintainer of most of the packaged ICU4X crates in Fedora.

We don’t have the *entire* ICU4X stack packaged, so we don’t have a package for the top-level `icu` meta-crate. Even if we did, we would probably try to avoid circular test dependencies to make it easier to bootstrap everything. So while I do try to run as many tests as I reasonably can in each crate, I need to disable those that use the `icu` crate.

I’ve notice that in *almost all* cases, the `icu` meta-crate is used only in doctests, while tests that won’t appear in rustdoc output use the individual `icu_*` crates directly. This makes sense presentation-wise, and it is a really helpful convention for me, because it gives me the option to sidestep the `icu` test dependency by omitting doctests, and just building and running everything else.

This PR therefore “fixes” the sole use of the `icu` crate in `icu_time`’s lib-tests.

### DIFF
--- a/components/time/src/zone/windows.rs
+++ b/components/time/src/zone/windows.rs
@@ -155,7 +155,7 @@ impl WindowsParserBorrowed<'_> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use icu::locale::subtags::subtag;
+    use icu_locale_core::subtags::subtag;
 
     #[test]
     fn basic_windows_tz_lookup() {


musicinmybrain:
> Can you confirm for which other crates this currently works?

Hmm, this is a bit of a manual auditing exercise. It looks like it works for at least `icu_calendar`, `icu_collator`, `icu_decimal`, `icu_locale`, `icu_locale_core`, `icu_normalizer`, `icu_provider`, and `icu_segmenter`.

On the other hand, it looks it doesn’t work (I am carrying a patch to avoid the `icu` dependency in at least one non-doc test) for `icu_collections` and `icu_properties`, so this might not be quite as consistent a pattern as I had thought and claimed.

If a crate isn’t mentioned in either list above, it may be because the tests don’t use `icu`, or just because we don’t have it packaged.

If you would like a similar PR for `icu_collections` and `icu_properties`, I’m happy to follow up with one. If you are *not* interested in more PR’s like this, that’s OK too. Just let me know.

robertbastian:
If you want to avoid carrying patches we'd appreciate PRs for the other crates.

musicinmybrain:
> If you want to avoid carrying patches we'd appreciate PRs for the other crates.

Ok, thanks! I’ll follow up on these, probably in the coming day or so.

PR #7743, opened 2026-03-06
Remove `YearInfo: PartialEq` bound

robertbastian:
There's code that compares whole `YearInfo`s when it should be comparing extended years.

### DIFF
--- a/components/calendar/src/cal/east_asian_traditional.rs
+++ b/components/calendar/src/cal/east_asian_traditional.rs
@@ -841,7 +841,7 @@ impl<A: AsCalendar<Calendar = ChineseTraditional>> Date<A> {
 }
 
 /// Information about a [`EastAsianTraditional`] year.
-#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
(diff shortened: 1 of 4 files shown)


components/calendar/src/cal/east_asian_traditional.rs
```diff
@@ -841,7 +841,7 @@ impl<A: AsCalendar<Calendar = ChineseTraditional>> Date<A> {
 }
 
 /// Information about a [`EastAsianTraditional`] year.
-#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
+#[derive(Copy, Clone, Debug)]
 // TODO(#3933): potentially make this smaller
 pub struct EastAsianTraditionalYear {
```
sffc, line 846: This seems to be the only public type whose impls are changing, and it is new in 2.2

Manishearth: Also HijriYear, which is public but i think unstable

robertbastian: both public YearInfos are unstable.

PR #7739, opened 2026-03-05
Improve the performance of `ArithmeticDate.until()`

poulsbo:
Progress on #7077

Improve the performnace of the calendar `until()` function.

- Tighten the estimation of the number of months between two dates in order to reduce the number of incremental `surpasses()` tests. This is done by introducing a new DateFieldsResolver function, `min_months_from()`, which the Coptic, Ethiopic, and Hebrew calendar implementations customize.

- Eliminate the `surpasses()` loop for the final days delta calculation. Instead compute an intermediate date, which is the start date plus the years, months, and weeks computed so far. Subtract the Rata Die of the intermediate date from that of the target date to get the days field of the DateDuration.

- Fix minor typos in function names and comments.

Performance, as measured using criterion, is improved for Years and Months intervals. Improvements on the order of 20% to 80%, or 100s of ns to >1 ms, are shown, for base times of 500 ns to 5 ms.

For Weeks and Days, which are already computed efficiently using RD subtraction, criterion measures a slight slowdown (on the order of 1 ns, for a 15-20 ns operation). This is most likely due to a change in compiler inlining, optimization, register usage, etc., since the code for Days/Weeks has not changed.

## Changelog: N/A

(covered by other PRs)

PR #7746, opened 2026-03-06
More documentation for min_months_from_inner

Manishearth:
Improving docs from https://github.com/unicode-org/icu4x/pull/7739

### DIFF
--- a/components/calendar/src/cal/east_asian_traditional.rs
+++ b/components/calendar/src/cal/east_asian_traditional.rs
@@ -156,11 +156,6 @@ pub struct EastAsianTraditional<R>(pub R);
 ///
 /// This crate currently provides [`Rules`] for [`China`] and [`Korea`].
 ///
-/// Currently, implementations of this trait must produce calendars that
-/// have a leap month at least once every three years. This is used in optimizing
-/// `until()`, and may cause errors and debug assertions there if violated. We may
-/// change this to not be needed in the future.
-///
 /// <div class="stab unstable">
 /// 🚫 This trait is sealed; it should not be implemented by user code. If an API requests an item that implements this
 /// trait, please consider using a type from the implementors listed below.
@@ -594,7 +589,7 @@ impl<R: Rules> DateFieldsResolver for EastAsianTraditional<R> {
 
     #[inline]
     fn min_months_from_inner(_start: Self::YearInfo, years: i64) -> i64 {
-        // By `Rules` invariant
+        // TODO(#7077) This is not verified, and the error is not constant bounded
         12 * years + (years / 3)
     }
 


components/calendar/src/cal/east_asian_traditional.rs
```diff
@@ -591,6 +591,12 @@ impl<R: Rules> DateFieldsResolver for EastAsianTraditional<R> {
     fn min_months_from_inner(_start: Self::YearInfo, years: i64) -> i64 {
         // A lunisolar leap month is inserted at least every 3 years. Reingold denies
         // that the Chinese calendar determines leap years using the 19-year Metonic cycle.
+        //
+        // This still drifts from the actual month value so months calculations will still be
+        // linear time in the year diff being considered (just with a much smaller constant factor).
+        //
+        // TODO(#7077): Investigate the Chinese calendar and find a calculation
+        // here that can result in a bounded error bar.
         12 * years + (years / 3)
```
robertbastian, line 593: #7739 was merged before I had a chance to review it, so I'll leave my comment here:

there are a two issues with this:

* `EastAsianTraditional` is generic in `R: Rules`, and the rules can do whatever they want. We either need to add a requirement that the rules insert a leap month at least every three years, or delegate this logic to the `Rules` implementation
* Even for our concrete calendars, I'm not sure that this is correct. Both `China` and `Korea` use three different rules (1900-1912, 1912-2050/2100, simple approximation otherwise), so this needs to be shown to hold for all three (and for the transitions). For the hardcoded data we can write a test that this holds, but for the simple approximation I would like to see some kind of reasoning.

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -235,7 +235,11 @@ pub(crate) trait DateFieldsResolver: Calendar {
         12
     }
 
-    /// The minimum number of months over `years` years, starting from the given year.
+    /// A minimum bound for the number of months over `years` years, starting from the given year.
```
```suggestion
    /// A lower bound for the number of months in `years` years, starting from the beginning of this year.
```

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -235,7 +235,11 @@ pub(crate) trait DateFieldsResolver: Calendar {
...
+    ///
+    /// This may be equal to the number of months, but should never be over. Ideally,
```
```suggestion
    /// This may be equal to the number of months, but should never be larger. Ideally,
```

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -235,7 +235,11 @@ pub(crate) trait DateFieldsResolver: Calendar {
...
+    /// implementations should be close enough to the actual answer such that the error
+    /// is bounded by a constant.
```
```suggestion
    /// is bounded by a constant with respect to `years`.
```

components/calendar/src/cal/hebrew.rs
```diff
@@ -136,7 +136,30 @@ impl DateFieldsResolver for Hebrew {
...
     fn min_months_from_inner(_start: HebrewYear, years: i64) -> i64 {
-        // There are 7 leap years in every 19-year Metonic cycle.
+        // The Hebrew Metonic cycle is 7 leap years every 19 years,
+        // which comes out to 235 months per 19 years.
+        //
+        // We need to ensure that this is always *lower or equal to* the number of
+        // months in a given year span.
+        //
+        // Firstly, note that this math will produce exactly the number of months in any given cycle
+        // that is a multiple of 19 years. Note that we are only performing integer
```
robertbastian:
```suggestion
        // Firstly, note that this math will produce exactly the number of months in any given period
        // that spans a whole number of cycles. Note that we are only performing integer
```

components/calendar/src/cal/hebrew.rs
```diff
@@ -136,7 +136,30 @@ impl DateFieldsResolver for Hebrew {
...
+        // ops here, and our SAFE_YEAR_RANGE is well within the range of allowed values
+        // for multiplying by 235.
+        //
+        // So we only need to verify that this math produces the right results within a single cycle.
+        //
+        // The Hebrew Metonic cycle has leap years in year 3, 6, 8, 11, 14, 17, and 19,
+        // i.e., leap year gaps of +3, +3, +2, +3, +3, +3, +2.
+        //
+        // 235 / 19 is 12.368 months per year, which leads to one leap month every three years plus 0.1
```
robertbastian: please use fractions instead of rounded decimals

components/calendar/src/cal/hebrew.rs
```diff
@@ -136,7 +136,30 @@ impl DateFieldsResolver for Hebrew {
...
+        //
+        // Firstly, note that this math will produce exactly the number of months in any given cycle
+        // that is a multiple of 19 years. Note that we are only performing integer
+        // ops here, and our SAFE_YEAR_RANGE is well within the range of allowed values
+        // for multiplying by 235.
+        //
+        // So we only need to verify that this math produces the right results within a single cycle.
+        //
+        // The Hebrew Metonic cycle has leap years in year 3, 6, 8, 11, 14, 17, and 19,
+        // i.e., leap year gaps of +3, +3, +2, +3, +3, +3, +2.
```
robertbastian: be explicit about the boundary, either add year 0 or year 22

components/calendar/src/cal/hebrew.rs
```diff
@@ -136,7 +136,30 @@ impl DateFieldsResolver for Hebrew {
...
+        //
+        // 235 / 19 is 12.368 months per year, which leads to one leap month every three years plus 0.1
+        // month left over. So this is "in bounds" as long as it does not predict a leap month in 2 years
```
robertbastian:
```suggestion
        // month left over. So this is correct as long as it does not predict a leap month in 2 years
```

components/calendar/src/cal/hebrew.rs
```diff
@@ -136,7 +136,30 @@ impl DateFieldsResolver for Hebrew {
...
+        // where the Hebrew calendar expects one in 3.
+        //
+        // The longest sequence of "three year leap months" in the Hebrew calendar
+        // is 4: year 9->11->14->17. In that time the error will accumulate to about 0.3, which is not
```
robertbastian:
```suggestion
        // is 4: year 8->11->14->17. In that time the error will accumulate to about 0.3, which is not
```

components/calendar/src/cal/east_asian_traditional.rs
poulsbo: suggestion: Merge these observations into the TODO block, so they "travel with the code" until such time as we are able to address them.

question: Is the issue serious enough that we should back this off to return $12y$ until they are resolved?

robertbastian: I think this is serious, because even though our long-term approximation isn't "correct" (whatever that means), if 12y+y/3 doesn't hold, we are probably breaking some Temporal invariants between `add` and `until`.

components/calendar/src/cal/hebrew.rs
Manishearth: Year 19 is the boundary. This is every 19 years.

components/calendar/src/cal/hebrew.rs
Manishearth: Added an approx equals. I think this is clearer when you can see the actual numbers.

components/calendar/src/cal/east_asian_traditional.rs
Manishearth: > We either need to add a requirement that the rules insert a leap month at least every three years

I think this is fine as a temporary measure and we can think more about this when we try and stabilize Rules.

One thing we can do, btw, is tweak the min months approximation so that in the rare case it fails we just fall back to the old iterative code. Then it's totally fine for this to be wrong.

components/calendar/src/cal/hebrew.rs
Manishearth: Eh, fractions works too

components/calendar/src/cal/east_asian_traditional.rs
```diff
@@ -156,6 +156,11 @@ pub struct EastAsianTraditional<R>(pub R);
 ///
 /// This crate currently provides [`Rules`] for [`China`] and [`Korea`].
 ///
+/// Currently, implementations of this trait must produce calendars that
```
robertbastian: ok, now show that `China` and `Korea` obey this invariant

components/calendar/src/cal/east_asian_traditional.rs
```diff
@@ -591,6 +596,12 @@ impl<R: Rules> DateFieldsResolver for EastAsianTraditional<R> {
     fn min_months_from_inner(_start: Self::YearInfo, years: i64) -> i64 {
         // A lunisolar leap month is inserted at least every 3 years. Reingold denies
         // that the Chinese calendar determines leap years using the 19-year Metonic cycle.
+        //
+        // This still drifts from the actual month value so months calculations will still be
+        // linear time in the year diff being considered (just with a much smaller constant factor).
+        //
+        // TODO(#7077): Investigate the Chinese calendar and find a calculation
+        // here that can result in a bounded error bar.
```
robertbastian:
```suggestion
// By `Rules` invariant
```

components/calendar/src/cal/east_asian_traditional.rs
Manishearth: No.

I am _mostly_ convinced the code currently upholds this property. I am not convinced enough to "prove" it, I am convinced from a read through of the code, enough that I am satisfied with landing that code.

My original plan was to properly sit down and figure out China and Korea (such that there is a _constant time_ error). I haven't had time to do that (and didn't plan to do that for 2.2), but Alan opened his PR in the meantime. I am doing followups for his PR because I _have_ thought about the Hebrew stuff and wanted to write it down. Me choosing to do those followups is not me signing up for more work here, if you want to disagree with that PR feel free to disagree with it and discuss further, or open a PR to undo the Chinese part of it.

I already recommended another route above: instead of debug asserting after computing `min_months`, we can set `min_months` to `sign` when that assertion fails. That is the only thing that needs to change to turn this from a hard requirement into a soft "optimal behavior" one. If you care about this, go ahead and do that.

robertbastian: Well you merged Alan's PR before it was thoroughly reviewed, so you're kind of on the hook for this now.

Now you're adding an invariant to the type without showing that it holds. The safe thing to do would be to remove the invariant, go back to returning 12 * years and figure this out later.

> I already recommended another route above: instead of debug asserting after computing min_months, we can set min_months to sign when that assertion fails. That is the only thing that needs to change to turn this from a hard requirement into a soft "optimal behavior" one.

Cool, we could have discussed that on the PR that added this code. I don't want to do this work either, so let's revert #7739.

Manishearth: I am fine doing small changes here. Please do not ask me to do large changes as a part of a followup PR that I am making as a courtesy.

robertbastian (approved):

robertbastian:
> No, the last commit pushed a few seconds ago regresses performance relative to what landed on main.

Yes but it's provably correct. Please let's not go back to yoloing code with debug assertions, that has been painful enough to clean up in the past.

Manishearth:
I am going to undo my undoing so that this PR is purely a docs fix. We should land this, and we can separately discuss the chinese invariant. I'll make a separate PR for that.

I am rather frustrated with this ping ponging on a small courtesy PR.

Manishearth:
> Yes but it's provably correct. Please let's not go back to yoloing code with debug assertions, that has been painful enough to clean up in the past.



And it's scope creep for this PR.

Manishearth:
Especially since the fix to avoid the debug assertions is not that big a deal.

robertbastian:
> I am rather frustrated with this ping ponging on a small courtesy PR.

I'm rather frustrated with https://github.com/unicode-org/icu4x/pull/7739 being merged without my review. When I left work on Friday that PR was still actively being worked on, and I had drafted review comments. Now I'm the bad guy for asking for these changes to be reverted.

Manishearth:
> Now I'm the bad guy for asking for these changes to be reverted.

No. You could still post the review comments.

I'm just frustrated that _I'm_ being asked to fix these in _this_ PR.

components/calendar/src/cal/east_asian_traditional.rs
sffc: I'm quite confident that Pinqi follows this invariant, because it adds a leap month with an exact period, I think every 33 months.

robertbastian:
> I'm just frustrated that _I'm_ being asked to fix these in _this_ PR.

*You're* adding invariants that might not actually be true, thereby legitimizing problematic code. I'm happy to approve this without the east asian changes.

Manishearth:
> You're adding invariants that might not actually be true, thereby legitimizing problematic code.

You said: "We either need to add a requirement that the rules insert a leap month at least every three years".

I followed that feedback. Turned out you wanted the associated proof as well, which I do not have right now. Okay, fine, I'll undo that change there.

components/calendar/src/cal/east_asian_traditional.rs
robertbastian: > I'm quite confident that Pinqi follows this invariant, because it adds a leap month with an exact period, I think every 33 months.

Please don't call the simple approximation Pinqi, it's not that because it's based on the Gregorian year.

I'm quite confident too, but I don't want to find out when Temporal is experiencing panics in half a year. We use several approximations in that code.

sffc:
I think the version of the PR that documented the invariant was better, because we should write down our assumptions, but the PR here is still an improvement over the status quo so I approve

PR #7747, opened 2026-03-07
Add and use PluralCategoryExtended

### LINKED ISSUE #5369: Suggestion: Use `PluralRules` instead of `Count`
### Summary
Currently, the enum `Count` is being used in multiple places around `icu4x`, such as `units`, `currency`, `decimal` ... etc.

-  I suggest implement in `PluralRules` place an enum called l `PluralRuleCount`, which would contain all the CLDR counts: 

```rust
pub enum PluralRulesCount {

  Zero,
  One,
  Two,
... etc.
}
```

- All data providers should adopt this standardized `PluralRuleCount`.
- If users need to use extra counts or variations, they should extend the existing `PluralRuleCount`.
     -  Example: ```rust
     - pub enum CompactDecimalCount {
     - PluralRules(PluralRulesCount)
     - Explecit1,
     - Explecit0,
     - }
```

### PR DESCRIPTION
**Pull request by Manishearth**
Fixes #5369

Agent driven, was experimenting with throwing agents at our tech debt issues during SEW.

### DIFF
--- a/components/plurals/src/lib.rs
+++ b/components/plurals/src/lib.rs
@@ -1150,22 +1150,6 @@ impl<T: PartialEq> PluralElements<T> {
         })
     }
 
-    /// Sets the value for `Explicit0`.
-    pub fn with_explicit_zero_value(self, explicit_zero: Option<T>) -> Self {
-        Self(PluralElementsInner {
-            explicit_zero: explicit_zero.filter(|t| *t != self.0.other),
-            ..self.0
-        })
-    }
-
-    /// Sets the value for `Explicit1`.
-    pub fn with_explicit_one_value(self, explicit_one: Option<T>) -> Self {
-        Self(PluralElementsInner {
-            explicit_one: explicit_one.filter(|t| *t != self.0.other),
-            ..self.0
-        })
-    }
-
     /// Sets the value for [`PluralCategory::One`].
     pub fn with_one_value(self, one: Option<T>) -> Self {
         Self(PluralElementsInner {
@@ -1194,6 +1178,22 @@ impl<T: PartialEq> PluralElements<T> {
     pub fn with_many_value(self, many: Option<T>) -> Self {
         Self(PluralElementsInner {
             many: many.filter(|t| *t != self.0.other),
+            ..self.0
+        })
+    }
+
+    /// Sets the value for `Explicit0`.
+    pub fn with_explicit_zero_value(self, explicit_zero: Option<T>) -> Self {
+        Self(PluralElementsInner {
+            explicit_zero: explicit_zero.filter(|t| *t != self.0.other),
+            ..self.0
+        })
+    }
+
+    /// Sets the value for `Explicit1`.
+    pub fn with_explicit_one_value(self, explicit_one: Option<T>) -> Self {
+        Self(PluralElementsInner {
+            explicit_one: explicit_one.filter(|t| *t != self.0.other),
             ..self.0
         })
     }
--- a/ffi/capi/tests/missing_apis.txt
+++ b/ffi/capi/tests/missing_apis.txt
@@ -27,6 +27,9 @@ icu::collator::CollatorBorrowed::compare_latin1_utf16#FnInStruct
 icu::collator::CollatorBorrowed::write_sort_key_to#FnInStruct
 icu::collator::CollatorBorrowed::write_sort_key_utf16_to#FnInStruct
 icu::collator::CollatorBorrowed::write_sort_key_utf8_to#FnInStruct
+icu::plurals::PluralCategoryExtended#Enum
+icu::plurals::PluralCategoryExtendedULE#Struct
+icu::plurals::PluralRules::extended_category_for#FnInStruct
 icu::properties::script::HarfbuzzScriptData#Struct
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: components/plurals/src/lib.rs [hunk 1]
```diff
@@ -183,6 +183,49 @@ pub enum PluralCategory {
...
+    /// CLDR "two" plural category.
+    Two = 2,
+    /// CLDR "few" plural category.
+    Few = 3,
+    /// CLDR "many" plural category.
+    Many = 4,
+    /// CLDR "other" plural category.
+    Other = 5,
+    /// Explicit 0 rule.
+    Explicit0 = 6,
```
Manishearth, line 211: question: Should this be 6 or -1? I think 6 is fine.

components/plurals/src/lib.rs
```diff
@@ -473,6 +516,42 @@ impl PluralRules {
...
+    /// ```
+    /// use icu::locale::locale;
+    /// use icu::plurals::{PluralCategoryExtended, PluralRules};
+    ///
+    /// let pr = PluralRules::try_new(locale!("en").into(), Default::default())
+    ///     .expect("locale should be present");
+    ///
+    /// assert_eq!(pr.extended_category_for(0_usize), PluralCategoryExtended::Explicit0);
+    /// assert_eq!(pr.extended_category_for(1_usize), PluralCategoryExtended::Explicit1);
+    /// assert_eq!(pr.extended_category_for(2_usize), PluralCategoryExtended::Other);
```
Manishearth, line 553: question: more tests?

components/plurals/src/lib.rs
```diff
@@ -226,6 +226,22 @@ impl From<PluralCategory> for PluralCategoryExtended {
     }
 }
 
+impl From<u8> for PluralCategoryExtended {
```
Manishearth, line 229: question: do we want this impl? Should it be TryFrom?

provider/source/src/units/essentials.rs
```diff
@@ -55,44 +55,51 @@ impl DataProvider<UnitsEssentialsV1> for SourceDataProvider {
...
-                prefixes.insert(
-                    PatternKey::Power {
-                        power: power_value,
-                        count: *count,
-                    },
-                    pattern.to_string(),
-                );
-            });
+            ];
+            for (pattern, count) in patterns {
```
Manishearth, line 92: thought: good cleanup

components/plurals/src/provider.rs
```diff
@@ -532,33 +535,44 @@ where
     pub fn get<'a>(&'a self, op: PluralOperands, rules: &PluralRules) -> (FourBitMetadata, &'a V) {
         let parts = self.as_parts();
 
-        let category = rules.category_for(op);
-
-        match parts.specials {
-            Some(specials) => {
-                if op.is_exactly_zero() {
+        if let Some(specials) = parts.specials {
```
Manishearth, line 538: this code I haven't looked too closely at. looks right.

ffi/capi/tests/missing_apis.txt
```diff
@@ -27,6 +27,9 @@ icu::collator::CollatorBorrowed::compare_latin1_utf16#FnInStruct
 icu::collator::CollatorBorrowed::write_sort_key_to#FnInStruct
 icu::collator::CollatorBorrowed::write_sort_key_utf16_to#FnInStruct
 icu::collator::CollatorBorrowed::write_sort_key_utf8_to#FnInStruct
+icu::plurals::PluralCategoryExtended#Enum
```
Manishearth, line 30: question: Should we expose this over FFI? Should this live under `provider` and be internal/unstable? I'm not really sure...

Manishearth:
PluralElements seems to be different from this, though?

[PluralElements](https://unicode-org.github.io/icu4x/rustdoc/icu/plurals/struct.PluralElements.html) is the bag of per-plural-case values. Whereas [PluralCategory](https://unicode-org.github.io/icu4x/rustdoc/icu/plurals/enum.PluralCategory.html) is the "category" of a given number.

This is extending the latter. We should also migrate dimension over to PluralElements.

robertbastian:
The user of a pluralized value should never need to have a plural category, this type really is an encoding detail. `PluralElements` abstracts both the storage and the lookup.

Manishearth:
Okay, but the dimension code internally needs a type like this, and currently it has its own copy. It shouldn't have to, that is what that issue is proposing.

robertbastian:
The current modelling of the units essentials data needs a type like this. I'm not convinced it inherently needs a type like this. An alternative design would be:

```rust
    /// The SI prefixes, in order
    /// * ki, Mi, Gi, Ti, Pi, Ei, Zi, Yi, Ri, Qi
    /// * da, h, k, M, G, T, P, E, Z, Y, R, Q
    /// * d, c, m, µ, n, p, f, a, z, y, r, q
    pub si_prefixes: VarZeroVec<'data, str>,

    pub power_prefixes: VarZeroCow<'data, PluralElementsPackedULE<str>>,
```

`PluralElementsPackedULE` allow storing four bits of "metadata" for free, this can also be used to store whether it's a power of two or three.

Manishearth:
Okay, I think that issue should not be tech debt priority then.

Issue #7725, opened 2026-02-27
ByteStr in zerotrie is theoretically unsound


Manishearth:
Discovered in https://github.com/unicode-org/icu4x/pull/6915#discussion_r2866656987

https://github.com/unicode-org/icu4x/blob/0b926a3a2d452170de0799315e6c0db5c11be51a/utils/zerotrie/src/builder/bytestr.rs#L16-L28

Copying over my comment with edits

`(&ByteStr, usize)` is not safe to transmute from `(&str, usize)`, at least not according to my current understanding of unsafe rules. In practice I don't expect this to be a real problem anytime soon.

`str` guarantees its own layout, which is sufficient for _direct_ transmutation. However, within a `repr(Rust)` type, all bets are off: Rust is e.g. allowed to treat `(i8, usize)` and `(u8, usize)` as having different layouts. People sometimes test with `-Zrandomize-layout`. I _believe_ this can't happen when the difference is just two `repr(transparent)` types[^1], but it can happen for two types with the same layout.

It is worth asking Rust for clarity on this. Rust has historically been very reluctant to make commitments about what layout optimizations it is allowed to do, but this might be something where they're fine doing so, specifically around tuples. Current rustc doesn't do too much here but it could choose to in the future. I plan to engage upstream on this.

I'm also not really convinced this whole thing is necessary: why not use an `AsRef<[u8]>` impl internally? Is it the code size we're worried about? There may be a halfway method to do this.

 [^1]: I'm not actually sure of this either: I think Rust is allowed to give `struct Foo(u8)` and `struct Bar(u8)` different layouts if it really wants; but tuples are different, and `transparent` is likely meaningful there.

sffc:
> I'm also not really convinced this whole thing is necessary: why not use an `AsRef<[u8]>` impl internally? Is it the code size we're worried about? There may be a halfway method to do this.

I think it's because this is `const` code which can't use traits. I'd also be a bit worried about codesize.

Manishearth:
Ahhhh.

I think an enum model would work without traits and likely optimize to the same thing.

Manishearth:
(besides, it's builder code, I think we care less about perf)

sffc:
yeah I think an enum with helper functions would be a design that would remove the need for this particular unsafe code.

Manishearth:
Since this is a soundness issue, I am marking this as tech debt priority.

sffc:
[...]

PR #7748, opened 2026-03-07
Replace ByteStr with SliceWithIndices and SerdeByteStr

Manishearth:
Fixes https://github.com/unicode-org/icu4x/issues/7725

This was primarily agent-driven.

ByteStr is used in `const` code to abstract over `&[(T, usize)]` where `T` can be `&str` or `&[u8]`. `SliceWithIndices` is the same concept represented as an enum. This is builder code, so it doesn't matter that much.

ByteStr was also used in serde code to allow serializing/deserializing from a byte sequence or a string. I'm not 100% clear if that is actually still necessary: can we just ser/de the store. I didn't want to break anything, so I kept the existing behavior. I'm hoping @sffc can have a look and perform further cleanups if necessary.

<s>I did need to add an `either` dependency to write a convenient iterator type. It's internal, so I can also choose to not do that and iterate manually instead.</s> Figured out a cleaner way to do it.

## Changelog

zerotrie:
 - Fix minor unsoundness due to assumptions on the layout of tuples

### DIFF
--- a/utils/zerotrie/src/serde.rs
+++ b/utils/zerotrie/src/serde.rs
@@ -44,22 +44,6 @@ impl<'de> Visitor<'de> for ByteStrVisitor {
     }
 }
 
-struct SerdeByteStr<'a>(&'a [u8]);
-
-impl Serialize for SerdeByteStr<'_> {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: Serializer,
-    {
-        if serializer.is_human_readable() {
-            if let Ok(s) = core::str::from_utf8(self.0) {
-                return serializer.serialize_str(s);
-            }
-        }
-        serializer.serialize_bytes(self.0)
-    }
-}
-
 #[derive(PartialEq, Eq, PartialOrd, Ord)]
 #[repr(transparent)]
 pub(crate) struct SerdeByteStrOwned(pub(crate) Box<[u8]>);
@@ -97,6 +81,22 @@ impl Serialize for SerdeByteStrOwned {
             }
         }
         serializer.serialize_bytes(bytes)
+    }
+}
+
+struct SerdeByteStr<'a>(&'a [u8]);
+
+impl Serialize for SerdeByteStr<'_> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        if serializer.is_human_readable() {
+            if let Ok(s) = core::str::from_utf8(self.0) {
+                return serializer.serialize_str(s);
+            }
+        }
+        serializer.serialize_bytes(self.0)
     }
 }
 


utils/zerotrie/src/builder/konst/builder.rs
```diff
@@ -139,10 +138,8 @@ impl<const N: usize> ZeroTrieBuilderConst<N> {
 
     /// The actual builder algorithm. For an explanation, see [`crate::builder`].
     #[must_use]
-    const fn create_or_panic<const K: usize>(
-        &mut self,
-        all_items: ConstSlice<(&ByteStr, usize)>,
-    ) -> usize {
+    #[expect(clippy::indexing_slicing)] // lots of indexing, should be in range
```
robertbastian:
```suggestion
    #[expect(clippy::indexing_slicing)] // documented panic
```

utils/zerotrie/src/builder/bytestr.rs
```diff
@@ -2,131 +2,113 @@
...
-
-    #[allow(unsafe_code)] // transparent newtype casts are documented
-    pub fn from_bytes(input: &[u8]) -> &Self {
-        // Safety: [u8] and ByteStr have the same layout and invariants
-        unsafe { &*(input as *const [u8] as *const Self) }
+#[cfg(feature = "alloc")]
+use alloc::vec::Vec;
+
+/// Const function to evaluate `a < b` lexicographically.
+#[allow(clippy::indexing_slicing)] // in-range loop conditions
```
robertbastian: allow scopes should be as small as possible, put this on the loop

utils/zerotrie/src/builder/bytestr.rs
```diff
@@ -2,131 +2,113 @@
...
-    #[allow(unsafe_code)] // transparent newtype casts are documented
-    pub fn from_boxed_bytes(input: Box<[u8]>) -> Box<Self> {
-        // Safety: [u8] and ByteStr have the same layout and invariants
-        unsafe { core::mem::transmute::<Box<[u8]>, Box<Self>>(input) }
+/// Const function to evaluate `a[..prefix_len] == b[..prefix_len]`
+///
+/// # Panics
+///
+/// Panics if `prefix_len` is longer than either `a` or `b`.
+#[allow(clippy::indexing_slicing)] // in-range loop conditions
```
robertbastian: allow scopes should be as small as possible, put this on the if statement

utils/zerotrie/src/builder/bytestr.rs
```diff
@@ -2,131 +2,113 @@
...
+}
 
-    #[allow(dead_code)] // may want this in the future
-    pub fn from_str(input: &str) -> &Self {
-        Self::from_bytes(input.as_bytes())
-    }
+/// An abstraction over a slice of key-value pairs that can have either `&[u8]`
+/// or `&str` keys. This is used in the builder to avoid unsound casts.
+#[derive(Copy, Clone)]
+pub(crate) enum SliceWithIndices<'a> {
```
robertbastian: this is not generic, it should be `ByteSliceWithIndices`

utils/zerotrie/src/builder/bytestr.rs
```diff
@@ -2,131 +2,113 @@
...
+            Self::Str(s) => s.len(),
+        }
     }
 
-    #[allow(dead_code)] // not used in all features
-    pub fn is_all_ascii(&self) -> bool {
-        for byte in self.0.iter() {
-            if !byte.is_ascii() {
-                return false;
+    #[allow(clippy::indexing_slicing)]
```
robertbastian: justify?

utils/zerotrie/src/builder/bytestr.rs
```diff
@@ -2,131 +2,113 @@
...
-        assert!(prefix_len <= self.len());
-        assert!(prefix_len <= other.len());
-        let mut i = 0;
-        while i < prefix_len {
-            if self.0[i] != other.0[i] {
-                return false;
+    #[cfg(feature = "alloc")]
+    pub fn is_all_ascii(&self) -> bool {
+        /// Const function to evaluate if all bytes in `s` are ASCII.
+        #[allow(clippy::indexing_slicing)] // in-range loop conditions
```
robertbastian: same

utils/zerotrie/src/builder/bytestr.rs
```diff
@@ -2,131 +2,113 @@
...
+        #[cfg(feature = "alloc")]
```
robertbastian: I don't think you need nested cfgs, and I also don't think this actually needs alloc

utils/zerotrie/src/builder/slice_indices.rs
```diff
@@ -2,131 +2,113 @@
...
+        const fn is_all_ascii(s: &[u8]) -> bool {
+            let mut i = 0;
+            while i < s.len() {
+                if !s[i].is_ascii() {
+                    return false;
+                }
+                i += 1;
             }
-            i += 1;
+            true
```
robertbastian:
```suggestion
            s.iter().all(|c| c.is_ascii())
```

utils/zerotrie/src/zerotrie.rs
```diff
@@ -453,8 +454,8 @@ macro_rules! impl_zerotrie_subtype {
                     .iter()
                     .map(|(k, v)| (k.borrow().as_bytes(), *v))
                     .collect();
-                let byte_str_slice = ByteStr::from_byte_slice_with_value(&tuples);
-                Self::try_from_tuple_slice(byte_str_slice)
+                let slice = SliceWithIndices::from_byte_slice(&tuples);
```
robertbastian: nit: `byte_str_slice` was a better name. even though it was just named after the type, it gave more information than "slice"

utils/zerotrie/src/builder/konst/builder.rs
```diff
@@ -105,20 +104,22 @@ impl<const N: usize> ZeroTrieBuilderConst<N> {
...
+                    if !is_less_than(prev, ascii_str) {
+                        panic!("Strings in ZeroTrie constructor are not sorted");
                     }
                 }
             };
-            prev = Some(ascii_str)
-        });
+            prev = Some(ascii_str);
+            i += 1;
+        }
```
robertbastian: move this to the slice type, `assert_sorted`

utils/zerotrie/src/builder/konst/builder.rs
```diff
@@ -128,9 +129,7 @@ impl<const N: usize> ZeroTrieBuilderConst<N> {
     ///
     /// `K` is the stack size of the lengths stack. If you get an error such as
     /// `AsciiTrie Builder: Need more stack`, try increasing `K`.
-    pub const fn from_sorted_const_tuple_slice<const K: usize>(
-        items: ConstSlice<(&ByteStr, usize)>,
-    ) -> Self {
+    pub const fn from_sorted_const_tuple_slice<const K: usize>(items: SliceWithIndices) -> Self {
```
robertbastian: issue: the name is not correct anymore

utils/zerotrie/src/builder/konst/builder.rs
```diff
@@ -155,9 +152,9 @@ impl<const N: usize> ZeroTrieBuilderConst<N> {
         let mut current_len = 0;
         // Start the main loop.
         loop {
-            let item_i = all_items.get_or_panic(i);
-            let item_j = all_items.get_or_panic(j - 1);
-            debug_assert!(item_i.0.prefix_eq(item_j.0, prefix_len));
+            let item_i = all_items.get(i);
+            let item_j = all_items.get(j - 1);
+            debug_assert!(prefix_eq(item_i.0, item_j.0, prefix_len));
```
robertbastian: nit: don't import free functions, qualify with module name

sffc (commented):
This aligns with the design I previously discussed with @Manishearth. I'll take another look after @robertbastian's suggestions are landed, but you don't need to wait for me.

utils/zerotrie/src/builder/bytestr.rs
Manishearth: It needs alloc in order to not be unused.

utils/zerotrie/src/builder/slice_indices.rs
Manishearth: const.

utils/zerotrie/src/builder/konst/builder.rs
```diff
@@ -138,8 +127,11 @@ impl<const N: usize> ZeroTrieBuilderConst<N> {
 
     /// The actual builder algorithm. For an explanation, see [`crate::builder`].
     #[must_use]
-    #[expect(clippy::indexing_slicing)] // lots of indexing, should be in range
-    const fn create_or_panic<const K: usize>(&mut self, all_items: SliceWithIndices) -> usize {
+    #[expect(
+        clippy::indexing_slicing,
+        reason = "panic documented on from_tuple_slice"
```
robertbastian: uhm now I'm confused. is this method only used in one place? if so, I'd prefer inlining so that the panic documentation and the allow are next to one another

utils/zerotrie/src/builder/slice_indices.rs
```diff
@@ -64,7 +62,12 @@ impl<'a> SliceWithIndices<'a> {
         }
     }
 
-    #[allow(clippy::indexing_slicing)]
+    /// Gets the bytes and index at `index`
+    ///
+    /// # Panics
+    ///
+    /// Panics when `index >= self.len()`
```
robertbastian, line 69: hmm, documented panics for internal functions just dissipate, because clippy doesn't force us to allow them at call sites

utils/zerotrie/src/builder/slice_indices.rs
```diff
@@ -91,11 +94,29 @@ impl<'a> SliceWithIndices<'a> {
         }
     }
 
+    pub const fn is_sorted(self) -> bool {
+        let mut i = 0;
+        let mut prev: Option<&'a [u8]> = None;
+
+        while i < self.len() {
+            let (ascii_str, _) = self.get(i);
```
robertbastian: i.e. this should have a comment why it doesn't panic, but it doesn't

utils/zerotrie/src/builder/slice_indices.rs
robertbastian: doesn't need to be const because only ever used in non-const code

utils/zerotrie/src/zerotrie.rs
robertbastian: also everywhere else

utils/zerotrie/src/builder/bytestr.rs
robertbastian: but it's nested in a cfg alloc method already?!

utils/zerotrie/src/builder/konst/builder.rs
Manishearth: Transitively, and it's more than one function. I went ahead and just added panic docs everywhere

utils/zerotrie/src/builder/slice_indices.rs
Manishearth: Yes. Transitive analyses are hard.

We use these lints mostly because in 90% of ICU4X code we actually want no panics period (in release mode), limiting panics to things checked with local invariants. The lints work well for that use case.

However in datagen/zerotrie builder code/etc we are more loose about panics, so you get a lot of stuff like this. We're not trying to be perfect here, just have _some_ amount of breadcrumbs.


People have in the past discussed transitive panic lints in clippy where you have a `clippy::panics` annotation that gets detected. Still won't be perfect, but could be useful.

utils/zerotrie/src/builder/slice_indices.rs
Manishearth: Ah. this probably changed.

utils/zerotrie/src/builder/bytestr.rs
Manishearth: Yes, that was a response to the second half of your comment.

Manishearth: as in, "the outer function needs alloc because of unused lints"

Manishearth:
Yeah, and that's preexisting. This is builder code, by and large we don't care.

utils/zerotrie/src/builder/konst/builder.rs
```diff
@@ -128,21 +118,25 @@ impl<const N: usize> ZeroTrieBuilderConst<N> {
     ///
     /// `K` is the stack size of the lengths stack. If you get an error such as
     /// `AsciiTrie Builder: Need more stack`, try increasing `K`.
-    pub const fn from_sorted_const_tuple_slice<const K: usize>(
-        items: ConstSlice<(&ByteStr, usize)>,
-    ) -> Self {
+    ///
+    /// # Panics
+    ///
+    /// Panics if `items` is not sorted.
```
sffc: Nit: If you are adding panic documentation, please ensure that it is enforced. Previously it said "Assumes that the items are sorted. If they are not, unexpected behavior may occur." but not a panic. If you are changing it to a guaranteed panic, please update the docs and the code.

utils/zerotrie/src/builder/konst/builder.rs
```diff
@@ -128,21 +118,25 @@ impl<const N: usize> ZeroTrieBuilderConst<N> {
...
         let total_size = result.create_or_panic::<K>(items);
         debug_assert!(total_size == result.data.len());
         result
     }
 
     /// The actual builder algorithm. For an explanation, see [`crate::builder`].
+    ///
+    /// # Panics
+    ///
+    /// Panics if `all_items` is not sorted.
```
sffc: Issue: Don't document a panic unless you actually panic in that situation.

utils/zerotrie/src/builder/konst/builder.rs
```diff
@@ -128,21 +118,25 @@ impl<const N: usize> ZeroTrieBuilderConst<N> {
...
     #[must_use]
-    const fn create_or_panic<const K: usize>(
-        &mut self,
-        all_items: ConstSlice<(&ByteStr, usize)>,
-    ) -> usize {
+    #[expect(clippy::indexing_slicing, reason = "documented panic")]
```
sffc: Issue: Your use of indexing_slicing is not related to the documented panic. Please change this to say something more like "all call sites are well-commented and will not panic".

utils/zerotrie/src/builder/slice_indices.rs
sffc: Issue: it was previously called `get_or_panic`. Please change it back.

utils/zerotrie/src/builder/konst/builder.rs
```diff
@@ -192,7 +186,7 @@ impl<const N: usize> ZeroTrieBuilderConst<N> {
                     // A string that equals the prefix does not take part in the branch node.
                     break;
                 }
-                let candidate = candidate.byte_at_or_panic(prefix_len);
+                let candidate = candidate[prefix_len];
```
sffc, line 191: Suggestion: since you are adding more types here, you can avoid the indexing_slicing by matching on an `Option` and merging this with the if statement above.

utils/zerotrie/src/builder/konst/builder.rs
```diff
@@ -170,20 +164,20 @@ impl<const N: usize> ZeroTrieBuilderConst<N> {
             prefix_len -= 1;
             let mut new_i = i;
             let mut new_j = j;
-            let mut ascii_i = item_i.0.byte_at_or_panic(prefix_len);
-            let mut ascii_j = item_j.0.byte_at_or_panic(prefix_len);
+            let mut ascii_i = item_i.0[prefix_len];
+            let mut ascii_j = item_j.0[prefix_len];
```
sffc, line 169: Nit (optional): add a comment saying that `prefix_len` is guaranteed to be shorter than the two strings. It is easy to tell for `item_i.0`. It is a bit more involved to prove it for `item_j.0`; it is a loop invariant.

utils/zerotrie/src/builder/slice_indices.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+    a.len() < b.len()
+}
+
+/// Const function to evaluate `a[..prefix_len] == b[..prefix_len]`
+///
+/// # Panics
+///
+/// Panics if `prefix_len` is longer than either `a` or `b`.
+pub(crate) const fn prefix_eq(a: &[u8], b: &[u8], prefix_len: usize) -> bool {
```
sffc: Nit: in this codebase, please name functions that have panic invariants with `_or_panic`

utils/zerotrie/src/serde.rs
```diff
@@ -2,7 +2,6 @@
 // called LICENSE at the top level of the ICU4X source tree
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
```
sffc, line 4: Grumble to self: I thought I was done reviewing this PR but now I need to rethink how we do serde, too

utils/zerotrie/src/serde.rs
```diff
@@ -45,45 +54,49 @@ impl<'de> Visitor<'de> for ByteStrVisitor {
     }
 }
 
-impl<'data, 'de: 'data> Deserialize<'de> for &'data ByteStr {
+impl<'de> Deserialize<'de> for SerdeByteStrOwned {
```
sffc, line 57: Nit: Please make the diff smaller by putting your impls in the same order that they were before. This one should be the impl for `SerdeByteStr`.

utils/zerotrie/src/serde.rs
```diff
@@ -45,45 +54,49 @@ impl<'de> Visitor<'de> for ByteStrVisitor {
...
+            }
         }
+        serializer.serialize_bytes(bytes)
     }
 }
 
-impl Serialize for &ByteStr {
+struct SerdeByteStr<'a>(&'a [u8]);
+
+impl Serialize for SerdeByteStr<'_> {
```
sffc, line 89: Suggestion: Should this be using `PotentialUtf8` instead? I think that type has these same types of impls. I think PotentialUtf8 came around after ByteStr.

utils/zerotrie/src/zerotrie.rs
```diff
@@ -597,9 +602,10 @@ macro_rules! impl_zerotrie_subtype {
         impl $name<Vec<u8>>
         {
             #[cfg(feature = "serde")]
-            pub(crate) fn try_from_serde_litemap(items: &LiteMap<Box<ByteStr>, usize>) -> Result<Self, ZeroTrieBuildError> {
-                let lm_borrowed: LiteMap<&ByteStr, usize> = items.to_borrowed_keys();
-                Self::try_from_tuple_slice(lm_borrowed.as_slice())
+            pub(crate) fn try_from_serde_litemap(items: &LiteMap<crate::serde::SerdeByteStrOwned, usize>) -> Result<Self, ZeroTrieBuildError> {
+                let tuples: Vec<(&[u8], usize)> = items.iter().map(|(k, v)| (k.as_bytes(), *v)).collect();
+                let byte_str_slice = ByteSliceWithIndices::from_byte_slice(&tuples);
```
sffc, line 607: Observation: `.to_borrowed_keys()` allocates, too, so I think there are no "new" allocations here.

utils/zerotrie/src/builder/konst/builder.rs
Manishearth: I left a TODO for you for j. Might be a good place to just do a checked indexing operation.

utils/zerotrie/src/builder/konst/builder.rs
Manishearth: `get()` isn't stable in const.

utils/zerotrie/src/serde.rs
Manishearth: Interesting idea, I don't want to risk changing behavior in this PR, but we should investigate that as a followup.

utils/zerotrie/src/serde.rs
Manishearth: I had the same feeling working on this; the serde stuff was unexpected.

utils/zerotrie/src/builder/konst/builder.rs
sffc: Suggestion (optional): use a helper function that returns `Option` in const. Internally it uses indexing_slicing with an easily provable invariant. I feel like we already had such a function.

utils/zerotrie/src/builder/konst/builder.rs
```diff
@@ -101,25 +100,12 @@ impl<const N: usize> ZeroTrieBuilderConst<N> {
     ///
     /// `K` is the stack size of the lengths stack. If you get an error such as
     /// `AsciiTrie Builder: Need more stack`, try increasing `K`.
-    ///
-    /// # Panics
-    ///
-    /// Panics if the items are not sorted
```
sffc, line 107: Issue: This fn still panics. Please leave the docs

utils/zerotrie/src/builder/konst/builder.rs
Manishearth: A helper function on what? This is a stdlib slice type.

Do you want a free function here? I think a free function is a bit strange for just one callsite. That just moves the invariant check around.

We should file an issue for cleaning up the indexing here and do that separately.

Manishearth: > We should file an issue for cleaning up the indexing here and do that separately.

Specifically, I think the code here should just always use checked indexing.

utils/zerotrie/src/serde.rs
```diff
@@ -45,45 +44,59 @@ impl<'de> Visitor<'de> for ByteStrVisitor {
...
         } else {
             let s = Vec::<u8>::deserialize(deserializer)?;
-            Ok(ByteStr::from_boxed_bytes(s.into_boxed_slice()))
+            Ok(SerdeByteStrOwned(s.into_boxed_slice()))
         }
     }
 }
 
-impl Serialize for &ByteStr {
+impl Serialize for SerdeByteStrOwned {
```
sffc, line 72: Observation: Before it seems we didn't have `impl Serialize for Box<ByteStr>`, only for the borrowed version? Why do we need it now when we didn't need it before?

utils/zerotrie/src/serde.rs
```diff
@@ -45,45 +44,59 @@ impl<'de> Visitor<'de> for ByteStrVisitor {
...
 
-impl<'data, 'de: 'data> Deserialize<'de> for &'data ByteStr {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+struct SerdeByteStr<'a>(&'a [u8]);
+
+impl Serialize for SerdeByteStr<'_> {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
-        D: Deserializer<'de>,
+        S: Serializer,
```
sffc: Nit: The impls still aren't in the same order as before

Manishearth: They are. We just don't need the `&'data ByteStr` deserialize impl. We weren't doing zero copy deserialization anyway.

utils/zerotrie/src/serde.rs
Manishearth: Because `Box<T>` has a forwarding Serialize impl, but `SerdeByteStrOwned` does not

utils/zerotrie/src/serde.rs
Manishearth: Oh, I guess I can make them line up better if I reorder the borrowed and owned ones.

sffc:
I care about getting deserialization right because, well, we actually use ZeroTries in production in serde-like ways. I'm not sure whether we use these impls, but we could land code that uses them at any time.

Manishearth:
Like I said, we weren't doing a zero copy deserialization anyway. The deserialization code for ZeroTries uses a completely different codepath for the borrowed case: it's not using ByteStr. ByteStr is specifically a helper for the human readable case.

Go ahead and try it, you can remove that impl on main and the code still compiles.

> I'm not sure whether we use these impls, but we could land code that uses them at any time

Great, and we can add them again. But I think it's unlikely, for the reason I gave above.

This is an internal serialization helper type: me removing this impl does not change the Deserialize impls available on LiteMaps, if that is your concern.

PR #7750, opened 2026-03-09
In component crates, limit use of the icu crate to doctests

musicinmybrain:
As maintainer of most of the ICU4X stack in Fedora, this practice makes it easier for me to run as many tests as possible without downstream patching, and to avoid dependency cycles that interfere with bootstrapping. This is a follow-up to https://github.com/unicode-org/icu4x/pull/7742, where the idea was discussed in more detail, and in which @robertbastian indicated that this change would be welcome.

I tried to cover *all* of the crates, not just the ones we have packaged in Fedora so far. Each commit was tested with `cargo test --all-features` in the directory corresponding to the crate in question.

### DIFF
--- a/components/segmenter/tests/spec_test.rs
+++ b/components/segmenter/tests/spec_test.rs
@@ -117,7 +117,7 @@ fn line_break_test(file: &'static str) {
             test.break_result_utf8.insert(0, 0);
         }
         if result != test.break_result_utf8 {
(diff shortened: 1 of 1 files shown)


musicinmybrain:
Thank you!

sffc (commented):
This is fine and good, but do we have a story to prevent regressing?

PR #7751, opened 2026-03-09
Remove circular meta crate dependency

robertbastian:
It's very annoying how all our tests depend on *everything*, which leads to long build times during local iteration.

This is a proof-of-concept of how we could remove this dependency.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2028,7 +2028,6 @@ dependencies = [
  "chrono-tz",
  "databake",
  "displaydoc",
- "icu",
  "icu_calendar",
  "icu_datetime",
  "icu_locale_core",
--- a/components/time/Cargo.toml
+++ b/components/time/Cargo.toml
@@ -43,7 +43,6 @@ chrono-tz = { workspace = true, optional = true }
 icu_time_data = { workspace = true, optional = true }
 
 [dev-dependencies]
-icu = { path = "../../components/icu", default-features = false }
 icu_datetime = { path = "../../components/datetime", features = ["compiled_data"] }
 icu_provider_blob = { path = "../../provider/blob" }
 tinystr = { workspace = true }
--- a/components/time/src/ixdtf.rs
+++ b/components/time/src/ixdtf.rs
@@ -56,6 +56,7 @@ pub enum ParseError {
     ///
     /// # Example
     /// ```
+    #[doc = crate::setup_metacrate!(time, calendar)]
     /// use icu::calendar::Iso;
     /// use icu::time::{ZonedDateTime, ParseError, zone::IanaParser};
     ///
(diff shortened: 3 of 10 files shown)


Manishearth:
Clever hack. Not opposed.

sffc (commented):
In my experience, the metacrate isn't the bottleneck; it's that changing the docstests causes the library crate to get dirty, since they are in the same file, and then it needs to be rebuilt along with its data before you can run the docstests. I've been trying to use integration tests instead of unit tests more for this reason, but there are reasons like `test_all_cals!` why I can't use integration tests as much as I would like.

hsivonen:
Do we need a macro to set up doctests that use `icu::foo`? Wouldn't it be better to change the docs to use `icu_foo`, since it's very annoying to have to adjust copypaste from the docs to replace `::` with `_`. Using the metacrate in an app is an anti-pattern for both the build time and vendoring footprint of the app.

I don't recall why we have the meta crate, but it seems bad to guide app developers towards the meta crate.

hsivonen:
FWIW, in Firefox, we even have [tooling to explicitly prohibit the introduction of the meta crate](https://searchfox.org/firefox-main/rev/7b1a35d6c70c08d2966c6c8dfc9aa9c04319f721/python/mozbuild/mozbuild/vendor/vendor_rust.py#86).

robertbastian:
> it's that changing the docstests causes the library crate to get dirty, since they are in the same file, and then it needs to be rebuilt along with its data before you can run the docstests.

and with the metacrate, that also causes the metacrate to get dirty, and that depends on everything else which might be dirty if e.g. baked data was regenerated

> Do we need a macro to set up doctests that use `icu::foo`? Wouldn't it be better to change the docs to use `icu_foo`,

So far we've viewed using the individual crates as more of a power user feature.

Manishearth:
Yeah, I think early on in ICU4X's lifecycle we decided that we wanted to have our primary advertised API to be the "single library". I did consider that to be not super Rusty (most Rust users will prune deps where possible, dependency bloat is hard enough already), but it wasn't an irreversible decision so I thought it would be fine. ICU4C's single library approach makes it easier to talk about.

Given the experience of having the library bake for many years, I think it might be time to revisit that decision. But I don't think it's urgent.

One thing to note is that the `icu::` crate docs are really nice to navigate the entire library, and if those docs mostly mention _other_ crates in their examples it might get frustrating for users.

But users visiting `icu_foo` docs have the same problem in the other direction with the `icu::` imports anyway, and we're encouraging worse practice by having people import the metacrate.

robertbastian:
Do I close or merge this

Manishearth:
I think we should not do this for 2.2, but I'd like to have the broader discussion _at some point_ about whether we should still advertise the metacrate as the primary way of using ICU4X.

So probably won't merge soon, you could close and file a discussion issue or leave open/draft.

PR #7753, opened 2026-03-09
Make min_months invariant no longer a hard invariant for Chinese/Korean

Manishearth:
This is one way of resolving (some of?) @robertbastian's concerns from https://github.com/unicode-org/icu4x/pull/7739

### DIFF
--- a/components/calendar/src/calendar_arithmetic.rs
+++ b/components/calendar/src/calendar_arithmetic.rs
@@ -189,7 +189,7 @@ pub(crate) enum MinMonths {
     ///
     /// It is ok for ICU4X to `debug_assert` on things deriving from the correctness of this calculation
     Guaranteed(i64),
-    /// This number is not guaranteed. ICU4X should check that it is actually in range, and if not,
+    /// This number is not guaranteed. Code should check that it is not too large, and if not,
     /// fall back to the `guarantee` value (which *is* guaranteed).
     Guessed { guess: i64, guarantee: i64 },
 }


robertbastian:
I don't think a debug assert that will be hit by some poor caller at some point is a solution

Manishearth:
@robertbastian I'm confused.

We currently have a debug assertion. The invariant is proven for all calendars except Chinese/Dangi. This PR opts those two calendars out of the assertion.

Are you worried about the correctness of the assertion for non-Chinese?

Manishearth:
Though in my code I did accidentally invert the `if`, fixed now

Manishearth:
I could remove the assertion completely until we have tests for this. But I think the Hebrew/Ethiopic math is fine and well proven.

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -1005,13 +1010,18 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
...
+                let mut min_months = self.min_months_from(min_years);
+                if self.surpasses(
                     other,
                     DateDuration::from_signed_ymwd(years, min_months, 0, 0),
                     sign,
                     cal,
-                ));
+                ) {
+                    // TODO(#7077) We should try to have a hard invariant here for everyone.
+                    debug_assert!(matches!(cal.calendar_algorithm(), Some(CalendarAlgorithm::Chinese | CalendarAlgorithm::Dangi)),
```
robertbastian: I don't like code that's conditional on the calendar algorithm, this is likely to be missed when we stabilise the rules trait. We could make `min_months_from` return an enum that is `Guaranteed` or `Guessed` or something. that would also avoid the extra surpasses call for other calendars

robertbastian: actually, could the surpasses check be moved inside the EAT implementation? then it would satisfy the method invariant, and this code wouldn't have to special-case a calendar

Manishearth: clever idea, but some of the strategies for optimizing surpasses would make it harder to cache there

I think Guarantee/Guess is a cheap and easy option!

robertbastian: actually you shouldn't fall back to `sign`, we have a much better lower bound: `years * 12`. So `Guess` should return a guess and a guaranteed bound to fall back to

Manishearth: Good idea. Done

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -1005,11 +1020,25 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
...
+                let min_months = match self.min_months_from(min_years) {
+                    MinMonths::Guaranteed(m) => m,
+                    // In case it's a guess, check that the guess is in range,
+                    // and if it's not, return the guarantee
+                    MinMonths::Guessed { guess, guarantee } => {
+                        if self.surpasses(
+                            other,
+                            DateDuration::from_signed_ymwd(years, guess, 0, 0),
+                            sign,
+                            cal,
```
sffc, line 1037: Nit: Combine this with the main `surpasses` call to reduce the number of call sites and reduce code size if the compiler decides to inline it. Also, `surpasses` is expensive, so we shouldn't call it more than we need to.

sffc (commented):
I would rather put the max-3-year-gap-between-leap-months requirement on the Rules trait than add this guessing thing, but 🤷 as long as it doesn't impact performance.

If we implement the guessing thing, then I feel like we might want to support "guess that is expected to sometimes overshoot", "guess that should be below but we're just not 1000% sure", and "guess that is guaranteed to be below"

components/calendar/src/calendar_arithmetic.rs
Manishearth: I doubt the compiler will inline it, it's called 4 times in this function.

What do you mean by "combine this with the main surpasses call"? This is somewhat deep within a conditional, and it's only going to get more complicated. I'm not really sure what control flow you are envisioning here.

I don't really wish to complicate this code.

Manishearth:
> If we implement the guessing thing, then I feel like we might want to support "guess that is expected to sometimes overshoot", "guess that should be below but we're just not 1000% sure", and "guess that is guaranteed to be below"

Yeah I considered that for a bit, I think it's a promising route. Then for Chinese we can just pick the lunation ration and write a small backoff routine. I was previously proposing a binary search style thing anyway, that would be similar.

components/calendar/src/calendar_arithmetic.rs
sffc: The code is complicated; that ship is sailing even more with this PR

I was thinking you would make a new stateful local variable of type `Option<u32>` (or whatever the type is). If surpasses returns `false`, clear the option. If it returns `true` and the option is there, reset the candidate months to the option value, clear the option, and continue the loop.

```rust
let mut candidate_months = sign;
let mut min_months = None;
{
    // ... logic to get the min months ...
    candidate_months = calendar_guess_months;
    min_months = Some(calendar_min_months);
}
loop {
    let surpasses = // ...
    if surpasses {
        if let Some(min_months) = min_months.take() {
            candidate_months = min_months;
            continue;
        }
        break;
    } else {
        months = candidate_months;
        candidate_months += sign;
        min_months.clear();
    }
}
```

Manishearth: > The code is complicated; that ship is sailing even more with this PR

Yes, and I'd like to not do that without more motivation. I don't think "the compiler might choose to inline it" is compelling.

This PR does not change the shape of the code too much, the actual specced code is untouched. In all of my changes here so far I have been trying to optimize for keeping the _specced_ part of the code looking the same, with some additional non-specced code in between steps. This is a considered choice I have been making, and I don't want to go against that here, _especially_ not until the rest of the refactors happen here.

I would prefer to land this as is.

I pushed some comments highlighting the non-spec optimization segmetns. As you can see, they are pretty well separated from the rest of it, they are each a simple block of code that can be (sort of) skipped over when comparing against the spec.

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -183,6 +183,17 @@ impl PackWithMD for i32 {
     }
 }
 
+/// A lower bound for the number of months in `years` years, for use in `until`.
+pub(crate) enum MinMonths {
+    /// This number is guaranteed to be correct based on documented checking of invariants
+    ///
+    /// It is ok for ICU4X to `debug_assert` on things deriving from the correctness of this calculation
```
robertbastian, line 190: this is a fairly pointless comment, of course if it's "guaranteed to be correct" we can assume it is correct

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -183,6 +183,17 @@ impl PackWithMD for i32 {
...
+    Guaranteed(i64),
+    /// This number is not guaranteed. ICU4X should check that it is actually in range, and if not,
```
```suggestion
    /// This number is not guaranteed. Code should check that it is not too large, and if not,
```

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -242,9 +253,13 @@ pub(crate) trait DateFieldsResolver: Calendar {
     /// is bounded by a constant with respect to `years`.
     ///
     /// The default impl is for non-lunisolar calendars with 12 months!
+    ///
+    /// `until()` will debug assert if this ever returns a value greater than the
```
robertbastian, line 257: I don't like documenting debug assert behaviour. I should be allowed to add debug assertions anywhere at any time, these are not API contracts

I think this behaviour is sufficiently documented on the `MinMonths` type that it doesn't have to be repeated here.

Manishearth:
I am merging this since this addresses open comments. @sffc can continue the discussion with me if he disagrees with calling `surpasses` twice the way we do now.

PR #7755, opened 2026-03-10
Change nesting of `atTime` and zone `appendItems`

robertbastian:
Follow up https://github.com/unicode-org/icu4x/pull/7416

The time zone should be appended to the time, not to the joined `atTime` result: https://unicode-org.atlassian.net/browse/CLDR-19226


Manishearth:
Not my wheelhouse but can secondary review if needed.

provider/source/src/datetime/names.rs
```diff
@@ -530,7 +530,9 @@ fn datetimepattern_convert(
         GlueType::DateTime => at_time.parse(),
         GlueType::DateZone => append_tz.interpolate_to_string(["{1}", "{2}"]).parse(),
         GlueType::TimeZone => append_tz.interpolate_to_string(["{0}", "{2}"]).parse(),
-        GlueType::DateTimeZone => append_tz.interpolate_to_string([at_time, "{2}"]).parse(),
+        GlueType::DateTimeZone => at_time
+            .replace("{0}", &append_tz.interpolate_to_string(["{0}", "{2}"]))
+            .parse(),
```
sffc, line 535: Suggestion (optional): use the pattern parser instead of string replace

```suggestion
        GlueType::DateTimeZone => at_time
            .parse::<DoublePlaceholderPattern>()
            .interpolate_to_string([&append_tz.interpolate_to_string(["{0}", "{2}"]), "{1}"])
            .parse(),
```

robertbastian: `icu_pattern::Pattern` does not implement `FromStr`, its construction is more complicated

PR #7759, opened 2026-03-10
Polish compact decimal API

robertbastian:
https://github.com/unicode-org/icu4x/issues/3647

## Changelog: N/A


Manishearth:
Changelog is fine because experimental. In the final 2.2 changelog we'll probably have _something_ about compact decimal.

PR #7761, opened 2026-03-10
Add docs about input types for dynamic field sets

sffc:
#7758

Depends on #7763

CC @rizzen-yazston

## Changelog

N/A


rizzen-yazston:
The expanded table is very clear on what is supported. Thanks for this documentation improvement.

sffc:
Replaced by #7842

Issue #7512, opened 2026-01-27
Consider replacing Date::try_new_from_codes with something better


Manishearth:
[`Date::try_new_from_codes`](https://unicode-org.github.io/icu4x/rustdoc/icu/calendar/struct.Date.html#method.try_new_from_codes) isn't the best API, for two reasons:

 - We're not happy with its return type and can do better, see https://github.com/unicode-org/icu4x/issues/7010#issuecomment-3802706892
 - It takes in a MonthCode, when we now have a very nice type-safe non-stringy [`Month`](https://unicode-org.github.io/icu4x/rustdoc/icu/calendar/types/struct.Month.html) type now

Would be good to try and replace it.

 - What should it be called? `new_with_era_and_month`?
 - What should its returned error be? Probably a subset of DateFromFieldsError based on whatever we decide in https://github.com/unicode-org/icu4x/issues/7010?
 - Are there any other changes we should make to the API?

cc @robertbastian @sffc

Manishearth:
Marking as Discuss but I would like to see some discussion here first.

sffc:
We should use this to reconsider the shape in other ways, too. I want to position the replacement of try_new_from_codes to be the convenient/ergonomic API, but it doesn't need all of the functionality since we can load everything else we need into `try_from_fields`. For example, we could drop the `era` option entirely and always take the extended year. Clients who need to provide eras in a cross-calendar way can use `try_from_fields`.

Manishearth:
Hmm, an interesting proposition. I somewhat _like_ the era input: it's optional so you can still use extended year, but if you have an era date you can provide it. OTOH `try_from_fields` works but it has more error cases to consider. But we can consider this. Weak preference for keeping era around as an optional parameter.

sffc:
I don't like how when I read `Date::try_new_from_codes(None, ...)`, I have no idea what `None` means.

We could do `Date::try_new_bikeshed(YearType::Extended(...), ...)` but that requires importing an extra type and it gets verbose. Maybe we should consider `Date::try_new_bikeshed(123.into(), ...)` ??

sffc:
We've traditionally been allergic to using trait bounds in these functions (can you remind me the list of why not?), but since we have date_from_fields for power users, maybe we can make the year type be `impl Into<YearType>` or something, and then you can pass `2025` or `("reiwa", 6)`, etc.

Manishearth:
[...]

PR #7764, opened 2026-03-10
Add Date::try_from_codes, the replacement of Date::try_new_from_codes

Manishearth:
Fixes https://github.com/unicode-org/icu4x/issues/7512

<s>I tried to upgrade the error type used internally, which works out well, but it means that we lose context when producing RangeErrors. If desired, I could instead introduce a complex conversion so that that behavior is preserved.</s> The original behavior is now preserved.

First three commits are agent generated.

I have marked this API as insta-stable since we planned it for 2.2. I'm about to start gearing up to unflag these APIs, so no point making this unstable now.

## Changelog

icu_calendar: Add `Date::try_from_codes`, which replaces `Date::try_new_from_codes`, and takes typed year/month values.
  - New methods: `Date::try_from_codes` (and primarily-internal `Calendar::from_codes2`)
  - New types: `InputYear`, `DateFromCodesError`

### DIFF
--- a/components/calendar/src/error.rs
+++ b/components/calendar/src/error.rs
@@ -645,7 +645,7 @@ pub enum DateFromCodesError {
     InvalidEra,
 
     /// The year was out of range
-    #[displaydoc("Invalid year for `Date::try_from_codes()`, must be within -9999..=9999")]
+    #[displaydoc("Invalid year, must be within -9999..=9999")]
     InvalidYear,
 }
 


components/calendar/src/calendar.rs
```diff
@@ -43,14 +47,57 @@ pub trait Calendar: crate::cal::scaffold::UnstableSealed {
...
+    }
+
+    /// Construct a date from month codes and [`InputYear`].
+    ///
+    /// This is used by [`Date::try_from_codes()`].
+    #[expect(clippy::wrong_self_convention)]
+    fn from_codes2(
+        &self,
+        year: InputYear,
+        month_code: types::MonthCode,
```
robertbastian: why `MonthCode`?

components/calendar/src/types.rs
```diff
@@ -248,6 +251,28 @@ impl fmt::Debug for DateFields<'_> {
     }
 }
 
+/// Year information to be used as input to [`Date::try_from_codes()`].
+///
+/// Note: The input variants represent different ways of specifying a year;
+/// see [`YearInfo`] for the output formats.
+#[derive(Copy, Clone, Debug, PartialEq)]
+#[non_exhaustive]
+pub enum InputYear<'a> {
```
robertbastian: can we bikeshed this? `InputYear` is inconsistent with `Month` as the other input type.

components/calendar/benches/date.rs
```diff
@@ -168,7 +168,7 @@ fn date_benches(c: &mut Criterion) {
         "calendar/hebrew",
         &fxs,
         icu::calendar::cal::Hebrew,
-        |y, m, d, c| Date::try_new_from_codes(None, y, types::Month::new(m).code(), d, c).unwrap(),
+        |y, m, d, c| Date::try_from_codes(y.into(), types::Month::new(m), d, c).unwrap(),
```
robertbastian: can we please be explicit here

```suggestion
        |y, m, d, c| Date::try_from_codes(types::InputYear::ExtendedYear(y), types::Month::new(m), d, c).unwrap(),
```

components/calendar/src/types.rs
```diff
@@ -248,6 +251,28 @@ impl fmt::Debug for DateFields<'_> {
...
+    /// An "extended year", which is a single number representing the year.
+    ///
+    /// For many calendars this matches the year number, but for some it is a year
+    /// relative to some specific epoch.
+    ExtendedYear(i32),
```
robertbastian: `InputYear::ExtendedYear` is repetitive, it should just be `InputYear::Extended`

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -312,33 +312,44 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
...
+                calendar
+                    .extended_year_from_era_year_unchecked(era.as_bytes(), y)
+                    .map_err(|_| DateFromCodesError::InvalidEra)?
+            }
         };
         let year = calendar.year_info_from_extended(extended_year);
 
         let validated = Month::try_from_utf8(month_code.0.as_bytes()).map_err(|e| match e {
-            MonthCodeParseError::InvalidSyntax => DateError::UnknownMonthCode(month_code),
+            MonthCodeParseError::InvalidSyntax => DateFromCodesError::MonthNotInCalendar,
```
robertbastian: this is not the correct error type

components/calendar/src/error.rs
```diff
@@ -611,6 +621,41 @@ mod unstable {
...
+    MonthNotInCalendar,
+    /// The specified month exists in this calendar, but not in the specified year.
+    #[displaydoc("The specified month exists in this calendar, but not for this year")]
+    MonthNotInYear,
+    /// The era code is invalid for the calendar.
+    #[displaydoc("Unknown era or invalid syntax")]
+    InvalidEra,
+    /// The date is out of range.
+    #[displaydoc("Result out of range")]
+    Overflow,
```
robertbastian: Overflow is for rata-die range errors, which is not the right error case here. According to https://github.com/unicode-org/icu4x/issues/7010#issuecomment-3892880221 this should be `InvalidYear`.

components/calendar/src/types.rs
robertbastian: also in the discussion we used `YearInput`

components/calendar/src/types.rs
```diff
@@ -248,6 +251,28 @@ impl fmt::Debug for DateFields<'_> {
...
+///
+/// Note: The input variants represent different ways of specifying a year;
+/// see [`YearInfo`] for the output formats.
+#[derive(Copy, Clone, Debug, PartialEq)]
+#[non_exhaustive]
+pub enum InputYear<'a> {
+    /// An "extended year", which is a single number representing the year.
+    ///
+    /// For many calendars this matches the year number, but for some it is a year
+    /// relative to some specific epoch.
```
robertbastian: this is a confusing sentence, the year number is always relative to some specific epoch, even for era years. link to our existing extended year documentation on `YearInfo::extended_year`.

components/calendar/src/types.rs
```diff
@@ -248,6 +251,28 @@ impl fmt::Debug for DateFields<'_> {
...
+    ExtendedYear(i32),
+    /// A year specified by an era code and the year in that era.
+    EraYear(&'a str, i32),
+}
+
+impl From<i32> for InputYear<'_> {
```
robertbastian: this implementation makes me want to veto the whole `YearInput` approach. `y.into()` is not clearer than `None, y`

ffi/capi/src/date.rs
```diff
@@ -406,15 +408,19 @@ pub mod ffi {
             day: u8,
             calendar: &Calendar,
         ) -> Result<Box<Date>, CalendarError> {
-            let era = if !era_code.is_empty() {
-                Some(core::str::from_utf8(era_code).map_err(|_| CalendarError::UnknownEra)?)
+            use icu_calendar::types::InputYear;
+            let input_year = if !era_code.is_empty() {
```
robertbastian, line 412: huh, this already has the empty string behaviour that we discarded in the issue. I think keeping this consistent over FFI is an advantage of that option

ffi/capi/src/date.rs
```diff
@@ -406,15 +408,19 @@ pub mod ffi {
             day: u8,
             calendar: &Calendar,
         ) -> Result<Box<Date>, CalendarError> {
```
robertbastian, line 410: observation: this was a different API than the Rust API before, and is a different API than the Rust API after

ffi/capi/src/errors.rs
```diff
@@ -285,6 +286,20 @@ impl From<icu_calendar::error::DateFromFieldsError> for CalendarDateFromFieldsEr
     }
 }
 
+#[cfg(feature = "calendar")]
+impl From<icu_calendar::error::DateFromCodesError> for CalendarError {
```
robertbastian, line 290: observation: we're still forced to use the old error type

components/calendar/src/calendar.rs
Manishearth: ... oops.

Manishearth: Leftover in the step by step refactor. Shouldn't be hard to fix

components/calendar/src/types.rs
Manishearth: Yeah I was just matching the discussion. I agree about the naming.

components/calendar/src/types.rs
Manishearth: In the discussion we used both. I wasn't sure which one we picked.

components/calendar/src/types.rs
Manishearth: This was already touched on in the discussion, and I don't see any opposition then.

https://github.com/unicode-org/icu4x/issues/7512#issuecomment-3968827955

yes, this is not your preferred option. We should not be rehashing those discussions here.

components/calendar/src/calendar_arithmetic.rs
Manishearth: Got fixed by this code getting moved out.

components/calendar/src/types.rs
Manishearth: I switched it since we were mostly saying YearInput. Start a discussion on the issue if you want to bikeshed it, I am trying to implement the discussion consensus here and don't want to end up ping ponging between reviewers who disagree.

components/calendar/benches/date.rs
Manishearth: Going to try and update test code to this at least.

components/calendar/src/calendar.rs
```diff
@@ -88,14 +92,14 @@ pub trait Calendar: crate::cal::scaffold::UnstableSealed {
         }
     }
 
-    /// Construct a date from month codes and [`InputYear`].
+    /// Construct a date from a [`Month`] and [`YearInput`].
```
robertbastian: ... and a day

components/calendar/src/error.rs
```diff
@@ -651,9 +643,18 @@ pub enum DateFromCodesError {
     /// The era code is invalid for the calendar.
     #[displaydoc("Unknown era or invalid syntax")]
     InvalidEra,
-    /// The date is out of range.
-    #[displaydoc("Result out of range")]
-    Overflow,
+
+    /// The year was out of range
+    #[displaydoc("Invalid year for `Date::try_from_codes()`, must be within -9999..=9999")]
```
robertbastian: nit: none of the other display strings say "for `Date::try_from_codes()`"

components/calendar/src/types.rs
Manishearth: FWIW: I think it's fine to realize you care about something more than previously after seeing it in practice. But I want to highlight that I am implementing based on the existing discussion, and this was already brought up.

So if you want, perhaps open a targeted discussion on that issue about the Into impl. Overall I think it's pretty clear, and nice to have. Most calendars don't care about eras.

Manishearth: But I do see the argument "why not also do it for Month", which is fair.

ffi/capi/src/date.rs
Manishearth: good observation, though I think with FFI we are forced to avoid typed APIs so I don't consider this a strong signal

components/calendar/src/types.rs
robertbastian: If we also do this for months, and embrace the `into`, this API could become a lot less code-y:

```rust

pub fn try_new(y: impl Into<YearInput>, m: impl Into<Month>, d: u8, c: C) -> ...

Date::try_new(2025, 8, 10, Gregorian);

Date::try_new(YearInput::EraYear("bce", 843), 10, 4, Julian);

Date::try_new(5038, Month::leap(2), 4, Hebrew);
```

Manishearth: Hmmmmmmmmmmmm

interesting prospect. I don't hate it, but I also don't like having `impl Into` on APIs (it pessimizes code size pretty badly)

https://doc.rust-lang.org/unstable-book/language-features/doc-notable-trait.html helps with this but not quite with Into.

Manishearth: One of the reasons I think Into on month is worse is that it implicitly assumes that it is a normal month, which isn't a big deal but worth highlighting.

But I just realized we already decided to call the API Month::new not new_normal, so we're okay with that.

So Into on both makes sense. Maybe should file a followup issue about both.

Manishearth: I have filed https://github.com/unicode-org/icu4x/issues/7769

components/calendar/src/error.rs
Manishearth: Deliberate choice, since this is a from-codes-specific range, unlike the others.

components/calendar/src/types.rs
Manishearth: (FTR I think `Year` is fine but I don't want to make changes I'll then get told to undo)

components/calendar/src/error.rs
robertbastian: Right, but this is a from-codes-specific error already

PR #7765, opened 2026-03-11
Note leap month behavior

Manishearth:
Leftover note from Markus' review

## Changelog: N/A

### DIFF
--- a/components/calendar/src/types.rs
+++ b/components/calendar/src/types.rs
@@ -475,6 +475,10 @@ impl fmt::Display for MonthCode {
 /// * `Month::new(7)` = `M07`
 /// * `Month::leap(2)` = `M02L`
 ///
+/// Note that in Temporal's model, "leap month 2" (M02L) means the month that occurs after the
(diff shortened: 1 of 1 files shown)


components/calendar/src/types.rs
```diff
@@ -475,6 +475,10 @@ impl fmt::Display for MonthCode {
 /// * `Month::new(7)` = `M07`
 /// * `Month::leap(2)` = `M02L`
 ///
+/// Note that in Temporal's model, "leap month 2" (M02L) means the month that occurs after the
```
robertbastian, line 478:
```suggestion
/// Note that in this model, "leap month 2" (M02L) means the month that occurs after the
```

robertbastian: also move this to after the first paragraph

Manishearth: Then we can't use month codes in this note.

But sure.

robertbastian: we can use the API

PR #7767, opened 2026-03-11
Robert's feedback on month docs

Manishearth:
https://github.com/unicode-org/icu4x/pull/7765#discussion_r2917097813


components/calendar/src/types.rs
```diff
@@ -467,6 +467,10 @@ impl fmt::Display for MonthCode {
 /// (`Month::new(7)`) might be the 7th or 8th month of the year, depending if the month
 /// Adar was repeated or not.
 ///
+/// In this model, "leap month 2" means the month that occurs after the
+/// non-leap month numbered 2, even if the calendar considers the month to be a variant of
```
```suggestion
/// In this model, `Month::leap(2)` is the month that occurs after `Month::new(2)`, even if the calendar considers the month to be a variant of
```

PR #7768, opened 2026-03-11
CodePointTrie support for normalizer and collator perf improvements

hsivonen:
Split out of #7526 and #7600. The code here needs to be published to crates.io, before those changes can land, because the `utf8_iter` and `utf16_iter` crates need to depend on a version `icu_collections` that has this code on crates.io.

## Changelog

icu_collections: Add `CodePointTrie` getters for fusing lookup into iterating over text: getters by Latin1, ASCII, two-byte UTF-8, and three-byte UTF-8.
  - New methods: `get8()`, `get7()`, `get_utf8_two_byte()`, `get_utf8_three_byte()` on `CodePointTrie` and `TypedCodePointTrie`
  - New trait: `AbstractCodePointTrie`

icu_collections: Serde and databake support for typed `CodePointTrie`s
 - New impls: `databake::Bake`, `BakeSize`, `serde::Serialize`, `serde::Deserialize` for `FastCodePointTrie` and `SmallCodePointTrie`

icu_collections: Iterators by `char` and `TrieValue` pairs for Latin1,`str`, and delegate iterator over `char`.
 - New types: `CharIndicesWithTrie`, `CharIndicesWithTrieDefaultForAscii`, `CharIterWithTrie`, `CharsWithTrie`, `CharsWithTrieDefaultForAscii`, `CharsWithTrieEx`, `Latin1CharIndicesWithTrie`, `Latin1CharsWithTrie`,
 - New trait: `WithTrie`
 - New extension traits on `str`: `CharsWithTrieDefaultForAsciiEx`, `Latin1CharsWithTrieEx`

### DIFF
--- a/components/collections/src/codepointinvliststringlist/mod.rs
+++ b/components/collections/src/codepointinvliststringlist/mod.rs
@@ -175,16 +175,11 @@ impl<'data> CodePointInversionListAndStringList<'data> {
 
     /// See [`Self::contains_str`]
     pub fn contains_utf8(&self, s: &[u8]) -> bool {
-        use utf8_iter::Utf8CharsEx;
-        let mut chars = s.chars();
-        if let Some(first_char) = chars.next() {
-            if chars.next().is_none() {
-                return self.contains(first_char);
-            }
+        if let Ok(well_formed) = core::str::from_utf8(s) {
+            self.contains_str(well_formed)
+        } else {
+            false
         }
-        self.str_list
-            .binary_search_by(|t| t.as_bytes().cmp(s))
-            .is_ok()
     }
 
     ///
--- a/components/collections/src/codepointinvliststringlist/mod.rs
+++ b/components/collections/src/codepointinvliststringlist/mod.rs
@@ -175,11 +175,25 @@ impl<'data> CodePointInversionListAndStringList<'data> {
 
     /// See [`Self::contains_str`]
     pub fn contains_utf8(&self, s: &[u8]) -> bool {
-        if let Ok(well_formed) = core::str::from_utf8(s) {
-            self.contains_str(well_formed)
-        } else {
-            false
+        let single_char = s.get(0).and_then(|&first_byte| {
+            if first_byte.is_ascii() && s.len() == 1 {
+                return Some(first_byte as char);
+            }
+            let utf8_len = first_byte.leading_ones() as usize;
+            if utf8_len != s.len() {
+                return None;
+            }
+
+            str::from_utf8(&s[..utf8_len]).ok()?.chars().next()
+        });
+
+        if let Some(single_char) = single_char {
+            return self.contains(first_char);
         }
+
+        self.str_list
+            .binary_search_by(|t| t.as_bytes().cmp(s))
+            .is_ok()
     }
 
     ///
--- a/components/collections/src/codepointinvliststringlist/mod.rs
+++ b/components/collections/src/codepointinvliststringlist/mod.rs
@@ -188,7 +188,7 @@ impl<'data> CodePointInversionListAndStringList<'data> {
         });
 
         if let Some(single_char) = single_char {
-            return self.contains(first_char);
+            return self.contains(single_char);
         }
 
         self.str_list
--- a/components/collections/src/codepointtrie/cptrie.rs
+++ b/components/collections/src/codepointtrie/cptrie.rs
@@ -385,6 +385,12 @@ impl<'trie, T: TrieValue> CodePointTrie<'trie, T> {
             return Err(Error::DataTooShortForFastAccess);
         }
 
+        // The builder is supposed to support direct indexing to the data array
+        // by ASCII.
+        if data.len() < 128 {
+            return Err(Error::DataTooShortForAsciiAccess);
+        }
+
         // Invariant upheld for `data`: If we got this far, the length of `data`
         // satisfies `data`'s length invariant on the assumption that the contents
         // of `fast_index` subslice of `index` and `header.trie_type` will not
@@ -395,7 +401,9 @@ impl<'trie, T: TrieValue> CodePointTrie<'trie, T> {
 
     /// Turns this trie into a version whose trie type is encoded in the Rust type.
     #[inline]
-    pub fn to_typed(self) -> Typed<FastCodePointTrie<'trie, T>, SmallCodePointTrie<'trie, T>> {
+    pub const fn to_typed(
+        self,
+    ) -> Typed<FastCodePointTrie<'trie, T>, SmallCodePointTrie<'trie, T>> {
         match self.header.trie_type {
             TrieType::Fast => Typed::Fast(FastCodePointTrie { inner: self }),
             TrieType::Small => Typed::Small(SmallCodePointTrie { inner: self }),
@@ -577,6 +585,16 @@ impl<'trie, T: TrieValue> CodePointTrie<'trie, T> {
         );
 
         let bit_prefix = (code_point as usize) >> FAST_TYPE_SHIFT;
+        let bit_suffix = (code_point & FAST_TYPE_DATA_MASK) as usize;
+        self.get_bit_prefix_suffix_assuming_fast_index(bit_prefix, bit_suffix)
+    }
+
+    #[inline(always)]
+    unsafe fn get_bit_prefix_suffix_assuming_fast_index(
+        &self,
+        bit_prefix: usize,
+        bit_suffix: usize,
+    ) -> T {
         debug_assert!(bit_prefix < self.index.len());
         // SAFETY: Relying on the length invariant of `self.index` having
         // been checked and on the unchangedness invariant of `self.index`
@@ -584,7 +602,6 @@ impl<'trie, T: TrieValue> CodePointTrie<'trie, T> {
         let base_offset_to_data: usize = usize::from(u16::from_unaligned(*unsafe {
             self.index.as_ule_slice().get_unchecked(bit_prefix)
         }));
-        let bit_suffix = (code_point & FAST_TYPE_DATA_MASK) as usize;
         // SAFETY: Cannot overflow with supported (32-bit and 64-bit) `usize`
         // sizes, since `base_offset_to_data` was extended from `u16` and
         // `bit_suffix` is at most `FAST_TYPE_DATA_MASK`, which is well
@@ -692,6 +709,97 @@ impl<'trie, T: TrieValue> CodePointTrie<'trie, T> {
             v
         } else {
             self.get32_by_small_index_cold(code_point)
+        }
+    }
+
+    /// Returns the value that is associated with `latin1` in this [`CodePointTrie`].
+    #[inline(always)]
+    pub fn get8(&self, latin1: u8) -> T {
+        let code_point = u32::from(latin1);
+        debug_assert!(code_point <= SMALL_TYPE_FAST_INDEXING_MAX);
+        // SAFETY: `u8` is always below `SMALL_TYPE_FAST_INDEXING_MAX` and,
+        // therefore, belowe `FAST_TYPE_FAST_INDEXING_MAX`.
+        unsafe { self.get32_assuming_fast_index(code_point) }
+    }
+
+    /// Returns the value that is associated with `ascii` in this [`CodePointTrie`].
+    ///
+    /// # Safety
+    ///
+    /// `ascii` must be less than 128.
+    #[inline(always)]
+    pub unsafe fn get7(&self, ascii: u8) -> T {
+        debug_assert!(ascii < 128);
+        debug_assert!((ascii as usize) < self.data.len());
+        // SAFETY: Length of `self.data` checked in the constructor.
+        T::from_unaligned(*unsafe { self.data.as_ule_slice().get_unchecked(ascii as usize) })
+    }
+
+    /// Returns the value that is associated with a two-byte UTF-8 sequence in this [`CodePointTrie`].
+    ///
+    /// `high_five` is the low five bits of the lead byte of a two-byte UTF-8 sequence.
+    /// `low_six` is the low six bits of the trail byte of a two-byte UTF-8 sequence.
+    ///
+    /// # Safety
+    ///
+    /// `high_five` must not have bit positions other than the lowest 5 set to 1.
+    /// `low_six` must not have bit positions other than the lowest 6 set to 1.
+    ///
+    /// # Panics
+    ///
+    /// With debug assertions enabled, panics if the above safety invariants are
+    /// violated or `high_five` represents non-shortest form.
+    #[inline(always)]
+    pub unsafe fn get_utf8_two_byte(&self, high_five: u32, low_six: u32) -> T {
+        debug_assert!(low_six <= 0b111_111); // Safety invariant.
+        debug_assert!(high_five <= 0b11_111); // Safety invariant.
+        debug_assert!(high_five > 0b1); // Non-shortest form; not safety invariant.
+                                        // SAFETY: The highest character representable as a two-byte
+                                        // UTF-8 sequence is U+07FF, eleven binary ones, which is below
+                                        // both `SMALL_TYPE_FAST_INDEXING_MAX` and `FAST_TYPE_FAST_INDEXING_MAX`.
+        self.get_bit_prefix_suffix_assuming_fast_index(high_five as usize, low_six as usize)
+    }
+
+    /// Returns the value that is associated with a three-byte UTF-8 or WTF-8 sequence in this [`CodePointTrie`].
+    ///
+    /// `high_ten` is the low four bits of the lead byte of three-byte UTF-8 or WTF-8 sequence shifted left by 6 followed by the low six bits of the first trail byte.
+    /// `low_six` is the low six bits of the last trail byte of a three-byte UTF-8 or WTF-8 sequence.
+    ///
+    /// Sequences representing surrogates (WTF-8) are allowed.
+    ///
+    /// # Safety
+    ///
+    /// `low_six` must not have bit positions other than the lowest 6 set to 1.
+    ///
+    /// # Intended Invariant
+    ///
+    /// `high_ten` must not have bit positions other than the lowest 10 set to 1.
+    ///
+    /// # Panics
+    ///
+    /// With debug assertions enabled, panics if the above safety invariant is
+    /// violated or `high_ten` is out of range for three-byte WTF-8 (or UTF-8)
+    /// sequence.
+    #[inline(always)]
+    #[allow(clippy::unusual_byte_groupings)]
+    pub unsafe fn get_utf8_three_byte(&self, high_ten: u32, low_six: u32) -> T {
+        debug_assert!(low_six <= 0b111_111); // Safety invariant.
+        debug_assert!(high_ten <= 0b1111_111_111); // Not actually a _safety_ invariant for this impl.
+        debug_assert!(high_ten > 0b11_111); // Non-shortest form; not safety invariant.
+
+        let fast_max = match self.header.trie_type {
+            TrieType::Fast => FAST_TYPE_FAST_INDEXING_MAX,
+            TrieType::Small => SMALL_TYPE_FAST_INDEXING_MAX,
+        };
+        // Keep only the prefix bits:
+        let max_bit_prefix = fast_max >> FAST_TYPE_SHIFT;
+        if high_ten <= max_bit_prefix {
+            // SAFETY: The caller is responsible for upholding the safety
+            // invariant for `low_six` and we just checked the safety
+            // invariant of `high_ten`.
+            self.get_bit_prefix_suffix_assuming_fast_index(high_ten as usize, low_six as usize)
+        } else {
+            self.get32_by_small_index_cold((high_ten << 6) | low_six)
         }
     }
 
@@ -1432,6 +1540,8 @@ impl<T: TrieValue> Iterator for CodePointMapRangeIterator<'_, T> {
 /// All implementations of `TypedCodePointTrie` are reviewable in this module.
 trait Seal {}
 
+impl<'trie, T: TrieValue> Seal for CodePointTrie<'trie, T> {}
+
 /// Trait for writing trait bounds for monomorphizing over either
 /// `FastCodePointTrie` or `SmallCodePointTrie`.
 #[allow(private_bounds)] // Permit sealing
@@ -1461,6 +1571,22 @@ pub trait TypedCodePointTrie<'trie, T: TrieValue>: Seal {
         } else {
             self.as_untyped_ref().get32_by_small_index_cold(code_point)
         }
+    }
+
+    /// Lookup trie value by Latin1 Code Point without branching on trie type.
+    #[inline(always)]
+    fn get8(&self, latin1: u8) -> T {
+        self.as_untyped_ref().get8(latin1)
+    }
+
+    /// Lookup trie value by ASCII Code Point without branching on trie type.
+    ///
+    /// # Safety
+    ///
+    /// `ascii` must be less than 128.
+    #[inline(always)]
+    unsafe fn get7(&self, ascii: u8) -> T {
+        self.as_untyped_ref().get7(ascii)
     }
 
     /// Lookup trie value by non-Basic Multilingual Plane Scalar Value without branching on trie type.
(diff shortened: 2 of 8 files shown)


components/collections/src/codepointtrie/cptrie.rs
```diff
@@ -577,14 +585,23 @@ impl<'trie, T: TrieValue> CodePointTrie<'trie, T> {
         );
 
         let bit_prefix = (code_point as usize) >> FAST_TYPE_SHIFT;
+        let bit_suffix = (code_point & FAST_TYPE_DATA_MASK) as usize;
+        self.get_bit_prefix_suffix_assuming_fast_index(bit_prefix, bit_suffix)
+    }
+
+    #[inline(always)]
+    unsafe fn get_bit_prefix_suffix_assuming_fast_index(
```
Manishearth, line 594: issue: please document safety invariants (even if it's obvious from the function name)

edit: it's not; because there are invariants on `bit_prefix` and `bit_suffix`.

We should document this and ensure it's upheld by the callers.

components/collections/src/codepointtrie/cptrie.rs
```diff
@@ -695,6 +712,97 @@ impl<'trie, T: TrieValue> CodePointTrie<'trie, T> {
...
+    /// Returns the value that is associated with `ascii` in this [`CodePointTrie`].
+    ///
+    /// # Safety
+    ///
+    /// `ascii` must be less than 128.
+    #[inline(always)]
+    pub unsafe fn get7(&self, ascii: u8) -> T {
+        debug_assert!(ascii < 128);
+        debug_assert!((ascii as usize) < self.data.len());
+        // SAFETY: Length of `self.data` checked in the constructor.
```
Manishearth: issue: We may add more ctors in the future. This should reference the safety invariants on `data`, just say `// SAFETY: Allowed by `data`s safety invariant`, updating `data`'s invariant to require that it has at least 128 elements and updating the constructor validation to saying something like `// `data` safety invariant upheld here`

components/collections/src/codepointtrie/cptrie.rs
```diff
@@ -695,6 +712,97 @@ impl<'trie, T: TrieValue> CodePointTrie<'trie, T> {
...
+    /// # Panics
+    ///
+    /// With debug assertions enabled, panics if the above safety invariants are
+    /// violated or `high_five` represents non-shortest form.
+    #[inline(always)]
+    pub unsafe fn get_utf8_two_byte(&self, high_five: u32, low_six: u32) -> T {
+        debug_assert!(low_six <= 0b111_111); // Safety invariant.
+        debug_assert!(high_five <= 0b11_111); // Safety invariant.
+        debug_assert!(high_five > 0b1); // Non-shortest form; not safety invariant.
+                                        // SAFETY: The highest character representable as a two-byte
```
Manishearth: nit: maybe introduce a newline so  that this formats better

Manishearth (commented):
Will take some time to properly review

components/collections/src/codepointtrie/iter.rs
```diff
@@ -0,0 +1,1240 @@
+// This file is part of ICU4X. For terms of use, please see the file
```
Manishearth, line 1: This file is a lot of unsafe code and I'm not convinced it is justified. Can we reduce the amount of unsafe code in this PR by writing these iterators to wrap CharIndices? There will still be unsafe in this file, but it will be around CPT invariants rather than _also_ around UTF8 decoding.

Separately we can try and justify additional unsafe using benchmarks if needed, and that would be a nice scoped PR that can be easily reviewed and benchmarked.

components/collections/src/codepointtrie/iter.rs
```diff
@@ -0,0 +1,1240 @@
...
+        }
+    }
+
+    /// Obtains the remainder of the iterator as a string slice.
+    #[inline]
+    pub fn as_str(&self) -> &'slice str {
+        // SAFETY: OK, because `delegate` came from `str` and is always
+        // advanced in a way that leaves the iterator at an UTF-8 sequence
+        // boundary.
+        unsafe { core::str::from_utf8_unchecked(self.delegate.as_slice()) }
```
Manishearth, line 64: issue: this should be an invariant of the delegate field, and every time it is mutated there should be a safety comment explaining how it is upheld

(and then all consumers of the invariant can just say "safe to treat as utf8 from field invariant")

but also: I suggest getting rid of this unsafe above.

components/collections/src/codepointinvliststringlist/mod.rs
```diff
@@ -175,16 +175,11 @@ impl<'data> CodePointInversionListAndStringList<'data> {
 
     /// See [`Self::contains_str`]
     pub fn contains_utf8(&self, s: &[u8]) -> bool {
-        use utf8_iter::Utf8CharsEx;
-        let mut chars = s.chars();
-        if let Some(first_char) = chars.next() {
-            if chars.next().is_none() {
-                return self.contains(first_char);
-            }
+        if let Ok(well_formed) = core::str::from_utf8(s) {
```
Manishearth: thought: this could probably be optimized to not need to UTF8 check, but perhaps not worth it.

components/collections/src/codepointtrie/cptrie.rs
```diff
@@ -385,6 +385,12 @@ impl<'trie, T: TrieValue> CodePointTrie<'trie, T> {
             return Err(Error::DataTooShortForFastAccess);
         }
 
+        // The builder is supposed to support direct indexing to the data array
```
Manishearth: issue: as mentioned below; let's make this a field invariant of `data` and then here say "Safety: Data field length invariant upheld here."

components/collections/src/codepointtrie/cptrie.rs
```diff
@@ -695,6 +712,97 @@ impl<'trie, T: TrieValue> CodePointTrie<'trie, T> {
         }
     }
 
+    /// Returns the value that is associated with `latin1` in this [`CodePointTrie`].
+    #[inline(always)]
+    pub fn get8(&self, latin1: u8) -> T {
+        let code_point = u32::from(latin1);
+        debug_assert!(code_point <= SMALL_TYPE_FAST_INDEXING_MAX);
+        // SAFETY: `u8` is always below `SMALL_TYPE_FAST_INDEXING_MAX` and,
```
Manishearth, line 721: suggestion (non blocking): worth documenting on those two constants that their precise values are extremely safety relevant and relied upon by many different checks in this file

components/collections/src/codepointtrie/cptrie.rs
Manishearth: issue: the safety invariants on this function are not currently documented, but once they are, this comment should be in terms of those invariants

components/collections/src/codepointtrie/cptrie.rs
```diff
@@ -695,6 +712,97 @@ impl<'trie, T: TrieValue> CodePointTrie<'trie, T> {
...
+    /// `high_ten` is the low four bits of the lead byte of three-byte UTF-8 or WTF-8 sequence shifted left by 6 followed by the low six bits of the first trail byte.
+    /// `low_six` is the low six bits of the last trail byte of a three-byte UTF-8 or WTF-8 sequence.
+    ///
+    /// Sequences representing surrogates (WTF-8) are allowed.
+    ///
+    /// # Safety
+    ///
+    /// `low_six` must not have bit positions other than the lowest 6 set to 1.
+    ///
+    /// # Intended Invariant
```
Manishearth, line 777: question: what is this? Is this a non-safety-relevant invariant?

Perhaps explicitly say it is non-safety relevant.

components/collections/src/codepointinvliststringlist/mod.rs
robertbastian: well that's what the previous code did, apparently the up-front check is faster?

components/collections/src/codepointinvliststringlist/mod.rs
```diff
@@ -175,16 +175,11 @@ impl<'data> CodePointInversionListAndStringList<'data> {
...
+            self.contains_str(well_formed)
+        } else {
+            false
         }
```
robertbastian: Let's not validate the whole string just to get rid of the dependency. It's a bit more complex to only validate the first character (I'm surprised the Rust standard library doesn't have any helpers), but it's not too complicated

```suggestion
        let single_char = s.get(0).and_then(|&first_byte| {
            if first_byte.is_ascii() && s.len() == 1 {
                return Some(first_byte as char);
            }
            let utf8_len = first_byte.leading_ones() as usize;
            if utf8_len != s.len() {
                return None;
            }

            str::from_utf8(&s[..utf8_len]).ok()?.chars().next()
        });

        if let Some(single_char) = single_char {
            return self.contains(first_char);
        }

        self.str_list
            .binary_search_by(|t| t.as_bytes().cmp(s))
            .is_ok()
```

components/collections/src/codepointtrie/iter.rs
hsivonen: Fusing the trie lookup into UTF-8 decoding is the key point of this changeset: CPT in ICU4C has been designed so that its bit split lines up with the bits in the last UTF-8 trail byte, and we've been using it pessimally in ICU4X.

I guess I will need to port the UTF-16 NFC to NFD throughput benchmark to `str` and then get exact numbers for the effect here.

Manishearth: Hmm, I see. I feel like using CharIndices (especially with its offset function) you might still be able to get the same benefits, but I understand if that was the point of this change.

In that case we should probably have more careful tracking of the invariant on the contained iterator whenever it is advanced.

components/collections/src/codepointtrie/iter.rs
hsivonen: I reworked the safety remarks to use the field invariant pattern,

components/collections/src/codepointinvliststringlist/mod.rs
hsivonen: Used @robertbastian 's suggestion with a trivial fix to make it compile.

components/collections/src/codepointtrie/cptrie.rs
hsivonen: Changed as requested.

components/collections/src/codepointtrie/cptrie.rs
hsivonen: Added a line break.

Manishearth:
Great, thanks for updating all those invariants, this is looking much better! I'll try and finish review today, so we can get this in for 2.2.

components/collections/src/codepointtrie/cptrie.rs
```diff
@@ -1573,6 +1792,37 @@ impl<'trie, T: TrieValue> TypedCodePointTrie<'trie, T> for FastCodePointTrie<'tr
...
+    /// Returns the value that is associated with a three-byte UTF-8 or WTF-8 sequence.
+    ///
+    /// `high_ten` is the low four bits of the lead byte of three-byte UTF-8 or WTF-8 sequence shifted left by 6 followed by the low six bits of the first trail byte.
+    /// `low_six` is the low six bits of the last trail byte of a three-byte UTF-8 or WTF-8 sequence.
+    ///
+    /// Sequences representing surrogates (WTF-8) are allowed.
+    ///
+    /// # Safety
+    ///
+    /// `high_ten` must not have bit positions other than the lowest 10 set to 1.
```
Manishearth, line 1805: thought: we should consider using different types here, like u16 and u8. We can then `as` cast.

non blocking

components/collections/src/codepointtrie/cptrie.rs
```diff
@@ -1676,6 +1989,194 @@ pub enum Typed<F, S> {
...
+/// Method naming intentionally differs from the method naming on
+/// those types in order to disambiguate.
+#[allow(private_bounds)] // Permit sealing
+pub trait AbstractCodePointTrie<'trie, T: TrieValue>: Seal {
+    /// Look up trie value by an ASCII character.
+    ///
+    /// # Safety
+    ///
+    /// `ascii` must be less than 128.
+    unsafe fn ascii(&self, ascii: u8) -> T;
```
Manishearth, line 2004: question: should these be `get_*`?

I think they don't have to be, but this is a new public API so we should think about it.

components/collections/src/codepointtrie/iter.rs
```diff
@@ -0,0 +1,1288 @@
...
+    V: TrieValue + Default,
+    T: AbstractCodePointTrie<'trie, V>,
+{
+    #[inline]
+    fn trie(&self) -> &'trie T {
+        self.trie
+    }
+}
+
+impl<'slice, 'trie, T, V> Iterator for CharsWithTrieDefaultForAscii<'slice, 'trie, T, V>
```
Manishearth, line 510: question: any hope of sharing code between the CharsWithTrie and CharsWithTrieDefaultForAscii impls?

Manishearth: Seems straightforward: have a type `CharsWithTrieWithDefaultHandling` that has `.next<F>()` where it calls `f(lead)` in the default case.

Manishearth (approved):
Overall this looks good.

I am in favor of landing this for the 2.2 release, which we're hoping to make next week. I have convinced myself there is no code where the safety is _worse_ than before (the code for `get_bit_prefix_suffix_assuming_fast_index` was already lacking justification), and the newly introduced unsafe code is well done. Further cleanups/documentation can be performed (I can file followups for the main unsafe issues)

It would be nice to have as much of these comments addressed as possible, but we should also try to land this before next week.

components/collections/src/codepointtrie/cptrie.rs
Manishearth: If these are mostly internal facing we should namespace them as `abstract_cpt_ascii` or something.

Manishearth:
One note: This adds a lot of new APIs, and if we want to land this as new public APIs we will need to make sure they are 

This adds:

 - New iterator types
 - A trait for abstracting over those iterator types
 - New get methods on CPT (`get7`, etc), some of them unsafe
 - A trait for abstracting over CPT types, with many methods

How much of this is needed by `utf8_iter`? Can we mark most of these new APIs as unstable for now and still make a release? I think I see a fair number of naming questions that we should spend time on (marked as `discuss-priority` to see if we can handle them Thursday)

cc @sffc to look at new APIs and naming as well

components/collections/src/codepointtrie/cptrie.rs
```diff
@@ -695,6 +713,99 @@ impl<'trie, T: TrieValue> CodePointTrie<'trie, T> {
...
+    ///
+    /// `high_five` must not have bit positions other than the lowest 5 set to 1.
+    /// `low_six` must not have bit positions other than the lowest 6 set to 1.
+    ///
+    /// # Panics
+    ///
+    /// With debug assertions enabled, panics if the above safety invariants are
+    /// violated or `high_five` represents non-shortest form.
+    #[inline(always)]
+    pub unsafe fn get_utf8_two_byte(&self, high_five: u32, low_six: u32) -> T {
```
Manishearth, line 755: This is a new public function. Should these parameters be narrower types?

components/collections/src/codepointtrie/cptrie.rs
```diff
@@ -1676,6 +1989,194 @@ pub enum Typed<F, S> {
     Small(S),
 }
 
+/// Trait for writing trait bounds for monomorphizing over either
+/// `CodePointTrie`, `FastCodePointTrie`, or `SmallCodePointTrie`.
+///
+/// Method naming intentionally differs from the method naming on
+/// those types in order to disambiguate.
+#[allow(private_bounds)] // Permit sealing
+pub trait AbstractCodePointTrie<'trie, T: TrieValue>: Seal {
```
Manishearth, line 1998: So we already have a `TypedCodePointTrie` trait. Can we avoid duplicating a similar trait? How much is this needed? I understand that `TypedCodePointTrie` cannot abstract over `CodePointTrie`, but what is the use case for abstracting over all three here?

In retrospect, `TypedCodePointTrie` should have been designed as a trait with a single associated constant that inherited from `AbstractCodePointTrie` (probably `CodePointTrieLike`).

Manishearth (changes requested):
Given the large number of new APIs I'm actually going to wait for us to discuss them more. Might not be worth trying to make this land for 2.2, but perhaps we can resolve everything by Thursday.

If there's an MVP set of changes that enable `utf8_iter` integration we should try for that.

But also I'm open to doing out-of-cycle releases for this. I dislike doing nontrivial API additions in an out of cycle release but ..... I don't want to rush this release either.

components/collections/src/codepointtrie/cptrie.rs
sffc, line 755: Issue, here and below: public functions with confusing invariants are not great to advertise to clients (although if they need to be cross-crate, it's better to have them documented than not). I see that you have these functions on a trait, too; I would prefer if you would keep them only on the trait and not export the concrete fns.

components/collections/src/codepointtrie/cptrie.rs
```diff
@@ -1605,14 +1855,59 @@ impl<'trie, T: TrieValue> TryFrom<CodePointTrie<'trie, T>> for FastCodePointTrie
...
     inner: CodePointTrie<'trie, T>,
 }
 
+impl<'trie, T: TrieValue> SmallCodePointTrie<'trie, T> {
+    #[doc(hidden)] // databake internal
+    /// # Safety
+    ///
+    /// `header.trie_type`, `index`, and `data` must
+    /// satisfy the invariants for the fields of the
+    /// same names on `CodePointTrie`.
```
sffc, line 1890: Suggestion: I prefer the safety invariant on parts constructors to be more like

```suggestion
    /// The parameters must have been returned from `SmallCodePointTrie::to_parts()`
```

because it is easier to verify this.

components/collections/src/codepointtrie/cptrie.rs
sffc: +1; this is a scaffolding trait, and we've almost always been disappointed when we give scaffolding trait functions nice names. So many editors are happy to create an import when you type `my_cpt.ascii()`, but we don't want to make unsafe functions so easy to accidentally stumble upon. Clients should always start with the safe functions by default, and where unsafe functions deliver significant gains, they are available for clients who need them.

(here and elsewhere in this trait)

sffc (commented):
Thanks. Preface: I entrust @Manishearth and others to judge the tradeoffs of the new unsafe abstractions. My comments below are under the assumption that the abstractions are well motivated.

components/collections/src/codepointtrie/iter.rs
hsivonen: Do you mean having `CharsWithTrieWithDefaultHandling` as the inner type of `repr(transparent)` `CharsWithTrie` and `CharsWithTrieDefaultForAscii`.

It seems to me that a macro would be simpler. Would that be OK?

components/collections/src/codepointtrie/cptrie.rs
hsivonen: The use case for abstracting over all three is to make the code that uses `AbstractCodePointTrie` work not only in the non-`serde` config where UTS 46 trie is small and UAX 15 tries are fast but also in the `serde` case where both are untyped.

components/collections/src/codepointtrie/cptrie.rs
hsivonen: I deliberately picked non-`get` naming for these to make it clear without turbofishes what's from the trait and what's not.

These are public in the sense that they need to be visible to the `utf8_iter` and `utf16_iter` crates. I think it's not harmful for other code to call through the trait when calling through the trait isn't strictly necessary, so I'd prefer not to obfuscate these for that reason.

The niceness of naming doesn't affect safety: The non-trait `get` counterparts are safe or `unsafe` in the same situations as these. The names for the two and three-byte UTF-8 accessors are already rather obscure.

Do I understand correctly that the main issue is that `ascii` is nice, but `ascii` is also `unsafe` (as `unsafe` as `get7`)? It would be possible to say `utf8_one_byte` instead of `ascii`, but wouldn't that just be weird?

It's very frustrating that simple things remain perma-undecided/unstable in the Rust standard library: `ascii` could be safe if https://github.com/rust-lang/rust/issues/110998 had been stabilized already.

components/collections/src/codepointtrie/iter.rs
hsivonen: I don't see how `CharIndices` would allow avoiding redundant math in the CPT queries.

components/collections/src/codepointtrie/cptrie.rs
sffc: If you prefix the trait fns with `abstract_cpt_`, then it is "clear without turbofishes what's from the trait and what's not", and it is more forward-compatible, too, because when https://github.com/rust-lang/rust/issues/110998 eventually lands, you can add a safe version of the fn.

Manishearth:
Some answers to questions on usage:

The trait is used by `utf8_iter` to abstract over all CPTs: https://github.com/hsivonen/utf8_iter/compare/main...cptrie

`utf8_two_byte`, etc are used inside manual UTF8 decoding code in utf8_iter. It is not used by normalizer/collator code as far as I can tell.

The utf8_iter code is mostly more iterators with unsafe decoding code, code that looks rather similar to the iterator code here already. In terms of net amount of unsafe code, that's a lot of unsafe that _probably_ could be abstracted over: we have `next()` and `next_back()` impls for each type of encoding, in a "with default" and "normal" mode. The "with default" and "normal" mode are definitely similar enough to be abstracted over, maybe with an internal macro. `next()` and `next_back()`  are not. The unvalidated and validated Utf8 code might be; I'm not sure.



------


We're already supporting additional encodings here with the Latin-1 iterator. It is an ICU4X norm to support potentially-invalid utf8/utf16 and often Latin-1. So I think _if_ we are choosing to have all these iterators, it's fine for them to live in `icu_collections`. This will also mean you don't need to do the release dance: you can add this to `icu_collections` and immediately start using it in normalizer/collator without any.

With those changes, I _think_ we no longer need any public non-hidden methods on AbstractCPT, nor do we need any of the public methods on the concrete types. We'll only need:

 - `AbstractCodePointTrie` (sealed, no visible methods)
 - All the iterators and extension traits.

This is a publiclly-safe addition to the public API, and much more scoped since it's just the iterators.


With this, it might be nice to figure out a way to share code between the different iterator impls. My preference is still for this unsafe code to live here rather than split across `icu_collections` and `utf8_iter` regardless of whether we reduce the code, but I would like to investigate codesharing. I might try some things.


I _also_ think we can get rid of the extension traits if we instead switch to having methods on the CPT types, like `iter_str()`, `iter_utf8()` (maybe `zip_str()`, or `chars_with_str()`?). Curious what @sffc thinks about that.

sffc:
> I _also_ think we can get rid of the extension traits if we instead switch to having methods on the CPT types, like `iter_str()`, `iter_utf8()` (maybe `zip_str()`, or `chars_with_str()`?). Curious what @sffc thinks about that.

My understanding based on @Manishearth's comment, which could be incorrect: these methods are public and used only by the `utf8_iter` crate. The proposal is to create an iterator that instead lives in `icu_collections`, such that these unsafe methods don't need to be exported across crate boundaries, and the exported APIs become more safe.

If my understanding is correct, then I am supportive in principle.

Manishearth:
Copying a suggestion from chat:

Another way of doing this would be to extend `Utf8Chars` with a `Utf8CharsWithHandler` type that takes in a Handler trait that has `unsafe fn handle_one_byte(ascii) -> V`, `unsafe fun handle_two_byte(high, low, point) -> V` (may not actually have to be unsafe!) , etc. Then CPT can have a private `CPTHandler` type that it uses: just wrapping `Utf8CharsWithHandler<CPTHandler>` lets you make all the iterators you want.

I'd very much support this extension: As I commented earlier I disliked the fact that the UTF8 code was being duplicated a bunch of times, and would love to see it being refactored.

Overall I do not think we have really explored the space of encapsulating the unsafe code in ways that reduce copy-pasting of the same unsafe code.

Manishearth:
And based on later discussion, I'm hoping we can try doing that work in `utf8_iter`. I have a pretty clear idea of what we need at this point, but won't be able to look at this for a few weeks.

sffc:
We briefly discussed this today. I am okay so long as the unsafe code is appropriately packaged, without crossing crate boundaries, while deferring to @Manishearth on concerns about duplicating the unsafe blocks between icu_collections and utf8_iter. @hsivonen is waiting on @Manishearth for feedback on how to better structure the unsafe code to reduce the duplication with utf8_iter.

Manishearth:
Hmm, I did not realize @hsivonen was waiting for me. I did offer to do some of the work here myself to get it started but that was if @hsivonen didn't have time to do it himself. I guess this is an indicator that he doesn't, or needs help understanding what I was going for? In that case, yes I can do this, but it won't be soon.

(and in the meantime we shouldn't block PRs like https://github.com/unicode-org/icu4x/pull/7878 on that)

sffc:
I think @hsivonen was assuming you were going to post a more detailed reply based on this comment:

> I have a pretty clear idea of what we need at this point, but won't be able to look at this for a few weeks.

Manishearth:
Right, that comment was an update for the rest of the team posted after Henri and I had had a pretty in depth conversation on this on a call. I have some interest in trying to sketch a solution but I was also under the impression that I had given Henri enough to go on.

Either way not super important, I'll get to it eventually ...

Manishearth:
https://github.com/hsivonen/utf8_iter/pull/2

PR #7770, opened 2026-03-11
Add EastAsianTraditional min_months_from invariant

robertbastian:
## Changelog: N/A

### DIFF
--- a/components/calendar/src/cal/east_asian_traditional.rs
+++ b/components/calendar/src/cal/east_asian_traditional.rs
@@ -2,9 +2,7 @@
 // called LICENSE at the top level of the ICU4X source tree
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
-use crate::calendar_arithmetic::{
-    ArithmeticDate, DateFieldsResolver, MinMonths, PackWithMD, ToExtendedYear,
-};
+use crate::calendar_arithmetic::{ArithmeticDate, DateFieldsResolver, PackWithMD, ToExtendedYear};
 use crate::error::{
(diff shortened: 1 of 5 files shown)


components/calendar/src/calendar_arithmetic.rs
```diff
@@ -1105,11 +1077,6 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
         // 1. Return ! CreateDateDurationRecord(_years_, _months_, _weeks_, _days_).
         DateDuration::from_signed_ymwd(years, months, weeks, days)
     }
-
-    /// The minimum number of months over `years` years, starting from `self.year()`.
-    pub(crate) fn min_months_from(self, years: i64) -> MinMonths {
```
Manishearth, line 1110: heh, i noticed this in the other change but decided to leave it. thanks

components/calendar/src/cal/east_asian_traditional.rs
```diff
@@ -155,6 +153,9 @@ pub struct EastAsianTraditional<R>(pub R);
 /// The rules for how to perform these calculations, as well as how local
 /// time is determined differ between countries and have changed over time.
 ///
+/// Implementations of this trait must produce calendars that have a leap month
```
Manishearth, line 156: observation: this doesn't let us update to a more precise approximation in the future

that is fine, we can add a trait method if we ever need to.

robertbastian: we can do whatever we want as long as the trait is unstable

Manishearth: Right, mostly just observing that the docs would need to be redone, and we can't stabilize before settling on this.

PR #7772, opened 2026-03-12
Simplify `AnyCalendar` implementation

robertbastian:
* Simplifies the `make_any_cal` macro by not special-casing deprecated variants
* Removes implementations that are not currently released (`From<$cal> for AnyCalendar`, `TryFrom<AnyCalendar> for $cal`, `PartialEq<$cal> for AnyCalendar`).
* Doesn't store the calendar identity in `AnyDateInner` anymore – the concrete calendars handle this now
* Re-reordering `AnyCalendarKind` variants that cargo-semver-checks currently complains about

## Changelog: N/A

### DIFF
--- a/components/calendar/src/any_calendar.rs
+++ b/components/calendar/src/any_calendar.rs
@@ -46,17 +46,16 @@ macro_rules! make_any_calendar {
         pub enum $any_date_ident {
             $(
                 #[doc = concat!("A date for a [`", stringify!($variant), "`] calendar")]
-                $variant(<$ty as $crate::Calendar>::DateInner, <$ty as AnyCalendarable>::Identity),
+                $variant(<$ty as $crate::Calendar>::DateInner),
             )+
         }
 
         impl PartialOrd for $any_date_ident {
-            #[rustfmt::skip]
             fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
                 use $any_date_ident::*;
                 match (self, other) {
                     $(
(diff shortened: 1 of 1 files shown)


components/calendar/src/any_calendar.rs
```diff
@@ -401,53 +285,62 @@ macro_rules! make_any_calendar {
...
-                        #[allow(deprecated)]
-                        Self::$deprecated_variant(ref c) => c.calendar_algorithm(),
-                    )*
                 }
             }
         }
+    };
+}
 
+macro_rules! make_any_calendar_impls {
```
Manishearth, line 294: nit: document this macro and where it should be used?

Manishearth:
I want to defer to Shane on reviewing the macro parts. What I can understand from this PR makes sense: we used to store calendar identity in ACInner, and now we don't. This simplifies all of our match statements.

Are we sure that the Hijri/Ethiopic impls are well behaved here?

Removing impls seems good.

This also splits `make_any_calendar` into two macros.

components/calendar/src/any_calendar.rs
robertbastian: nowhere, unlike the `make_any_cal` I don't think this should ever be public

Manishearth: Still would like documentation, but doesn't have to be much of it.

Non blocking nit.

sffc:
Can you confirm that the `is_date_compatible` thing is covering the ethiopic/hijri logic that was previously handled with the DateInner field?

robertbastian:
Ethiopian had identity `()` before, and the Hijri tabular identity is now handled by `is_date_compatible`.

Issue #7769, opened 2026-03-11
Add From<integer> for Month and YearInput


Manishearth:
This was mentioned in the discussion for https://github.com/unicode-org/icu4x/issues/7512, and again in its implementation PR: https://github.com/unicode-org/icu4x/pull/7764

Should `Month` be `From<u8>` (producing a normal month) and `YearInput` be `From<i32>`  (producing an extended year)?

Note that `Month` already has [`Month::new()`](https://unicode-org.github.io/icu4x/rustdoc/icu/calendar/types/struct.Month.html#method.new)

This mostly makes it easier for callsites to be like `Date::try_from_codes(1234.into(), 1.into(), 6, cal)`.

A followup question is that if we have this, should `try_from_codes` take `Into<YearInput>`, `Into<Month>` instead? My inclination is to avoid this: this tends to pessimize codesize since Rust doesn't (yet) do outlining optimizations. That said, `Date::try_from_codes` is a thin wrapper around another function, so it's _probably_ fine?

Manishearth:
This is 2.2-blocking in case https://github.com/unicode-org/icu4x/pull/7764 lands for 2.2 with an Into impl. If it doesn't, then this can be moved out of the milestone. 

cc @sffc @robertbastian @zbraniecki

robertbastian:
I think this is 2.2 blocking because if we do this, `try_from_codes` is maybe not the right name anymore. This will be very general purpose, so it might deserve the `try_new` name.

sffc:
I feel we should not implement `From<int> for Month` because it is ambiguous between ordinal months and month numbers. I could be convinced otherwise with a compelling argument. I don't feel that this is an issue for `From<int> for YearInput` because all calendars have a well defined extended year.

I'm okay naming the function `try_new`.

robertbastian:
A month is a non-leap month by default, if it's a leap month you need to be explicit,  e.g. using `3` as the month value is clearly not a leap month, whereas `Month::leap(3)` is.

The concept or ordinal month is really an implementation detail, it's not something that is commonly used, and as such users will not expect that interpretation. Nobody using the Hebrew calendar would think the 10 ever means Sivan.

Manishearth:
- @sffc Tried to think of cases where `From<int> for Month` would cause bugs and I can't identify any right now. I like the ergonomics so it's probably fine.
[...]

PR #7773, opened 2026-03-12
Rename Date::try_from_codes to Date::try_new

Manishearth:
Fixes https://github.com/unicode-org/icu4x/issues/7769

Mostly just implements the result of the discussion. Nothing interesting, though I did notice that FFI still has `from_codes_in_calendar`. I think keeping that is actually the right call, since over FFI we do accept string codes.

## Changelog: 

This is a modification of the changelog in https://github.com/unicode-org/icu4x/pull/7764, both PRs should be included in the changelog entry

icu_calendar: Add `Date::try_new`, which replaces `Date::try_new_from_codes`, and takes typed year/month values.
  - New methods: `Date::try_new` (and primarily-internal `Calendar::new_date`)
  - New types: `InputYear`, `DateNewError`


components/calendar/benches/date.rs
```diff
@@ -152,25 +152,23 @@ fn date_benches(c: &mut Criterion) {
...
 
     bench_calendar(
         &mut group,
         "calendar/hebrew",
         &fxs,
         icu::calendar::cal::Hebrew,
-        |y, m, d, c| {
-            Date::try_from_codes(types::YearInput::Extended(y), types::Month::new(m), d, c).unwrap()
-        },
+        |y, m, d, c| Date::try_new(y.into(), m.into(), d, c).unwrap(),
```
robertbastian: hm this should have probably been changed to the new Hebrew convenience constructor

components/calendar/src/cal/east_asian_traditional.rs
```diff
@@ -1656,19 +1656,19 @@ mod test {
         };
         let year = cal.year_info_from_extended(2023);
         for (ordinal, month) in [
-            Month::new(1),
-            Month::new(2),
+            1.into(),
+            2.into(),
```
robertbastian: would've probable left this untouched for consistency but whatever

components/calendar/src/tests/exhaustive.rs
```diff
@@ -34,7 +34,7 @@ super::test_all_cals!(
     #[ignore] // takes about 90 seconds in release-with-assertions
     fn check_from_fields<C: Calendar + Copy>(cal: C) {
         let months = (1..19)
-            .flat_map(|i| [types::Month::new(i), types::Month::leap(i)].into_iter())
+            .flat_map(|i| [i.into(), types::Month::leap(i)].into_iter())
```
robertbastian: also

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -447,7 +447,7 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
                         if ref_year.err() == Some(EcmaReferenceYearError::UseRegularIfConstrain)
                             && options.overflow == Some(Overflow::Constrain)
                         {
-                            let new_valid_month = Month::new(m.number());
+                            let new_valid_month = m.number().into();
```
robertbastian: I think readability suffers if you use `.into()` in places where the type is not clear

components/calendar/src/types.rs
```diff
@@ -504,6 +508,9 @@ impl fmt::Display for MonthCode {
 /// * `Month::new(7)` = `M07`
 /// * `Month::leap(2)` = `M02L`
 ///
+/// This type implements `From<u32>` producing an non-leap month, so you can simply
```
```suggestion
/// This type implements `From<u8>` producing an non-leap month, so you can simply
```

components/calendar/src/cal/east_asian_traditional.rs
Manishearth: yeah i noticed this, i think both are fine.

Manishearth:
After doing the bulk refactor I undid a bunch of Month::new cases that I noticed but most of the ones you pointed out make sense. I think the east_asian_traditional one is fine though

sffc (commented):
I agree that I would prefer for most of the `Month::new(x)` -> `x.into()` to not be changed by this PR. I want to favor advertising the `Month` type in docs. We can show the `Into` impl in a couple key places.

Manishearth:
@sffc I don't think that's very actionable feedback...

This PR did a bulk refactor, and then I undid some of the cases where it needed explicit typing (`let x: Month = x.into()`) and where `Month::new()` is next to `Month::leap()`.

Most of the changes are in tests; and I don't think it matters that much in tests. Are you suggesting I reduce the usage in tests?

As for docs, only a handful of places changed:

 - `try_new()`, which is why we added this in the first place
 - `try_new_{chinese,korean}_traditional()`
 - Error docs

Which locations do you take issue with? I can see an argument for `east_asian_traditional` since it is nice to be clear we take a month, but I also think if you are looking at the docs for a function you can see that it takes a Month anyway. Plus `5.into()` and `Month::new(5)` are both equally bad at telling you "you can pass a leap month here", you have to click through the docs either way.

On the error stuff I think conciseness on the "setup" is actually good: it's more important to see how it all comes together than to understand that a Month type is in place.

I addressed the callsites Robert took issue with, except for one which I mildly prefer using `.into()`. Can you please give me more to go on with than a vague sense that this adds `.into()` too much?

(perhaps this was a partial review and you intended to give further feedback later, which is fine)

components/calendar/tests/arithmetic.rs
```diff
@@ -204,21 +204,18 @@ fn test_tricky_leap_months() {
...
-        Date::try_new_hebrew_v2(5784, Month::new(6), 20).unwrap()
-    );
+    assert_eq!(date1, Date::try_new_hebrew_v2(5784, 6.into(), 20).unwrap());
     let duration0_actual = date0.try_until_with_options(&date1, until_options).unwrap();
     assert_eq!(duration0_actual, duration0);
 
     // M02L until M02 = 12mo
     let cdate0 = Date::try_new_chinese_traditional(2023, Month::leap(2), 1).unwrap();
-    let cdate1 = Date::try_new_chinese_traditional(2024, Month::new(2), 1).unwrap();
+    let cdate1 = Date::try_new_chinese_traditional(2024, 2.into(), 1).unwrap();
```
sffc: I find `Month::leap(2)` followed by `2.into()` to be profoundly confusing...

sffc:
> Which locations do you take issue with?

I can comment on each one but I would be commenting on about 80% of the lines in this PR.

components/calendar/tests/arithmetic.rs
Manishearth: Yes, this is Robert's feedback, I can fix the cases where the new was used next to leap.

Manishearth:
> I can comment on each one but I would be commenting on about 80% of the lines in this PR.

You can also give me a set of guidelines to follow. Robert's guidelines were that we shouldn't use the into impl next to `Month::leap()` and I fixed most of those cases but clearly missed a few. I could look for more. Are there other cases?

Manishearth:
I reverted the commit that moved us to using Into everywhere for month. APIs unchanged.

components/calendar/benches/date.rs.orig
```diff
@@ -0,0 +1,233 @@
+// This file is part of ICU4X. For terms of use, please see the file
```
sffc: nit: remove the merge file

components/calendar/src/tests/arithmetic.rs.orig
```diff
@@ -0,0 +1,346 @@
+// This file is part of ICU4X. For terms of use, please see the file
```
sffc: nit: another merge file

components/calendar/src/tests/extended_year.rs.orig
```diff
@@ -0,0 +1,83 @@
+// This file is part of ICU4X. For terms of use, please see the file
```
sffc: nit: another merge file

Manishearth:
Ugh. My `git` workflow is set up to not add untracked files, but now that I use `jj` it doesn't do that and I'm not sure i actually want to use those settings, but it takes some getting used to...

Manishearth:
Did `**` not work? hm.

PR #7776, opened 2026-03-13
Add docs for `impl From<u8> for Month`

sffc:
Re-written piece of #7773

## Changelog: N/A


Manishearth:
I removed the usage of into from https://github.com/unicode-org/icu4x/pull/7773 since that PR does more things. This PR can simply use the into impl where it wants to after merging that.

PR #7778, opened 2026-03-13
Rename LeapBase to Base

### LINKED ISSUE #7691: Membrane between MonthInfo and icu_datetime for formatting leap
We added `LeapStatus` in https://github.com/unicode-org/icu4x/pull/7667. Shane and I approved it with "some concerns about the naming", but we should resolve that before.

Currently, LeapStatus is the following:

```rust
enum LeapStatus {
    Normal,
    Leap,
    StandardAfterLeap,
}
```

StandardAfterLeap is for standard months occurring after leap months which format differently: true for Hebrew.

and icu_datetime checks `leap_status != Normal` before using leap codes (Hebrew is the only thing that uses StandardAfterLeap, and Hebrew stores its leap / "formatting leap" formatting info the same way).

@robertbastian has further work to try and tweak what icu_datetime does in https://github.com/unicode-org/icu4x/pull/7666.

We should pin down this membrane. I think a LeapStatus enum is mostly correct, but I don't think the current set of variants is correct.

Manishearth:
Fixes #7691 

## Changelog: N/A

Change to unreleased API

### DIFF
--- a/components/calendar/src/types.rs
+++ b/components/calendar/src/types.rs
@@ -523,7 +523,7 @@ pub struct Month {
 #[non_exhaustive]
 #[derive(Copy, Clone, Debug, PartialEq, Hash, Eq)]
 pub enum LeapStatus {
-    /// Not a leap month, aka a "normal", "common", or "standard" month.
+    /// Not a leap month, aka a "normal", "common", "ordinary", or "standard" month.
     Normal,
     /// A leap month.
     Leap,


components/calendar/src/types.rs
```diff
@@ -523,18 +523,18 @@ pub struct Month {
 #[non_exhaustive]
 #[derive(Copy, Clone, Debug, PartialEq, Hash, Eq)]
 pub enum LeapStatus {
-    /// Not a leap month.
+    /// Not a leap month, aka a "normal", "common", or "standard" month.
```
sffc:
```suggestion
    /// Not a leap month, aka a "normal", "common", "ordinary", or "standard" month.
```

PR #7779, opened 2026-03-14
Fix UB: use || instead of && 

sayrer:
in cast_bytes_to_slice alignment/length check

The condition guarding `from_raw_parts` joined the alignment check and length check with `&&`, meaning the error was only returned when *both* were wrong. A misaligned-but-correctly-sized buffer passed the check, then `from_raw_parts` dereferences misaligned data — undefined behavior.

Added tests that verify each condition is rejected independently.

## Changelog

resb: Fix UB around alignment check

### DIFF
--- a/utils/resb/src/binary/helpers.rs
+++ b/utils/resb/src/binary/helpers.rs
@@ -119,9 +119,8 @@ mod tests {
     #[test]
     fn cast_aligned_correct_length() {
         let data: [u32; 2] = [1, 2];
-        let bytes: &[u8] = unsafe {
-            core::slice::from_raw_parts(data.as_ptr() as *const u8, size_of_val(&data))
-        };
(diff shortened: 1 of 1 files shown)


Manishearth:
Please run `cargo fmt`

Manishearth:
Oh and you'll have to sign the CLA unfortunately.

sayrer:
Oh, I don't care about the CLA. It's 6pm here, I'll figure it out tomorrow.

Manishearth:
Yeah you just have to go click a button on link on the failing ci task.

sayrer:
> Nice catch, thanks! I'm pretty sure the tests can be written in a way that doesn't need unsafe but I'll look into that later.

Maybe so, I never use it, so my knowledge is just from FFI wrappers and the Rust Book.

PR #7783, opened 2026-03-16
Fix UB: validate replacement byte in TinyAsciiStr lossy constructors

sayrer:
```
from_utf8_lossy and from_utf16_lossy accept a `replacement: u8` that is
used to substitute invalid code units. This byte was not validated to be
ASCII, so passing >= 0x80 would create an invalid AsciiByte enum
discriminant (UB) and then feed non-UTF-8 data to from_utf8_unchecked.

Add an assert that replacement is a non-null ASCII byte (1..=127). The exploit
chain: TinyAsciiStr::<4>::from_utf8_lossy(b"\xFF", 0xFF) → 0xFF used as
replacement → stored in out array → AsciiByte::to_ascii_byte_array creates
invalid enum discriminant (UB) → .as_str() calls from_utf8_unchecked on
[0xFF] (UB). Two UBs for the price of one public API call.
```

I think this one is right. A panic is fine here, as you're clearly dealing
with adversarial input.

## Changelog

tinystr: Fix UB in `TinyAsciiStr::from_utf8_lossy` and `from_utf16_lossy`: validate that the replacement byte is ASCII

### DIFF
--- a/utils/tinystr/src/ascii.rs
+++ b/utils/tinystr/src/ascii.rs
@@ -50,7 +50,15 @@ impl<const N: usize> TinyAsciiStr<N> {
     /// will be replaced with the replacement byte.
     ///
     /// The input slice will be truncated if its length exceeds `N`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `replacement` is not a non-null ASCII byte (i.e. not in `1..=127`).
     pub const fn from_utf8_lossy(code_units: &[u8], replacement: u8) -> Self {
+        assert!(
+            replacement > 0 && replacement < 0x80,
+            "replacement must be a non-null ASCII byte (1..=127)"
+        );
         let mut out = [0; N];
         let mut i = 0;
         // Ord is not available in const, so no `.min(N)`
@@ -85,7 +93,15 @@ impl<const N: usize> TinyAsciiStr<N> {
     /// will be replaced with the replacement byte.
     ///
     /// The input slice will be truncated if its length exceeds `N`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `replacement` is not a non-null ASCII byte (i.e. not in `1..=127`).
     pub const fn from_utf16_lossy(code_units: &[u16], replacement: u8) -> Self {
+        assert!(
+            replacement > 0 && replacement < 0x80,
+            "replacement must be a non-null ASCII byte (1..=127)"
+        );
         let mut out = [0; N];
         let mut i = 0;
         // Ord is not available in const, so no `.min(N)`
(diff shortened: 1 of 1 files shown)


sayrer:
No shade thrown at our icu4x maintainers:

❯ That's a really surprising bug, honestly. Like the Unicode experts didn't spot that? I know it well enough, but not as well as they do.

```It's the kind of bug that hides in plain sight precisely because the domain experts are focused on Unicode correctness — decomposition, normalization, collation, all the hard stuff. A u8  parameter on a helper function just doesn't register as dangerous when you're thinking about Hangul syllable composition.```

``` And the function was probably written by someone who only ever intended b'?' to go there. It works, the tests pass, you move on to the actual i18n problem. The type signature is the bug — u8 when it should have been something constrained — but Rust doesn't have a AsciiChar in stable const contexts, so u8 was the pragmatic choice and the validation just got missed.```

❯ Yeah, that makes sense. They'll take the patch.

```Good. It's a one-line assert, no API change, no breakage for anyone passing a valid byte. Easy review.```

❯ maybe a bigger problem: Rust actually should have AsciiChar

```
It does — core::ascii::Char has been unstable since 1.72. It's been sitting
there for almost three years now. If it were stable and usable in const
contexts, this function signature would  have been replacement: ascii::Char
and the bug wouldn't exist.

The broader issue is that Rust has great type-level safety for UTF-8 (str vs
[u8]) but nothing stable for the ASCII subset. So every crate that deals with
ASCII ends up rolling its own — tinystr has AsciiByte, httparse has its own
thing, etc. And every one of those is a place where a u8 slips through
without validation.
```

❯ that's really bad though. ASCII falls out of UTF-8 with little effort. What's the problem?

```Good question. Let me check where it actually stands.```

```
Web Search("rust ascii::Char stabilization tracking issue 2026")
  ⎿  Did 1 search in 4s

Fetch(https://github.com/rust-lang/rust/issues/110998)
  ⎿  Received 372.4KB (200 OK)
```

```
Naming bikeshed and as-cast semantics. Three years of unstable over whether to
call it Char vs something else, and whether ascii_val as u8 should work.

  Meanwhile every crate in the ecosystem that handles ASCII is writing u8
  parameters and hoping callers stay in range. The
  https://github.com/rust-lang/rust/issues/110998 has four open questions,
  none of which are technical blockers — they're API design taste debates.

  The irony is that str exists specifically because "bytes that happen to be
  UTF-8" needed a type. ASCII is a strict subset with an even simpler
  invariant (one bit!), and it's been sitting in nightly since 2023.

  Sources:
  - https://github.com/rust-lang/rust/issues/110998
  ```

sayrer:
let's link up https://github.com/rust-lang/rust/issues/110998

Manishearth:
Good catch! And yeah, this is the type of thing that's easily missed because it's sort of an "additional" mutating API on top of the other stuff.

We should spend more time documenting and tracking invariants on tinystr.

PR #7788, opened 2026-03-18
Add zerofrom::transparent! macro

sffc:
Latest attempt at #7607

Looking for feedback on the general approach, and then on the concrete implementation.

## Changelog

zerofrom: Adds macro to derive ZeroFrom and concrete functions for repr(transparent) conversions
- New macro: `zerofrom::transparent!`

### DIFF
--- a/utils/zerofrom/src/zf_transparent.rs
+++ b/utils/zerofrom/src/zf_transparent.rs
@@ -10,11 +10,13 @@ use alloc::boxed::Box;
 ///
 /// # Safety
 ///
-/// `Outer` is `repr(transparent)` and has one non-zero-sized field
-/// of type `Inner`.
+/// `Outer` is `repr(transparent)` and has one non-zero-sized field of type `Inner`.
+///
+/// Suggestion: explicitly setting the generic parameters to two types satisfying this invariant
+/// makes the fn safe to call.
 #[cfg(feature = "alloc")]
 #[inline(always)]
-pub unsafe fn cast_transparent_box<Outer, Inner>(inner: Box<Inner>) -> Box<Outer> {
+pub unsafe fn cast_transparent_box<Inner, Outer>(inner: Box<Inner>) -> Box<Outer> {
     // Safety:
     //
     // - Both boxes have the same allocator (the global allocator).
@@ -82,7 +84,7 @@ macro_rules! transparent {
 		$(
 			impl<'zf> $crate::ZeroFrom<'zf, $inner_zf> for &'zf $outer {
 				fn zero_from(inner: &'zf $inner) -> Self {
-					unsafe { core::mem::transmute(inner) }
+					unsafe { core::mem::transmute::<&$inner, &$outer>(inner) }
 				}
 			}
 		)?
@@ -90,21 +92,20 @@ macro_rules! transparent {
 			$(
 				$(#[$meta_ref])*
 				$vis_ref fn $fn_ref(inner: &$inner_ref) -> &Self {
-					unsafe { core::mem::transmute(inner) }
+					unsafe { core::mem::transmute::<&$inner, &$outer>(inner) }
 				}
 			)?
 			$(
 				$(#[$meta_slice])*
 				$vis_slice fn $fn_slice(inner: &[$inner_slice]) -> &[Self] {
-					unsafe { core::mem::transmute(inner) }
+					unsafe { core::mem::transmute::<&[$inner], &[$outer]>(inner) }
 				}
 			)?
 			$(
 				$(#[$meta_box])*
 				$vis_box fn $fn_box(inner: $crate::internal::Box<$inner_box>) -> $crate::internal::Box<Self> {
 					// Safety: $outer is repr(transparent) over $inner.
-					// TODO: Enforce that $inner is the same as $inner_box
-					unsafe { $crate::internal::cast_transparent_box(inner) }
+					unsafe { $crate::internal::cast_transparent_box::<$inner, $outer>(inner) }
 				}
 			)?
 			$(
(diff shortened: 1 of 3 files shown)


Manishearth (commented):
In favor of the general implementation. I think the way you handle method docs/pub/etc is clever.

This is exactly along the lines as what I was envisioning for transparent casts for a long time, and is significantly better than my imagined level of complexity.

sffc:
Thanks; if this direction looks good to you then I'll clean up this PR and get it ready for merging.

utils/zerofrom/src/zf_transparent.rs
```diff
@@ -0,0 +1,95 @@
+// This file is part of ICU4X. For terms of use, please see the file
+// called LICENSE at the top level of the ICU4X source tree
+// (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
+/// Implements [`ZeroFrom`](crate::ZeroFrom) on a transparent type
+/// from a reference to the inner type.
+///
+/// Also supports creating concrete functions.
```
Manishearth: nit:

```suggestion
/// Also supports adding concrete conversion functions between various reference types.
```

utils/zerofrom/src/zf_transparent.rs
```diff
@@ -0,0 +1,95 @@
...
+/// let s = "hello";
+/// let wrap = <&StrWrap>::zero_from(s);
+///
+/// assert_eq!(&wrap.0, "hello");
+/// ```
+#[macro_export]
+macro_rules! transparent {
+	(
+		#[repr(transparent)]
+		$(#[$meta:meta])*
```
robertbastian, line 53: this is sensitive to attribute ordering, which it shouldn't be

utils/zerofrom/src/zf_transparent.rs
```diff
@@ -0,0 +1,95 @@
...
+				$(#[$meta_slice:meta])*
+				$vis_slice:vis fn $fn_slice:ident(&[$type_slice:ty]) -> &[Self];
+			)?
+			$(
+				@box
+				$(#[$meta_box:meta])*
+				$vis_box:vis fn $fn_box:ident(Box<$type_box:ty>) -> Box<Self>;
+			)?
+			$(
+				@rc
```
robertbastian, line 75: let's try to keep custom syntax to a minimum. it should suffice to match on the argument

utils/zerofrom/tests/test_transparent.rs
```diff
@@ -0,0 +1,29 @@
...
+// called LICENSE at the top level of the ICU4X source tree
+// (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
+use zerofrom::transparent;
+
+transparent!(
+    #[repr(transparent)]
+    /// hello world
+    #[derive(Debug)]
+    pub(crate) struct Foo([u8; 3]);
```
robertbastian, line 11: I don't like type declarations living inside macro invocations. From a call site it's not known that this code is passed through without modifications, the macro could do anything to it (change the repr, add fields, wrap fields, etc).

Manishearth: So this is why I've overall been very lukewarm on #7607 so far. But I think this is the most maintainable and low-impact version.

We can document what the macro does, and it's pretty easy to verify it, which is why I think this is still a good idea overall.

utils/zerofrom/src/zf_transparent.rs
sffc: `$(#[$meta:meta])*` is greedy and there isn't a clean way to make it non-greedy so I need to put the `#[repr(transparent)]` at the top.

utils/zerofrom/tests/test_transparent.rs
sffc: I don't love this approach, either. It was not my first or second or third choice at solving the problem. But @Manishearth identified flaws in all the other approaches. My conclusion is that this is the least bad, and better than not solving the problem.

utils/zerofrom/src/zf_transparent.rs
Manishearth: A different approach is to have the macro generate the repr but that's less obvious to the reader.

sffc: right. As currently written the macro does not add anything to the struct definition. It just echoes it back out. The only thing is does is add impls which it claims are safe due to the shape of the struct.

utils/zerofrom/src/zf_transparent.rs
```diff
@@ -0,0 +1,95 @@
...
+			$(
+				$(#[$meta_slice])*
+				$vis_slice fn $fn_slice(inner: &[$type_slice]) -> &[Self] {
+					unsafe { core::mem::transmute(inner) }
+				}
+			)?
+			$(
+				$(#[$meta_box])*
+				$vis_box fn $fn_box(inner: $crate::internal::Box<$type_box>) -> $crate::internal::Box<Self> {
+					unsafe { core::mem::transmute(inner) }
```
sffc: Reminder for self: I was lazy and wrote `core::mem::transmute` here but we should not use `transmute` on boxes. `Box::from_raw` is what we should use instead. (#7871)

robertbastian: Same for the `Rc`s

utils/zerofrom/src/zf_transparent.rs
sffc: I'll try again, but if I match on the argument then every token up to the argument can't be itself a match. macro_rules doesn't do fancy lookahead matching.

sffc: Here's a simplified example if you want to see if you can get it to work:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=8931afd392ba47737907ee2fdb37a2fb

sffc: Discussion with @robertbastian: if we remove the meta, vis, and function name, then we can match on the argument type. This seems fine.

sffc:
Should we put this in the same crate as other potential internal unstable APIs like BoundLocaleDataProvider?

https://github.com/unicode-org/icu4x/issues/4467

- @robertbastian says: it is justifiable to put it in a low-level crate like zerofrom since it is at the crate root, but that doesn't mean the macro should be zerofrom-specific.

Issue #7785, opened 2026-03-18
[currency] Feature request: Add `currencyDisplay` option to currency formatting


younies:
### Summary

Add support for a `currencyDisplay` option in ICU4X currency formatting APIs, enabling control over how currencies are rendered (symbol, narrow symbol, code, or name), similar to ECMA-402 `Intl.NumberFormat`.

---

### Background

In ECMA-402, currency formatting exposes the following option:

- `currencyDisplay`: "symbol" | "narrowSymbol" | "code" | "name"

Defined in:
https://tc39.es/ecma402/#sec-setnumberformatunitoptions

This option allows developers to choose between different representations of a currency (e.g., `$`, `US$`, `USD`, `US dollars`).

---

### Problem

Currently, ICU4X currency formatting APIs (including experimental ones) do not expose a unified way to select the currency display style.

While the underlying CLDR data supports:
- standard symbols
- narrow symbols
- ISO codes
- localized display names

there is no high-level API to select between them.

This limits:
- feature parity with ECMA-402
- consistency across platforms using ICU4X

---

### Proposal

Introduce a `CurrencyDisplay` enum in ICU4X:

```rust
pub enum CurrencyDisplay {
    Symbol,
    NarrowSymbol,
    Code,
    Name,
}

sffc:
I think we should drop the `width` field and use the ECMA-402 naming for these options.

PR #7789, opened 2026-03-18
feat: Add currency display options to CurrencyFormatter (#7785)

younies:
Enhance the CurrencyFormatterOptions struct to include a `currency_display` field, allowing for different display styles (Symbol, NarrowSymbol, Code, Name). 

Implement tests for each display style in the format module and adding todo to implement these test cases.

## Changelog

icu_experimental/currency: Add options for selecting currency format style
- New enum: `CurrencyDisplay`
- New field `CurrencyFormatterOptions::currency_display`
- New impl: `impl From<CurrencyDisplay> for CurrencyFormatterOptions`

### DIFF
--- a/components/experimental/src/dimension/currency/options.rs
+++ b/components/experimental/src/dimension/currency/options.rs
@@ -15,12 +15,46 @@ use serde::{Deserialize, Serialize};
 pub struct CurrencyFormatterOptions {
     /// The width of the currency format.
     pub width: Width,
+
+    /// The display style for currencies.
+    /// Default is [`CurrencyDisplay::Symbol`].
+    pub currency_display: CurrencyDisplay,
 }
 
 impl From<Width> for CurrencyFormatterOptions {
     fn from(width: Width) -> Self {
-        Self { width }
+        Self {
+            width,
+            currency_display: CurrencyDisplay::default(),
+        }
     }
+}
+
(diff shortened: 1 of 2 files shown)


younies:
@sffc 

question: As mentioned in https://github.com/unicode-org/icu4x/issues/7785, ECMA-402 uses display options for currency formatting. However, this seems to overlap or potentially conflict with the `width` option.

For example, if a user sets `width: Full/Long` and `display_option: symbol`, what should the expected result be?

components/experimental/src/dimension/currency/options.rs
```diff
@@ -15,14 +15,48 @@ use serde::{Deserialize, Serialize};
 pub struct CurrencyFormatterOptions {
     /// The width of the currency format.
     pub width: Width,
+
+    /// The display style for currencies.
+    /// Default is [`CurrencyDisplay::Symbol`].
+    pub currency_display: CurrencyDisplay,
```
sffc, line 21: As noted in the other PR, options should be optional:

```suggestion
    pub currency_display: Option<CurrencyDisplay>,
```

sffc (approved):
I think we should remove the `width` option, since it is covered by the `currency_display` option.

robertbastian:
@younies is this PR moot with the current design?

younies:
i think we can close this PR for now, I am going to send more PRs to implement our current design

PR #7790, opened 2026-03-18
Add APIs and tests for currency sign options to CurrencyFormatter

younies:
## Summary

- Introduced `CurrencySign` enum to control formatting of negative currency values, supporting both standard and accounting styles.
- Updated `CurrencyFormatterOptions` to include a `currency_sign` field, defaulting to `Standard`.
- Added tests for accounting style formatting in `test_en_us_accounting` to implement the new behaviour in the next PR

## Changelog

icu_experimental/currency: Add options for accounting currency sign
- New enum: `CurrencySign`
- New field `CurrencyFormatterOptions::currency_sign`
- New impl: `impl From<CurrencySign> for CurrencyFormatterOptions`

### DIFF
--- a/components/experimental/src/dimension/currency/options.rs
+++ b/components/experimental/src/dimension/currency/options.rs
@@ -15,11 +15,27 @@ use serde::{Deserialize, Serialize};
 pub struct CurrencyFormatterOptions {
     /// The width of the currency format.
     pub width: Width,
+
+    /// The sign style for negative currency values.
+    /// Default is [`CurrencySign::Standard`].
+    pub currency_sign: CurrencySign,
 }
 
 impl From<Width> for CurrencyFormatterOptions {
     fn from(width: Width) -> Self {
-        Self { width }
+        Self {
+            width,
+            currency_sign: CurrencySign::default(),
+        }
+    }
+}
+
+impl From<CurrencySign> for CurrencyFormatterOptions {
+    fn from(currency_sign: CurrencySign) -> Self {
+        Self {
+            width: Width::default(),
+            currency_sign,
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


components/experimental/src/dimension/currency/options.rs
```diff
@@ -15,11 +15,27 @@ use serde::{Deserialize, Serialize};
 pub struct CurrencyFormatterOptions {
     /// The width of the currency format.
     pub width: Width,
+
+    /// The sign style for negative currency values.
+    /// Default is [`CurrencySign::Standard`].
+    pub currency_sign: CurrencySign,
```
sffc, line 21: Nit: All options should be optional

```suggestion
    pub currency_sign: Option<CurrencySign>,
```

Probably `width` should be optional, too.

sffc (approved):

sffc:
Note: the changelog needs to follow a specific format. You had put:

> - Introduced `CurrencySign` enum to support standard and accounting currency sign styles.
> - Added `currency_sign` to `CurrencyFormatterOptions`, with `Standard` as the default.
> - Added `en-US` accounting-style formatting tests.

I changed it, according to https://github.com/unicode-org/icu4x/blob/main/documents/process/changelog.md

Also, now that we have the `## Changelog` section with the specific APIs, your `## Summary` section (which doesn't need a header) should describe what the PR accomplishes at a high level.

robertbastian (commented):
I want to raise my concern about this style of work. Adding API options that do nothing, and tests that are allowed to fail gives a false sense of how far along implementation of this functionality is. This kind of dead code is also in the way when working on this component, as I've encountered when I converted the currency data to use the compact data struct.

The accounting style formatting is going to need considerable work, because you will have to untangle the `DecimalFormatter` from the `CurrencyFormatter`, decompose the accounting currency patterns into currency and decimal patterns in datagen, and then combine them back at runtime.

Tests should be added in the PR that actually change behaviour, because then you will know which edge cases even exist and need testing.

Tests should only be annotated with `#[should_panic]` if they are testing an API that is documented as panicking. `#[should_panic]` should not be used to allow failed assertions, either use `assert_ne!`, change the expected value to what *the code actually returns*, or just don't add this test in the first place.

PR #7792, opened 2026-03-18
Update to CLDR 48.2.0

sffc:
https://github.com/unicode-org/cldr-json/pull/99

## Changelog

meta: updated to CLDR 48.2

### DIFF
--- a/provider/source/tests/data/cldr/cldr-core/supplemental/metaZones.json
+++ b/provider/source/tests/data/cldr/cldr-core/supplemental/metaZones.json
@@ -2338,7 +2338,9 @@
             "Vancouver": [
               {
                 "usesMetazone": {
-                  "_mzone": "America_Pacific"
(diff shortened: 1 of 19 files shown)


provider/source/tests/data/cldr/cldr-core/supplemental/metaZones.json
```diff
@@ -2338,7 +2338,9 @@
             "Vancouver": [
               {
                 "usesMetazone": {
-                  "_mzone": "America_Pacific"
+                  "_mzone": "America_Pacific",
```
echeran, line 2341: There should be a test to assert that ICU4X behavior supports the new Vancouver time zone after the corresponding ICU4X data update here.

sffc: #7793

PR #7795, opened 2026-03-19
Handle possible Overflow values on individual calendars

### LINKED ISSUE #7794: Adding 1 year to Adar I fails in non-leap year
When adding one year to a Hebrew date in Adar I of a leap year, the operation fails if the following year is not a leap year.

In the Hebrew calendar, leap years contain an extra month, Adar I. If a date falls in Adar I and you add one year, the next year may not contain Adar I. In that case, the operation currently fails with:

```
called `Result::unwrap()` on an `Err` value: UnknownMonthCode(MonthCode("M05L"))
```
I would expect this to resolve to the regular Adar in the following year instead.

## Reproduction

```rust
#[test]
fn add_year_to_date_with_adar_i() {
    // Adar I exists only in a leap year
    let adar_i = MonthCode::new_leap(5).unwrap();

    // 10 Adar I 5787
    // 5787 is a leap year; 5788 is not
    let mut date = Date::try_new_from_codes(None, 5787, adar_i, 10, Hebrew).unwrap();

    // Panics with:
    // called `Result::unwrap()` on an `Err` value: UnknownMonthCode(MonthCode("M05L"))
    date.try_add_with_options(DateDuration::for_years(1), DateAddOptions::default())
        .unwrap();
}
```

I'm on version 2.1.1 of `icu_calendar`

---

Thank you so much for working on this library. It has helped me tremendously.

Manishearth:
Fixes #7794.

`overflow` takes three values: `Constrain`, `Reject`, and `None`. `None` is equivalent to `Constrain`, but we don't have a `ResolvedOptions` type to fix that distinction.

_Almost_ all of `overflow`-consuming code is matching the Temporal spec, and the Temporal spec handles unset values correctly, so this is by and large not a problem.

However, _calendar-specific_ code is not specced by Temporal, and any place where we handle overflows there needs to handle all cases.

@sffc and I discussed this and using `match` statements seemed to be the most straightforward option. We could in theory introduce `ResolvedOptions` or pre-handle the None but most of the code is written to match the spec so this would actually potentially make the spec-matching less clear.

## Changelog

icu_calendar: Handle possible Overflow values on individual calendars

### DIFF
--- a/components/calendar/src/cal/east_asian_traditional.rs
+++ b/components/calendar/src/cal/east_asian_traditional.rs
@@ -647,7 +647,7 @@ impl<R: Rules> DateFieldsResolver for EastAsianTraditional<R> {
         &self,
         year: Self::YearInfo,
         month: types::Month,
-        options: DateFromFieldsOptions,
+        overflow: Overflow,
     ) -> Result<u8, MonthError> {
         let (number @ 1..=12, leap) = (month.number(), month.is_leap()) else {
             return Err(MonthError::NotInCalendar);
@@ -665,10 +665,10 @@ impl<R: Rules> DateFieldsResolver for EastAsianTraditional<R> {
 
         if leap {
             // This leap month doesn't exist in the year, reject if needed
-            match options.overflow {
-                Some(Overflow::Reject) => return Err(MonthError::NotInYear),
+            match overflow {
+                Overflow::Reject => return Err(MonthError::NotInYear),
                 // Written as a match for exhaustiveness
-                Some(Overflow::Constrain) | None => (),
+                Overflow::Constrain => (),
             }
         }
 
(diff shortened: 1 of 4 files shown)


components/calendar/src/date.rs
```diff
@@ -793,4 +793,33 @@ mod tests {
...
+        //
+        let mut date = Date::try_new(
+            2025.into(),
+            Month::leap(6),
+            1,
+            ChineseTraditional::default(),
+        )
+        .unwrap();
+        date.try_add_with_options(DateDuration::for_years(1), DateAddOptions::default())
+            .unwrap();
```
sffc, line 820:
```suggestion
        // 10 Adar I 5787
        // 5787 is a leap year; 5788 is not
        let mut date = Date::try_new(5787.into(), Month::leap(5), 10, Hebrew).unwrap();
        date.try_add_with_options(DateDuration::for_years(1), DateAddOptions::default())
            .unwrap();
        assert_eq!(date.month().to_input(), Month::new(6));

        // Leap Month 6, day 1, 2025.
        // 2025 is a leap year; 2026 is not
        let mut date = Date::try_new(
            2025.into(),
            Month::leap(6),
            1,
            ChineseTraditional::default(),
        )
        .unwrap();
        date.try_add_with_options(DateDuration::for_years(1), DateAddOptions::default())
            .unwrap();
```

PR #7796, opened 2026-03-20
Use 32-bit arithmetic for days

### LINKED ISSUE #7771: Arithmetic code should use 32-bit maths
The largest difference between two of our `Date`s is the size of the `VALID_RD_RANGE`, which is smaller than 2 * 1 000 000 * 366, which comfortably fits in an `i32`. `DateDuration::days` should be `u32` instead of `u64`, and all arithmetic code should use `i32` instead of `i64`s internally. This should significantly speed up this code on 32-bit systems, including WASM.

Manishearth:
Fixes #7771

## Changelog: N/A

Unreleased API, internal change.

### DIFF
--- a/components/calendar/src/tests/extrema.rs
+++ b/components/calendar/src/tests/extrema.rs
@@ -186,7 +186,8 @@ super::test_all_cals!(
 
             for generous_duration in [generous_max_duration, generous_max_neg_duration] {
                 for overflow in [CONSTRAIN, REJECT] {
-                    assert!(start_date.try_added_with_options(generous_duration, overflow).is_err(), "Adding durations from the generously-large range should always fail (but never panic)");
+                    assert_eq!(start_date.try_added_with_options(generous_duration, overflow), Err(DateAddError::Overflow),
+                               "Adding durations from the generously-large range should always fail (but never panic)");
                 }
             }
         }


components/calendar/src/cal/east_asian_traditional.rs
```diff
@@ -1219,8 +1219,8 @@ mod test {
                 (largest.to_extended_year() - smallest.to_extended_year() + 1) * 12 + num_leap;
             let approximated = EastAsianTraditional::<R>::min_months_from(
                 smallest,
-                (largest.to_extended_year() - smallest.to_extended_year()) as i64 + 1,
-            ) as i32;
+                (largest.to_extended_year() - smallest.to_extended_year()) + 1,
```
Manishearth, line 1219: yay

components/calendar/src/tests/arithmetic.rs
```diff
@@ -273,12 +273,12 @@ super::test_all_cals!(
...
                     let expected_rd = date
                         .try_added_with_options(
-                            DateDuration::for_months(signed_months as i32),
+                            DateDuration::for_months(signed_months),
                             add_constrain,
                         )
                         .unwrap()
                         .to_rata_die()
-                        + signed_months.signum();
+                        + i64::from(signed_months.signum());
```
Manishearth, line 281: boo (but actually, it's fine, this is RD)

components/calendar/src/tests/arithmetic.rs
```diff
@@ -312,7 +312,7 @@ super::test_all_cals!(
                         .try_added_with_options(year_month_duration, add_constrain)
                         .unwrap()
                         .to_rata_die()
-                        + if duration.is_negative { -1 } else { 1 } as i64;
+                        + if duration.is_negative { -1i64 } else { 1i64 };
```
Manishearth, line 315: good job, agent

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -629,14 +629,14 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
             //   1. Set _daysInMonth_ to CalendarDaysInMonth(_calendar_, _resolvedYear_, _resolvedMonth_).
             //   1. Set _resolvedDay_ to _resolvedDay_ + _daysInMonth_.
             days_in_month = C::days_in_provided_month(resolved_year, resolved_month);
-            resolved_day += i64::from(days_in_month);
+            resolved_day += days_in_month as i32;
         }
         // 1. Repeat, while _resolvedDay_ &gt; _daysInMonth_,
-        while resolved_day > i64::from(days_in_month) {
+        while resolved_day > days_in_month as i32 {
```
Manishearth: I don't like these `as i32`s. Asking the agent to fix that.

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -802,12 +796,7 @@ impl<C: DateFieldsResolver> ArithmeticDate<C> {
             return false;
         }
         // 1. Let _endOfMonth_ be BalanceNonISODate(_calendar_, _monthsAdded_.[[Year]], _monthsAdded_.[[Month]] + 1, 0).
-        let end_of_month = Self::new_balanced(
-            months_added.year,
-            i64::from(months_added.ordinal_month) + 1,
-            0,
-            cal,
-        );
+        let end_of_month = Self::new_balanced(y0, duration.add_months_to(m0) + 1, 0, cal);
```
sffc: Question/Issue: Why did you change the parameters to this function? They don't align with the spec text any more.

components/calendar/src/duration.rs
```diff
@@ -350,32 +326,32 @@ impl DateDuration {
     }
 
     #[inline]
-    pub(crate) fn add_months_to(&self, month: u8) -> i64 {
+    pub(crate) fn add_months_to(&self, month: u8) -> i32 {
         if !self.is_negative {
-            i64::from(month) + i64::from(self.months)
+            i32::from(month) + (self.months as i32)
```
sffc, line 332: Nit: When using an `as` cast, please debug assert that it is safe

components/calendar/src/calendar_arithmetic.rs
```diff
@@ -1231,8 +1227,8 @@ impl<'a, C: DateFieldsResolver> SurpassesChecker<'a, C> {
         // 15. Return CompareSurpasses(sign, balancedDate.[[Year]], balancedDate.[[Month]], balancedDate.[[Day]], calDate2).
         let balanced_date = ArithmeticDate::<C>::new_balanced(
             self.end_of_month.year,
-            i64::from(self.end_of_month.ordinal_month),
-            7 * weeks + i64::from(self.regulated_day),
+            i32::from(self.end_of_month.ordinal_month),
+            7 * weeks + i32::from(self.regulated_day),
```
sffc, line 1241: Praise: Thanks for using `i32::from` instead of `as i32` where possible

components/calendar/src/calendar_arithmetic.rs
sffc, line 869: Observation: This lets through a duration with maximal values for years, months, weeks, and days together. Are there tests that check whether this results in non-overflowing behavior in all calendars? If there isn't such a test, it should be easy to write. I would consider such a test a prerequisite for landing the internal optimization part of this PR, since the value proposition of 2.2 is that we are making icu_calendar less panicky.

components/calendar/src/calendar_arithmetic.rs
Manishearth: Huh. Didn't notice this during review, my bad. This is the type of mistake that I am surprised the agent made; and I'm going to update the types of special scrutiny I have on agent-driven changes.

components/calendar/src/tests/extrema.rs
```diff
@@ -170,6 +168,31 @@ super::test_all_cals!(
...
+
+        let mut generous_max_neg_duration = generous_max_duration;
+        generous_max_neg_duration.is_negative = true;
+
+        for start_date in RDS_TO_TEST {
+            let start_date = Date::from_rata_die(*start_date, cal);
+
+            for generous_duration in [generous_max_duration, generous_max_neg_duration] {
+                for overflow in [CONSTRAIN, REJECT] {
+                    assert!(start_date.try_added_with_options(generous_duration, overflow).is_err(), "Adding durations from the generously-large range should always fail (but never panic)");
```
sffc: Nit (optional): assert that it is an Overflow error. I think it should always be?

sffc (approved):

PR #7800, opened 2026-03-22
Add UCD source to datagen

opnuub:
Issue: #6941 

@sffc 
Added IdentifierStatus.txt as reference for radicals codepoint trie to make it more robust

## Changelog

- icu_provider_source: Filter Unihan radical data using UCD `IdentifierStatus.txt`
  * New methods: `SourceDataProvider::with_ucd()`, `SourceDataProvider::is_missing_ucd_error()`
  * Behavior change: Unihan radical data now excludes code points not present in `IdentifierStatus.txt`

- icu4x-datagen: Add `--ucd-root` for UCD `IdentifierStatus.txt` input used by radical generation


provider/icu4x-datagen/src/main.rs
```diff
@@ -185,6 +185,11 @@ struct Cli {
     #[cfg(feature = "provider")]
     unihan_root: Option<PathBuf>,
 
+    #[arg(long, value_name = "PATH")]
+    #[arg(help = "Path to a local UCD directory containing IdentifierStatus.txt.")]
+    #[cfg(feature = "provider")]
+    ucd_root: Option<PathBuf>,
```
Manishearth, line 193: question: is there a reason we can't use `ucd-tag`? I think it makes sense to support both, but I'd prefer if we do that as seaprate changes, especially this close to the release (since `ucd-root` is a new public API in the form of a CLI flag)

provider/icu4x-datagen/src/main.rs
```diff
@@ -441,6 +446,8 @@ fn run(cli: Cli) -> eyre::Result<()> {
             eyre::bail!(
                 "Unihan data is required for this invocation, set --unihan-root or --ucd-tag"
             );
+        } else if SourceDataProvider::is_missing_ucd_error(e) {
+            eyre::bail!("UCD data is required for this invocation, set --ucd-root");
```
sffc:
```suggestion
            eyre::bail!("UCD data is required for this invocation, set --ucd-root or --ucd-tag");
```

provider/icu4x-datagen/src/main.rs
sffc: We previously decided to have 3 flags

1. `--unihan-root`
2. `--ucd-root`
3. `--ucd-tag`

Where `--ucd-tag` is the tag used for both of the other two. We hadn't added `--ucd-root` because we didn't need any data files yet from UCD.

provider/icu4x-datagen/src/main.rs
```diff
@@ -534,6 +539,11 @@ fn run(cli: Cli) -> eyre::Result<()> {
                 (None, _) => p,
             };
 
+            p = match cli.ucd_root {
+                Some(path) => p.with_ucd(&path)?,
+                None => p,
+            };
+
```
sffc: Please add `.with_ucd_for_tag` here, reading from `--ucd-tag`

sffc (approved):

provider/source/src/lib.rs
```diff
@@ -318,6 +333,22 @@ impl SourceDataProvider {
...
+    /// ✨ *Enabled with the `networking` Cargo feature.*
+    #[cfg(feature = "networking")]
+    pub fn with_ucd_for_tag(self, tag: &str) -> Self {
+        Self {
+            ucd_paths: Some(Arc::new(AbstractFs::new_from_url(format!(
+                "https://www.unicode.org/Public/{tag}/"
+            )))),
+            ..self
+        }
+    }
```
robertbastian, line 350: this does not work, and is also untested in this PR

provider/source/src/lib.rs
```diff
@@ -230,6 +236,15 @@ impl SourceDataProvider {
         })
     }
 
+    /// Adds UCD source data to the provider. The path should point to a
+    /// directory containing `security/IdentifierStatus.txt`.
+    pub fn with_ucd(self, root: &Path) -> Result<Self, DataError> {
```
robertbastian, line 241: The proposed name `--ucd-root` (https://github.com/unicode-org/icu4x/issues/7510) was meant to refer to `/ucd/`, i.e. `/ucd/UCD.zip`. Making it refer to "a directory containing `security/IdentifierStatus.txt`", i.e. `/`, was definitely never agreed.

robertbastian (commented):
I wish I'd been around to push back on this, and I'm disappointed that this just got waved through, especially so close to the release.

robertbastian:
Also for the purpose of @opnuub's unstable, test-only segmenter work, there was no need to make any of this public API.

PR #7801, opened 2026-03-23
Write smarter changelog tool in Rust

Manishearth:
Our changelog process still calls for manual collation, but there's the boring part (collating by crate and reformatting) and the interesting part (actually prioritizing items in the changelog).

I made the script do most of the boring work.

It produces a report that looks like this: https://hackmd.io/XKCDISO_RhmAH7Ib1zqhjw. The first section of the report contains changelog entries organized by crate. After that the report includes information on additional "notes" found in some PRs that were not a part of the changelog entries but should be considered in the final collation, as well as a list of PRs without changelog information and PRs marked N/A for spot-checking.

A neat thing is that this separates the github fetching from the changelog generation: so you can fetch github state once to JSON (you can even update state for individual PRs in the json), and then run the tool repeatedly.

It currently uses the `gh` tool because I didn't want to set up a proper login flow.

This is a run-every-few-months tool, so I would prefer to land this as is, with improvements happening over time, without extensive code style/structure review. The maintenance burden of such a tool is low, and if it stops working, it stops working; it is not critical to anything.

## Changelog: N/A

Internal

### DIFF
--- a/tools/changelog/Cargo.toml
+++ b/tools/changelog/Cargo.toml
@@ -1,3 +1,7 @@
+# This file is part of ICU4X. For terms of use, please see the file
+# called LICENSE at the top level of the ICU4X source tree
+# (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
 [package]
 name = "changelog"
 version.workspace = true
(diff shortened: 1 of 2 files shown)


Manishearth:
Oops, pushed before noticing the review.

PR #7807, opened 2026-03-25
RegionDisplayName::try_new_short

sffc:
#3260

I used an AI, reviewed manually at first, prompted the AI to fix things, and I'm happy enough with the result that I am asking for a review.

## Changelog

N/A (new methods on a type that was newly added in #7697)

### DIFF
--- a/components/experimental/src/displaynames/displaynames.rs
+++ b/components/experimental/src/displaynames/displaynames.rs
@@ -66,8 +66,7 @@ impl RegionDisplayName {
         /// use icu::locale::{locale, subtags::region};
         /// use writeable::assert_writeable_eq;
         ///
-        /// let mut prefs = DisplayNamesPreferences::default();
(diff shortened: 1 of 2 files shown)


components/experimental/src/displaynames/displaynames.rs
```diff
@@ -100,6 +99,72 @@ impl RegionDisplayName {
             .payload;
         Ok(Self { payload })
     }
+
+    icu_provider::gen_buffer_data_constructors!(
+        (prefs: DisplayNamesPreferences, region: Region) -> result: Result<Self, DataError>,
+        /// Loads the short region display name for a given region and locale using compiled data.
+        /// It will fall back to the long name if the short name is not available.
```
Manishearth, line 106: this was basically the only thing i was going to suggest we do, praise: it's already done

Manishearth (approved):
Yeah, this makes sense. I presume you're going to follow this template for the rest.

PR #7816, opened 2026-03-26
Update Cargo versions for the release

Manishearth:
https://github.com/unicode-org/icu4x/issues/7809

## Changelog: N/A


utils/zoneinfo64/src/lib.rs
```diff
@@ -78,7 +78,6 @@ const SECONDS_IN_UTC_DAY: i64 = 24 * 60 * 60;
 
 /// An offset from UTC time (stored to seconds precision)
 #[derive(Debug, Clone, Copy, PartialEq, Default)]
-#[non_exhaustive] // newtype
```
Manishearth, line 81: This has a private field, nonexhaustive is unnecessary

sffc (approved):

PR #7818, opened 2026-03-26
Language script

sffc:
#3260

## Changelog

icu_experimental/displaynames:
- New types: `LanguageDisplayName`, `RegionDisplayName`


sffc:
I'm going to do this one without an agent

Issue #7134, opened 2025-10-21
Update line segmenter (UAX#14) to Unicode 17


makotokato:

Manishearth:
@makotokato any update on this? Do you have a rough timeline for this getting completed.

Manishearth:
(We can add it to the 2.2 milestone but would like to understand timelines for that)

Manishearth:
https://github.com/unicode-org/icu4x/issues/7828

Manishearth:
Reopening since we're reverting this for the release. @makotokato, can you re-make your PR and let @eggrobin review it? We can try to have it in the 2.3 milestone.

https://github.com/unicode-org/icu4x/pull/7830

PR #5218, opened 2024-07-10
Support Unicode 15.1 for line segmenter

makotokato:
This fix supports Unicode 15.1 for line segmenter. (a part of https://github.com/unicode-org/icu4x/issues/3255)

- Added LB15a handling to engine code since LB15a references previous breaking rules.
- Adding fallback handling into intermediate status for LB28a
- Modified tests like https://github.com/unicode-org/icu4x/pull/4389

Issue #1637, opened 2022-02-25
line segmenter's DataProvider should have some line break property


makotokato:
Actually line segmenter code references AI, AL, BA, CJ, CM, EX, H2, H3, HY, ID, IN, JL, JT, JV, NS, NU, PO_EAW, PR_EAW, SA and XX from generated state machine table directly, I think we should move this to `DataProvider`/`LineBreakDataV1`.  Word segmenter (rule segmenter) has  some property index such as SOT into `RuleBreakDataV1`.

CC: @aethanyc

aethanyc:
Yup, I also think about this problem recently. These symbol should be defined in `LineBreakDataV1`.

The numeric value of these symbols AI, AL, ... are tight to the data in `line.toml`, and can be referenced by utilities like https://github.com/unicode-org/icu4x/blob/4c22ef7d6aa78aa8d1bc4f7193619ccd03f65976/experimental/segmenter/src/line.rs#L319-L343

The most straightforward way is to add more fields in
```
pub struct LineBreakDataV1 {
  ...
  pub ai_property: u8,
  pub al_property: u8,
  ...
}
```

But, this soon become tedious. I'm not sure if there's a better way.

aethanyc:
Maybe we can store the numeric values in `line.toml` so that it can decouple from the numeric value in [`impl LineBreak`]( https://github.com/unicode-org/icu4x/blob/92a15945dbea37e761e9c8be1eab1edc41f05f9c/components/properties./src/props.rs#L535-L579)

```
[[tables]]
name = "AI"
value = 1

[[tables]]
name = "AL"
value = 2
```

makotokato:
Since this is hard-coded, when updating line.toml for newer spec, we have to increment version of `LineBreakData` etc.

PR #7823, opened 2026-03-30
Update line segmenter to Unicode 17

makotokato:
Fix https://github.com/unicode-org/icu4x/issues/7134.

Current line segmenter has multiple properties for East Asian language, but we handle it by LSTM, so I modify original `LineBreakTest.txt`.

## Changelog

icu_segmenter: Update line segmenter to Unicode 17.0.0


components/segmenter/src/line.rs
```diff
@@ -21,111 +21,139 @@ const UNKNOWN: u8 = 0;
 #[allow(dead_code)]
 const AI: u8 = 1;
 #[allow(dead_code)]
-const AK: u8 = 2;
+const AI_EASTASIAN: u8 = 2;
```
Manishearth, line 24: observation: this change is due to https://github.com/unicode-org/icu4x/issues/1637. Not great, we should fix that, but fine for now.

Manishearth:
> Current line segmenter has multiple properties for East Asian language, but we handle it by LSTM, so I modify original LineBreakTest.txt.



I'd prefer if we didn't modify the original test file, instead storing the original test file and having some way of skipping tests manually, so it's clear which tests we don't pass. Ok for followup.

Manishearth:
Actually, we should do something like this: https://github.com/unicode-org/icu4x/pull/7827

eggrobin:
This is out of date and needs to be updated:

https://github.com/unicode-org/icu4x/blob/90b7762e80e4b8871d13e3d7aec7681fc1b84b2e/components/segmenter/tests/testdata/LineBreakRandomTest.txt#L7-L9

@makotokato, How many random tests did you run locally to test this? Please indicate that in the PRs as I did in https://github.com/unicode-org/icu4x/pull/5218#issuecomment-2329760147.

eggrobin:
> instead storing the original test file and having some way of skipping tests manually, so it's clear which tests we don't pass. Ok for followup.

That is fundamentally unmaintainable; these tests very numerous, machine-generated, and not human-readable. @makotokato is following our existing practice here of having a readable diff of the code that generates those tests.

Manishearth:
> That is fundamentally unmaintainable; these tests very numerous, machine-generated, and not human-readable. @makotokato is following our existing practice here of having a readable diff of the code that generates those tests.

I see! That works fine, then.

Given that ICU4X exposes "run LBA without LSTM/Dictionary" is there actually a reason we should be tweaking these tests, though?

Manishearth:
Given that @eggrobin wants to review before this gets released, I've reverted these changes in https://github.com/unicode-org/icu4x/pull/7830

eggrobin:
LineBreakRandomTest notes
https://github.com/unicode-org/icu4x/blob/90b7762e80e4b8871d13e3d7aec7681fc1b84b2e/components/segmenter/tests/testdata/LineBreakRandomTest.txt#L1-L9

> How many random tests did you run locally to test this?

I think the answer was « exactly 200 » (and « not enough »), because when running a larger set of random tests, I see a failure on test case test case #&zwnj;202, with an unexpected break in OP NU CL ÷ IS AL
```
  | A | E | Code pt. | Line_Break         | General_Category   | East_Asian_Width | Literal
  | ÷ | ÷ |     000A |          Line_Feed |            Control |          Neutral |
  | ÷ | ÷ |     0029 |  Close_Parenthesis |  Close_Punctuation |           Narrow | )
  | ÷ | ÷ |     20C2 |     Prefix_Numeric |         Unassigned |          Neutral | ⃂
  | × | × |     FE5D |   Open_Punctuation |   Open_Punctuation |             Wide | ﹝
  | × | × |    110F2 |            Numeric |     Decimal_Number |          Neutral | 𑃲
  | × | × |     2773 |  Close_Punctuation |  Close_Punctuation |          Neutral | ❳
😭| ÷ | × |     003B |      Infix_Numeric |  Other_Punctuation |           Narrow | ;
  | × | × |    1457B |         Alphabetic |       Other_Letter |          Neutral | 𔕻
  | ÷ | ÷ |    1F3FF |         E_Modifier |    Modifier_Symbol |             Wide | 🏿
  | ÷ | ÷ |    11020 |             Aksara |       Other_Letter |          Neutral | 𑀠
```

eggrobin:
To be clear, while the random example is of course nonsense, the bug it uncovers can occur in normal text; for instance, the state table in this PR yields an inappropriate line break opportunity before the full stop in
> The set of solutions is {1, 2}.

It is not feasible to try to discern which failures are acceptable; we need to implement the algorithm conformantly, and check that with millions, not hundreds, of random tests. Even if we implement a tailoring, the tailoring should be expressed in terms of changes to the UAX14 rules, not in terms of which tests don’t pass.

PR #7826, opened 2026-03-30
Split experimental displaynames into modules

sffc:
#3260

## Changelog

icu_experimental/displaynames: Split formatting types into `multi` and `single` modules; seeking feedback
- Moved to `displaynames::multi`: `LanguageDisplayNames`, `LocaleDisplayNamesFormatter`, `RegionDisplayNames`, `ScriptDisplayNames`, `VariantDisplayNames`
- Introduced in `displaynames::single`: `RegionDisplayName`, `ScriptDisplayName`

### DIFF
--- a/components/experimental/src/displaynames/mod.rs
+++ b/components/experimental/src/displaynames/mod.rs
@@ -39,7 +39,7 @@
 //! let gb = RegionDisplayName::try_new(locale, region!("GB")).unwrap();
 //! assert_writeable_eq!(us, "United States");
 //! assert_writeable_eq!(gb, "United Kingdom");
-//! 
+//! ```
 
 // TODO: expand documentation
 


components/experimental/src/displaynames/mod.rs
```diff
@@ -11,15 +50,44 @@ mod options;
...
+    //! ### Status
+    //!
+    //! Currently, this module has limited support. It supports regions and scripts,
+    //! but support for languages, locales, and variants is currently missing.
+    //! More features are on their way.
+    //!
+    //! If you have any feedback, please let us know at
+    //! <https://github.com/unicode-org/icu4x/issues/7825>.
+    //!
+    //! See [`mod@super`] for a comparison of single and multi.
```
Manishearth: nit: `[the parent module](mod@super)`, otherwise this will show `mod@super` to the user.

components/experimental/src/displaynames/mod.rs
```diff
@@ -3,6 +3,45 @@
...
+//! let us = RegionDisplayName::try_new(locale, region!("US")).unwrap();
+//! let gb = RegionDisplayName::try_new(locale, region!("GB")).unwrap();
+//! assert_eq!(us.to_string(), "United States");
+//! assert_eq!(gb.to_string(), "United Kingdom");
+//! ```
+//!
+//! Currently, the data between the two modules is NOT shared, but we hope to
+//! make it shared in the future.
+//!
+//! We are not sure which design is better for users. If you have any feedback
```
Manishearth: thought: I guess this isn't _quite_ the question: since we want users to have access to both "designs", for a broad pair of sets of designs, it's the specific choice of designs we're not sure about. I think this is fine for now, though.

Maybe: We're not sure if we've picked the right set of tradeoffs here, if you have feedback (or just want to tell us about your use case), ..

components/experimental/src/displaynames/mod.rs
```diff
@@ -3,6 +3,45 @@
...
 //! Display names for languages and regions.
+//!
+//! There are currently two designs for how to use this component:
+//!
+//! 1. [`multi`]: Load multiple display names at once.
+//! 2. [`single`]: Load a single display name at a time.
+//!
+//! See the documentation for those submodules for more information.
+//!
+//! ### Comparison of Single and Multi
```
sffc:
```suggestion
//! ## Examples
```

components/experimental/src/displaynames/mod.rs
sffc: I'm still not convinced "single" is a use case that needs its own type. "multi" can get a single name. Conversely, @robertbastian would argue that "multi" does not need its own type, since you can get multiple single names.

components/experimental/src/displaynames/mod.rs
```diff
@@ -3,6 +3,45 @@
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
 //! Display names for languages and regions.
+//!
+//! There are currently two designs for how to use this component:
+//!
+//! 1. [`multi`]: Load multiple display names at once.
+//! 2. [`single`]: Load a single display name at a time.
+//!
+//! See the documentation for those submodules for more information.
```
sffc:
```suggestion
//! There are multiple use cases for this type, and we are not yet committed to either of
//! these designs being the "primary" design. Please share feedback at
//! <https://github.com/unicode-org/icu4x/issues/7824>.
//!
//! Note: Currently, the data between the two modules is NOT being shared.
```

components/experimental/src/displaynames/mod.rs
```diff
@@ -3,6 +3,45 @@
...
+//! // Multi: Load a formatter that can display many regions.
+//! let multi = RegionDisplayNames::try_new(locale, DisplayNamesOptions::default()).unwrap();
+//! assert_eq!(multi.of(region!("US")), Some("United States"));
+//! assert_eq!(multi.of(region!("GB")), Some("United Kingdom"));
+//!
+//! // Single: Load only the region(s) we need.
+//! let us = RegionDisplayName::try_new(locale, region!("US")).unwrap();
+//! let gb = RegionDisplayName::try_new(locale, region!("GB")).unwrap();
+//! assert_eq!(us.to_string(), "United States");
+//! assert_eq!(gb.to_string(), "United Kingdom");
```
sffc:
```suggestion
//! assert_writeable_eq!(us, "United States");
//! assert_writeable_eq!(gb, "United Kingdom");
```

components/experimental/src/displaynames/mod.rs
```diff
@@ -3,6 +3,45 @@
...
+//! use icu::experimental::displaynames::single::RegionDisplayName;
+//! use icu::experimental::displaynames::DisplayNamesOptions;
+//! use icu::locale::{locale, subtags::region};
+//!
+//! let locale = locale!("en").into();
+//!
+//! // Multi: Load a formatter that can display many regions.
+//! let multi = RegionDisplayNames::try_new(locale, DisplayNamesOptions::default()).unwrap();
+//! assert_eq!(multi.of(region!("US")), Some("United States"));
+//! assert_eq!(multi.of(region!("GB")), Some("United Kingdom"));
```
sffc:
```suggestion
//! assert_writeable_eq!(multi.of(region!("US")).unwrap(), "United States");
//! assert_writeable_eq!(multi.of(region!("GB")).unwrap(), "United Kingdom");
```

components/experimental/src/displaynames/mod.rs
```diff
@@ -3,6 +3,49 @@
...
+//! assert_writeable_eq!(us, "United States");
+//! assert_writeable_eq!(gb, "United Kingdom");
+//! ```
+//!
+//! Currently, the data between the two modules is NOT shared, but we hope to
+//! make it shared in the future.
+//!
+//! We are not sure which design is better for users. If you have any feedback
+//! on the design of this component, please let us know at
+//! <https://github.com/unicode-org/icu4x/issues/7824>.
```
sffc:
```suggestion
//! ```
```

components/experimental/src/displaynames/mod.rs
```diff
@@ -11,15 +54,44 @@ mod options;
...
+    //! ### Status
+    //!
+    //! Currently, this module has limited support. It supports regions and scripts,
+    //! but support for languages, locales, and variants is currently missing.
+    //! More features are on their way.
+    //!
+    //! If you have any feedback, please let us know at
+    //! <https://github.com/unicode-org/icu4x/issues/7825>.
+    //!
+    //! See [`mod@super`] for a comparison of single and multi.
```
sffc:
```suggestion
    //! See [the parent module](mod@super) for a comparison of single and multi.
```

components/experimental/src/displaynames/mod.rs
sffc: sorry for making you comment on this AI slop; I should have fixed this before sending it for review.

Does `mod@super` actually work? I haven't seen that syntax before.

components/experimental/src/displaynames/mod.rs
```diff
@@ -11,15 +54,44 @@ mod options;
...
+    pub use displaynames::VariantDisplayNames;
+}
+
+pub mod single {
+    //! Types for loading a single display name at a time.
+    //!
+    //! This submodule is useful for applications that only need to display one or
+    //! two specific names, such as the name of the current region.
+    //!
+    //! See [`mod@super`] for a comparison of single and multi.
```
sffc:
```suggestion
    //! See [the parent module](mod@super) for a comparison of single and multi.
```

components/experimental/src/displaynames/mod.rs
Manishearth: Okay, that's fine then.

components/experimental/src/displaynames/mod.rs
Manishearth: It's an item namespace disambiguator, which is 99% of the time unnecessary.

Useful for dealing with e.g. `enum Foo {}` and `fn Foo()` in the same module, you can then `enum@Foo` or `fn@Foo` to disambiguate.

Rust's casing norms avoid cross-namespace clashes (except for the macro namespace) so this basically never comes up. I think `super` should work fine here too.

components/experimental/src/displaynames/mod.rs
```diff
@@ -38,14 +38,7 @@
 //! let gb = RegionDisplayName::try_new(locale, region!("GB")).unwrap();
 //! assert_writeable_eq!(us, "United States");
 //! assert_writeable_eq!(gb, "United Kingdom");
-//! ```
-//!
-//! Currently, the data between the two modules is NOT shared, but we hope to
```
Manishearth: did you intend to remove this completely? I think we want this and the next paragraph

sffc: I moved it in a different commit

Manishearth: Ah, the commits had the same name and I missed one by accident

Manishearth (approved):

sffc:
The docs are wrong. The comment about limited support is in `multi` but it should be in `single`.

PR #7827, opened 2026-03-30
Revert to spec test, change test code to not use LSTM/dictionary

Manishearth:
https://github.com/unicode-org/icu4x/issues/7134 diverged the line segmenter test code from the original, because our code has LSTM/dictionary.

I don't think we should do that: we should include the original test, and test against our specced code, without LSTM/dictionary.

I made this as a quick proof of concept for what I think we should do.  I don't have the time to fix the failing tests, @makotokato @aethanyc can you look into that?

## Changelog: TBD

tbd, I don't plan on finishing this PR


Manishearth:
https://github.com/unicode-org/icu4x/issues/7828

PR #7829, opened 2026-03-30
Update example lockfiles

Manishearth:
https://github.com/unicode-org/icu4x/issues/7809


Manishearth:
Yeah, the others don't seem to get updated with the script in the release steps. I need to investigate.

PR #7863, opened 2026-04-09
Fix TrailingCase::Unchanged handling for Dutch

### LINKED ISSUE #7861: Trailing case adjustment doesn't handle Dutch
```rust
    let cm = CaseMapper::new();
    let nl = langid!("nl");
    let mut options = TitlecaseOptions::default();
    options.trailing_case = Some(TrailingCase::Lower);
    assert_eq!(
        cm.titlecase_segment_with_only_case_data_to_string("ij", &nl, default_options),
        "IJ"
    );
```

Instead if `IJ`, we get `Ij`. This is because the way we implement `TrailingCase::Unchanged` is that we switch over to lowercasing once we process the first character: https://github.com/unicode-org/icu4x/blob/78c809c1f0ce6df1653b5d25ca24fe22775e5dc9/components/casemap/src/internals.rs#L72-L75

AIUI, Dutch is the only tailoring that ever cares about more than the first character when titlecasing; so we mostly just need to ensure the trailing case handling can check for this case.

Manishearth:
Fixes #7861 

I couldn't figure out an easy way to keep the dutch handling within `full_helper` (where all other locale-sensitive stuff lives), since this is the one case that actually affects the titlecasing uppercase-to-lowercase state change.

I think this implementation works nicely. The `dutch_i_at_beginning` and `dutch_ij_pair_at_beginning_count` split comes from when I was trying to retain the old code and refactor it to be reusable, but I realized that I could do this without retaining the old code. I can merge these two functions, but I kind of like how it turned out. They're more testable this way.

They're not implemented on CaseMapContext since they run at the beginning of a string.

## Changelog

icu_casemapping: Fix `TrailingCase::Unchanged` handling for Dutch

### DIFF
--- a/components/casemap/src/internals.rs
+++ b/components/casemap/src/internals.rs
@@ -65,11 +65,38 @@ impl<'a, const IS_TITLE_CONTEXT: bool> Writeable for FullCaseWriteable<'a, '_, I
         let src = self.src;
         let mut mapping = self.mapping;
         let mut iter = src.char_indices();
+        // Number of characters to titlecase at the beginning
+        // for the dutch IJ case.
+        //
+        // This cannot be handled in full_helper: this affects the trailing case logic. All other special casing
+        // situations are able to cope by only titlecasing the first character, Dutch is the only place where
+        // the first two base characters get uppercased.
+        let mut dutch_titlecase_count = if IS_TITLE_CONTEXT && self.locale == CaseMapLocale::Dutch {
+            dutch_ij_pair_at_beginning_count(src, self.data)
+        } else {
+            None
+        };
         for (i, c) in &mut iter {
             let context = ContextIterator::new(&src[..i], &src[i..]);
             self.data
                 .full_helper::<IS_TITLE_CONTEXT, W>(c, context, self.locale, mapping, sink)?;
             if IS_TITLE_CONTEXT {
+                // Check if we're uppercasing a dutch IJ
+                if let Some(count) = dutch_titlecase_count {
+                    // If we are, we want to wait `count` characters
+                    // before we switch to lowercasing (or TrailingCase::Unchanged)
+                    if count > 1 {
+                        // We still have code points to process
+                        dutch_titlecase_count = Some(count - 1);
+                        // Continue the loop to skip the mode switching code below
+                        continue;
+                    } else {
+                        // We would have been down to zero. Time to continue the loop as normal.
+                        dutch_titlecase_count = None;
+                    }
+                }
+
+                // If titlecasing, switch the mode to lowercasing/TrailingCase::Unchanged
                 if self.titlecase_tail_casing == TrailingCase::Lower {
                     mapping = MappingKind::Lower;
                 } else {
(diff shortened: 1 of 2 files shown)


components/casemap/src/internals.rs
sffc, line 104: The code path with `dutch_titlecase_count` is a little confusing but _fine_

Optional: I would have expected something more like

```suggestion
} else if i == 0 && is_dutch && is_ij_start(c) {
    // pass
} else if i == 1 && is_dutch && is_ij_end(c) && self.titlecase_tail_casing == TrailingCase::Lower {
    mapping = MappingKind::Lower;
} else {
    break;
}
```

components/casemap/src/internals.rs
```diff
@@ -809,56 +824,172 @@ impl<'a> ContextIterator<'a> {
...
-            //
-            // See https://unicode-org.atlassian.net/browse/ICU-22429
-            match mapping.dot_type(c) {
-                // Not a combining character; ccc = 0
-                DotType::NoDot | DotType::SoftDotted => break,
-                // found combining character, bail
-                _ => return false,
+/// Is there an i at the beginning of the string which may be relevant
+/// for Dutch titlecasing?
+fn dutch_i_at_beginning(s: &'_ str) -> Option<DutchIData<'_>> {
```
sffc, line 841: Optional:

```suggestion
fn dutch_i_at_beginning(s: &'_ str) -> Option<DutchIData<'_>> {
    let mut chars = s.chars();
    let has_accent = match chars.next() {
        Some('i') | Some('I') => match chars.peek() {
            Some(ACUTE) => {
                chars.next();
                true
            }
            _ => false
        },
        Some('í') | Some('Í') => true,
        _ => return None
    }
    Some(DutchIData { /* ... */ }
)
```

components/casemap/src/internals.rs
```diff
@@ -809,56 +824,172 @@ impl<'a> ContextIterator<'a> {
...
+/// This checks for a Dutch-relevant IJ pair at the beginning of a string.
+/// This is an I followed by a J, with any casing, and no accents other than acute.
+/// Acute accents must be on both or neither.
+///
+/// This returns the number of characters (codepoint-wise, not code unit-wise)
+/// in the IJ pair, not including any combining characters on the J.
+///
+/// In dutch titlecasing mode, the first N characters should be uppercased:
+/// `ijabc` should titlecase to `IJabc`.
+fn dutch_ij_pair_at_beginning_count(s: &str, mapping: &CaseMap) -> Option<usize> {
```
sffc, line 887: Suggestion (optional): make a helper function that takes a `&str` and returns a struct similar to DutchIData but works for either I or J, so you can re-use the logic. It can take `const` generics for the four magic chars `i`, `I`, `í`, `Í`

sffc (approved):

components/casemap/src/internals.rs
Manishearth: It's not reusable logic, the j logic is pretty different

components/casemap/src/internals.rs
Manishearth: I considered this, and waffled on choice for a while

Firstly, `.peek()` doesn't work that way, you have to call `.peekable()` to make an intermediate iterator and then peek.

I think

```rust
let peekable = chars.peekable();
match peekable.peek() {
   Some(ACUTE) => peekable.next(), true,
   ...
}
```

is not very different from having a `rest` local variable. The main improvement is that my version is a little bit more verbose with the structs, which I don't like but I think it's _fine_. I like not having to track control flow in my version because each branch returns. Sometimes I wish Rust had field-order construction for local structs....

components/casemap/src/internals.rs
Manishearth: there's a lot more state in between ij_start and ij_end because you have to check for accents and stuff.

PR #7867, opened 2026-04-13
Handle the `sr-Cyrl-ME` -> `sr-ME` collation-only fallback

### LINKED ISSUE #3287: Consume the new component parents from CLDR JSON
CLDR 43 adds the following to parentLocales.json:

```
      "collations": {
        "bs-Cyrl": "bs",
        "sr-Latn": "sr",
        "sr-Cyrl-ME": "sr-ME",
        "zh-Hant": "zh",
        "yue-CN": "zh-Hans",
        "yue-Hans": "zh-Hans",
        "yue-Hans-CN": "zh-Hans",
        "yue": "zh-Hant",
        "yue-Hant": "zh-Hant"
      },
      "segmentations": {
        "zh-Hant": "zh"
      }
```

We should be able to consume that in ICU4X.

robertbastian:
Fixes #3287

This handles the `sr-Cyrl-ME` -> `sr-ME` fallback the same way we handle the other collation-only fallbacks: by explicitly adding the data for `sr-Cyrl-ME` so we don't go through the default fallback mechanism for that locale.

## Changelog

N/A


sffc (approved):
Would be nice for the test to actually run the Collator constructor but it is a bit tricky... good for a follow-up.

hsivonen:
Do I understand correctly that explicitly asking for Cyrillic collation for Montenegrin now results in the Latin collation similarly to how asking for Cyrillic collation for Croatian results in the Latin collation ... but unlike Bosnian and Serbian, which do allow explicit request for the non-default script.

Why? What's the upstream CLDR issue motivating this change?

For reference, here's the Firefox/SpiderMonkey test case that documents the situation before this PR: https://searchfox.org/firefox-main/rev/23974e2d947e31e4ae42ae2758a4416c9a6d8671/js/src/tests/non262/Intl/Collator/bcms.js

Also, AFAICT, there is no technical reason why we couldn't merge the Latin and Cyrillic collation data for Bosnian-Croatian-Montenegrin-Serbian and make Latn vs. Cyrl a matter of script reordering on top.

Are users of Bosnian-Croatian-Montenegrin-Serbian actually better served by having the other script collate according to root as opposed to having the other script also collate according to language-specific rules?

robertbastian:
Apparently these come from upstreaming ICU behaviour to CLDR: https://github.com/unicode-org/cldr/pull/2664, https://github.com/unicode-org/cldr/pull/3504.

I believe it was initially added to ICU in [icu4c/source/data/icu-coll-deprecates.xml](https://github.com/unicode-org/icu/blob/7525392d15814c40ba4e8238e0a53dfbf89bd00f/icu4c/source/data/icu-coll-deprecates.xml), which was commited as "Merge CLDR25 data into trunk". However, I cannot find any reference to `sr_Cyrl_ME` in CLDR 25, so I believe that file was handwritten. `sr_ME` is listed with the other `sr` variants there (and other multi-script languages), and it's the only one where the added script tag *is not the likely one*[^1]. It looks suspiciously like a typo.

[^1]: Note that `Cyrl` *was* the likely script for `sr-ME` until [CLDR-2203](https://unicode-org.atlassian.net/browse/CLDR-2203), but that was way before CLDR 25

The initial aliases from that file have since evolved through
* [tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/IcuConverterConfig.java](https://github.com/unicode-org/icu/blob/b702db31df551c8a607da5e6f4ff4238c61efbe0/tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/IcuConverterConfig.java#L172-L186)
* [tools/cldr/cldr-to-icu/build-icu-data.xml](https://github.com/unicode-org/icu/blob/cf4ce535414f3739375023f77fa42a7a10e32c4f/tools/cldr/cldr-to-icu/build-icu-data.xml#L83-L85)
* [tools/cldr/cldr-to-icu/config.xml](https://github.com/unicode-org/icu/blob/ed0e09fcac6dd2bf3a844152f25214503001fd50/tools/cldr/cldr-to-icu/config.xml#L136-L137)

and the ones that are just likely subtags have disappeared, leaving just `sr-Cyrl-ME` -> `sr-ME`. Along the way, helpful comments like 

> It is not at all clear why this is being done (we expect "sr_Latn_ME" normally).

have been added and removed. It still says 

> TODO: Find out and document this properly

today, but that work is not being tracked anywhere, and apparently wasn't enough to have someone look at this before upstreaming it into CLDR.

PR #7871, opened 2026-04-14
Remove a potential source of UB from potential_utf

weiznich:
This commit removes a transmute from `PotentialUtf8::from_boxed_bytes` as this might rely on undefined behaviour.

It is only sound to transmute between two types like that if their layout is guranteed to be the same. For this particular case it's guranteed that `[u8]` and `PotentialUtf8` have the same layout via the `#[repr(transparent)]` attribute on `PotentialUtf8`. This gurantee doesn't seem to extend on `Box` as:

* `Box` itself is not marked as `#[repr(transparent)]` (or `#[repr(C)]`)
* The [memory layout](https://doc.rust-lang.org/std/boxed/index.html#memory-layout) section of the standard library documentation for `Box` only gurantees a certain layout for sized types:
> So long as T: Sized, a Box<T> is guaranteed to be represented as a single pointer and is also ABI-compatible with C pointers (i.e. the C type T*).
* This specific case involves unsized types, so that gurantee doesn't help us here.

Given that it seems safer to just perform a pointer cast there instead which only involves types with a known compatible layout.

Practically speaking I wouldn't expect that to ever become a problem, but it's better to use a way that's guranteed to work.

## Changelog

potential_utf: Use `Box::from_raw()` instead of `transmute` for converting unsized transparent boxes.

### DIFF
--- a/utils/potential_utf/src/ustr.rs
+++ b/utils/potential_utf/src/ustr.rs
@@ -76,8 +76,18 @@ impl PotentialUtf8 {
     #[inline]
     #[cfg(feature = "alloc")]
     pub fn from_boxed_bytes(other: Box<[u8]>) -> Box<Self> {
-        // Safety: PotentialUtf8 is transparent over [u8]
-        unsafe { core::mem::transmute::<Box<[u8]>, Box<Self>>(other) }
+        // Safety: PotentialUtf8 is transparent over [u8] therefore
+        // the cast between [u8] and Self is sound.
+        // Box::into_raw returns a well aligned pointer that is not
+        // null. The cast then changes the type, but the pointer
+        // is still well aligned, not null and created by the
+        // same global allocator, so it's safe to use Box::from_raw
+        // to create a new Box instance out of it.
+        //
+        // We cannot directly transmute the `Box` here as
+        // there is no gurantee about the layout of `Box` with
+        // unsized types.
+        unsafe { Box::from_raw(Box::into_raw(other) as *mut Self) }
     }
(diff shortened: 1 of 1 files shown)


robertbastian (commented):
thanks for finding

please add a safety comment for the usage of `Box::from_raw`. I think you want to say that `PotentialUtf8` and `[u8]` have the same memory layout, and both `Box` types use the same allocator

robertbastian:
we seem to have three other `Box` transmutes in the repo that we also need to fix.

@sffc this also needs to be fixed in your PR https://github.com/unicode-org/icu4x/pull/7788

Manishearth (commented):
Yeah, this is a good idea, even though I agree that this is practically not a problem. If we asked, Rust would probably document this for us.

Still, for unsafe code, it is good to be technically correct as much as possible, for future-proofness and also to improve the quality of the code in the eyes of reviewers.

r+ with Robert's comments addressed.

weiznich:
Hopefully these changes fix the safety comment.

For reference: I've also asked about this on zulip: [#t-opsem > Is transmuting between &#96;Box<A&gt;&#96; and &#96;Box<B&gt;&#96; UB? @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/136281-t-opsem/topic/Is.20transmuting.20between.20.60Box.3CA.3E.60.20and.20.60Box.3CB.3E.60.20UB.3F/near/585350243).

Manishearth (approved):

PR #7872, opened 2026-04-14
Evolve `CollationSpecialPrimariesV1` data struct

robertbastian:
Now that we can change the baked representation, this evolves the data struct while keeping the serialized representation stable, removing the hacks we had added for this.

## Changelog

`icu_collator`: Changes to the `CollationSpecialPrimariesV1` data struct

### DIFF
--- a/components/collator/src/comparison.rs
+++ b/components/collator/src/comparison.rs
@@ -38,8 +38,8 @@ use crate::provider::CollationMetadataV1;
 use crate::provider::CollationReordering;
 use crate::provider::CollationReorderingV1;
 use crate::provider::CollationRootV1;
+use crate::provider::CollationSpecialPrimaries;
 use crate::provider::CollationSpecialPrimariesV1;
-use crate::provider::CollationSpecialPrimariesValidated;
 use crate::provider::CollationTailoringV1;
 use core::cmp::Ordering;
 use core::convert::Infallible;
@@ -49,7 +49,6 @@ use icu_normalizer::provider::NormalizerNfdDataV1;
 use icu_normalizer::provider::NormalizerNfdTablesV1;
 use icu_normalizer::DecomposingNormalizerBorrowed;
 use icu_normalizer::Decomposition;
-use icu_provider::marker::ErasedMarker;
 use icu_provider::prelude::*;
 use smallvec::SmallVec;
 use utf16_iter::Utf16CharsEx;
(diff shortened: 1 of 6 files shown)


sffc (commented):
I like how the code is structured better with this PR.

components/collator/src/comparison.rs
```diff
@@ -874,27 +772,25 @@ impl CollatorBorrowed<'static> {
...
 }
 
-macro_rules! collation_elements {
-    ($self:expr, $chars:expr, $tailoring:expr, $numeric_primary:expr) => {{
-        let jamo = <&[<u32 as AsULE>::ULE; JAMO_COUNT]>::try_from($self.jamo.ce32s.as_ule_slice());
-
-        let jamo = jamo.unwrap();
-
+impl<'a> CollatorBorrowed<'a> {
+    fn collation_elements<C: Iterator<Item = char>>(
```
hsivonen: Did you re-run benchmarks to see that this doesn't regress perfomance? From the code change, it looks like it shouldn't, but there have been surprises here previously.

hsivonen: Specifically, we need to make sure that rustc keeps initializing `CollationElements`directly into its eventual memory location instead of initializing it first and then moving it.

robertbastian: Well, the `::new` call is a way more complex function than this, so if anything doesn't get inlined it's that (and that's preexisting). But I can run benchmarks to confirm, is it just `cargo bench`?

hsivonen: `cargo criterion`, rather, with measures to make the clock stable (CPU governor set to `performance` and Intel turbo or AMD boost turned off). I even do `setarch -R cargo criterion` just in case, but I'm not sure if that bit matters.

(And, yes, it _looks_ OK, but there's previously been a regression here that looked OK from source.)

robertbastian: ok I've inlined this instead. this is now code-equivalent to the previous macro version, with the upside that we have types and stuff

robertbastian:
@hsivonen bump

sffc (commented):
Still fine and good I think; I'm waiting on @hsivonen for the full approval unless you want me to do that

robertbastian:
I've pinged him multiple times on this already, so I'm not sure we'll get a review from him

sffc (approved):

PR #7874, opened 2026-04-15
Backport UAX #14-aware line-break oracle to 1.4 (issue #7218)

kenton-r:
Backport of #7232 to `release/1.4`.

Fixes issue #7218 — Asian text (Khmer/Thai/Lao) produced double line
breaks around spaces. The dict/LSTM complex segmenter emits *word*
boundaries, and `line.rs` was treating those as line-break opportunities.
When a word boundary landed next to a space, LB7 (× SP) plus the
following break-after-space rule produced two adjacent breaks, bumping
the space to the start of the next line.

## Approach

Split `complex/mod.rs` into two named entry points:

- `complex_language_segment_str` / `_utf16` — **unchanged.** Word-boundary
  oracle used by `word.rs`; never was a line-break function.
- `complex_language_line_breaks_str` / `_utf16` — **new.** Runs the same
  dict/LSTM, then filters out breaks whose following char has a UAX #14
  class that forbids a break before it (LB7 × SP, LB21 × BA/HY/NS,
  LB12a × WJ/GL, LB13 × CL/CP/EX/IS/SY, LB22 × IN, LB9 × CM/ZWJ). For
  the terminal break at `s.len()`, `line.rs` passes the next external
  char so the filter works across SA chunk boundaries.

Filtering lives in the oracle (not `line.rs`) so it runs once per
dict/LSTM invocation — `line.rs` is on the width/justification hot path
and cannot afford per-candidate rescanning.

## 1.4-specific adaptations vs. #7232

- `complex/mod.rs` entry points are free functions taking
  `&ComplexPayloads` (1.4 shape), not methods on a borrowed payload type.
- 1.4's `LineBreakOptions` fields are already resolved
  (`LineBreakStrictness` / `LineBreakWordOption`), so no `.resolve()`
  step is needed in the predicate closure.
- UTF-16 `CharType` is `u32` in 1.4; predicate wraps the BMP u16 code
  unit captured from the iterator.

## Tests

- New (`line.rs`): `khmer/thai/lao_line_break_with_spaces`,
  `sa_sp_boundary_filtered_at_every_run`, `non_ascii_whitespace_around_sa`
  (U+3000, U+2009, U+0009, U+00A0), `lstm_vs_dictionary_no_double_break`,
  `word_segmenter_thai_unchanged`.
- Updated (`tests/complex_word.rs`): `word_line_th_wikipedia_auto` —
  offsets 27, 47, 82, 113, 220, 281 (each immediately before an ASCII
  space) are now correctly suppressed under LB7. `breakpoints_word_utf8`
  / `_utf16` unchanged, proving the word-segmenter path is untouched.

All `cargo test -p icu_segmenter` and `--features serde` suites pass.

### DIFF
--- a/components/segmenter/src/complex/mod.rs
+++ b/components/segmenter/src/complex/mod.rs
@@ -339,7 +339,13 @@ fn try_load<M: KeyedDataMarker, P: DataProvider<M> + ?Sized>(
     }
 }
 
-/// Return UTF-16 segment offset array using dictionary or lstm segmenter.
+/// Return UTF-16 **word-boundary** offsets for complex-script input.
+///
+/// This is the word-boundary entry point used by `word.rs`. It runs the
+/// dictionary or LSTM segmenter over each SA (Southeast Asian) run and
+/// returns every word boundary the engine identifies. It performs **no**
+/// UAX #14 line-break filtering — callers that need line-break
+/// opportunities must use [`complex_language_line_breaks_utf16`] instead.
 pub(crate) fn complex_language_segment_utf16(
     payloads: &ComplexPayloads,
     input: &[u16],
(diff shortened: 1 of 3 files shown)


sffc:
We need to approve and land #7232 first.

Any reason you are putting this on the 1.4 branch? If you need this in the 1.x stream, 1.5 would be better.

robertbastian:
Closing this as a 1.5 backport should be sufficient

PR #7875, opened 2026-04-15
Update box transmutes

robertbastian:
See #7871 

## Changelog

N/A

### DIFF
--- a/components/pattern/src/frontend/mod.rs
+++ b/components/pattern/src/frontend/mod.rs
@@ -145,7 +145,7 @@ impl<B: PatternBackend> Pattern<B> {
     #[cfg(feature = "alloc")]
     pub(crate) fn from_boxed_store_unchecked(store: Box<B::Store>) -> Box<Self> {
         // Safety: Box::into_raw fulfils Box::from_raw's requirements, as Pattern<B> is
-        // repr(transparent) over B::Store
+        // repr(transparent) over B::Store, and does not have further validity constraints
         unsafe { Box::from_raw(Box::into_raw(store) as *mut Self) }
     }
 
--- a/components/pattern/src/frontend/mod.rs
+++ b/components/pattern/src/frontend/mod.rs
@@ -143,9 +143,10 @@ where
 
 impl<B: PatternBackend> Pattern<B> {
     #[cfg(feature = "alloc")]
-    pub(crate) const fn from_boxed_store_unchecked(store: Box<B::Store>) -> Box<Self> {
-        // Safety: Pattern is repr(transparent) over B::Store
-        unsafe { core::mem::transmute::<Box<B::Store>, Box<Self>>(store) }
+    pub(crate) fn from_boxed_store_unchecked(store: Box<B::Store>) -> Box<Self> {
+        // Safety: Box::into_raw fulfils Box::from_raw's requirements, as Pattern<B> is
+        // repr(transparent) over B::Store
+        unsafe { Box::from_raw(Box::into_raw(store) as *mut Self) }
     }
 
     #[doc(hidden)] // databake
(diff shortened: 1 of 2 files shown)


sffc (commented):
Let's use this opportunity to make these safety comments bullet-proof, and then I will use them in #7788. To do that we should wait for @Manishearth's review.

components/pattern/src/frontend/mod.rs
```diff
@@ -143,9 +143,10 @@ where
 
 impl<B: PatternBackend> Pattern<B> {
     #[cfg(feature = "alloc")]
-    pub(crate) const fn from_boxed_store_unchecked(store: Box<B::Store>) -> Box<Self> {
-        // Safety: Pattern is repr(transparent) over B::Store
-        unsafe { core::mem::transmute::<Box<B::Store>, Box<Self>>(store) }
+    pub(crate) fn from_boxed_store_unchecked(store: Box<B::Store>) -> Box<Self> {
+        // Safety: Box::into_raw fulfils Box::from_raw's requirements, as Pattern<B> is
+        // repr(transparent) over B::Store
```
Manishearth:
```suggestion
        // repr(transparent) over B::Store, and does not have further validity constraints
```

provider/core/src/request.rs
```diff
@@ -323,8 +323,10 @@ impl DataMarkerAttributes {
             return Err(AttributeParseError);
         };
 
-        // SAFETY: `Self` has the same layout as `str`
-        Ok(unsafe { core::mem::transmute::<Box<str>, Box<Self>>(s.into_boxed_str()) })
+        let boxed = s.into_boxed_str();
+        // Safety: Box::into_raw fulfils Box::from_raw's requirements, as DataMarkerAttributes is
+        // repr(transparent) over str
```
Manishearth:
```suggestion
        // repr(transparent) over str, and its (non-safety) validity constraints were validated above
```

provider/core/src/request.rs
```diff
@@ -355,8 +357,10 @@ impl DataMarkerAttributes {
 impl ToOwned for DataMarkerAttributes {
     type Owned = Box<Self>;
     fn to_owned(&self) -> Self::Owned {
-        // SAFETY: `Self` has the same layout as `str`
-        unsafe { core::mem::transmute::<Box<str>, Box<Self>>(self.as_str().to_boxed()) }
+        let boxed = self.as_str().to_boxed();
+        // Safety: Box::into_raw fulfils Box::from_raw's requirements, as DataMarkerAttributes is
+        // repr(transparent) over str
```
Manishearth:
```suggestion
        // repr(transparent) over str, and `str` has strictly fewer validity constraints than DataMarkerAttributes
```

Manishearth (approved):
Hmm. So my main issue with this was that `Box::from_raw` doesn't _have_ enumerated requirements, which kind of sucks. Normally a good safety comment actually talks about how it upholds each enumerated requirement. But the docs there talk a lot about layout (and layout is only a part of validity).

Given that I think just mentioning validity constraints in context is good.

sffc:
Yeah I hit the lack of enumerated requirements too when trying to use `Box::from_raw`. I think I listed out things like: the layout (size and alignment) is correct, the pointer is non-null, ... some of these things are covered by `Box::into_raw` but not all. For example, we need `transparent` to prove the layout. Is there anything else that `transparent` gives us, or is saying "layout" sufficient?

Manishearth:
Transparent only gives you the layout. The thing you also need to be clear on is the validity, e.g. `[u8]` and `str` have the same layout but not the same overall validity. Unfortunately there's no generic way to talk about validity in Rust, so it becomes type specific.

PR #7876, opened 2026-04-15
Revert the sr-Cyrl-ME -> sr-ME collation-only fallback

robertbastian:
Reverts #7867 

## Changelog

N/A


sffc (approved):
ok. I hope to someday see tests for what the expected behavior is, because at this point I don't know what the intention is.

robertbastian:
The intention here is to follow CLDR

PR #7878, opened 2026-04-16
Update to 2024 edition where possible

robertbastian:
A lot of changes are due to rustfmt, but they are nice changes.

## Changelog

N/A


sffc:
Rust edition 2024 was released in 1.85, is that compatible with MSRV? (take into account utils crates that have lower pinned MSRV)

components/casemap/src/casemapper.rs
```diff
@@ -108,7 +108,11 @@ impl<'a> CaseMapperBorrowed<'a> {
     ///
     /// See [`Self::lowercase_to_string()`] for the equivalent convenience function that returns a string,
     /// as well as for an example.
-    pub fn lowercase(self, src: &'a str, langid: &LanguageIdentifier) -> impl Writeable + 'a {
+    pub fn lowercase(
+        self,
+        src: &'a str,
+        langid: &LanguageIdentifier,
+    ) -> impl Writeable + 'a + use<'a> {
```
sffc, line 115: what does `use<'a>` mean

components/pattern/src/frontend/zerovec.rs
```diff
@@ -46,7 +46,9 @@ where
     }
 
     unsafe fn from_bytes_unchecked(bytes: &[u8]) -> &Self {
-        let store = S::from_bytes_unchecked(bytes);
+        // Safety: validate_bytes succeeded, so S::parse_bytes succeeded,
+        // so S::validate_bytes succeeded, so S::from_bytes_unchecked is safe.
+        let store = unsafe { S::from_bytes_unchecked(bytes) };
```
sffc, line 51: is the doubly-nested `unsafe` a new requirement?

robertbastian: in edition 2024, unsafe functions (which is a signal to *callers*) don't let you implicitly use unsafe code inside anymore. this is a major improvement

robertbastian: i.e. this unsafe call now requires a safety comment where it didn't before

components/casemap/src/casemapper.rs
robertbastian: https://blog.rust-lang.org/2024/10/17/Rust-1.82.0/#precise-capturing-use-syntax

robertbastian:
> Rust edition 2024 was released in 1.85, is that compatible with MSRV? (take into account utils crates that have lower pinned MSRV)

This is why the PR is titled "Update to 2024 edition *where possible*". Only crates with MSRV >= 1.85 are updated

components/casemap/src/casemapper.rs
robertbastian: (this was generated by `cargo fix --edition`)

sffc: ick. It sounds like they changed the default lifetime capturing behavior in order to reduce callsite boilerplate, but for us it makes it worse because we take some non-capturing borrowed arguments.

I think the `+ 'a` is no longer necessary?

components/decimal/src/compact_formatter.rs
```diff
@@ -347,7 +347,7 @@ impl CompactDecimalFormatter {
     ///     "999K"
     /// );
     /// ```
-    pub fn format(&self, value: &Decimal) -> impl Writeable + Display + '_ {
+    pub fn format(&self, value: &Decimal) -> impl Writeable + Display + '_ + use<'_> {
```
sffc, line 350: There's only one lifetime here, so I think edition 2024's default behavior is what we want

components/experimental/src/unicodeset_parse/parse.rs
```diff
@@ -127,7 +127,7 @@ impl ParseError {
     ///     r"[\N← error: unimplemented"
     /// );
     /// ```
-    pub fn fmt_with_source(&self, source: &str) -> impl Display {
+    pub fn fmt_with_source(&self, source: &str) -> impl Display + use<> {
```
sffc, line 130: Thought: This is a very unintuitive syntax to say that the `impl Display` does not capture the input lifetime.

hsivonen:
Can we defer changes of this kind to collections, collator, and normalizer until the in-flight perf improvements have landed?

robertbastian:
I would like to land the first commit at least, which shouldn't impact those open PRs.

The second commit is just an invocation of `cargo fmt` with the 2024 edition, we can do that at any time.

components/decimal/src/compact_formatter.rs
robertbastian: I've reverted most of the changes done by `cargo fix --edition`, as they mostly affected macro expression matching and if-let drop order, where if it compiles it compiles. However, this is a public API bound, and what `cargo fix --edition` generates keeps the bounds exactly equivalent. I'd rather not touch this.

components/casemap/src/casemapper.rs
robertbastian: I think removing the `'a` might be breaking

components/decimal/src/compact_formatter.rs
sffc: fine as a follow-up (please file one)

provider/source/src/locale/aliases.rs
```diff
@@ -37,24 +38,24 @@ impl crate::IterableDataProviderCached<LocaleAliasesV1> for SourceDataProvider {
...
-    (
-        -union_size,
-        langid.language,
-        langid.script,
-        langid.region,
-        langid.variants.clone(),
-    )
+fn appendix_c_cmp(a: &LanguageIdentifier, b: &LanguageIdentifier) -> Ordering {
+    let a_union_size = a.variants.len() as u8
+        + (!a.language.is_unknown()) as u8
```
sffc, line 43: Issue: Please do not cast booleans to integers. This code was more readable before and now is less readable

provider/source/src/time_zones/mod.rs
```diff
@@ -634,7 +634,7 @@ impl IterableDataProviderCached<TimezonePeriodsV1> for SourceDataProvider {
 }
 
 impl crate::source::Tzdb {
-    fn transitions<'a>(&'a self, iana: &str) -> Option<impl Iterator<Item = Transition> + 'a> {
+    fn transitions(&self, iana: &str) -> Option<impl Iterator<Item = Transition> + use<>> {
```
sffc, line 637: Issue: this looks like a change; we should borrow from `'a` but now you are not borrowing from anything

rustfmt.toml
```diff
@@ -0,0 +1,5 @@
+# This file is part of ICU4X. For terms of use, please see the file
+# called LICENSE at the top level of the ICU4X source tree
+# (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
+style_edition = "2021"
```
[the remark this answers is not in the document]
robertbastian: stupid

provider/source/src/time_zones/mod.rs
robertbastian: and what's the issue? this compiles

provider/source/src/locale/aliases.rs
robertbastian: this code is more readable now because there is no mutable state anymore

provider/source/src/time_zones/mod.rs
sffc: It's private so I guess the change is fine

provider/source/src/locale/aliases.rs
sffc: Boolean casts to integers are never readable.

A mutable integer local to a function is readable; I find both forms acceptable

```rust
// Local mutable
let mut a = 10;
if condition {
    a += 5;
}

// Inline if
let a = 10 +
    if condition { 1 } else { 0 };
```

I won't block on this if someone else approves but I'm not comfortable approving myself because Boolean casts are not good style and this PR introduces one unnecessarily in pre existing code

sffc (commented):
Unresolved comment

components/casemap/src/casemapper.rs
Manishearth: If it compiles, it shouldn't be breaking.

`Foo + 'a` is a narrower type than `Foo`, and this is not a trait method, so widening the type is backwards compatible.

Manishearth: > ick. It sounds like they changed the default lifetime capturing behavior in order to reduce callsite boilerplate, but for us it makes it worse because we take some non-capturing borrowed arguments.

I don't think this reduces callsite boilerplate, it reduces def-site boilerplate since by default you won't have to write `+ 'a` unless you want to opt out of borrowing specific lifetimes.

components/decimal/src/compact_formatter.rs
Manishearth: > There's only one lifetime here, so I think edition 2024's default behavior is what we want

No? There's `&self` and `&Decimal`

components/pattern/src/frontend/zerovec.rs
Manishearth, line 51: huh, for some reason I thought we already had this lint enabled. It's a useful one.

provider/source/src/locale/aliases.rs
Manishearth: I find the boolean cast more readable than the inline expanded if.

utils/crlify/Cargo.toml
```diff
@@ -10,7 +10,7 @@ description = "A std::io::Write wrapper that replaces \n with \r\n on Windows."
 documentation = "https://docs.rs/litemap"
 
 authors.workspace = true
-edition.workspace = true
+edition = "2021"
```
Manishearth, line 13: nit: have a comment on each opt out explaining why

rustfmt.toml
Manishearth, line 5: I assume this is for a future PR

Manishearth (approved):

rustfmt.toml
robertbastian: yes, it makes merges easier if the conflict is just with formatting

PR #7880, opened 2026-04-16
Update toolchain to 1.95

robertbastian:
Picking up two releases' worth of clippy lints.

## Changelog

N/A

### DIFF
--- a/components/collections/src/iterator_utils.rs
+++ b/components/collections/src/iterator_utils.rs
@@ -31,11 +31,8 @@ where
         // range from last time, or the next range
         let mut ret = if let Some(peek) = self.peek.take() {
             peek
-        } else if let Some(next) = self.iter.next() {
-            next
         } else {
-            // No ranges, exit early
-            return None;
+            self.iter.next()?
         };
 
         // Keep pulling ranges
--- a/components/collections/src/iterator_utils.rs
+++ b/components/collections/src/iterator_utils.rs
@@ -32,6 +32,7 @@ where
         let mut ret = if let Some(peek) = self.peek.take() {
             peek
         } else {
+            // Exit early when there are no ranges
             self.iter.next()?
         };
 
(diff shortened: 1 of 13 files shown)


components/collections/src/iterator_utils.rs
```diff
@@ -31,11 +31,8 @@ where
         // range from last time, or the next range
         let mut ret = if let Some(peek) = self.peek.take() {
             peek
-        } else if let Some(next) = self.iter.next() {
-            next
         } else {
-            // No ranges, exit early
-            return None;
+            self.iter.next()?
```
Manishearth, line 36: nit: retain comment. "Exit early when there are no ranges"

provider/source/src/segmenter/mod.rs
```diff
@@ -379,35 +379,33 @@ fn generate_rule_break_data(
                     {
                         for cp in 0..(CODEPOINT_TABLE_LEN as u32) {
                             match lb.get32(cp) {
-                                LineBreak::OpenPunctuation => {
-                                    if (p.name == "OP_OP30"
+                                LineBreak::OpenPunctuation
```
Manishearth, line 382: observation: these changes make you have to go all the way down to  the _ branch to check if it's no-op, which is kind of annoying. But I think it's still an overall cleaner change

provider/source/src/segmenter/mod.rs
```diff
@@ -676,28 +670,26 @@ fn generate_rule_break_data_override(
...
             match &*segmenter.segmenter_type {
-                "word" => {
+                "word"
                     // UAX29 defines the colon as MidLetter, but ICU4C's
                     // English data doesn't.
                     // See https://unicode-org.atlassian.net/browse/ICU-22112
                     //
                     // TODO: We have to consider this definition from CLDR instead.
-                    if p.name == "MidLetter" {
+                    if p.name == "MidLetter" => {
```
Manishearth: nit: move the comment (here and below) to above the pattern

Manishearth (approved):

Manishearth (approved):

PR #7887, opened 2026-04-20
Various safety fixes and defense-in-depth

Manishearth:
This fixes a bunch of issues found by @sayrer by running Claude on ICU4X to look for vulnerabilities (reported in https://github.com/unicode-org/icu4x/security/advisories/GHSA-c8m9-943c-5xmw).

Most of the fixes here are defense in depth: the code was already safe, but we can add assertions/etc to make it safer. In other cases the issues found were potential correctness issues, not safety issues.

The only real safety issue found was the multiplication overflows in VarZeroVec validation code: in theory, a 64 bit platform might validate a VZV that is invalid on a 32 bit platform. These have been fixed. They aren't severe enough to warrant a patch release / advisory: the only time this can cause a security issue is if you are trusting VZV data blobs constructed on a 64 bit platform on a 32 bit platform, which only happens with baked data, and 4GB-size VZVs aren't really a thing there.

Thanks, @sayrer !

The first commit was agent assisted but in the end mostly rewritten by me.

Changes not included in changelog:

icu_list: Prevent correctness errors from out-of-range indices in list patterns.

## Changelog

zerovec: Fix minor soundness issue around unchecked multiplication, add defense in depth against other overflow situations.

resb: Add defense-in-depth around checked multiplication.

icu_provider_fs: Add defense-in-depth against path traversal.

### DIFF
--- a/components/list/src/patterns.rs
+++ b/components/list/src/patterns.rs
@@ -95,7 +95,8 @@ impl<'data> ListJoinerPattern<'data> {
                     && (allow_prefix || index_0 == 0)
                     && (allow_suffix || index_1 == pattern.len() - 3) =>
             {
-                if (index_0 > 0 && !cfg!(test)) || index_1 - 3 >= 256 {
+                let index1_minus_three = index_1.wrapping_sub(3);
+                if (index_0 > 0 && !cfg!(test)) || index_0 >= 256 || index1_minus_three >= 256 {
                     return Err(DataError::custom(
                         "Found valid pattern that cannot be stored in ListFormatterPatterns",
                     )
@@ -112,7 +113,7 @@ impl<'data> ListJoinerPattern<'data> {
                         .into_boxed_str(),
                     ),
                     index_0: index_0 as u8,
-                    index_1: (index_1 - 3) as u8,
+                    index_1: index1_minus_three as u8,
                 })
             }
             _ => Err(DataError::custom("Invalid list pattern").with_debug_context(pattern)),
--- a/provider/fs/src/fs_data_provider.rs
+++ b/provider/fs/src/fs_data_provider.rs
@@ -93,7 +93,12 @@ impl FsDataProvider {
                             })?,
                     );
                 } else {
-                    path.push(req.id.marker_attributes.as_str());
+                    let attr_path = req.id.marker_attributes.as_str();
+                    // Data marker attributes are already validated to be
+                    // alphanumeric + underscore/dash. This is defense-in-depth
+                    // against potential path traversal attacks.
+                    assert!(!attr_path.contains(['/', '\\']));
+                    path.push(attr_path);
                 }
             }
             let mut string_path = path.into_os_string();
--- a/utils/zerovec/src/hashmap/algorithms.rs
+++ b/utils/zerovec/src/hashmap/algorithms.rs
@@ -62,11 +62,12 @@ pub fn compute_index(f: (u32, u32), d: (u32, u32), m: u32) -> Option<usize> {
 /// # Arguments
 ///
 /// * `key_hashes` - [`ExactSizeIterator`] over the hashed key values
-#[expect(clippy::indexing_slicing, clippy::unwrap_used)]
+#[expect(clippy::indexing_slicing, clippy::unwrap_used, clippy::expect_used)]
 pub fn compute_displacements(
     key_hashes: impl ExactSizeIterator<Item = u64>,
 ) -> (Vec<(u32, u32)>, Vec<usize>) {
     let len = key_hashes.len();
+    let len_u32 = u32::try_from(len).expect("Hashmap too large for u32");
 
     // A vector to track the size of buckets for sorting.
     let mut bucket_sizes = vec![0; len];
@@ -122,8 +123,8 @@ pub fn compute_displacements(
         // start, end - 1 are always within bounds of `bucket_sizes`
         let buckets = &bucket_flatten[start..end];
 
-        'd0: for d0 in 0..len as u32 {
-            'd1: for d1 in 0..len as u32 {
+        'd0: for d0 in 0..len_u32 {
+            'd1: for d1 in 0..len_u32 {
                 if (d0, d1) == (0, 0) {
                     continue;
                 }
@@ -131,7 +132,7 @@ pub fn compute_displacements(
                 generation += 1;
 
                 for ((_, f0, f1), _) in buckets {
-                    let displacement_idx = compute_index((*f0, *f1), (d0, d1), len as u32).unwrap();
+                    let displacement_idx = compute_index((*f0, *f1), (d0, d1), len_u32).unwrap();
 
                     // displacement_idx is always within bounds
                     if occupied[displacement_idx] || assignments[displacement_idx] == generation {
--- a/utils/zerovec/src/ule/option.rs
+++ b/utils/zerovec/src/ule/option.rs
@@ -197,6 +197,7 @@ unsafe impl<U: VarULE + ?Sized> VarULE for OptionVarULE<U> {
 
     #[inline]
     unsafe fn from_bytes_unchecked(bytes: &[u8]) -> &Self {
+        debug_assert!(!bytes.is_empty());
         let entire_struct_as_slice: *const [u8] =
             ::core::ptr::slice_from_raw_parts(bytes.as_ptr(), bytes.len() - 1);
         &*(entire_struct_as_slice as *const Self)
@@ -260,5 +261,19 @@ impl<U: VarULE + ?Sized + PartialOrd> PartialOrd for OptionVarULE<U> {
 impl<U: VarULE + ?Sized + Ord> Ord for OptionVarULE<U> {
     fn cmp(&self, other: &Self) -> Ordering {
         self.as_ref().cmp(&other.as_ref())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    #[cfg(debug_assertions)]
+    #[should_panic]
+    fn test_option_varule_underflow() {
+        unsafe {
+            let _ = OptionVarULE::<str>::from_bytes_unchecked(&[]);
+        }
     }
 }
--- a/utils/zerovec/src/varzerovec/components.rs
+++ b/utils/zerovec/src/varzerovec/components.rs
@@ -279,11 +279,19 @@ impl<'a, T: VarULE + ?Sized, F: VarZeroVecFormat> VarZeroVecComponents<'a, T, F>
         };
         // The indices array is one element shorter since the first index is always 0,
         // so we use len_minus_one
+        //
+        // We do the math in u32 space so that we can be sure that VZVs constructed
+        // on 64 bit computers can still be used on 32 bit ones.
+        let indices_len = u32::try_from(F::Index::SIZE)
+            .ok()
+            .and_then(|x| x.checked_mul(len_minus_one))
+            .and_then(|x| usize::try_from(x).ok())
+            .ok_or(VarZeroVecFormatError::Metadata)?;
         let indices_bytes = slice
-            .get(..F::Index::SIZE * (len_minus_one as usize))
+            .get(..indices_len)
             .ok_or(VarZeroVecFormatError::Metadata)?;
         let things = slice
-            .get(F::Index::SIZE * (len_minus_one as usize)..)
+            .get(indices_len..)
             .ok_or(VarZeroVecFormatError::Metadata)?;
 
         let borrowed = VarZeroVecComponents {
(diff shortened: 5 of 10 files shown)


provider/fs/src/fs_data_provider.rs
```diff
@@ -93,7 +93,12 @@ impl FsDataProvider {
                             })?,
                     );
                 } else {
-                    path.push(req.id.marker_attributes.as_str());
+                    let attr_path = req.id.marker_attributes.as_str();
+                    // Data marker attributes are already validated to be
+                    // alphanumeric + underscore/dash. This is defense-in-depth
+                    // against potential path traversal attacks.
```
sffc, line 99: Where do we do the claimed validation? I think we allow attributes to contain `/`, such as `FR/long` to be the long French display name, which should by design create a directory in FsDataProvider.

Consider forbidding other characters like `.` and `:` to prevent the path traversal

utils/zerovec/src/ule/option.rs
```diff
@@ -197,6 +197,7 @@ unsafe impl<U: VarULE + ?Sized> VarULE for OptionVarULE<U> {
 
     #[inline]
     unsafe fn from_bytes_unchecked(bytes: &[u8]) -> &Self {
+        debug_assert!(!bytes.is_empty());
```
sffc, line 200: Thought: this one is very 🤷 because the trait's safety invariants have us handled via the check on line 179

utils/zerovec/src/varzerovec/components.rs
```diff
@@ -351,8 +360,10 @@ impl<'a, T: VarULE + ?Sized, F: VarZeroVecFormat> VarZeroVecComponents<'a, T, F>
         };
         // The indices array is one element shorter since the first index is always 0,
         // so we use len_minus_one
-        let indices_bytes = slice.get_unchecked(..F::Index::SIZE * (len_minus_one as usize));
-        let things = slice.get_unchecked(F::Index::SIZE * (len_minus_one as usize)..);
+        let indices_len = F::Index::SIZE.wrapping_mul(len_minus_one as usize);
+        debug_assert!(F::Index::SIZE.checked_mul(len_minus_one as usize).is_some());
```
sffc, line 364: Observation: this is equivalent to the `*` operator you had before I think

utils/zerovec/src/varzerovec/owned.rs
```diff
@@ -134,14 +134,23 @@ impl<T: VarULE + ?Sized, F: VarZeroVecFormat> VarZeroVecOwned<T, F> {
...
         } else if let Some(idx) = self.index_data(idx) {
             idx.iule_to_usize()
         } else {
             0
         };
-        debug_assert!(out + F::Len::SIZE + (len - 1) * F::Index::SIZE <= self.entire_slice.len());
+        let indices_size = F::Index::SIZE
+            .checked_mul(len - 1)
+            .expect(F::Index::TOO_LARGE_ERROR);
+        debug_assert!(out + F::Len::SIZE + indices_size <= self.entire_slice.len());
```
sffc, line 153: Note, here and elsewhere: I'm not thinking too hard in this review about why you are doing a checked_mul on some and a debug assertion on others. I assume that when you assert, it is a safety thing that is technically reachable, and debug assert is an invariant that is not reachable. I can go back through this again if you want me to.

provider/fs/src/fs_data_provider.rs
Manishearth: `FR/long` is not an attribute. We have the key `long` and the attribute `fr`.

My initial code here forbade `.` and string-initial `/` and `\`, but then I checked the attributes docs and realized they forbid slashes anyway.


https://unicode-org.github.io/icu4x/rustdoc/icu_provider/struct.DataMarkerAttributes.html#method.try_from_str

utils/zerovec/src/ule/option.rs
Manishearth: I'm basically always in favor of adding cheap debug assertions that double-validate things unless it's going to be a big mess.

I guess I wouldn't want this everywhere.

I'm going to lean slightly towards keeping this since the non-emptiness is relevant to type-specific invariants.

utils/zerovec/src/varzerovec/components.rs
Manishearth: Correct, but now it's intentional :smile:

utils/zerovec/src/varzerovec/owned.rs
Manishearth: Yes. There are three ways I've done things in this PR:

 - Assertions for safety things that are reachable in an API that is allowed to panic (like VZV construction here). I consider FS data provider to be "allowed to panic" in case of a path traversal; that's pretty serious. Happy to downgrade to debug assert if you don't want panics there, though. I view FS code as very fragile and panicky anyway.
 - A debug assertion for safety things that are not reachable but I'm being cautious.
 - An error for actual reachable safety issues

 We only have one real reachable safety issue here: the one with 32 bit overflow in VZV components.

utils/zerovec/src/varzerovec/components.rs
sffc: I don't think I like doing this in general in the icu4x codebase, but since this is in one our two most sensitive files (the other being yoke)... ok to be extra explicit I guess

provider/fs/src/fs_data_provider.rs
sffc: oof, this breaks part of my DisplayNames design, I guess a part I haven't implemented yet but assumed would be implementable. We will want to go back and revisit this.

I have a slight preference for you to pull this assumption out to a separate PR because we should discuss this on the DataMarkerAttributes level, but if you want to land this to assert on the status quo, I guess that's okay.

sffc (approved):
This is landable but I requested an LLM review to double check that the changes you're making to math expressions are no-ops, which is easier for an LLM to verify than a human.

utils/zerovec/src/ule/option.rs
sayrer: @sffc is correct that "this one is very 🤷". However, it requires constant vigilance. Maybe this:

```
#[cfg(test)]
mod misuse_tests {
    use super::OptionULE;
    use crate::ule::ULE;

    /// Direct misuse of the unsafe contract: `from_bytes_unchecked` documents
    /// that callers must pass a slice that has already passed `validate_bytes`,
    /// which rejects empty input. Calling it with `&[]` violates that contract.
    ///
    /// In debug builds the `debug_assert!(!bytes.is_empty())` catches this at
    /// the construction site. In release builds this same call is undefined
    /// behavior (the returned `&OptionULE<u8>` would read an out-of-bounds
    /// discriminant byte on its next use), which is why the test is gated to
    /// debug assertions.
    #[test]
    #[should_panic]
    #[cfg(debug_assertions)]
    fn from_bytes_unchecked_empty_slice_trips_debug_assert() {
        // SAFETY: intentionally unsafe — this test exists to verify that the
        // defense-in-depth assertion fires when the caller violates the
        // documented precondition.
        let _ = unsafe { <OptionULE<u8> as ULE>::from_bytes_unchecked(&[]) };
    }

    /// Same shape, but routed through a "helpful wrapper" — the realistic
    /// refactor hazard. A future author writes a convenience function that
    /// forwards a slice to `from_bytes_unchecked` and forgets the length
    /// check that lived in the original call site. The assertion catches it.
    #[test]
    #[should_panic]
    #[cfg(debug_assertions)]
    fn wrapper_forwarding_empty_slice_trips_debug_assert() {
        unsafe fn construct_without_validating(bytes: &[u8]) -> &OptionULE<u8> {
            // SAFETY: (intentionally unsound — simulates a refactor that drops
            // the validate_bytes call upstream.)
            unsafe { <OptionULE<u8> as ULE>::from_bytes_unchecked(bytes) }
        }

        // SAFETY: see above — deliberately violating the contract.
        let _ = unsafe { construct_without_validating(&[]) };
    }

    /// Sanity check: the *contract-respecting* path does not trip the
    /// assertion. A single-byte slice `[0]` encodes `None` for `OptionULE<u8>`
    /// (discriminant 0, inner byte ignored — here we supply two bytes so the
    /// layout is well-formed regardless of U::SIZE assumptions).
    #[test]
    fn from_bytes_unchecked_valid_slice_is_fine() {
        let bytes = [0u8, 0u8];
        // SAFETY: validate_bytes would accept this slice (length == 1 + u8::SIZE,
        // discriminant is 0 meaning None, inner byte is a valid u8 ULE).
        let opt = unsafe { <OptionULE<u8> as ULE>::from_bytes_unchecked(&bytes) };
        // Just prove we can touch it without UB in debug.
        let _ = opt;
    }
}
```

utils/zerovec/src/varzerovec/components.rs
Manishearth: Right. Also I think this is not the first time someone will run AI tooling to search for vulns, so some defense in depth is good in that it prevents duplicate reports.

utils/zerovec/src/varzerovec/components.rs
```diff
@@ -729,7 +740,14 @@ where
     // idx_offset = offset from the start of the buffer for the next index
     let mut idx_offset: usize = 0;
     // first_dat_offset = offset from the start of the buffer of the first data block
-    let first_dat_offset: usize = idx_offset + (elements.len() - 1) * F::Index::SIZE;
+    #[expect(
+        clippy::expect_used,
+        reason = "Function contract allows panicky behavior"
+    )]
+    let indices_size = F::Index::SIZE
+        .checked_mul(elements.len() - 1)
```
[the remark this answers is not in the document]
Manishearth: We have checked that elements is nonempty already

provider/fs/src/fs_data_provider.rs
```diff
@@ -93,7 +93,12 @@ impl FsDataProvider {
...
+                    assert!(!attr_path.contains(['/', '\\']));
```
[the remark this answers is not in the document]
Manishearth: This should not be hit, and DOS is better than path traversal.

provider/fs/src/fs_data_provider.rs
sffc: https://github.com/unicode-org/icu4x/pull/7890

components/list/src/patterns.rs
```diff
@@ -95,7 +95,8 @@ impl<'data> ListJoinerPattern<'data> {
                     && (allow_prefix || index_0 == 0)
                     && (allow_suffix || index_1 == pattern.len() - 3) =>
             {
-                if (index_0 > 0 && !cfg!(test)) || index_1 - 3 >= 256 {
+                let index1_minus_three = index_1.wrapping_sub(3);
```
robertbastian, line 98: this is a needless change, if the string contains both `{0}` and `{1}`, and `index_0` < `index_1`, then `index_1` is at least 3. I can see that when reading the code, we should not complicate our code just because AIs are too stupid to understand such simple dependencies

utils/zerovec/src/ule/option.rs
robertbastian: `debug_assert!(!bytes.is_empty());` could be added to most `from_bytes_unchecked` implementations then

utils/zerovec/src/hashmap/algorithms.rs
```diff
@@ -62,11 +62,12 @@ pub fn compute_index(f: (u32, u32), d: (u32, u32), m: u32) -> Option<usize> {
 /// # Arguments
 ///
 /// * `key_hashes` - [`ExactSizeIterator`] over the hashed key values
-#[expect(clippy::indexing_slicing, clippy::unwrap_used)]
+#[expect(clippy::indexing_slicing, clippy::unwrap_used, clippy::expect_used)]
```
robertbastian, line 65: justification?

utils/zerovec/src/ule/option.rs
Manishearth: I know, this is a case where there's some relevant arithmetic just after it.

From an unsafe review perspective this is nicer, though I probably should have added an assertion message.

utils/zerovec/src/hashmap/algorithms.rs
Manishearth: Same justification as the preexisting ones? This is a construction API allowed to panic. I could add justification there just wasn't any already so I didn't.

components/list/src/patterns.rs
Manishearth: No, I made this change because it's pretty non-obvious to humans too. It took me a while to realize this was checked. An alternate way to do this is to have comments about the invariants. I'm happy to switch to that if you prefer.

This isn't unsafe code, but I was reviewing all of this with my unsafe review hat, where you _really_ want things to be dead obvious.

PR #7890, opened 2026-04-21
Allow slashes in DataMarkerAttributes

sffc:
This is for multi-layered attributes, like "FR/long", which we may want for DisplayNames.

I made this with a heavily-prompted agent.

## Changelog

icu_provider: allow slashes in DataMarkerAttributes


provider/fs/src/fs_data_provider.rs
```diff
@@ -94,10 +94,14 @@ impl FsDataProvider {
...
-                    // alphanumeric + underscore/dash. This is defense-in-depth
-                    // against potential path traversal attacks.
-                    assert!(!attr_path.contains(['/', '\\']));
+                    // Data marker attributes are validated to be alphanumeric
+                    // + underscore/dash/slash. This is defense-in-depth
+                    // against potential path traversal attacks: we ensure that
+                    // the path does not start with a slash, and does not contain
+                    // '..' or backslashes.
+                    assert!(!attr_path.starts_with('/'));
+                    assert!(!attr_path.contains(".."));
```
Manishearth: nit: just testing for `'.'` will be faster due to the memchr optimization

Manishearth: And you can then merge the two `contains` calls into `.contains(['\\', '.'])`

provider/core/src/request.rs
```diff
@@ -268,19 +268,29 @@ impl DataMarkerAttributes {
     /// Safety-usable invariant: validated bytes are ASCII only
     const fn validate(s: &[u8]) -> Result<(), AttributeParseError> {
         let mut i = 0;
+        let mut prev_was_slash = true;
```
Manishearth, line 275: nit: comment that setting this to true disallows leading slashes

Manishearth (commented):
just a code review. not yet sure if we should support this but ... probably? I thought we didn't end up picking the nested attribute designs for DisplayNames, and I have a slight preference to only make this change when we actually need it. But maybe we need it now, and I'm not strongly opposed to doing this proactively.

sffc:
> just a code review. not yet sure if we should support this but ... probably? I thought we didn't end up picking the nested attribute designs for DisplayNames

We said that length and region could be nested in the attributes, but I had already implemented length on the key level and didn't go back to revisit it yet.

provider/core/src/request.rs
```diff
@@ -234,7 +234,7 @@ impl Default for DataIdentifierCow<'_> {
 #[derive(PartialEq, Eq, Ord, PartialOrd, Hash)]
 #[repr(transparent)]
 pub struct DataMarkerAttributes {
-    // Validated to be non-empty ASCII alphanumeric + hyphen + underscore
+    // Validated to be non-empty ASCII alphanumeric + hyphen + underscore + forward slash. Disallows leading and double slashes.
```
robertbastian: it should also disallow trailing slashes

Manishearth: I agree, otherwise this leads to multiple attributes with the same fs path.

Manishearth (approved):
If we have a current use case, sure. LGTM with the trailing slashes fixed

Manishearth (approved):

PR #7895, opened 2026-04-21
Use an enum for currency rounding modes

robertbastian:
Follow-up #7278.

Previously, the values 0-14,50 were allowed for rounding modes. CLDR currently only uses 1,5,50, and with v49 will use 1,5,20,50, so this code will break.

## Changelog

N/A


younies (approved):

sffc (commented):
Note: The values in ECMAScript are 1, 2, 5, and 25, since 50 is 5 with extra rules around fraction digits

robertbastian:
The DTD allows 1-5 for `rounding`, and 1,5,20,50 for `cashRounding`. I don't think 3 or 4 make sense 🤷

robertbastian:
Should we store 50 as 5 and 20 as 2?

sffc:
I think for internal purposes we should definitely have a code path for only `2` and `5`, not `20` and `50`. When reading from CLDR we obviously need to support `20` and `50`. I'll leave it as an open question whether to support `20` and `50` in the user-facing API.

PR #7896, opened 2026-04-21
Add EastAsianWidth support to unicode sets

robertbastian:
## Changelog

icu_experimental/unicodeset: Add support for `:EastAsianWidth=:` to unicode set parser


Manishearth (approved):
Seems like we just support every property we have here, yes?

robertbastian:
I don't know exactly what we need to support, but I need EAW for my segmenter work.

sffc (commented):
lgtm, we should figure out property sets when stabilizing the unicodeset component

Manishearth:
Would also be cool if we could codegen or macro generate all this code

PR #7898, opened 2026-04-21
Segmenter proof of concept

robertbastian:
This uses the data from https://github.com/unicode-org/unicodetools/pull/1321.

## Changelog

N/A


components/segmenter/src/provider/mod.rs
```diff
@@ -15,6 +15,11 @@
 // Provider structs must be stable
 #![allow(clippy::exhaustive_structs, clippy::exhaustive_enums)]
 
+#[cfg(feature = "unstable")]
+#[allow(missing_docs)]
+mod neo;
```
Manishearth, line 20: observation: My first reaction to this was "wait, why are we touching datetime stuff"

(this naming is fine and is I think what I've suggested we do here, I'm just amused)

components/segmenter/src/provider/neo.rs
```diff
@@ -0,0 +1,88 @@
...
+    pub const TRASH_STATE: State = State::MAX;
+    /// The end-of-text class. This is a dummy class that only appears at the end of the input,
+    /// and allows the state machine to have special transitions on end-of-text.
+    pub const EOT_CLASS: Class = 0;
+}
+
+icu_provider::data_marker!(
+    /// `SegmenterBreakLineV2`
+    SegmenterBreakLineV2,
+    "segmenter/break/line/v2",
```
Manishearth, line 85: observation: BreakLineV2 sounds a bit strange but it's fine

components/segmenter/src/provider/neo.rs
```diff
@@ -0,0 +1,88 @@
...
+
+#[derive(Debug, yoke::Yokeable, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize))]
+#[cfg_attr(feature = "datagen", derive(serde::Serialize, databake::Bake))]
+#[cfg_attr(feature = "datagen", databake(path = icu_segmenter::provider))]
+pub struct SegmenterStateMachine<'data> {
+    // A map from Unicode scalar values to their segmentation classes
+    #[cfg_attr(feature = "serde", serde(borrow))]
+    pub classes: CodePointTrie<'data, Class>,
+    // A dense map of states
```
Manishearth, line 54: for future: eventually we should have detailed docs on what all of these entries really are (map of states from where to where?)

Manishearth (approved):
Approving so progress can be made. Cursory code review lgtm, I think we can do a proper code review once this is done being experimented with.

components/segmenter/src/provider/neo.rs
robertbastian: The current one is called `SegmenterBreakLineV1`. I don't want to invent new names

sffc (commented):
Praise: Easy to read, few lines of code. Some thoughts:

- Final API shape TBD. We probably want LineSegmenterBorrowed.
- Word and Sentence segmenters have locale-specific tailorings. Do we have a plan to handle those? (OK if the answer is, we'll figure it out later)

sffc (approved):

PR #7899, opened 2026-04-22
Test various date construction patterns

Manishearth:
Fixes https://github.com/unicode-org/icu4x/issues/7853

The code here is entirely agent generated, but done step by step with heavy prompting on strategies to use. I like the final result quite a bit.

## Changelog (N/A)

### DIFF
--- a/components/calendar/src/tests/constructor_roundtrip.rs
+++ b/components/calendar/src/tests/constructor_roundtrip.rs
@@ -1,0 +1,482 @@
+// This file is part of ICU4X. For terms of use, please see the file
+// called LICENSE at the top level of the ICU4X source tree
+// (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
+use crate::error::DateFromFieldsError;
+use crate::options::{DateFromFieldsOptions, Overflow};
+use crate::types::{DateFields, Month};
+use crate::Calendar;
+use crate::Date;
+use calendrical_calculations::rata_die::RataDie;
+
+// We could use the `bitflags` crate here, but we are using raw constants
+// to avoid pulling in an extra dependency.
+type Flags = u32;
+const EXTENDED_YEAR: u32 = 1 << 0;
+const ERA: u32 = 1 << 1;
+const ERA_YEAR: u32 = 1 << 2;
+const MONTH: u32 = 1 << 3;
+const MONTH_CODE: u32 = 1 << 4;
+const ORDINAL_MONTH: u32 = 1 << 5;
+const DAY: u32 = 1 << 6;
+const MAKE_INCONSISTENT_YEAR: u32 = 1 << 7;
+const MAKE_INCONSISTENT_MONTH: u32 = 1 << 8;
+
+fn try_from_fields_helper<C: Calendar + Copy>(
+    flags: Flags,
+    rd: RataDie,
+    options: DateFromFieldsOptions,
+    cal: C,
+) -> Result<Date<C>, DateFromFieldsError> {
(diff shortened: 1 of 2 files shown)


components/calendar/src/tests/constructor_roundtrip.rs
```diff
@@ -0,0 +1,482 @@
...
+    cal: C,
+) -> Result<Date<C>, DateFromFieldsError> {
+    let date = Date::from_rata_die(rd, cal);
+    let mut fields = DateFields::default();
+
+    if flags & EXTENDED_YEAR != 0 {
+        fields.extended_year = Some(date.year().extended_year());
+    }
+
+    let era_bytes = date.year().era().map(|ey| ey.era);
```
sffc, line 38: Nit move variables used by only one branch inside it

components/calendar/src/tests/constructor_roundtrip.rs
```diff
@@ -0,0 +1,482 @@
...
+                        rd
+                    );
+                }
+            }
+        }
+    }
+);
+
+#[test]
+fn test_calendar_specific_constructors() {
```
sffc, line 278: Suggestion: move these into docs tests on each calendar

sffc (approved):
Good application of an agent

components/calendar/src/tests/constructor_roundtrip.rs
Manishearth: We can't share the rd generation function though.

The calendar specific ctors already have doctests.

components/calendar/src/tests/constructor_roundtrip.rs
Manishearth: No, these are borrowed. They need to be around for the lifetime.

components/calendar/src/tests/constructor_roundtrip.rs
sffc: My comment was more about that I don't like these tests here possibly getting out of sync with the list of calendar constructors. Putting the tests in the individual calendar files makes them closer to the source, and if someone adds a new calendar, they copy all the tests from the previous calendar file. If you make it a unit test, you can import the rd generation.

Manishearth: Reasonable. But I don't want to make our documentation poorer with things that are really tests, not examples. The whole rd set and everything else here is pretty inside baseball.

A unit test would be fine. We could also maybe refactor this to use `test_all_cals` and then have some type-level conditional work. Ugly, but would have all the properties requested here.

sffc: Unit tests in each file or making it work with `test_all_cals` are both more desirable to me than what is currently in the PR

sffc: > Unit tests in each file or making it work with `test_all_cals` are both more desirable to me than what is currently in the PR

do you plan to follow up on this?

Manishearth: I wasn't planning on it unless you really wanted me to. I don't really see that having these per file will help them not get out of sync; it depends on how people are duplicating their files when adding calendars, and it's such a rare operation that I don't care to spend time optimizing for this. If this is something you want to do, I'd approve it, and I could probably just ask an agent for this, but I wasn't _a priori_ planning to work on this.

sffc: https://github.com/unicode-org/icu4x/issues/7921

Issue #3935, opened 2023-08-24
All combinatorics on provider::Baked


sffc:
Users should be able to chain provider::Baked providers to make a multi-crate AnyProvider.

@Manishearth suggested `aked.add(otherprovider::Baked)` making a fork provider.

robertbastian:
`crate::provider::Baked` is an implementation detail in the unstable provider module. Users can use compiled data by depending on the data crates and creating their own baked providers:

```rust

struct MyBakedProvider;

icu_datetime_data::impl_datetime_chinese_datelengths_v1!(MyBakedProvider);
icu_datetime_data::impl_datetime_chinese_datesymbols_v1!(MyBakedProvider);
icu_list_data::impl_and_list_v1!(MyBakedProvider);
```

sffc:
Discuss with:

- @Manishearth 
- @robertbastian 
- @sffc

sffc:
Discussion: revisit this in the context of #3947, when we remove AnyProvider. Leaving some breadcrumbs for my mental model of this and #1893:

- We should have a macro `impl_forking_keyed_provider!([key1, key2, key3], Provider1, Provider2)` where the list of keys go to `Provider1` and any key not in the list goes to `Provider2`
- The `provider::Baked` impls can have a function that internally contains this match statement for all of the keys it supports.

sffc:
Another use case here is if you want your language packs to contain only non-singleton keys and get the singleton keys from compiled data with the non-singleton keys from dynamically loaded data.

A few approaches here:

1. Use `with_any_provider` constructors and a registry macro that forks between the singleton compiled data and the blob dynamic data
2. Use `_unstable` constructors (but then we don't handle older blob data)
3. Use a new type of constructor called `_unstable_with_backwards_compatibility`
4. Use `_with_buffer_provider` and export the singleton keys as blobs from the data crates
5. In 2.0 make `_with_provider` constructors that take an enum between BufferProvider and AnyProvider (or equivalent) and fork between the singleton static data and the blob data

sffc:
Probably doesn't block 2.0 so removing it from that milestone, but it is still important.

PR #7901, opened 2026-04-22
Add `impl_blanket_data_provider!` to baked exporter

sffc:
https://github.com/unicode-org/icu4x/issues/3935

This is an example of an approach that would allow for more ergonomic forking between DataProvider impls. It allows two providers with different sets of markers to be forked with the DataProvider trait. DCE should "just work" since the if statement can be resolved at compile time.

## Changelog

TODO

### DIFF
--- a/provider/baked/src/export.rs
+++ b/provider/baked/src/export.rs
@@ -835,6 +835,11 @@ impl DataExporter for BakedExporter {
 
         let required_crates_list = required_crates.iter().map(|c| format!(" * `{c}`"));
 
+        let marker_bakes = data
+            .keys()
+            .map(|marker_info| bake_marker(*marker_info))
+            .collect::<Vec<_>>();
+
         // mod.rs is the interface for built-in data. It exposes one macro per marker.
         self.write_to_file(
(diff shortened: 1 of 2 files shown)


provider/baked/src/export.rs
```diff
@@ -883,9 +888,34 @@ impl DataExporter for BakedExporter {
...
+                    ($wrapper:ty, $inner_provider:ty) => {
+                        impl<M> icu_provider::DataProvider<M> for $wrapper
+                        where
+                            M: icu_provider::DataMarker,
+                        {
+                            fn load(&self, req: icu_provider::DataRequest) -> Result<icu_provider::DataResponse<M>, icu_provider::DataError> {
+                                use core::any::TypeId;
+                                #(
+                                    if TypeId::of::<M>() == TypeId::of::<#marker_bakes>() {
+                                        let res = icu_provider::DataProvider::<#marker_bakes>::load(&self.0, req)?;
```
sffc, line 903: Note: This code won't compile unless the inner provider implements DataProvider on the specified marker, i.e., that all markers from the bake crate are implemented.

Issue #1225, opened 2021-10-28
Support alt variants of skeletons


sffc:
CLDR JSON has skeletons such as

```
                "Md": "MM-dd",
                "Md-alt-variant": "d/M",
```

We should figure out how to handle this data.

sffc:
Discussion 2022-03-31:

- @sffc - This can be handled by the work on display context.
- @Manishearth - There are also alt forms of eras.

sffc:
CC @younies

sffc:
My current thinking is that each of the alt variants are fairly subject-specific. Clients have a specific need to enable a certain set of alt variants, like alt=ascii.

Therefore, I think it makes sense to give clients an API for each individual alt variant.

The setter could take a bitmask or something, but I think it's simplest to explain and document to just have

```rust
source_provider.with_alt_ascii_time_formats()
```

and

```rust
icu4x-datagen --alt-ascii-time-formats
```

We have precedent for optionless flags setting booleans: `--no-internal-fallback` sets a boolean in https://unicode-org.github.io/icu4x/rustdoc/icu_provider_export/baked_exporter/struct.Options.html, for example.

sffc:
I'm now proposing: `--alt-variant datetime-ascii` which takes multiple arguments. We give each alt variant class a name.

Manishearth:
I like that choice. Easily extended.

PR #7902, opened 2026-04-22
Implement alt=ascii as a datagen flag

sffc:
https://github.com/unicode-org/icu4x/issues/1225

There appears to be demand from Chrome for having alt=ascii be a runtime toggle so that they can run experiments and such. To implement this, I added some datagen flags to enable alt=ascii and also export it as a small overlay provider.

To make this work better in a fork provider, I sketched out https://github.com/unicode-org/icu4x/pull/7901

I wrote this PR with a hand-held agent.

CC @hsivonen @jedel1043 @nekevss 

## Changelog

icu_provider_source: Support setting alt=ascii variant for datetime formatting
- New method: `SourceDataProvider::with_alt_ascii_datetime_formats`

icu_provider_export: Support filtering data based on whether it is an alt variant
- New enum: `AltVariantStrategy`
- New method: `DataExporter::with_alt_variant_strategy`

icu_provider: Add metadata to support the filtering on alt variants
- New enum: `AltVariantStatus`
- New field: `DataResponseMetadata::alt_variant_status`


provider/source/src/lib.rs
```diff
@@ -486,6 +488,15 @@ impl SourceDataProvider {
         }
     }
 
+    /// Instruct the provider to use patterns with `alt="ascii"` whenever available
+    /// for datetime formatting.
+    pub fn with_alt_ascii_datetime_formats(self) -> Self {
```
robertbastian, line 493: hard block. I do not want to add a method per data marker and alt value. this should be a flexible API, like the attributes filter

```rust
    pub fn with_alt_value(self, marker: DataMarkerInfo, alt_value: &str) -> Self
```

provider/core/src/response.rs
```diff
@@ -22,6 +22,19 @@ use alloc::rc::Rc as SelectedRc;
 #[cfg(feature = "sync")]
 use alloc::sync::Arc as SelectedRc;
 
+/// Status indicating whether the data returned is an alternative variant.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum AltVariantStatus {
```
robertbastian, line 28: I'm not convinced by the use case of leaking the alt concept out of the source data provider. It should be a config that drives the `SourceDataProvider`, but the rest of the provider infrastructure does not need to know whether a value comes from alt or not

provider/source/src/lib.rs
robertbastian: or

```rust
     pub fn with_alt_value(self, domain: &str, alt_value: &str) -> Self

**Review by robertbastian — CHANGES_REQUESTED** [review]

File: provider/core/src/response.rs (hunk 2 above)
Comment (sffc) [reply provider/core/src/response.rs:28]: This is how SourceDataProvider sends info to DataExporter in order to implement `--only-alt-variants`

Do you have a proposal for an alternative way to implement `--only-alt-variants`?

Comment (robertbastian) [reply provider/core/src/response.rs:28]: The flag should not be on the export driver, and the export driver should not use the data provider trait system to request alt variants. The flag should be on SourceDataProvider and it should just return alt variants where configured.

File: provider/source/src/lib.rs (hunk 1 above)
Comment (sffc) [reply provider/source/src/lib.rs:493]: `DataMarkerInfo` is not the right input because some of the alts impact multiple data markers and this should be transparent. I think alt=ascii impacts like 2-3 markers.

I considered something more like `domain`, either as a string or an enum, but it seems a bit confusing that the alt variant domain is not the same as the marker attribute domain. But maybe that's okay.

On the CLI it could look something like

```
--alt-variant datetime=ascii
```

@robertbastian Is this shape of solution acceptable? (string-based alt variant domains that are not the same as data marker attribute domains)

File: provider/core/src/response.rs (hunk 2 above)
Comment (sffc) [reply provider/core/src/response.rs:28]: > Do you have a proposal for an alternative way to implement `--only-alt-variants`?

Comment (robertbastian) [reply provider/core/src/response.rs:28]: The SourceDataProvider can change the behaviour of IterableDataProvider to achieve that.

File: provider/source/src/lib.rs (hunk 1 above)
Comment (robertbastian) [reply provider/source/src/lib.rs:493]: yes

Comment (robertbastian) [reply provider/source/src/lib.rs:493]: That seems to be the command line version of the Rust API I proposed

File: provider/core/src/response.rs (hunk 2 above)
Comment (sffc) [reply provider/core/src/response.rs:28]: One issue is, we need to plumb _all_ of the IterableDataProvider impls in SourceDataProvider to block data that isn't an alt variant, because the user doesn't know ahead of time which exact DataMarkers are impacted by the alt variant. The intended use case is they export `--all-markers --only-alt-variants`.

I expected that you would share the sentiment that SourceDataProvider would be focused on how to formulate the data being generated, and DataExporter would be focused on what data to export.

File: provider/source/src/lib.rs [hunk 3]
```diff
@@ -188,6 +189,7 @@ impl SourceDataProvider {
             )
             .unwrap(),
             collation_root_han: Default::default(),
+            use_alt_ascii_datetime_formats: false,
```
sffc, line 192: With a string-key-value-based formulation for the alt variants (https://github.com/unicode-org/icu4x/pull/7902/changes#r3122473919), we could consider modeling this as a field in the DataRequest, and then all of the API plumbing goes on the DataExporter.

The DataRequestMetadata could gain a field of type `&'a [(&'a str, &'a str)]` which is a sorted tuple vec of the alt variants. This would be a nice pairing with the DataResponseMetadata field.

@robertbastian

provider/source/src/lib.rs
Manishearth: FWIW I have not looked at the PR yet but "I want to ensure the flags are sufficiently general" was also my reaction to seeing the PR title.

Manishearth: Also the new flag should be mentioned in the changelog, or TBD while you iterate on the design.

provider/core/src/response.rs
sffc: (I'm waiting on an additional reply from @robertbastian on this thread before I update the PR)

robertbastian: Yes I share that sentiment. I don't share the sentiment that alt is something we need to expose in ICU4X. It's a CLDR concept that should be resolved at datagen time, it should not make its way into `DataProvider`.

robertbastian (commented):
.

provider/core/src/response.rs
sffc: I could expose it on `DataPayload<ExportMarker>` and add the trait scaffolding to support that?

The design where `DataProvider` is used for both exporting and runtime is a solid choice given the tradeoffs. It also means that there are not as many places to add exporter-specific functionality.

If we add it to DataResponseMetadata, a provider implementation could choose to retain it and return it. The field is optional and wouldn't be returned otherwise.

Do you have concrete suggestions to offer?

robertbastian: > One issue is, we need to plumb _all_ of the IterableDataProvider impls in SourceDataProvider to block data that isn't an alt variant, because the user doesn't know ahead of time which exact DataMarkers are impacted by the alt variant. The intended use case is they export `--all-markers --only-alt-variants`.

But this is a different issue than using alt data. The operation you're getting at is some kind of overlay provider: given providers A and B, only export structs from B that don't match A.

For that, we should add an option to `ExportDriver` to specify a base provider, and then it doesn't export data structs that match the base provider (the base provider doesn't have to be exportable, it can be baked/postcard). Basically, the inverse of the forking provider.

sffc: Hmm, a `DifferenceProvider`, ok. It's a bit more complex than I was thinking but it seems like a clean way to keep things out of `DataResponseMetadata` if that's a goal.

sffc: Is there a way to make two `SourceDataProvider`s share the same backing data (CldrCache, etc)?

robertbastian: just clone, all the caches are arcs

sffc:
I'm redoing this in two parts, one to add alt=ascii, the other to support the alt-only overlay provider thing.

Part 1: https://github.com/unicode-org/icu4x/pull/8025

PR #7903, opened 2026-04-22
Update docs on DataProvider support for forking providers

sffc:
The docs are not accurate. We implement `DataProvider<M>` since 0bec39c3b00.

## Changelog

N/A

### DIFF
--- a/provider/adapters/src/fork/by_error.rs
+++ b/provider/adapters/src/fork/by_error.rs
@@ -13,8 +13,12 @@ use icu_provider::prelude::*;
 /// This is an abstract forking provider that must be provided with a type implementing the
 /// [`ForkByErrorPredicate`] trait.
 ///
-/// [`ForkByErrorProvider`] does not support forking between [`DataProvider`]s. However, it
-/// supports forking between [`BufferProvider`], and [`DynamicDataProvider`].
+/// This provider supports any data provider trait as long as it is implemented by both
+/// child providers.
+///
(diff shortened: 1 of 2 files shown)


Manishearth (approved):

provider/adapters/src/fork/by_error.rs
```diff
@@ -13,8 +13,12 @@ use icu_provider::prelude::*;
...
 /// [`ForkByErrorPredicate`] trait.
 ///
-/// [`ForkByErrorProvider`] does not support forking between [`DataProvider`]s. However, it
-/// supports forking between [`BufferProvider`], and [`DynamicDataProvider`].
+/// This provider supports any data provider trait as long as it is implemented by both
+/// child providers.
+///
+/// Some traits like [`BufferProvider`] work on all markers. However, [`DataProvider<M>`]
+/// is specific to a single marker type `M`. For this reason, [`DataProvider<M>`] is only
+/// implemented if both child providers implement it for the same `M`.
```
robertbastian, line 21: I feel like that's kind of obvious, `DataProvider<A>` and `DataProvider<B>` are different traits

sffc: We've been told by even regular contributors that data provider traits are hard to grok so reinforcing this type of information is not harmful.

Manishearth: Also specifically around fork providers where you might try to fork between two providers where only one implements the right traits (if you wanted to fork by key)

PR #7904, opened 2026-04-22
Generate properties from the UCD

robertbastian:
#4602 

## Changelog

`icu_provider_source`:
* Compute properties directly from the `unicode` data source, instead of from `icuexport`


Manishearth:
Is there a reason we have a data diff at all here?

robertbastian:
The diff for the BidiClass and Script CPTs is weird, but it's the same diff you get if you rebuild CPTs with icuexportdata, see #7299. The removed/added values are the trie's default values, and iterating through the trie yields the same results, so I don't think this diff is relevant.

The diff for CCC parsing is because the UCD also defines numeric aliases, which I've added to the parser for good measure.

components/properties/src/trievalue.rs
```diff
@@ -28,6 +28,10 @@ impl TrieValue for CanonicalCombiningClass {
     fn to_u32(self) -> u32 {
         u32::from(self.0)
     }
+
+    fn default() -> Self {
+        Self::NotReordered
```
Manishearth: nit: would appreciate having links to _something_ that backs up where we got some of these defaults. Or is it just always the 0 value?

provider/source/src/properties/bin_cp_set.rs
```diff
@@ -5,45 +5,221 @@
...
+    pub(super) fn get_binary_prop(
+        &self,
+        name: &str,
+        short_name: &str,
+    ) -> Result<CodePointInversionList<'static>, DataError> {
         let mut builder = CodePointInversionListBuilder::new();
-        for (start, end) in &self.ranges {
-            builder.add_range32(start..=end);
+
+        // UTS #18 Annex C Compatibility Properties
```
Manishearth, line 53: thought: huh; I didn't realize we generate compound compat props.

That's data duplication. Though it's also not that important i guess.

provider/source/src/properties/bin_cp_set.rs
```diff
@@ -5,45 +5,221 @@
...
+        if name == "alnum" {
+            // \p{alpha}\p{digit} = \p{Alphabetic}\p{gc=Decimal_Number}
```
Manishearth, line 55: thought: would be very cool to use our unicodeset code to make this big function smaller

not actually a suggestion

Manishearth (commented):
General shape is promising.

provider/source/src/properties/bin_cp_set.rs
robertbastian: I'd love to get rid of them. We can still support them in the unicode set parser.

sffc:
Nit: "Changelog: N/A" is not correct. It is relevant to clients that their data comes from a `--ucd-root` instead of a `--icuexportdata-root`. Please fix

provider/source/data/debug/property/name/parse/canonical/combining/class/v1.json
```diff
@@ -1,5 +1,63 @@
 {
```
sffc, line 1: Observation: this probably should have been named `PropertyNameParseCanonicalcombiningclass`

sffc (commented):
I'm inclined to defer to @Manishearth's review, and ideally @markusicu could take a look, too. I'd like though if we can get this to a zero diff. I don't understand why there are diffs in bidi, script, and canonical combining class. Please try to make this a clean diff if possible or else write up in more detail why the diff is (1) benign and (2) unavoidable.

sffc:
> The diff for the BidiClass and Script CPTs is weird, but it's the same diff you get if you rebuild CPTs with icuexportdata, see #7299. The removed/added values are the trie's default values, and iterating through the trie yields the same results, so I don't think this diff is relevant.

Please try to land that PR before this one in order to avoid the data diff.

> The diff for CCC parsing is because the UCD also defines numeric aliases, which I've added to the parser for good measure.

I prefer also splitting this out into a separate PR.

robertbastian:
> Please try to land that PR before this one in order to avoid the data diff.

That PR is not landeable because of datagen performance. I'm not going to invest time improving that just to immediately replace it again.

provider/source/data/debug/property/name/parse/canonical/combining/class/v1.json
robertbastian: no. we decided that marker names don't have to conform to rigid structures, and the structure of the `PropertyNameParse` subtree is that there isn't one

sffc:
I don't understand: you are needing to build all the same CPTs here as you do in that other PR. So why does only one of them have performance problems?

robertbastian:
Because in this PR I've implemented caching inside the UCD data source.

Manishearth:
Would it be possible to make a small PR (maybe one that adds defaults to TrieValue) that has temporary hacks leading to the same diffs? It would be nic eto have a programmatic understanding of what's changed. But I don't think we should spend too much time on this.

robertbastian:
I've updated the Script default in the other PR. The diffs are now byte-identical.

Manishearth:
Nice macro.

provider/icu4x-datagen/src/main.rs
```diff
@@ -444,10 +444,6 @@ fn run(cli: Cli) -> eyre::Result<()> {
             );
         } else if SourceDataProvider::is_missing_segmenter_lstm_error(e) {
             eyre::bail!("Segmentation LSTM data is required for this invocation, set --segmenter-lstm-root or --segmenter-lstm-tag");
-        } else if SourceDataProvider::is_missing_unihan_error(e) {
```
Manishearth: thought: when doing a UCD folder people may wish to supply UCD-without-unihan since unihan is big. 

I don't actually think this matters for this change (since this change only affects unihan users, not non-users), but it's worth flagging.

provider/source/src/properties/bidi.rs
```diff
@@ -17,28 +17,79 @@ impl DataProvider<PropertyEnumBidiMirroringGlyphV1> for SourceDataProvider {
...
-        let bpt_lookup = bpt.values_to_names_long();
+                let typ = match parts.next().unwrap().trim() {
+                    "o" => BidiPairedBracketType::Open,
+                    "c" => BidiPairedBracketType::Close,
+                    "n" => BidiPairedBracketType::None,
+                    _ => unreachable!(),
+                };
+                Some((cp, typ))
+            })
+            .collect::<HashMap<_, _>>();
```
Manishearth, line 92: nit: BTreeMap, preferably, for determinism.

provider/source/src/properties/bin_cp_set.rs
```diff
@@ -5,45 +5,221 @@
...
+                | "Emoji_Modifier_Base"
+                | "Emoji_Modifier"
+                | "Emoji_Presentation"
+                | "Emoji"
+                | "Extended_Pictographic" => "ucd/emoji/emoji-data.txt",
+                "Bidi_Mirrored" => "ucd/extracted/DerivedBinaryProperties.txt",
+                _ => "ucd/PropList.txt",
+            };
+
+            for line in self.unicode()?.read_to_string(file)?.lines() {
```
Manishearth, line 202: issue: this type of UCD parsing code shows up in multiple spots: can we make a shared helper?
(or maybe two, one for binary and one for enum props)

Having a separate ucd.rs with all that code would be nice.

They are somewhat different formats (especially things like emoji-sequences) so maybe it's not easy to share the helper, but maybe they can all be in one file at least.

provider/source/src/properties/bin_cp_set.rs
```diff
@@ -5,45 +5,221 @@
...
+                let line = line.split('#').next().unwrap().trim();
+                if line.is_empty() {
```
Manishearth, line 204: issue: These files have `@missing` rules that are also supposed to be parsed. As far as I can tell, each file has only a single such rule per property, but I'm not sure if that's guaranteed, and I'm not sure if that rule always matches the enum default we have here. Probably.

We should at the very least think about this and add a comment here mentioning it. I've already asked Robin for clarity.

provider/source/src/properties/bidi.rs
robertbastian: this is not iterated over

provider/icu4x-datagen/src/main.rs
robertbastian: this is a change from the diffbase

provider/source/src/properties/bin_cp_set.rs
robertbastian: I though about this, but
* we can isolate the comment stripping, but apparently some comments are semantically relevant, so that might be confusing
* there are enough edge cases, like CCC, that this would get quite complicated
* I don't want to create intermediate data structures if possible

provider/source/src/properties/bin_cp_set.rs
robertbastian: afaict only `DerivedNormalizationProps.txt` has `@missing`. I'll add it

eggrobin: > As far as I can tell, each file has only a single such rule per property, 

Nope, and that’s explicitly documented: https://www.unicode.org/reports/tr44/#Missing_Conventions
> Starting with Version 15.0, some data files in the UCD may contain multiple @&zwnj;missing lines defined for the _same_ property.

And DerivedEastAsianWidth.txt, which is in this PR, is one such file.

robertbastian: That also says

> An `@missing` line is never provided for a binary property, because the default value for binary properties is always "N" and need not be defined redundantly for each binary property.

which doesn't match the data.

robertbastian: ah, the property I looked at wasn't binary, it was enumerated over Yes and No. of course

eggrobin: (It does, it’s just that two enumerated properties—NFD_QC and NFKD_QC—have exactly two values Y=Yes and N=No. Interestingly, they do not have the aliases T=True and F=False that real binary properties have.)

eggrobin: To @Manishearth’s original comment: The `@missing` support is there in this PR already, in enum_codepointtrie.rs‎ which is where it is applicable:

https://github.com/unicode-org/icu4x/blob/556232d1ec1d595898023684a85926f3cfec7334/provider/source/src/properties/enum_codepointtrie.rs#L115-L122

provider/source/src/properties/enum_codepointtrie.rs
```diff
@@ -226,28 +311,43 @@ macro_rules! expand {
...
-                    )?;
-                    let trie = data.build_codepointtrie()?;
 
-                    Ok(DataResponse {
-                        metadata: Default::default(),
-                        payload: DataPayload::from_owned(PropertyCodePointMap::CodePointTrie(trie)),
-                    })
+                    #[cfg(not(any(feature = "use_wasm", feature = "use_icu4c")))]
+                    return Err(DataError::custom(
+                        "icu_provider_source must be built with use_icu4c or use_wasm to build properties data",
```
Manishearth, line 317: question: is this change something we care about?

I think this is fine, kind of inevitable, wasm is default so I expect most users to just use this without noticing.

provider/source/src/properties/enum_codepointtrie.rs
```diff
@@ -258,6 +358,9 @@ macro_rules! expand {
                     }
                     let trie = map
                         .into_iter()
+                        // Filter CCC's numeric names.
+                        // TODO: Don't
```
Manishearth, line 362: nit: followup issue

provider/source/src/properties/script.rs
```diff
@@ -19,52 +24,119 @@ impl DataProvider<PropertyScriptWithExtensionsV1> for SourceDataProvider {
...
+                let mut script_sets_lookup = BTreeMap::new();
+
+                let mut char_with_extensions = HashMap::new();
+
+                for line in self
+                    .unicode()?
+                    .read_to_string("ucd/ScriptExtensions.txt")?
+                    .lines()
+                {
+                    let line = line.split('#').next().unwrap().trim();
```
Manishearth, line 62: observation: this code is more complex than the others

provider/source/src/properties/uprops_serde.rs
```diff
@@ -2,50 +2,32 @@
...
 }
 
 #[derive(serde::Deserialize)]
 pub(crate) struct PropertyValue<T> {
     pub(crate) discr: T,
-    pub(crate) long: String,
-    pub(crate) short: Option<String>,
-    #[serde(default)]
-    pub(crate) aliases: Vec<String>,
+    pub(crate) short: String,
```
Manishearth, line 22: observation: so we still use this data for ICU4C discriminants and the short name?

provider/source/src/segmenter/mod.rs
```diff
@@ -750,33 +791,36 @@ macro_rules! implement_override {
     }
 }
 
-fn hardcoded_segmenter_provider() -> SourceDataProvider {
+fn unicode_15_1() -> &'static SourceDataProvider {
```
Manishearth, line 794: nit: document as being for segmenter to use

provider/source/src/segmenter/mod.rs
```diff
@@ -200,25 +148,36 @@ fn generate_rule_break_data(
...
-            properties_names.push(p.name.clone());
-            (properties_names.len() - 1).try_into().unwrap()
-        } else {
-            continue;
-        };
-
-        if p.left.is_none() && p.right.is_none() {
-            // If any values aren't set, this is builtin type.
-            simple_properties_count += 1;
+    match &*segmenter.segmenter_type {
```
Manishearth, line 151: I like this new code

Manishearth (approved):
Overall I like this. I still would prefer if all UCD-format handling for loops were in a separate file in one place (even if they are half a dozen different functions with slightly different APIs), in case we need to fix bugs around that. Markus/Elango have indicated that there are nuances there, and `@missing` is one such nuance we messed up here once already, and things are liable to change in the future.

But I'm happy to do that myself as a followup.

provider/source/src/properties/enum_codepointtrie.rs
Manishearth: Shane also agrees this is fine.

PR #7905, opened 2026-04-22
Support numbering system overrides in datetime

Manishearth:
Fixes #1225

Multiple of these are used in a year or day context, so the original suggestion of "generate string month data for these" does not work. In theory, we _could_ handle most of these in one-off ways: `jpnyear` is a simple hardcoded `if`, the month-only ones like romanlow can go in data, etc. But since we need this infrastructure anyway for hanidec/hebr years, I don't think it's worth implementing this three different ways.

This is a pretty nontrivial PR that was also almost entirely agent-written. I spent a lot of work talking to the agent, linking it to docs, and getting it to diagnose various issues. In the actual code, most of my work was minor cleanups and reorganizing the `jj` history.

(I used an agent instead of doing this myself because I am attending UTC this week and I cannot do focused coding work but I can prompt agents)

The agent was pretty good at doing its own printf debugging to figure out the source of various bugs. The initial naïve implementation of just applying the preexisting pattern code was insufficient, there were a number of bugs, and it took a lot of back-and-forth with the agent to diagnose them. There were a lot of cycles of the agent implementing an incorrect fix (e.g. by force-overriding things in datagen) and me saying we don't want to do that, undoing the change, and trying again with more prompting.

A thing I did not do (that the agent wanted to do) was support explicit user-driven overrides, like `he-u-calendar-hebrew-nu-hebr`. I didn't do this because I'd need to support `hanidays` for a wider range. However, it's not a complicated change to make.

<s>I have been reviewing this code extensively as I go on. However, this is a draft PR because I haven't done a proper end to end review. I also want to triple check the hebrew code against sources: it matches my understanding of the formatting but it might not be 100% correct.</s>

## Changelog

icu_datetime: Support numbering system overrides for datetime patterns when found in data

PR #7906, opened 2026-04-23
Simplify test sources

robertbastian:
Split from #7882 

## Changelog

N/A


provider/source/src/segmenter/mod.rs
```diff
@@ -98,11 +99,8 @@ fn generate_rule_break_data(
     use icu::properties::{props::ExtendedPictographic, PropertyParser};
     use icu_codepointtrie_builder::CodePointTrieBuilder;
 
-    let segmenter = provider
-        .icuexport()
-        .unwrap()
-        .read_and_parse_toml::<SegmenterRuleTable>(rules_file)
-        .expect("The data should be valid!");
+    let segmenter =
```
Manishearth, line 100: question: are we not caching anymore? don't we want to cache?

Manishearth (commented):
Just looking at hte first commit, I'm not really sure what's going on. The `include_files` macro seems great (feel free to land separately if you'd like). I don't quite understand the rest of this "cleanup". What is being changed? Why? What's bad with what we have now?

provider/source/src/segmenter/mod.rs
robertbastian: We don't need caching because these are only loaded once, and even if not, the TOML parsing is only a tiny part of the work that's being done in the implementation.

These don't belong in the icuexportdata source, which I'm trying to slim down.

robertbastian:
I believe your question refers to the second commit, because I don't see what you wouldn't understand about the others. The second commit changes the download-repo-sources script to use the existing `AbstractFs` code, which deals with networking, unzipping, etc., so we don't have to maintain all of that twice.

Manishearth:
I mean, I was just reviewing the `simplify` commit in the commit-by-commit view; This PR has very little information on what it is trying to do; I have to piece it together from the code, which is fine but slower, and commit-by-commit view is nicer for that.

robertbastian:
simplify just simplifies the generated code, and makes some small changes here and there like introducing some internal constructors which really shouldn't be controversial

Manishearth (commented):
Alright, I think I see the pertinent changes.

This

 - Moves download-repo-sources into an icu_provider_source "test" so that it can use AbstractFS. I'm not a huge fan of using the test runner for binaries but I guess it works.
 - Uses AbstractFS
 - Adds a macro that makes filepahts betters
 - Stops using AbstractFS for one-off data loads. I'm still not fully sure of this but it's intentional.

Overall seems positive. Want to wait for @sffc's thoughts especially on turning download-repo-sources into a non-test. But I think this is fine.

robertbastian:
> I'm not a huge fan of using the test runner for binaries but I guess it works.

We already do this for make-testdata.

robertbastian:
This is blocking other PRs, can you not just approve?

Manishearth (approved):
ok. For future reference, if you want reviews to be faster, please, _please_ provide some context on what you're doing in the PR body. It was not easy to figure out what this PR was trying to do; and why.

Manishearth:
Hmm, this PR removed `haw/ca-gregory` data, which I was using in https://github.com/unicode-org/icu4x/pull/7905 for testing. Was that intentional?

sffc:
> Hmm, this PR removed `haw/ca-gregory` data, which I was using in #7905 for testing. Was that intentional?

https://github.com/unicode-org/icu4x/pull/7915

sffc (commented):
I looked at the first commit for a while and am not really sure what it is doing but 🤷.

Moving the tool to be a unit test tool is fine with me. We did it with make-testdata already. I consider it a flaw in Rust that there isn't an easy way to export APIs for local users that aren't included in a crates.io release.

The deleting of the "unused" data files should not have been in this PR. They exist for reasons. Maybe those reasons are flawed, but that should be debated in a standalone PR, not in a "refactor to be merged asap" PR.

PR #7884, opened 2026-04-20
messageformat: MF2 implementation, :currency CLDR accounting, and conformance fixtures

JustFly1984:
## Summary

Adds **Unicode MessageFormat 2** under `icu_experimental::messageformat` (parse, validate, format, built-in handlers, `format_to_parts`), vendors the **message-format-wg** JSON conformance suite at a pinned SHA, and wires the `messageformat_conformance` / `messageformat_test` targets. Adds ICU4X-maintained `functions/unit.json` for draft **`:unit`** until the WG ships the same path, with fixture schema and `sync-mf2-tests.sh` updates.

### `:currency` and CLDR accounting

- Routes **`currencySign=accounting`** through dimension currency formatters and `CurrencyEssentials`, including **`currencyDisplay=code`** / **`never`** via outer affixes when the resolved short pattern encodes the negative sign.
- **`icu_provider_source`**: **`#3838`** — `extract_currency_essentials` keys `currencyFormats` off **`numbers.json` `defaultNumberingSystem`** (Sindhi → `latn` per **`#5374`**, then `latn` fallback). **`#6064`** — `currency_pattern_selection` considers optional negative `;` subpatterns together with positive arms, with a merge heuristic when they disagree; unit tests cover the merge helper.
- **Short compact currency (`#6064`)**: log on negative subpatterns only when they disagree from positive under the same parse rules or fail to parse.
- **`icu_experimental_data`**: rebaked **`CurrencyEssentialsV1`** so compiled data matches the new datagen. Regression tests account for **`ar-EG`**: default numbering system **`arab`** uses CLDR rows where accounting equals standard (no ASCII parentheses), unlike **`latn`** rows for the same locale.

### Other changes

- **`documents/todo-inventory.md`**: workspace TODO/FIXME snapshot and pointers to follow-ups.
- **DisplayNames (experimental, breaking):** multi `Region` / `Script` / `Language` / `Variant` `::of` return `Option<Cow<'_, str>>` and respect **`DisplayNamesOptions::fallback`**; FFI and examples updated. Changelog entries under **`icu_experimental`**.
- **Calendar / personnames / hygiene:** `Date` **`Ord`** vs **`PartialOrd`** documentation; personnames derived initials capped; small **`#3709`** comment typo fix in calendrical calculations.
[...]

Issue #3838, opened 2023-08-10
in `provider/datagen/src/transform/cldr/currency/mod.rs`, the currency patterns might be numbering system dependent.


younies:

sffc:
Since this is an i18n correctness issue I consider this blocking the currency formatter.

JustFly1984:
Cross-linking work in progress: https://github.com/unicode-org/icu4x/pull/7884

This PR implements **locale-default** selection of `currencyFormats` for `CurrencyEssentials` datagen by keying off `numbers.json` `defaultNumberingSystem` (with the Sindhi → `latn` override and `latn` fallback described in the PR), rebakes `CurrencyEssentialsV1` data, and updates related tests/docs.

It **does not** necessarily cover everything originally envisioned for numbering-system–specific patterns (e.g. marker-scoped `-u-nu-*` overrides called out in the PR description). Maintainers may want to keep this issue open until that scope is explicitly done or split to follow-ups.

PR #7909, opened 2026-04-24
fix(provider_source): CurrencyEssentials uses default numbering system (#3838)

JustFly1984:
## Summary

`extract_currency_essentials` now keys `currencyFormats` off `numbers.json` `defaultNumberingSystem`, matching `currency/patterns.rs`, with:

- The Sindhi `latn` override (unicode-org#5374)
- A `latn` map fallback when the default key is missing

Rebake of `icu_experimental_data` `CurrencyEssentialsV1` picks up the new payloads — under CLDR 48.2 only `fa` changes (the Arabic-default locales use CLDR rows where `arab`'s `currencyFormats` equals `latn`'s, so their baked payloads are unchanged).

## Test plan

- [x] `cargo test -p icu_provider_source --features unstable --lib currency::essentials` — `currency::essentials::test_basic ... ok`
- [x] `cargo make bakeddata experimental` — only `fa` currency/essentials row changes in `fingerprints.csv`

## Related issues

- unicode-org#3838 — locale-default numbering system for `CurrencyEssentials` datagen
- unicode-org#5374 — Sindhi `latn` override

## Context

This is the first of several smaller PRs split off from #7884. That umbrella PR stays open and will be cross-linked as the pieces land.


sffc (changes requested):
I think we probably want to make the CurrencyEssentials data marker have a numbering system attribute, like the icu_decimal data markers. CC @younies

robertbastian (commented):
> I think we probably want to make the CurrencyEssentials data marker have a numbering system attribute, like the icu_decimal data markers. CC @younies

yes

PR #7910, opened 2026-04-24
fix(provider_source): merge currency PatternSelection across `;` subpatterns (#6064)

JustFly1984:
## Summary

- `currency_pattern_selection` now classifies both the positive and optional negative CLDR subpatterns and merges them via a small helper:
  - identical arms pass through,
  - when the two arms disagree, `StandardAlphaNextToNumber` wins so negative amounts render consistently until per-polarity selections exist in data.
- Unit tests cover the merge heuristic (identical + union-prefers-alpha cases).
- Short compact currency datagen parses the optional negative subpattern with the **same rules as positive** and only logs on mismatch or parse failure, instead of warning whenever a negative subpattern is present.

## Test plan

- [x] `cargo test -p icu_provider_source --features unstable --lib currency::` — 8 passed (including the two new `merge_currency_pattern_selections_tests`).

## Related issues

- unicode-org#6064 — currency negative subpatterns / pattern selection / compact warnings

## Context

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Changelog
 
 
+## Unreleased
+
+- Data
+    - `icu_provider_source`: `currency_pattern_selection` now considers the optional CLDR negative subpattern alongside the positive one; when the two arms disagree, a single `PatternSelection` is chosen so negative amounts render consistently (unicode-org#6064)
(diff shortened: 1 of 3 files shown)


sffc:
I think @robertbastian should review this one since he recently reviewed this code.

robertbastian (changes requested):
> * when the two arms disagree, `StandardAlphaNextToNumber` wins so negative amounts render consistently until per-polarity selections exist in data.

This is not the right thing to do. We know that the code does not implement this correctly (it's experimental code), but this change is not more correct. It adds a lot of AI-generated code, but doesn't solve the underlying issue of supporting different positive and negative patterns, or in fact generate any different data than before.

Issue #3077, opened 2023-02-03
Add person name formatting in ICU4X


sffc:
It would be nice to have an ICU4X implementation of the output of the CLDR Person Names WG, ideally in time for the Unicode person names event.

sffc:
- @zbraniecki - It may be a bit tricky because the spec is so new and there are probably loose ends.
- @sffc - It's more about the positioning of ICU4X as implementing the latest cutting-edge features.

JustFly1984:
Related PR: https://github.com/unicode-org/icu4x/pull/7884

That PR also includes a small **person names** adjustment: `derive_missing_initials` caps generated initials at **two** tokens when interpolating multi-part given names into initial patterns (aligned with two-slot `initial_sequence_pattern` usage), with tests updated accordingly.

This is a minor fix in the same branch as the larger MessageFormat / currency work; it does not change the overall umbrella scope tracked here.

PR #7911, opened 2026-04-24
fix(personnames): cap derived initials at two (#3077)

JustFly1984:
## Summary

- `personnames::specifications::derive_missing_initials` now caps derived initials at two for multi-part given names.
- Rationale: `initial_sequence_pattern` is a two-slot pattern, so interpolating further initials would append beyond the intended display width.
- The existing multi-initial test (previously gated with a \`TODO(#3077)\` and `assert_ne!`) now holds as `assert_eq!(result, \"M. J.\")`.

## Test plan

- [x] `cargo test -p icu_experimental --lib personnames::specifications::derive_missing_initials` — 3 passed (`test_multi_3_initial_should_still_only_be_2`, `test_multi_initial`, `test_single_initial`).

## Related issues

- unicode-org#3077 — `personnames` derived initials overflow

## Context

### DIFF
--- a/components/experimental/src/personnames/specifications/derive_missing_initials.rs
+++ b/components/experimental/src/personnames/specifications/derive_missing_initials.rs
@@ -34,7 +34,10 @@ pub fn derive_missing_initials(
                 modifier: requested_field.modifier.with_length(FieldLength::Auto),
             })
             .split(' ')
-            .filter_map(|s| s.trim().chars().next());
+            .filter_map(|s| s.trim().chars().next())
+            // Cap at two initials: `initial_sequence_pattern` is a two-slot pattern, and
+            // interpolating further initials would append beyond the intended display width.
+            .take(2);
 
         let mut interpolated_initials =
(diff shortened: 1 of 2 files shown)


components/experimental/src/personnames/specifications/derive_missing_initials.rs
```diff
@@ -34,7 +34,10 @@ pub fn derive_missing_initials(
                 modifier: requested_field.modifier.with_length(FieldLength::Auto),
             })
             .split(' ')
-            .filter_map(|s| s.trim().chars().next());
+            .filter_map(|s| s.trim().chars().next())
+            // Cap at two initials: `initial_sequence_pattern` is a two-slot pattern, and
+            // interpolating further initials would append beyond the intended display width.
+            .take(2);
```
sffc, line 40: Is the `.take(2)` backed by the person names spec text in UTS 35?

This smells like a change Claude made in order to make a test pass.

sffc (changes requested):

PR #7912, opened 2026-04-24
fix(displaynames): honor Fallback option in multi Region/Script/Language/Variant `of()`

JustFly1984:
## Summary

Breaking change in `icu_experimental` (`displaynames` is experimental): `RegionDisplayNames::of`, `ScriptDisplayNames::of`, `LanguageDisplayNames::of`, and `VariantDisplayNames::of` now return `Option<Cow<'_, str>>` and respect `DisplayNamesOptions::fallback`:

- `Fallback::Code` — return the subtag itself as an owned string when no localized name exists.
- `Fallback::None` — return `None`.

Doctests, the Rust example (`examples/cargo/experimental/src/main.rs`), and the FFI `RegionDisplayNames::of` shim in `icu_capi` are updated for the new return type.

## Test plan

- [x] `cargo test -p icu_experimental --features compiled_data --test displaynames_test` — 3 passed, including the two new tests:
  - `region_display_fallback_code_uses_subtag` — missing data under `Fallback::Code` returns the region code
  - `region_display_fallback_none_returns_none` — missing data under `Fallback::None` returns `None`
- [x] `cargo test -p icu_experimental --features compiled_data --doc displaynames` — 20 doctests pass
- [x] `cargo check -p icu_capi` — FFI shim compiles

## Context

### DIFF
--- a/components/experimental/src/displaynames/displaynames.rs
+++ b/components/experimental/src/displaynames/displaynames.rs
@@ -38,7 +38,7 @@ define_preferences!(
 /// let display_name = RegionDisplayNames::try_new(locale, options)
 ///     .expect("Data should load successfully");
 ///
-/// assert_eq!(display_name.of(region!("AE")), Some("United Arab Emirates"));
+/// assert_eq!(display_name.of(region!("AE")).as_deref(), Some("United Arab Emirates"));
 /// ```
 #[derive(Default, Debug)]
 pub struct RegionDisplayNames {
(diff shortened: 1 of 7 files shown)


components/experimental/src/displaynames/displaynames.rs
```diff
@@ -79,14 +79,24 @@ impl RegionDisplayNames {
...
             Some(Style::Short) => data.short_names.get(&region.to_tinystr().to_unvalidated()),
             _ => None,
         }
-        .or_else(|| data.names.get(&region.to_tinystr().to_unvalidated()))
-        // TODO: Respect options.fallback
+        .or_else(|| data.names.get(&region.to_tinystr().to_unvalidated()));
+        match name {
+            Some(s) => Some(Cow::Borrowed(s)),
+            None => match self.options.fallback {
+                Fallback::Code => Some(Cow::Owned(region.as_str().into())),
```
sffc, line 96: Issue: Don't allocate Strings, especially short strings like this. I've seen AIs love to allocate strings, but that is not the right solution.

I suggest perhaps taking `Region` by reference instead of by value so you can borrow from it directly.

sffc (changes requested):

PR #7913, opened 2026-04-24
docs(calendar): clarify Date Ord vs PartialOrd; fix #3709 typo

### LINKED ISSUE #3709: Chinese and Dangi incosistent with ICU implementations for extreme dates
The current implementation of the Chinese calendar, as well as the Dangi calendar in #3694, are not consistent with ICU for all dates; based on writing a number of manual test cases (see the aforementioned PR), this seems to only be an issue for dates very far in the past or far in the future (ex. year -3000 ISO). Additionally, the difference only seems to be off by 1 day in all cases I've tested.

Since the difference is very small, and only happens for dates which are not very important practically, @Manishearth and I agree this is a low priority problem. Additionally, looking at the existing ICU code (see #3703 for example), there are places in ICU which intentionally deviate from more accurate methods of determining dates. Furthermore, the ICU4X Chinese/Dangi and astronomy functions are newly-written and have several algorithms based on the most recent edition of _Calendrical Calculations_, while the existing ICU code seems to be from 2000, incorporating algorithms from the 1997 edition of _Calendrical Calculations_.

Combining these facts together means it is very probable that the new ICU4X code gives more accurate predictions than existing ICU. This is probably a very low priority issue, but it might be worthwhile to eventually determine whether this is worth addressing in either ICU4X or ICU, and if so, how. At least for now, I have added some comments into astronomy.rs and chinese_based.rs indicating that this is a known issue.

JustFly1984:
## Summary

- `icu_calendar::Date`: document the intended semantics of `Ord` when the calendar instances are incompatible. `PartialOrd` already returns `None` in that case; `Ord` must produce a total ordering, so we compare the raw `DateInner` representations — which is **not** a chronologically meaningful ordering across incompatible calendars. Prefer `PartialOrd` (or ensure compatible calendars) whenever semantic ordering matters.
  - Behaviorally identical to before; the `check_date_compatibility` match previously returned the same `inner().cmp(...)` in both arms, so it's simplified to a direct call.
- `calendrical_calculations::astronomy`: fix the spelling in the `TODO(#3709)` comment (\"inconcistencies\" → \"inconsistencies\").

## Test plan

- [x] `cargo build -p icu_calendar -p calendrical_calculations` — clean build.

## Related issues

- unicode-org#3709 — typo fix in comment only; does not change behavior

## Context

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 # Changelog
 
 
+## Unreleased
+
+- Components
+    - `icu_calendar`
+        - Document `Ord` vs `PartialOrd` semantics on `Date` when calendar instances are incompatible: `Ord` falls back to comparing the raw `DateInner` representations to satisfy the trait contract; `PartialOrd` remains the semantic ordering and returns `None` when calendars disagree.
+- Utils
+    - `calendrical_calculations`: fix spelling in `TODO(#3709)` comment.
+
 ## icu 2.2.x
 
 Several crates have had patch releases in the 2.2 stream:
--- a/components/calendar/src/date.rs
+++ b/components/calendar/src/date.rs
@@ -631,6 +631,19 @@ where
     }
 }
 
+/// Total ordering for [`Date`](crate::Date) values that share the same [`AsCalendar`](crate::AsCalendar)
+/// type parameter.
+///
+/// When [`Calendar::check_date_compatibility`](crate::Calendar::check_date_compatibility) succeeds,
+/// this compares the underlying [`Calendar::DateInner`](crate::Calendar::DateInner) values, matching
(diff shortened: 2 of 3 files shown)


sffc (changes requested):
I think the problem should be fixed, not documented away. For example, should the dates be converted to the same calendar and then compared? (I am not suggesting you do this, I am suggesting we have a discussion first)

robertbastian:
Yes this needs to be discussed.

sffc:
Some options moving forward:

1. Implement `Ord` only for `Calendar<DateCompatibilityError = Infallible>` (semver breaking)
2. Do a RD comparison in the incompatible case

What's not an option is to document and embrace the DateInner comparison, because that's just wrong.

PR #7916, opened 2026-04-25
Add unstable provider APIs to infallibly create a DecimalFormatter

sffc:
EDIT: Despite what I had written below, this PR is a standalone improvement to add a cross-crate internal API for creating a DecimalFormatter without a Result.

Original description: For @Manishearth to use in #7905

## Changelog

icu_decimal: added APIs in the provider module to unstably but infallibly create a DecimalFormatter

### DIFF
--- a/components/decimal/src/provider.rs
+++ b/components/decimal/src/provider.rs
@@ -304,28 +304,89 @@ impl<'a> core::ops::Deref for DecimalSymbols<'a> {
     }
 }
 
+const UND_LATN_DECIMAL_SYMBOLS_BYTES: &[u8] = &[
+    1, 1, 2, 2, 3, 4, b'-', b'+', b'.', b',', b'l', b'a', b't', b'n',
+];
+
 impl DecimalSymbols<'static> {
+    /// The symbols for the default locale in the default numbering system.
+    pub const UND_LATN: Self = Self {
+        // Safety: well-formedness tested in `test_decimal_symbols_und_latn_safety`
+        strings: unsafe { VarZeroCow::from_bytes_unchecked(UND_LATN_DECIMAL_SYMBOLS_BYTES) },
+        grouping_sizes: GroupingSizes {
+            primary: 3,
+            secondary: 3,
+            min_grouping: 1,
+        },
+    };
+
     /// Create a new en-US format for use in testing
     #[cfg(feature = "datagen")]
     pub fn new_en_for_testing() -> Self {
-        let strings = DecimalSymbolStrsBuilder {
-            minus_sign_prefix: VarZeroCow::new_borrowed("-"),
-            minus_sign_suffix: VarZeroCow::new_borrowed(""),
-            plus_sign_prefix: VarZeroCow::new_borrowed("+"),
-            plus_sign_suffix: VarZeroCow::new_borrowed(""),
-            decimal_separator: VarZeroCow::new_borrowed("."),
-            grouping_separator: VarZeroCow::new_borrowed(","),
-            numsys: VarZeroCow::new_borrowed("latn"),
-        };
-        Self {
-            strings: VarZeroCow::from_encodeable(&strings),
-            grouping_sizes: GroupingSizes {
-                primary: 3,
-                secondary: 3,
-                min_grouping: 1,
-            },
+        Self::UND_LATN
+    }
+
+    /// Infallibly create a formatter with these symbols and the specified digits.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use icu::decimal::provider::DecimalSymbols;
+    /// use icu::decimal::provider::DIGITS_HANIDEC;
+    /// use writeable::assert_writeable_eq;
+    ///
+    /// let formatter = DecimalSymbols::UND_LATN.make_formatter_unstable(DIGITS_HANIDEC, Default::default());
+    /// let decimal = 12345678.into();
+    ///
+    /// assert_writeable_eq!(
+    ///     formatter.format(&decimal),
+    ///     "一二,三四五,六七八"
+    /// );
+    /// ```
+    pub fn make_formatter_unstable(
+        self,
+        digits: [char; 10],
+        options: crate::options::DecimalFormatterOptions,
+    ) -> crate::DecimalFormatter {
+        crate::DecimalFormatter {
+            options,
+            symbols: DataPayload::from_owned(self),
+            digits: DataPayload::from_owned(digits),
         }
     }
+}
+
+#[test]
+fn test_decimal_symbols_und_latn_safety() {
+    let actual_und_latn: DataResponse<DecimalSymbolsV1> = Baked
+        .load(DataRequest {
+            id: DataIdentifierBorrowed::for_locale(&DataLocale::default()),
+            metadata: Default::default(),
+        })
+        .unwrap();
+    assert_eq!(&DecimalSymbols::UND_LATN, actual_und_latn.payload.get());
+    assert_eq!(
+        UND_LATN_DECIMAL_SYMBOLS_BYTES,
+        actual_und_latn.payload.get().strings.as_bytes()
+    );
+}
+
+/// The digits of the `hanidec` numbering system.
+///
+/// Compatible with [`DecimalDigitsV1`].
+pub const DIGITS_HANIDEC: [char; 10] = ['〇', '一', '二', '三', '四', '五', '六', '七', '八', '九'];
+
+#[test]
+fn test_hanidec_consistency() {
+    let actual_hanidec_digits: DataResponse<DecimalDigitsV1> = Baked
+        .load(DataRequest {
+            id: DataIdentifierBorrowed::for_marker_attributes(
+                DataMarkerAttributes::from_str_or_panic("hanidec"),
+            ),
+            metadata: Default::default(),
+        })
+        .unwrap();
+    assert_eq!(&DIGITS_HANIDEC, actual_hanidec_digits.payload.get());
 }
 
 #[cfg(feature = "unstable")]
--- a/components/decimal/src/decimal_formatter.rs
+++ b/components/decimal/src/decimal_formatter.rs
@@ -34,9 +34,9 @@ size_test!(DecimalFormatter, decimal_formatter_size, 96);
 #[doc = decimal_formatter_size!()]
 #[derive(Debug, Clone)]
 pub struct DecimalFormatter {
-    options: DecimalFormatterOptions,
-    symbols: DataPayload<DecimalSymbolsV1>,
-    digits: DataPayload<DecimalDigitsV1>,
+    pub(crate) options: DecimalFormatterOptions,
+    pub(crate) symbols: DataPayload<DecimalSymbolsV1>,
+    pub(crate) digits: DataPayload<DecimalDigitsV1>,
 }
 
 impl AsRef<DecimalFormatter> for DecimalFormatter {


components/decimal/src/provider.rs
```diff
@@ -304,30 +304,91 @@ impl<'a> core::ops::Deref for DecimalSymbols<'a> {
...
+    /// use icu::decimal::provider::DecimalSymbols;
+    /// use icu::decimal::provider::DIGITS_HANIDEC;
+    /// use writeable::assert_writeable_eq;
+    ///
+    /// let formatter = DecimalSymbols::UND_LATN.make_formatter_unstable(DIGITS_HANIDEC, Default::default());
+    /// let decimal = 12345678.into();
+    ///
+    /// assert_writeable_eq!(
+    ///     formatter.format(&decimal),
+    ///     "一二,三四五,六七八"
```
robertbastian, line 343: Using Han digits with Latin commas is bad i18n, so this is a bad example.

components/decimal/src/provider.rs
```diff
@@ -304,30 +304,91 @@ impl<'a> core::ops::Deref for DecimalSymbols<'a> {
...
+    /// );
+    /// ```
+    pub fn make_formatter_unstable(
```
robertbastian, line 346: the type is already unstable, no need to call a method `_unstable`

robertbastian, line 346: why does this logic live on `DecimalSymbols`? it should be a constructor on `DecimalFormatter`

robertbastian, line 346: this constructor breaks an (undocumented) invariant of `DecimalFormatter`, namely that the digits are loaded for `symbols.numsys`. We don't afaik use that invariant currently, but breaking it is bad i18n as shown above.

Manishearth:
As I note in https://github.com/unicode-org/icu4x/pull/7905#discussion_r3141764429, I don't really think these APIs pull their weight.

We shouldn't be using a full DecimalFormatter here; it's a heavier API that handles grouping/etc. What we just want is the logic for writing a decimal number (i.e. the stuff in `impl Writeable for <integer>`) with custom digits. I think that might be a useful API to expose somewhere (`write_decimal_with_digit_array([char; 10])`). Not sure.

An argument can be made about having decimal stuff live in `icu_decimal`, but the entire numeric_overrides module is decimal stuff; implementing various RBNFs manually.

We don't end up deduplicating anything here either. The decimal data is locked away in baked data, and we can at best access it via tests. `icu_datetime` can do that too, I can introduce a test that ensures that my `hanidec` digits are identical to `icu_decimal`'s baked hanidec digits.

sffc:
> We shouldn't be using a full DecimalFormatter here; it's a heavier API that handles grouping/etc. What we just want is the logic for writing a decimal number (i.e. the stuff in `impl Writeable for <integer>`) with custom digits. I think that might be a useful API to expose somewhere (`write_decimal_with_digit_array([char; 10])`). Not sure.

DecimalFormatter is not that heavy, and we already include DecimalFormatter code in datetime.

> An argument can be made about having decimal stuff live in `icu_decimal`, but the entire numeric_overrides module is decimal stuff; implementing various RBNFs manually.

No, most or all of these are non-decimal RBNFs. `romanlow` and friends are not decimal numbering systems. CLDR has a definition that a decimal numbering system is essentially one where you can substitute ASCII digits for numbering system digits, and all others are "algorithmic".

> We don't end up deduplicating anything here either. The decimal data is locked away in baked data, and we can at best access it via tests. `icu_datetime` can do that too, I can introduce a test that ensures that my `hanidec` digits are identical to `icu_decimal`'s baked hanidec digits.

We can delete the `n % 10, n / 10` type of logic, and the indexing_slicing.

Manishearth:
>  I can introduce a test that ensures that my hanidec digits are identical to icu_decimal's baked hanidec digits.

Went ahead and did that.

> We can delete the n % 10, n / 10 type of logic, and the indexing_slicing.

I'd rather go via something dedicated for simple digit formatting, then, not DecimalFormatter.

> No, most or all of these are non-decimal RBNFs. romanlow and friends are not decimal numbering systems. CLDR has a definition that a decimal numbering system is essentially one where you can substitute ASCII digits for numbering system digits, and all others are "algorithmic".

Ah, fair, I was loose about "numeric" vs "decimal".

> DecimalFormatter is not that heavy, and we already include DecimalFormatter code in datetime.

It's not a codesize thing, it's about calling an API that is more generic than needed.

We are _adding_ this unstable API just for the purposes of my PR. I don't see this API as well-motivated on its own, and I do not think my PR justifies this API. I might have held a different position if this already existed and was strongly motivated elsewhere.

I think my PR _may_ justify an API in the Writeable crate, though it would still be a little odd at least.

components/decimal/src/provider.rs
```diff
@@ -304,30 +304,91 @@ impl<'a> core::ops::Deref for DecimalSymbols<'a> {
     }
 }
 
+const UND_LATN_DECIMAL_SYMBOLS_BYTES: &[u8] = &[
+    1, 1, 2, 2, 3, 4, b'-', b'+', b'.', b',', b'l', b'a', b't', b'n',
```
Manishearth, line 308: issue: I also think this is strictly worse than what we had before.

We shouldn't be using `UND_LATN` decimal formatters anyway, we should use `Writeable for u32` if we want that (with, as you say, comments saying why it's okay, which we would need here too)

sffc:
A few things to unpack here.

This is a [cross-crate internal API](https://github.com/unicode-org/icu4x/blob/main/documents/process/style_guide.md#cross-crate-internal-apis--suggested). We went out of our way to _not_ state that cross-crate internal APIs needed to meet a higher motivational bar than private APIs. Of all the themes we debate as maintainers, none tires me more than debating the motivation of an API after already reaching for the `unstable` escape hatch which is supposed to short-circuit that debate.

I made this PR after _looking at_ your PR, but I did not make this _in response to_ your PR. I was at my computer one evening and decided to open a small friendly uncontroversial change to demonstrate how to use DecimalFormatter in an infallible way. I used `hanidec` not necessarily because it showed up in your PR but because it's the only decimal numbering system that requires an explicit `[char; 10]` since the digits are not in order. I mentioned #7905 in the description but this clearly muddied my message so I have edited the description.

I remain half opposed to, half confused by this statement

> I'd rather go via something dedicated for simple digit formatting, then, not DecimalFormatter.

DecimalFormatter _is_ "something dedicated for simple digit formatting". It's tiny; it's _literally_ the posterchild of ICU4X code size optimization. Sure, it spends a dozen of bytes of assembly on grouping separators.

I saw it as a flaw that DecimalFormatter, some of the first code I ever contributed in ICU4X, didn't serve your need for doing hanidec formatting, so I made this PR to show that it _can_ serve that need. We can separately debate whether it is the _right tool_ (for which I am strongly in favor and I think the facts are on my side), but that debate should not block this PR or your PR.

robertbastian:
> It's tiny; it's literally the posterchild of ICU4X code size optimization. Sure, it spends a dozen of bytes of assembly on grouping separators.

My main gripe is that the input type is not an integer but FixedDecimal, which is a Drop type that might allocate.

sffc:
> > It's tiny; it's literally the posterchild of ICU4X code size optimization. Sure, it spends a dozen of bytes of assembly on grouping separators.
> 
> My main gripe is that the input type is not an integer but FixedDecimal, which is a Drop type that might allocate.

That's fair. FixedDecimal is before we/I started parameterizing the storage type :). We could still do that, or make DecimalFormatter work via a trait, something we might want to do anyway with the fixed_decimal rewrite.

Manishearth:
> This is a [cross-crate internal API](https://github.com/unicode-org/icu4x/blob/main/documents/process/style_guide.md#cross-crate-internal-apis--suggested). We went out of our way to not state that cross-crate internal APIs needed to meet a higher motivational bar than private APIs.

I don't recall motivation as at all being a part of that discussion. But sure, they don't need to meet a higher motivational bar than private APIs. This PR doesn't meet the motivational bar for private APIs either, in my view.

If we have a private API that we don't use, then it gets deleted. Seems to me that if we are adding a private API, "are we going to actually use it" is a reasonable question to ask.

The current state of the API is not one I think we will use. I don't want to use it in my code (see below), and I am struggling to come up with use cases for this.

In most cases in ICU4X, we should be using localized decimal formatting, using a full, _loaded_ DecimalFormatter. There are some cases where we need ungrouped Latin formatting (usually error cases), which is already handled by `Writeable for u32`.

This enables "use und-latn symbols, with en-US grouping data, with arbitrary digits", a use case full of footguns:

 * You get Latin symbols, which could be wrong (especially if you use Arabic digits, iirc).
 * If you do use grouping, you're silently stuck on `en-US` grouping. I think that never makes sense if you're using custom digits. I think that only barely makes sense when using Latn digits, even: maybe for debugging output. In error cases I'd rather not use grouping.


I just don't see where we'd use this. This comes up in datetime because we want to do extra decimal formatting that we're hardcoding, but it seems like a very special case, and while I don't like having to have a decimal formatting for loop in that code, I also don't want to have this footgun-filled pattern there.

(I'm very much in favor of moving the decimal formatting loop elsewhere. Writeable would allow for code sharing though I do think it's a bit strange there. There may be a place for such a function in `icu_decimal` in a way that avoids code duplication)

----------

Finally, this adds unsafe code, which I will always want motivation for.


> DecimalFormatter is "something dedicated for simple digit formatting". It's tiny; it's literally the posterchild of ICU4X code size optimization. Sure, it spends a dozen of bytes of assembly on grouping separators.


I'm not talking about the code size here, and have never been talking about the code size in this entire discussion. I'm talking about the API being clunky.

To use this in my hanidec use case, I'd have to do:

```rust
let fd = FixedDecimal::from(number);
DecimalSymbols::UND_LATN.make_formatter_unstable(icu_decimal::provider::DIGITS_HANIDEC, DecimalFormatterOptions { grouping_strategy: Some(GroupingStrategy::None });
```

That's clunky, and has a bunch of things that need to be explained (what is UND_LATN doing here? Did we construct the FixedDecimal correctly?). I'm not a fan of that.

The Drop thing Robert notes is also a (minor) issue.

sffc:
> This enables "use und-latn symbols, with en-US grouping data, with arbitrary digits", a use case full of footguns:

No, it enables all use cases where you know your symbols and grouping data at compile time. I'm exposing `UND_LATN` as a convenience thing _that we already exposed in that file_, but I'm making it const-friendly. You can build your own symbols struct using the builder.

> Finally, this adds unsafe code, which I will always want motivation for.

I would be enthusiastically happy to add e.g. a `varzerovec!` proc macro that generates a VarULE at build time and avoid the unsafe, but have a strong expectation that you would push back against that saying that it replaces only a small bit of unsafe.

---

Since it seems `UND_LATN` is the main point of contention here, I am open to remove that change from this PR and only add the `make_formatter_unstable` function which is the thing that you can't currently do outside of the crate.

sffc:
> That's clunky, and has a bunch of things that need to be explained (what is UND_LATN doing here? Did we construct the FixedDecimal correctly?). I'm not a fan of that.

If you feel this way, then I am open to export a function that forces certain behavior. In the extreme case,

```rust
pub fn format_positive_int(value: u64, digits: [char; 10]) -> impl Writeable
```

I chose not to propose this design since it is a subset of the internal private power user version offered here.

Manishearth:
> I would be enthusiastically happy to add e.g. a varzerovec! proc macro that generates a VarULE at build time and avoid the unsafe, but have a strong expectation that you would push back against that saying that it replaces only a small bit of unsafe.

Correct. From the point of view of unsafe code, "here is some binary data that I guarantee is fine" is already iffy. Even with a test; as an unsafe reviewer this means I have to run the test, and that's not always easy.

But reviewing a proc macro is far, far, worse. Generally unsafe proc macros are very fishy and easy to get wrong. We have a couple of them, I do not want more without strong motivation.

I basically consider `zerovec` to be at the limit of unsafe complexity; it's getting very large and hard to review and I had to put a lot of effort into getting it reviewed by others. (Splitting it isn't a solution, an unsafe reviewer would still be reviewing the zerovecverse together, likely).


(This is a separate discussion but I want to try and make my position clear and predictable. You have already accurately predicted my position here)

> Since it seems UND_LATN is the main point of contention here, I am open to remove that change from this PR and only add the make_formatter_unstable function which is the thing that you can't currently do outside of the crate.

Yes, that would be fine. _That_ API I like, I think it's probably useful, though I don't know for sure what the use cases are.

Without UND_LATN the datetime numbering systems code can't use it without it being pretty clunky, though. I think? Are we okay with allocating a DecimalSymbols during number formatting?

I guess technically the numbering systems code can just use the `DecimalFormatter`'s `DecimalSymbols` object and format based on that? As long as you are formatting positive numbers with no grouping the actual object you use shouldn't matter.

I know you want to separate the discussion of this PR from the numbering systems use case, but that _is_ a place where I'd _like_ to use _some_ API, so I think it's relevant to at least see how various API proposals measure up.


> If you feel this way, then I am open to export a function that forces certain behavior. In the extreme case,

That's what I proposed in the other thread. If it lives in the `writeable` crate we can add it without any code duplication, but it would feel like a strange wart on that crate. If it lives in `icu_decimal` there may be code and codesize duplication which isn't great. I'm not sure if there's a way to structure it to avoid it.

sffc:
I don't plan to put more time into this right now but may revisit in the future.

PR #7918, opened 2026-04-27
Fix `get_likely_script`

### LINKED ISSUE #7866: Should LocaleDirectionality::get perform full Add-Likely-Subtags algorithm?
`LocaleDirectionality` doesn't add likely subtags to `und` in all cases. Is this expected? I've interpreted Marcus' comment in <https://github.com/unicode-org/icu4x/issues/3172#issuecomment-1462648659> that a missing script subtag is added by performing the Add-Likely-Subtags algorithm and this is also what then got spec'ed into ECMA402's [TextDirectionOfLocale](https://tc39.es/ecma402/#sec-textdirectionoflocale).

This test case prints `Some(RightToLeft)` for `und-SA`, but `None` for `und-US`. 

```rs
use icu_locale::LocaleDirectionality;
use icu_locale::langid;

fn main() {
  let ld = LocaleDirectionality::new_extended();
  dbg!(ld.get(&langid!("und-SA")));
  dbg!(ld.get(&langid!("und-US")));
}
```

robertbastian:
Fixes #7866

## Changelog

`icu_locale`:
* Fixes an issue in `LocaleDirectionality` for unknown languages

### DIFF
--- a/components/locale/src/expander.rs
+++ b/components/locale/src/expander.rs
@@ -374,6 +374,7 @@ impl LocaleExpander {
     /// ```
     pub fn maximize(&self, langid: &mut LanguageIdentifier) -> TransformResult {
         let data = self.as_borrowed();
+        let (und_l, und_s, und_r) = data.get_und();
 
         if !langid.language.is_unknown() && langid.script.is_some() && langid.region.is_some() {
             return TransformResult::Unmodified;
@@ -415,12 +416,10 @@ impl LocaleExpander {
         // We failed to find anything in the und-SR, und-S, or und-R tables,
         // to fall back to bare "und"
         debug_assert!(langid.language.is_unknown());
-        update_langid(
-            data.get_und().0,
-            Some(data.get_und().1),
-            Some(data.get_und().2),
-            langid,
-        )
+
+        // und expands to `und_l-und_s-und_r`
+        // TODO: remove this behavior: https://unicode-org.atlassian.net/browse/CLDR-14524
+        update_langid(und_l, Some(und_s), Some(und_r), langid)
     }
 
     /// This returns a new Locale that is the result of running the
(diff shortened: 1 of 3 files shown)


Manishearth:
Can review if I have to, but fallback code is not my forte.

components/locale/src/expander.rs
```diff
@@ -534,40 +534,41 @@ impl LocaleExpander {
...
-                return Some(script);
+        let (_, und_s, und_r) = self.as_borrowed().get_und();
+
+        if language.is_unknown() {
+            if region.is_none_or(|r| r == und_r) {
+                // If the region is unknown or `und`'s region
+                return Some(und_s);
             }
+            // Otherwise, infer the language and use that to infer the script
+            language = self.as_borrowed().get_r(region?)?.0;
```
sffc: Issue: Shouldn't `und` just return `None`? I don't feel like we should hallucinate `Latn` from it. It's weird that `und` gives an answer but `und-ZY` does not (where `ZY` is a region without a likely script subtag). Removing subtags should not cause you to get an answer where you previously got `None`.

robertbastian: I agree, but the problem is that the likely script subtag for `und` *is* `Latn`.

zbraniecki: is it because of CLDR data, or UTS 35 algo?

sffc: By this logic, then `und-XZ` should infer `Latn`, too. https://github.com/unicode-org/icu4x/pull/7919

sffc:
Discuss with CLDR DWG:

- Why are the likely subtags for und `en-Latn-US` instead of `und-Zzzz-ZZ`
- Even if there's a decent explanation for that, should the same logic apply to Locale Directionality?

sffc:
Awaiting feedback on the following CLDR issue: https://unicode-org.atlassian.net/browse/CLDR-14524

sffc (commented):
Aren't we still waiting on CLDR to make up its mind?

We currently implement the behavior that CLDR is considering adopting; I don't think we should implement the current spec behavior that no-one likes unless CLDR decides to keep it

sffc (changes requested):
^

robertbastian:
> We currently implement the behavior that CLDR is considering adopting

No we don't, we return `None` for `und-US` but `RightToLeft` for `und-SA`

sffc (approved):
ok

components/locale/src/expander.rs
```diff
@@ -603,4 +604,26 @@ mod tests {
...
+            Some(script!("Latn"))
+        );
+        assert_eq!(
+            lc.get_likely_script(&locale!("und-US").id),
+            Some(script!("Latn"))
+        );
+        assert_eq!(
+            lc.get_likely_script(&locale!("und-SA").id),
+            Some(script!("Arab"))
+        );
```
robertbastian, line 632: @sffc do you actually want to approve this behaviour? it's not what CLDR is planning

sffc: I'm not clear on what you are asking for my feedback on. "und-US" and "und-SA" are fixed by this PR, right? And "und" becomes more consistent with CLDR data, but the behavior is less good than before.

I think I lean toward merging this assuming it fixes an actual bug (und-US and und-SA).

robertbastian:
well let's just remove the `und` -> `en-Latn-US` fallback while we're here. CLDR is planning to do this, and ICU4X is a good candidate for trying this. doing this already revealed a bug in units

robertbastian:
I've split the bigger change into https://github.com/unicode-org/icu4x/pull/8110

components/locale/src/expander.rs
robertbastian: it should, for consistency with the locale expander.

sffc (approved):

PR #7919, opened 2026-04-28
Add locale expander tests for private use script and region

sffc:
## Changelog

N/A (test-only)


zbraniecki (approved):
ugh, I struggle a bit with `und`->`en`, but I guess we are not in the right place to revisit that.

robertbastian (commented):
well that's horrible

PR #7923, opened 2026-04-29
Implement AsRef for Locale and LanguageIdentifier

aumetra:
This PR adds `AsRef<LanguageIdentifier>` implementations for the `Locale` and `LanguageIdentifier` types of `icu_locale_core`.

The usecase for this change are trait bounds such as in `fluent-langneg`:

```rust
pub fn foobar<
    'a,
    R: 'a + AsRef<LanguageIdentifier>,
>(
    requested: &[R],
);
```

In this example a pure `From<T> for LanguageIdentifier` impl is not sufficient and would necessitate cloning the value, even for just a read-only access.

This is essentially just adding back trait impls that were removed from when the crate was created from `icu_locid`.

## Changelog

icu_locale_core: Add `AsRef<LanguageIdentifier>` impls
  - New impls: `AsRef<LanguageIdentifier> for Locale`, `AsRef<LanguageIdentifier> for LanguageIdentifier`

### DIFF
--- a/components/locale_core/src/langid.rs
+++ b/components/locale_core/src/langid.rs
@@ -511,6 +511,12 @@ impl LanguageIdentifier {
     }
 }
 
+impl AsRef<LanguageIdentifier> for LanguageIdentifier {
+    fn as_ref(&self) -> &LanguageIdentifier {
+        self
+    }
+}
+
 impl core::fmt::Debug for LanguageIdentifier {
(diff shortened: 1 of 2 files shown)


sffc (approved):

Manishearth (approved):
It's already public that Locale contains an LI, so this is possible to do without further constraining

PR #7927, opened 2026-05-01
Fix out-of-bounds panic in backward-secondary sort key path

rs-sac:
With collation "fr-CA" and an empty input string, the reverse-segment block panicked due to arithmetic underflow.  This is an error from the port.  ICU4C uses signed integers and caught this condition because the index went negative.

## Changelog

icu_collator: Fix a possible panic when generating a sort key for a collation that uses backward secondary level.


Manishearth (approved):
This makes sense. Sorry this took some time to review, feel free to ping me on PRs that languish in review.

cc @hsivonen for collator changes

PR #7928, opened 2026-05-01
Fix off-by-one Unihan range boundary in identical-level sort key

rs-sac:
ICU4C said `prev<0x4e00 || prev>=0xa000`, but the BOCSU port used an inclusive range, so U+A000 went into the wrong branch.

## Changelog

icu_collator: Fix generation of identical level sort keys containing the codepoint U+A000.

### DIFF
--- a/components/collator/src/comparison.rs
+++ b/components/collator/src/comparison.rs
@@ -2520,7 +2520,7 @@ where
     let mut prev = 0i32;
 
     for c in iter {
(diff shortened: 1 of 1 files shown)


Manishearth (approved):
Yep, this is correct. Thanks!

https://github.com/unicode-org/icu/blob/4ade0ece2ec3de769d060579cad1460a797db8f8/icu4c/source/i18n/bocsu.cpp#L119

PR #7930, opened 2026-05-01
Monomorphize sort key code based on reordering

rs-sac:
In many of the sort key benchmarks, performance improves by 3-4%.  In a few cases, the gain is over 10%.  A small number of cases regress by 2-3%.

## Changelog

icu_collator: The performance of sort key generation was tuned.


hsivonen (approved):
I'm curious what the binary size cost is, but I guess it's OK, since apps that want sort keys are going to want perf here, and LTO should be able to address the binary size for apps that don't use sort keys. Thank you and sorry about the delay.

PR #7934, opened 2026-05-04
Validate line break property constants

robertbastian:
The way datagen currently derives the numeric value for a property is by the ordering of the property in `line.toml`. This seems extremely brittle, and should be validated against the constants we use at runtime.

#1637

## Changelog

N/A


components/segmenter/src/line.rs
```diff
@@ -977,14 +910,40 @@ impl<Y: LineBreakType> Iterator for LineBreakIterator<'_, '_, Y> {
...
+                    | RuleBreakData::LINE_PROPERTY_SA,
+                    _,
+                ) => {
+                    left_prop = RuleBreakData::LINE_PROPERTY_ID;
                 }
                 //  typographic letter units shouldn't be break
                 (
                     LineBreakWordOption::KeepAll,
-                    AI | AL | ID | NU | HY | H2 | H3 | JL | JV | JT | CJ,
-                    AI | AL | ID | NU | HY | H2 | H3 | JL | JV | JT | CJ,
```
sffc, line 987: Nit (optional): I find the old formatting more readable because I can see that they are the same match

sffc (approved):
I think I have a weak preference for decoupling the renames from the refactor and improved tests but I don't feel strongly

Manishearth (approved):

Manishearth:
observation: RuleBReakData is a provider type, so we can add hidden things to it easily

Issue #3959, opened 2023-08-29
Stabilize UnicodeSet parsing


skius:
Tracking issue for UnicodeSet parsing stabilization

Decide which of these issues are blocking:
- #3556
- #3558
- #3559
- #3684
- #3893

(cc @robertbastian)

PR #7935, opened 2026-05-04
Move `unicodeset_parse` to `icu_properties`

robertbastian:
#3959

## Changelog

icu_properties:
* Add experimental UTS#35 Unicode set parsing


components/properties/src/lib.rs
```diff
@@ -103,3 +103,6 @@ mod private {
 
 #[cfg(feature = "harfbuzz_traits")]
 mod harfbuzz;
+
+#[cfg(feature = "unstable")]
+pub mod unicodeset_parse;
```
Manishearth, line 109: nit: should be called `unstable_`

also needs scary unstable header

Manishearth (approved):
Seems fine. This doesn't seem to be modifying `provider/source`, but when that happens we will need to access this code through `doc(hidden)` `pub mod unstable_internal` so that datagen doesn't need unstable features.

robertbastian:
> so that datagen doesn't need unstable features

datagen already uses unstable features.

Manishearth:
Datagen uses unstable APIs. It does not, and should not, use the unstable Cargo feature.

It is using them as "internal APIs" not "unstable APIs". As per the plan outlined in #6659 such APIs should be accessed via provider or unstable_internal paths.

robertbastian:
#6659 does not say anything about `icu_provider_source` not enabling the unstable feature

Manishearth:
Yes it does

> Cross-crate ICU4X-internal usages do similar things but use "internal" naming/modules and doc(hidden) instead of feature = unstable.


One of the goals there is that you should not accidentally get unstable code without typing "unstable" somewhere (ideally in Cargo.toml for "unstable", but acceptably in Rust code if it's a trait method or something), with a carve out for "provider" as a legacy naming we're not planning on redoing. I think we already have `unstable_internal` reexports in the code.

We do not ever want unstable features to be implicitly enabled by stable crates. I don't see why this is controversial.

robertbastian:
I fully understood this as "cross-component internal usages".

`icu_provider_source` enables `icu/unstable` behind its unstable feature. I don't see what the problem with that is.

Manishearth:
> icu_provider_source enables icu/unstable behind its unstable feature. 

Yes, unstable feature enabling unstable feature is fine.

`icu_segmenter`'s datagen code in `icu_provider_source` (stable), however, should never start enabling `icu_properties/unstable`.

Manishearth:
Oh, if the new segmenter datagen code is `icu_provider_source/unstable` then that's fine until it gets stabilized.

sffc (commented):
lgtm

PR #7937, opened 2026-05-05
Simplify segmenter implementations

robertbastian:
The goal here is to reduce the difference between the line segmenter implementations and the other segmenters, simplify code, and isolate tailoring logic.

Best reviewed commit by commit.

## Changelog

N/A


components/segmenter/src/line.rs
```diff
@@ -872,6 +872,18 @@ impl<Y: LineBreakType> Iterator for LineBreakIterator<'_, '_, Y> {
 
         'a: loop {
             debug_assert!(!self.is_eof());
+
+            // NOTE(egg): The special-casing of `LineBreakStrictness::Anywhere` allows us to pass
```
sffc, line 876: Not convinced that 85c4bbb275a58d2b8c8c7880126cd0812d0c4f38 is a no-op, but it seems harmless so long as we pass tests, and we can add more tests and fix things later

components/segmenter/src/lib.rs
```diff
@@ -158,9 +158,26 @@ pub mod options {
 
 /// Largely-internal scaffolding types (You should very rarely need to reference these directly)
 pub mod scaffold {
-    pub use crate::line::LineBreakType;
+    // TODO: These should have never been public
     pub use crate::rule_segmenter::{Latin1, PotentiallyIllFormedUtf8, RuleBreakType, Utf16, Utf8};
```
sffc, line 162: These ones are correctly public/exported in `scaffold` because they are in bounds. Their methods should be unstable/hidden.

components/segmenter/src/lib.rs
```diff
@@ -162,7 +162,19 @@ pub mod scaffold {
...
-    pub use crate::word::WordBreakType;
+    /// A trait allowing for [`WordBreakIterator`] to be generalized to multiple string iteration methods.
+    ///
+    /// This is implemented by ICU4X for several common string types.
+    ///
+    /// <div class="stab unstable">
+    /// 🚫 This trait is sealed; it cannot be implemented by user code. If an API requests an item that implements this
+    /// trait, please consider using a type from the implementors listed below.
+    /// </div>
+    pub trait WordBreakType: crate::private::Sealed + Sized + RuleBreakType {}
```
sffc, line 173: Note: It appears that WordBreakType is completely unused.

components/segmenter/src/word.rs
```diff
@@ -131,7 +125,11 @@ impl<'data, 's, Y: RuleBreakType> WordBreakIterator<'data, 's, Y> {
     /// Returns the word type of the segment preceding the current boundary.
     #[inline]
     pub fn word_type(&self) -> WordType {
-        self.0.word_type()
+        match self.0.rule_status() {
```
sffc, line 128: Thought: Good that we had a method here that you can easily rewire

sffc (approved):

components/segmenter/src/line.rs
robertbastian: `LineBreakStrictness::Anywhere` is broken *anyway*

sffc:
This PR, or perhaps a related PR landed within 1-2 days of this one (hard to tell for sure), appears to have caused a ~100% regression in the Latin1 segmenter performance benchmarks, and a smaller regression in the UTF-8 and UTF-16 performance benchmarks

https://unicode-org.github.io/icu4x/benchmarks/perf/components/segmenter/index.html

PR #7938, opened 2026-05-05
Validate CLDR ethiopic/ethiopic-amete-alem agreement

### LINKED ISSUE #7925: Validate that `ethiopic-amete-alem` data matches `ethiopic` data
We model both Ethiopic calendars as a single type, but CLDR has distinct data for Amete Mihret and Amete Alem. Currently SourceDataProvider uses the `-ethiopic` data and doesn't even look at `ethiopic-amete-alem` – we should verify that they match and report potential mismatches to CLDR.

When reading `ethiopic-amete-alem` files in SourceDataProvider, they have to be added to `globs.rs`, i.e partially revert https://github.com/unicode-org/icu4x/pull/7906/changes/8862cd2970bcba88e7118b00614327a9fd89d7af.

robertbastian:
Fixes #7925

```
ethiopic/ethiopic-amete-alem data mismatch: ff-Adlm
ethiopic/ethiopic-amete-alem data mismatch: ff-Adlm-BF
ethiopic/ethiopic-amete-alem data mismatch: ff-Adlm-CM
ethiopic/ethiopic-amete-alem data mismatch: ff-Adlm-GH
ethiopic/ethiopic-amete-alem data mismatch: ff-Adlm-GM
ethiopic/ethiopic-amete-alem data mismatch: ff-Adlm-GW
ethiopic/ethiopic-amete-alem data mismatch: ff-Adlm-LR
ethiopic/ethiopic-amete-alem data mismatch: ff-Adlm-MR
ethiopic/ethiopic-amete-alem data mismatch: ff-Adlm-NE
ethiopic/ethiopic-amete-alem data mismatch: ff-Adlm-NG
ethiopic/ethiopic-amete-alem data mismatch: ff-Adlm-SL
ethiopic/ethiopic-amete-alem data mismatch: ff-Adlm-SN
ethiopic/ethiopic-amete-alem data mismatch: sc
```

## Changelog

N/A


provider/source/src/datetime/mod.rs
```diff
@@ -108,7 +108,59 @@ impl SourceDataProvider {
...
+                &resource
+                    .eras
+                    .as_ref()
+                    .map(|e| (e.abbr.get("0"), e.names.get("0"), e.narrow.get("0"))),
+                &resource.month_patterns,
+                &resource.months,
+                &resource.time_formats,
+                &resource.time_skeletons,
+            ) {
+                log::warn!("ethiopic/ethiopic-amete-alem data mismatch: {locale}");
```
Manishearth, line 159: question: do we currently warn? Should we be erroring here?

datagen already outputs a lot of warnings. I would like to reduce those to a useful size, which is not a goal here, but if something shouldn't happen and doesn't currently happen we should consider using `log::error` or erroring.

Either way, this is better than nothing, so approving.

Manishearth (approved):

provider/source/src/datetime/mod.rs
robertbastian: > Should we be erroring here?

do you mean `DataError`, or `log::error!`? I don't want datagen to fail because of this, which it would.

the path to reducing datagen warnings is to fix things in CLDR.

Manishearth: Either. `log::error` is probably better. But this is fine.

sffc (commented):
log::warn is the correct behavior here in my opinion

robertbastian:
https://unicode-org.atlassian.net/browse/CLDR-19499

PR #7940, opened 2026-05-05
Support sparse enums in zerovec::make_ule

Manishearth:
We currently ban sparse enums: enum discriminants cannot have "gaps" since zerovec validation uses fast range checking. But that's not actually necessary, we can easily compile more complex enums down to `match`es over ranges (which LLVM is pretty good at optimizing). It's not going to be _as_ fast, but zerovec validation is allowed to be that complex.

This also fixes a soundness issue found by @shnatsel (thanks!) where you could create sparse enums without zerovec realizing.

https://github.com/unicode-org/icu4x/security/advisories/GHSA-mx24-jf73-wh55

## Changelog

zerovec_derive: Support sparse enums in `zerovec::make_ule`.

### DIFF
--- a/utils/zerovec/derive/src/make_ule.rs
+++ b/utils/zerovec/derive/src/make_ule.rs
@@ -6,7 +6,7 @@ use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
 
 use crate::utils::{self, FieldInfo, ZeroVecAttrs};
-use std::collections::HashSet;
+
 use syn::spanned::Spanned;
 use syn::{parse_quote, Data, DataEnum, DataStruct, DeriveInput, Error, Expr, Fields, Ident, Lit};
 
@@ -96,14 +96,9 @@ fn make_ule_enum_impl(
             .to_compile_error();
     }
 
-    // the smallest discriminant seen
-    let mut min = None;
-    // the largest discriminant seen
-    let mut max = None;
-    // Discriminants that have not been found in series (we might find them later)
-    let mut not_found = HashSet::new();
+    let mut discriminants = std::collections::BTreeSet::new();
 
-    for (i, variant) in enu.variants.iter().enumerate() {
+    for variant in enu.variants.iter() {
         if !matches!(variant.fields, Fields::Unit) {
             // This can be supported in the future, see zerovec/design_doc.md
             return Error::new(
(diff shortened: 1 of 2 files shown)


utils/zerovec/derive/src/make_ule.rs
```diff
@@ -166,16 +135,71 @@ fn make_ule_enum_impl(
...
+            }
+        );
+        let new_from_u8_impl = quote!(
+            if (#min..=#max).contains(&value) {
+                Some(zerovec::ule::AsULE::from_unaligned(#ule_name(value)))
+            } else {
+                None
+            }
+        );
+        (validate_bytes_impl, new_from_u8_impl)
```
robertbastian: why special-case the 1 case?

utils/zerovec/derive/src/make_ule.rs
```diff
@@ -236,11 +256,7 @@ fn make_ule_enum_impl(
             /// Attempt to construct the value from its corresponding integer,
             /// returning `None` if not possible
             pub(crate) fn new_from_u8(value: u8) -> Option<Self> {
-                if value <= #max {
-                    Some(zerovec::ule::AsULE::from_unaligned(#ule_name(value)))
-                } else {
-                    None
-                }
+                #new_from_u8_impl
```
robertbastian:
```suggestion
                if zerovec::ule::AsULE::validate_bytes(&[value]).is_ok() {
                    Some(zerovec::ule::AsULE::from_unaligned(#ule_name(value)))
                } else {
                    None
                }
```

utils/zerovec/derive/src/make_ule.rs
```diff
@@ -166,16 +135,71 @@ fn make_ule_enum_impl(
...
+        let mut start = first;
+        let mut end = first;
+        for &d in iter {
+            if d != end + 1 {
+                ranges.push(start..=end);
+                start = d;
+            }
+            end = d;
+        }
+        ranges.push(start..=end);
```
robertbastian: I'd prefer to produce the trivial match statement a la

```rust
1 | 3 | 4 => {},
_ => return Err(zerovec::ule::UleError::parse::<Self>())
```

and let the compiler figure out the ranges

Manishearth: I was worried this wouldn't optimize right but that's silly of me, match statements optimize great

https://rust.godbolt.org/z/Y1bdejMxn

Manishearth:
Redid the PR from scratch.

robertbastian (approved):

Issue #7932, opened 2026-05-04
Remove `LineBreakStrictness::Anywhere`


robertbastian:
The `anywhere` [CSS value](https://drafts.csswg.org/css-text-3/#valdef-line-break-anywhere) requires breaks between "[typographic character units](https://drafts.csswg.org/css-text-3/#typographic-character-unit)". These are defined as tailored extended grapheme clusters, but the spec does not consider actual tailoring in-scope. Therefore, to achieve `line-break: anywhere`, a `GraphemeClusterSegmenter` should be used.

Our current implementation of this is incorrect, it breaks after every code point. This is because we don't have grapheme segmentation loaded in the `LineSegmenter`, and we shouldn't do that, as it's only used optionally.

robertbastian:
We actually do have grapheme clustering loaded, through `ComplexPayloads`

Issue #4771, opened 2024-04-03
`LineBreakStrictness::Anywhere` gives the wrong breakpoints for Arabic in `icu_segmenter`


asibahi:
Hello.

I am aware this is probably a unicode spec issue, rather than a rust library issue, but I thought I would point it out regardless.

This is the minimal application I was using to test this behavior:

```rust
use icu_segmenter::{LineBreakOptions, LineBreakStrictness, LineSegmenter};
fn main() {
    let test = "الخيل والليل";

    let mut options = LineBreakOptions::default();
    options.strictness = LineBreakStrictness::Anywhere;

    let segmenter = LineSegmenter::new_auto_with_options(options);

    let breakpoints = segmenter.segment_str(test);

    for bp in breakpoints {
        println!("{bp}:  {}", &test[bp..]);
    }
}
```

This gives the following output:

```
0:  الخيل والليل
2:  لخيل والليل
4:  خيل والليل
6:  يل والليل
8:  ل والليل
10:   والليل
11:  والليل
13:  الليل
15:  لليل
17:  ليل
19:  يل
21:  ل
23:
 ```

as you can tell, it is breaking after every single letter, without respect to the letters' connections. However, as I am sure you are aware, the letters' connections are *not* optional. 

The output I expected is the following:

```
0:  الخيل والليل
2:  لخيل والليل
10:   والليل
11:  والليل
13:  الليل
15:  لليل
23:
 ```

Putting the break points across the visual boundaries of the letters. This is not the current orthodoxy, but any looser breaks than that and you'd be rendering the text illegible and unnatural.

Note: This is how old written manuscripts break their words.

sffc:
@eggrobin

eggrobin:
> I am aware this is probably a unicode spec issue, rather than a rust library issue

If it is an issue, it is a spec issue.
If it were a Unicode spec issue, I could perhaps help; but it is not a Unicode spec issue, it is a W3C spec issue.
See https://drafts.csswg.org/css-text-3/#valdef-line-break-anywhere. LineBreakStrictness::Strict comes straight from the Unicode Line Breaking Algorithm, but Normal and Loose are tailorings defined by CSS, and Anywhere is a set of rules defined by CSS and unrelated to the Unicode Line Breaking Algorithm (specifically, it would not be a conformant tailoring of the Unicode Line Breaking Algorithm, since it does not apply some of the non-tailorable rules).

asibahi:
[...]

PR #7941, opened 2026-05-06
Use grapheme segmentation for `LineBreakStrictness::Anywhere`

robertbastian:
#4771 #7932 

## Changelog

icu_segmenter:
* Use grapheme segmentation for `LineBreakStrictness:Anywhere`

### DIFF
--- a/components/segmenter/tests/css_line_break.rs
+++ b/components/segmenter/tests/css_line_break.rs
@@ -8,6 +8,7 @@ use icu_segmenter::options::LineBreakStrictness;
 use icu_segmenter::options::LineBreakWordOption;
 use icu_segmenter::LineSegmenter;
 
+#[track_caller]
 fn check_with_options(
     s: &str,
     mut expect_utf8: Vec<usize>,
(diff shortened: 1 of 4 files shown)


components/segmenter/tests/css_line_break.rs
```diff
@@ -200,15 +205,22 @@ fn linebreak_loose() {
...
+    anywhere(
+        "الخيل والليل",
+        false,
+        vec![2, 4, 6, 8, 10, 11, 13, 15, 17, 19, 21, 23],
+        vec![1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12],
+    );
+
     // css/css-text/line-break/line-break-anywhere-001.html
     anywhere(
         "aa-a.a)a,a) a\u{00A0}aa\u{2060}a\u{200D}a･a",
```
robertbastian, line 217: `a\u{200D}` is an extended grapheme cluster

sffc (approved):
ok. If LineBreakAnywhere is the same as a grapheme cluster segmenter, I think we should remove LineBreakAnywhere completely when we can, and ask browsers to use a grapheme segmenter when that option is given in CSS.

PR #7942, opened 2026-05-06
Full API for neo `LineSegmenter`, complex handling, tests

robertbastian:
https://github.com/unicode-org/icu4x/issues/7962

## Changelog

icu_segmenter:
* add experimental `_neo_` constructors


components/segmenter/src/lib.rs
```diff
@@ -206,3 +206,9 @@ pub(crate) mod private {
     /// implemented outside of the segmenter crate.
     pub trait Sealed {}
 }
+
+#[cfg(feature = "unstable")]
+/// Experimental reimplementations
+pub mod neo {
```
sffc, line 212: I think probably just use the name "unstable"

```suggestion
pub mod unstable {
```

sffc (approved):
Seems plausible; if it passes the tests then the complex script handling is probably correct. We can take another look once everything is landed

components/segmenter/src/lib.rs
robertbastian: I think it's clearer to have `neo::LineSegmenter` and `LineSegmenter` rather than `unstable::LineSegmenter` and `LineSegmenter`.

sffc: or `unstable::LineSegmenterNeo`

components/segmenter/src/line/neo.rs
```diff
@@ -2,75 +2,640 @@
...
 
-            if self
-                .data
-                .complex_classes
-                .iter()
-                .find(|&c| c == class)
-                .is_some()
-            {
-                // TODO
+            if Y::CAN_CONTAIN_SA && self.data.is_class_complex(class) {
```
Manishearth: not quite sure what's going on here but better looked at during a holistic review

Manishearth (approved):
(same reaction as shane)

appreciate the additional tests

Manishearth (approved):

Manishearth (approved):
CI seems sleepy

sffc (approved):

Manishearth (approved):

PR #7945, opened 2026-05-07
Fix weekday skeleton matching for Finnish (standalone weekday formats)

### LINKED ISSUE #5892: Fix weekday classical skeleton matching in datagen
See https://github.com/unicode-org/icu4x/pull/5826#issuecomment-2505702078

In the Finnish locale, we are resolving semantic skeletons to a pattern with `EEE` even though they should more likely be `ccc`. This is due to a potential limitation in the classical skeleton selection code.

Manishearth:
Fixes https://github.com/unicode-org/icu4x/issues/5892

This took a fair amount of time to debug, because the data I was looking at (latest CLDR) did not match the data ICU4X saw (CLDR 48.2). This had changed in https://github.com/unicode-org/cldr/pull/5524: CLDR no longer uses `c` in skeletons. ICU4X was ignoring `c`.

The main change here is to transform `c` to `E` in datagen (which is what https://github.com/unicode-org/cldr/pull/5524 does too), and then warn if we ever see both.

This will likely be fixed entirely by a future CLDR upgrade. It's questionable whether the code in this PR is necessary; if people think we should just wait for the release we can close this PR for now and resurrect its tests when we upgrade CLDR. At least there will be breadcrumbs for anyone looking at #5892 in the future. 

## Changelog

icu_datetime: Correctly handle CLDR 48.2 data with "c" in the skeletons.


components/datetime/src/provider/skeleton/reference.rs
```diff
@@ -162,7 +162,17 @@ impl TryFrom<&str> for Skeleton {
...
+            let mut field = Field::from((field_symbol, FieldLength::from_idx(field_length)?));
+
+            // CLDR skeletons in 48.2 use 'c' instead of 'E' (e.g. in Finnish).
+            // We treat them as equivalent for matching purposes by normalizing 'c' to 'E' here.
+            // This avoids missing matches when we request 'E' via synthesis in datagen.
+            //
+            // This will be obsoleted by future CLDR releases
+            // <https://github.com/unicode-org/cldr/pull/5524>
+            if let FieldSymbol::Weekday(fields::Weekday::StandAlone) = field.symbol {
+                field.symbol = FieldSymbol::Weekday(fields::Weekday::Format);
```
sffc, line 174: Nit: Add a `log::warn` here

```suggestion
                log::warn!("Found pattern-only weekday field 'c' in skeleton '{skeleton_string}' (see #5892)");
                field.symbol = FieldSymbol::Weekday(fields::Weekday::Format);
```

sffc (approved):

sffc:
it appears my suggestion didn't compile

PR #7949, opened 2026-05-08
Neo segmenter: use updated data files, generalise break status

robertbastian:
## Changelog

N/A


components/segmenter/src/line/neo.rs
```diff
@@ -695,17 +691,14 @@ impl<'s, Y: RuleBreakType> Iterator for LineBreakIterator<'_, 's, Y> {
...
-
         Some(break_index)
     }
 }
 
 impl<Y: RuleBreakType> LineBreakIterator<'_, '_, Y> {
     /// Returns whether the last break was mandatory
     pub fn is_mandatory(&self) -> bool {
-        self.last_accepting_mandatory
+        self.last_accepting_status == 1 || Y::is_empty(&self.remaining_input)
```
eggrobin, line 701: You might want to generate some constants (or even an enumeration type) based on the names in the States file, having to tie the 1 here to the 1 in 
https://github.com/unicode-org/icu4x/pull/7949/changes#diff-81725aab662eaea321904f966e93134e3db19a4f277e9a87f944214857e46742R914
makes this a bit hard to read (and would be brittle for word where there is more than one status and where that number is not set in stone—ICU has one more than ICU4X).

components/segmenter/src/provider/neo.rs
```diff
@@ -15,30 +15,27 @@ pub type Lookahead = u8;
 #[cfg_attr(feature = "datagen", derive(serde::Serialize, databake::Bake))]
 #[cfg_attr(feature = "datagen", databake(path = icu_segmenter::provider))]
 pub enum Acceptance {
-    Accept,
     Continue,
```
eggrobin, line 18: This seems like a somewhat unorthodox name, not that there is an orthodox name other than non-accepting state.

In the context of a lexer, it is true of both accepting and non-accepting states that you continue transitioning as long as you can, so it’s not a very descriptive name either.

Since we’re in an enum called Acceptance, I think I’d just call it None (Acceptance::None seems like a fairly transparent way of saying non-accepting).

components/segmenter/src/line/neo.rs
robertbastian: for word we have an existing enum which I'll reuse (I already have a branch with this)

for line I could define `pub enum Mandatory { No, Yes }`, but that seems overkill because the data struct itself does not use the type anyway, so we always have to go through `u8`

we cannot generate this type, as we don't generate API from data

components/segmenter/src/provider/neo.rs
robertbastian: let's bikeshed names later

eggrobin (approved):

PR #7952, opened 2026-05-08
Remove colon word break tailoring

robertbastian:
This was removed from ICU in https://unicode-org.atlassian.net/browse/ICU-22941, and from CLDR in https://github.com/unicode-org/cldr/pull/3973

## Changelog

icu_segmenter:
* Align word segmenter behaviour with ICU4C and UAX#29


sffc (changes requested):
From my reading of https://docs.google.com/document/d/1AOCdt1J8kwF_nMd-fkTfpv--NeRMK90iz8ipQLnKxu4/edit?tab=t.0#heading=h.drcymlu7ubpt, the currently implemented behavior is how we make the colon tailoring available until we have segmentation profiles, which we don't have yet.

sffc:
Isn't it still there in CLDR?

https://github.com/macchiati/cldr/blob/main/common/segments/fi.xml

sffc:
That's a stale branch. Here's a CLDR PR that actually deleted the files, ok

https://github.com/unicode-org/cldr/pull/4326

I'm still uncomfortable with this because I specifically remember deciding as a design group _to retain_ the special behavior, just that locale tailorings are probably not the right place, but they were okay until we have a better place to put them. The design doc linked above seems to align with this impression.

robertbastian:
This PR aims to reduce the diff between the current segmenter and the neo segmenter. With neo it will not be possible to keep this non-standard behaviour in the untailored segmenter.

eggrobin:
Well, we decided to remove it from the default in CLDR and ICU—and we did so—, and to add an option (which is not based on the locale, and not the default) to add it at a later date.

And, two years on, nobody seems interested enough in the option to actually add it.

In any case, it seems strange that the default here would be what UTC, ICU-TC, and CLDR-TC all decided shouldn’t be the default.

components/segmenter/src/word.rs
```diff
@@ -22,36 +22,6 @@ pub struct WordBreakOptions<'a> {
...
-    /// let mut options = WordBreakOptions::default();
-    /// let langid = &langid!("fi");
-    /// options.content_locale = Some(langid);
-    /// let segmenter = WordSegmenter::try_new_auto(options).unwrap();
-    ///
-    /// let breakpoints: Vec<usize> =
-    ///     segmenter.as_borrowed().segment_str("EU:ssa").collect();
-    /// assert_eq!(&breakpoints, &[0, 6]);
-    /// ```
     pub content_locale: Option<&'a LanguageIdentifier>,
```
sffc, line 25: Nit: please add a content_locale docs test to Sentence so that we keep content_locale being tested and demonstrated in docs

sffc (approved):

Issue #7957, opened 2026-05-11
Enable more features of ZeroFrom and maybe other crates on playground


sffc:
The following example does not compile on the Rust playground, even though it should

```rust
use zerofrom::ZeroFrom;

fn main() {
    let string: String = "abc".to_string();
    let _: &str = ZeroFrom::zero_from(&string);
}
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d5adac89c6aca956aa3d978e585b42f4

The root cause must be because the "alloc" feature is not enabled on the playground.

There is an FAQ about how to force more features to be enabled:

https://github.com/rust-lang/rust-playground/blob/main/CRATE_POLICY.md#i-am-a-crate-author-and-i-want-to-control-which-features-are-available

We should do this for ZeroFrom and maybe other crates.

CC @Manishearth

sffc:
I did this for zerofrom. https://github.com/unicode-org/icu4x/pull/7958

@Manishearth suggested we consider doing it for all the other crates. We already have similar metadata for docs.rs so this seems like it would be a harmless change.

Manishearth:
We should figure out which features we care about here. `all-features = true` is probably fine for most crates.

PR #7958, opened 2026-05-11
Enable all features of zerofrom in playground; bump version

sffc:
#7957

## Changelog

zerofrom: enable all features in playground

### DIFF
--- a/utils/zerofrom/Cargo.toml
+++ b/utils/zerofrom/Cargo.toml
@@ -5,11 +5,11 @@
 [package]
 name = "zerofrom"
 description = "ZeroFrom trait for constructing"
-version = "0.1.7"
-authors = ["Manish Goregaokar <manishsmail@gmail.com>"]
+version = "0.1.8"
 categories = ["data-structures", "caching", "no-std"]
 keywords = ["zero-copy", "serialization", "memory-efficiency"]
 
+authors.workspace = true
 edition = "2021"
 include.workspace = true
 license.workspace = true
@@ -20,6 +20,9 @@ rust-version = "1.71.1"
 independent = true
 
 [package.metadata.docs.rs]
+all-features = true
+
+[package.metadata.playground]
 all-features = true
 
 [features]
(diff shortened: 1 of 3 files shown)


sffc:
I would also like permission to push 0.1.8 of zerofrom, from the main branch. There are no other diffs, except for explicitly setting edition = "2021" instead of inheriting it from the workspace.

utils/zerofrom/Cargo.toml
```diff
@@ -5,11 +5,11 @@
 [package]
 name = "zerofrom"
 description = "ZeroFrom trait for constructing"
-version = "0.1.7"
-authors = ["Manish Goregaokar <manishsmail@gmail.com>"]
+version = "0.1.8"
```
Manishearth, line 8: thought: I'm still not fully clear on our workflow for updating versions in Cargo.toml.

https://github.com/unicode-org/icu4x/blob/main/documents/process/release.md has a full process that does not require crates to have been pre-updated.

The problem with pre-updating crates is that the next time someone touches this crate they have to check if it's already preupdated.

I would love for us to have a written down process for what we do mid-cycle here. Maybe we should be using `-dev` more. I don't know.

I think we started doing this a few years ago and have done it inconsistently in a way that doesn't end up being useful to release drivers.

This is not a blocking comment, just a thought.

Manishearth (approved):
LG

Should we just do this for everyone?

sffc:
> Should we just do this for everyone?

Maybe. This can be tracked in the issue. I just know I hit this issue with ZeroFrom so am doing the narrow fix for now.

sffc:
Still waiting on the playground update. 0.1.8 is published on crates.io:

https://crates.io/crates/zerofrom

But the playground still has 0.1.7: 

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=db3908ccb49667fc14554cc599741a5a

Manishearth:
Seems to be a semi-manual process: https://github.com/rust-lang/rust-playground/commits/main/compiler/base/Cargo.toml. I'd just wait.

PR #7960, opened 2026-05-12
Update neo line segmenter to Unicode 17

robertbastian:
## Changelog

N/A


eggrobin:
Most of these paragraphs in the header of LineBreakRandomTest.txt are obsolete:
```
# They do *not* cover all bugs that were found as part of https://github.com/unicode-org/icu4x/pull/5218,
# as the last bug encountered there was in test case #8208.
# When actually working on changes to the rules, much longer runs should be used.
#
# Note that starting with Unicode 16.0 and ICU 76, ICU’s rule-based line
# breaking implementation will no longer be tailored, so there should be no need
# to generate these tests from a fork going forward.
```

Just keep this:
```
# When actually working on changes to the rules, much longer runs should be used.
```

(although even that is a bit belt-and-suspenders at this point, since the same state machine will have been tested against a naïve implementation on millions of strings.)

sffc (approved):

sffc:
Praise: Good to have this in its own PR

Issue #487, opened 2021-02-09
Implement flexible day periods


sffc:
Flexible day periods, represented as `B` in UTS-35 datetime formats, represent periods of the day such as "in the morning", "in the afternoon", and so forth.  These day periods are locale-dependent, and they may take into account the latitude/longitude of the user as well.

sffc:
Here's an MDN example illustrating that this is in 402: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat#using_options

Manishearth:
Triage: correctly triaged as 2.x priority, it is an ecma402 issue

Manishearth:
Current status: we already have some day period names stored, see https://unicode-org.github.io/icu4x/rustdoc/icu/datetime/provider/names/struct.DayPeriodNamesV1.html, which stores AM, PM, `(noon)`, `(midnight)`. We could include `morning1`, `morning2`, `afternoon1`, `afternoon2`, `evening1`, `night1` if we desired.

We'd also need to load the `dayPeriodRules` from`dayPeriods.xml` and store them in a new marker.

We _probably_ want to only datagen day period rules for locales that actually use patterns with flexible day periods. Unless there is a way that someone can force flexible day periods via fieldset? (@sffc? Is it possible with the current API to e.g. force an English locale formatter to render `"morning"` (assuming we had day period support). Are there other users for this data?

sffc:
My mental model is that the flexible day periods should be stored in a new data marker. But, we could store them in DayPeriodNames. What we do now is, if the list is length 2, then it only contains AM and PM, and if it length 4, then it contains AM, PM, Noon, Midnight. We could potentially say that if it is longer, then the fifth element is morning1, the sixth is morning2, etc. I hadn't considered that design.

Currently, flexible day periods are only used when a datetimeSkeleton references them, like in zh-Hant. When we add field-specific length hints, then we can add more ways to request them. I don't think we will ever have a "force flexible day periods" option, because some locales simply don't support them.

Manishearth:
> My mental model is that the flexible day periods should be stored in a new data marker. But, we could store them in DayPeriodNames. 

[...]

PR #7964, opened 2026-05-12
Add DayPeriodRules data model and data

Manishearth:
Progress on https://github.com/unicode-org/icu4x/issues/487

This is just the rules data, not the symbols, which in the current design will be stored in the same symbols struct.

## Changelog: N/A

Changelog will be written on final PR.

### DIFF
--- a/components/datetime/src/provider/day_periods.rs
+++ b/components/datetime/src/provider/day_periods.rs
@@ -1,0 +1,263 @@
+// This file is part of ICU4X. For terms of use, please see the file
+// called LICENSE at the top level of the ICU4X source tree
+// (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
+//! Data structs and markers for day periods rules.
+
+use icu_provider::prelude::*;
+
+/// Day period rules representing the active period for each hour of the day.
+///
+/// Lookup assumes that at hour 0, the active period is the *last* present period in the
+/// sequence of present periods (sorted by enum value). This is because night periods
+/// often spill over past midnight (e.g., from 21:00 to 06:00), meaning hour 0 is
+/// typically covered by the last chronological period, which is usually a night period.
+/// Transitions move to the next present period in the sequence, wrapping around.
+#[derive(Debug, PartialEq, Clone, Copy, yoke::Yokeable, zerofrom::ZeroFrom)]
+#[cfg_attr(feature = "datagen", derive(serde::Serialize, databake::Bake))]
+#[cfg_attr(feature = "datagen", databake(path = icu_datetime::provider::day_periods))]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize))]
+#[yoke(prove_covariance_manually)]
+pub struct DayPeriodRules {
+    /// A bitmask of present day periods. Bit `i` is set if the period with
+    /// enum value `i` is present in the locale's rules.
+    pub presence: u8,
+    /// A 24-bit map (packed into 3 bytes) where bit `h` is set if a transition
+    /// to the next present period occurs at hour `h`.
+    pub transitions: [u8; 3],
+}
+
+icu_provider::data_struct!(
+    DayPeriodRules,
+    #[cfg(feature = "datagen")]
+);
+
+icu_provider::data_marker!(
+    /// `DayPeriodRulesV1` marker
+    DayPeriodRulesV1,
+    DayPeriodRules,
+);
+
+/// Day periods.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+#[cfg_attr(feature = "datagen", derive(serde::Serialize, databake::Bake))]
+#[cfg_attr(feature = "datagen", databake(path = icu_datetime::provider::day_periods))]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize))]
+pub enum DayPeriod {
+    /// Morning 1
+    Morning1 = 0,
+    /// Morning 2
+    Morning2 = 1,
+    /// Afternoon 1
+    Afternoon1 = 2,
+    /// Afternoon 2
+    Afternoon2 = 3,
+    /// Evening 1
+    Evening1 = 4,
+    /// Evening 2
+    Evening2 = 5,
+    /// Night 1
+    Night1 = 6,
+    /// Night 2
+    Night2 = 7,
+}
+
+impl DayPeriod {
+    /// Parses a CLDR day period name into a `DayPeriod`.
+    pub fn from_cldr_name(name: &str) -> Option<Self> {
+        match name {
+            "morning1" => Some(DayPeriod::Morning1),
+            "morning2" => Some(DayPeriod::Morning2),
+            "afternoon1" => Some(DayPeriod::Afternoon1),
+            "afternoon2" => Some(DayPeriod::Afternoon2),
+            "evening1" => Some(DayPeriod::Evening1),
+            "evening2" => Some(DayPeriod::Evening2),
+            "night1" => Some(DayPeriod::Night1),
+            "night2" => Some(DayPeriod::Night2),
+            _ => None,
+        }
+    }
+
+    /// Converts a u8 index to a `DayPeriod` enum.
+    pub fn from_u8(v: u8) -> Option<Self> {
+        match v {
+            0 => Some(DayPeriod::Morning1),
+            1 => Some(DayPeriod::Morning2),
+            2 => Some(DayPeriod::Afternoon1),
+            3 => Some(DayPeriod::Afternoon2),
+            4 => Some(DayPeriod::Evening1),
+            5 => Some(DayPeriod::Evening2),
+            6 => Some(DayPeriod::Night1),
+            7 => Some(DayPeriod::Night2),
+            _ => None,
+        }
+    }
+}
+
+impl DayPeriodRules {
+    /// Looks up the day period for a given hour (0-24).
+    pub fn lookup(&self, hour: u8) -> DayPeriod {
+        // Make the lookup cyclic and handle hour 24 (as hour 0 of next day).
+        let hour = hour % 24;
+
+        // GIGO: If presence is 0 (bad data), we return Morning1 as a safe fallback.
+        if self.presence == 0 {
+            debug_assert!(false, "presence must not be zero");
+            return DayPeriod::Morning1;
+        }
+
+        let count = self.presence.count_ones() as usize;
+
+        // Combine transitions bytes into a u32 for easier bit manipulation.
+        let transitions_u32 = u32::from_le_bytes([
+            self.transitions[0],
+            self.transitions[1],
+            self.transitions[2],
+            0,
+        ]);
+
+        // Create a mask for bits 0..=hour.
+        // For hour 0, mask is 0x1 (bit 0).
+        // For hour 23, mask is 0x00FF_FFFF (bits 0..23).
+        // Since hour < 24, (hour + 1) <= 24, so 1u32 << (hour + 1) fits in u32 without overflow.
+        let mask = (1u32 << (hour + 1)) - 1;
+        // Count how many transitions occurred up to `hour`.
+        let transitions = (transitions_u32 & mask).count_ones() as usize;
+
+        // Assume first period (at hour 0) is the last present period in the sorted sequence.
+        // Adding `transitions` moves us forward in the sequence.
+        // We subtract 1 and modulo `count` to start from the last period (index count - 1)
+        // and wrap around correctly.
+        let which_period = (transitions + count - 1) % count;
+
+        // Find the which_period-th set bit in presence (0-indexed).
+        let mut period_idx = 0u8;
+        let mut found_count = 0;
+        for i in 0..8 {
+            if (self.presence & (1 << i)) != 0 {
+                if found_count == which_period {
+                    period_idx = i;
+                    break;
+                }
+                found_count += 1;
+            }
+        }
+
+        if let Some(period) = DayPeriod::from_u8(period_idx) {
+            period
+        } else {
+            // This should be unreachable by construction as which_period < count.
+            debug_assert!(false, "Unreachable day period index: {}", period_idx);
+            DayPeriod::Morning1 // Fallback
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_lookup() {
+        // Create a dummy DayPeriodRules
+        // presence: Morning1 (0), Afternoon1 (2), Night1 (6)
+        // presence = (1<<0) | (1<<2) | (1<<6) = 1 | 4 | 64 = 69
+        // transitions: transitions at 6 (Morning1), 12 (Afternoon1), 18 (Night1)
+        // 6 is bit 6 of byte 0.
+        // 12 is bit 4 of byte 1.
+        // 18 is bit 2 of byte 2.
+
+        let mut transitions = [0u8; 3];
+        transitions[0] |= 1 << 6;
+        transitions[1] |= 1 << 4;
+        transitions[2] |= 1 << 2;
+
+        let rules = DayPeriodRules {
+            presence: 69,
+            transitions,
+        };
+
+        // Assume first period is Night1 (6) because it's the last night period.
+        // At hour 0, no transition, so it should be Night1.
+        assert_eq!(rules.lookup(0), DayPeriod::Night1);
+        assert_eq!(rules.lookup(5), DayPeriod::Night1);
+
+        // At hour 6, transition to Morning1 (0).
+        assert_eq!(rules.lookup(6), DayPeriod::Morning1);
+        assert_eq!(rules.lookup(11), DayPeriod::Morning1);
+
+        // At hour 12, transition to Afternoon1 (2).
+        assert_eq!(rules.lookup(12), DayPeriod::Afternoon1);
+        assert_eq!(rules.lookup(17), DayPeriod::Afternoon1);
+
+        // At hour 18, transition to Night1 (6).
+        assert_eq!(rules.lookup(18), DayPeriod::Night1);
+        assert_eq!(rules.lookup(23), DayPeriod::Night1);
+
+        // Hour 24 should be same as hour 0.
+        assert_eq!(rules.lookup(24), DayPeriod::Night1);
+    }
+
+    #[cfg(feature = "compiled_data")]
+    #[test]
+    fn test_compiled_data() {
+        use icu_provider::prelude::*;
+        let provider = crate::provider::Baked;
+        let rules: DataPayload<DayPeriodRulesV1> = provider
+            .load(DataRequest {
+                id: DataIdentifierBorrowed::for_locale(&icu_locale::langid!("en").into()),
+                ..Default::default()
+            })
+            .unwrap()
+            .payload;
+
+        assert_eq!(rules.get().lookup(0), DayPeriod::Morning1);
+        assert_eq!(rules.get().lookup(11), DayPeriod::Morning1);
+        assert_eq!(rules.get().lookup(12), DayPeriod::Afternoon1);
+        assert_eq!(rules.get().lookup(17), DayPeriod::Afternoon1);
+        assert_eq!(rules.get().lookup(18), DayPeriod::Evening1);
+        assert_eq!(rules.get().lookup(20), DayPeriod::Evening1);
+        assert_eq!(rules.get().lookup(21), DayPeriod::Night1);
+        assert_eq!(rules.get().lookup(23), DayPeriod::Night1);
+
+        // Test 'zh' (Chinese) rules
+        let rules_zh: DataPayload<DayPeriodRulesV1> = provider
+            .load(DataRequest {
+                id: DataIdentifierBorrowed::for_locale(&icu_locale::langid!("zh").into()),
+                ..Default::default()
+            })
+            .unwrap()
+            .payload;
+        assert_eq!(rules_zh.get().lookup(0), DayPeriod::Night1);
+        assert_eq!(rules_zh.get().lookup(4), DayPeriod::Night1);
+        assert_eq!(rules_zh.get().lookup(5), DayPeriod::Morning1);
+        assert_eq!(rules_zh.get().lookup(7), DayPeriod::Morning1);
+        assert_eq!(rules_zh.get().lookup(8), DayPeriod::Morning2);
+        assert_eq!(rules_zh.get().lookup(11), DayPeriod::Morning2);
+        assert_eq!(rules_zh.get().lookup(12), DayPeriod::Afternoon1);
+        assert_eq!(rules_zh.get().lookup(13), DayPeriod::Afternoon2);
+        assert_eq!(rules_zh.get().lookup(18), DayPeriod::Afternoon2);
+        assert_eq!(rules_zh.get().lookup(19), DayPeriod::Evening1);
+        assert_eq!(rules_zh.get().lookup(23), DayPeriod::Evening1);
+
+        // Test 'de' (German) rules
+        let rules_de: DataPayload<DayPeriodRulesV1> = provider
+            .load(DataRequest {
+                id: DataIdentifierBorrowed::for_locale(&icu_locale::langid!("de").into()),
+                ..Default::default()
+            })
+            .unwrap()
+            .payload;
+        assert_eq!(rules_de.get().lookup(0), DayPeriod::Night1);
+        assert_eq!(rules_de.get().lookup(4), DayPeriod::Night1);
+        assert_eq!(rules_de.get().lookup(5), DayPeriod::Morning1);
+        assert_eq!(rules_de.get().lookup(9), DayPeriod::Morning1);
+        assert_eq!(rules_de.get().lookup(10), DayPeriod::Morning2);
+        assert_eq!(rules_de.get().lookup(11), DayPeriod::Morning2);
+        assert_eq!(rules_de.get().lookup(12), DayPeriod::Afternoon1);
+        assert_eq!(rules_de.get().lookup(13), DayPeriod::Afternoon2);
+        assert_eq!(rules_de.get().lookup(17), DayPeriod::Afternoon2);
+        assert_eq!(rules_de.get().lookup(18), DayPeriod::Evening1);
+        assert_eq!(rules_de.get().lookup(23), DayPeriod::Evening1);
+    }
+}
--- a/components/datetime/src/provider/day_periods.rs
+++ b/components/datetime/src/provider/day_periods.rs
@@ -89,11 +89,9 @@ impl DayPeriod {
 }
 
 impl DayPeriodRules {
-    /// Looks up the day period for a given hour (0-24).
+    /// Looks up the day period for a given hour (0-23).
     pub fn lookup(&self, hour: u8) -> DayPeriod {
-        debug_assert!(hour <= 24, "hour must be in 0..=24");
-        // Make the lookup cyclic and handle hour 24 (as hour 0 of next day).
-        let hour = hour % 24;
+        debug_assert!(hour < 24, "hour must be in 0..24");
 
         // GIGO: If presence is 0 (bad data), we return Morning1 as a safe fallback.
         if self.presence == 0 {
--- a/components/datetime/src/provider/day_periods.rs
+++ b/components/datetime/src/provider/day_periods.rs
@@ -165,6 +165,11 @@ impl DayPeriodRules {
     ///
     /// Entries is a map from `(start_hour, end_hour)` tuple ranges to `DayPeriod`.
     /// Returns `None` if entries is empty, or if there are overlaps or gaps in 24-hour coverage.
+    #[allow(
+        clippy::indexing_slicing,
+        clippy::expect_used,
+        reason = "Datagen is allowed to panic"
+    )]
     pub fn from_cldr_rules(
         entries: &alloc::collections::BTreeMap<(u8, u8), DayPeriod>,
     ) -> Option<Self> {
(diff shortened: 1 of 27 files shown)


provider/source/src/datetime/day_periods.rs
```diff
@@ -0,0 +1,268 @@
...
+            .supplemental
+            .day_period_rule_set
+            .0
+            .get(&langid.to_string())
+            .ok_or_else(|| {
+                DataErrorKind::IdentifierNotFound
+                    .with_req(<DayPeriodRulesV1 as DataMarker>::INFO, req)
+            })?;
+
+        let data = compute_day_periods(rules, &req.id.locale.to_string()).ok_or_else(|| {
```
Manishearth, line 37: thought: I'm not fully sure what we should be doing in the None case. I think we should only generate data for day periods that have data. But this also returns None when there are no flex day periods.

In theory the data could do something like `zh: am, pm, noon, midnight, morning1, ...` and `zh-Hant: am, pm, noon, midnight` and the flex periods are missing, and in that case you might not actually want fallback to be triggered. Right now the data doesn't seem to do anything like that.

components/datetime/src/provider/day_periods.rs
```diff
@@ -0,0 +1,263 @@
...
+
+use icu_provider::prelude::*;
+
+/// Day period rules representing the active period for each hour of the day.
+///
+/// Lookup assumes that at hour 0, the active period is the *last* present period in the
+/// sequence of present periods (sorted by enum value). This is because night periods
+/// often spill over past midnight (e.g., from 21:00 to 06:00), meaning hour 0 is
+/// typically covered by the last chronological period, which is usually a night period.
+/// Transitions move to the next present period in the sequence, wrapping around.
```
robertbastian, line 15: would it make sense to have the night period have the lowest discriminant? days always start with night, so it'd be more "regular" if we didn't have to look at the last period to get the first period (but instead wrap around at the end)

components/datetime/src/provider/day_periods.rs
```diff
@@ -0,0 +1,263 @@
...
+#[derive(Debug, PartialEq, Clone, Copy, yoke::Yokeable, zerofrom::ZeroFrom)]
+#[cfg_attr(feature = "datagen", derive(serde::Serialize, databake::Bake))]
+#[cfg_attr(feature = "datagen", databake(path = icu_datetime::provider::day_periods))]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize))]
+#[yoke(prove_covariance_manually)]
+pub struct DayPeriodRules {
+    /// A bitmask of present day periods. Bit `i` is set if the period with
+    /// enum value `i` is present in the locale's rules.
```
robertbastian: there's no "locale" here

```suggestion
    /// enum value `i` is present.
```

components/datetime/src/provider/day_periods.rs
```diff
@@ -0,0 +1,263 @@
...
+    Evening2 = 5,
+    /// Night 1
+    Night1 = 6,
+    /// Night 2
+    Night2 = 7,
+}
+
+impl DayPeriod {
+    /// Parses a CLDR day period name into a `DayPeriod`.
+    pub fn from_cldr_name(name: &str) -> Option<Self> {
```
robertbastian, line 61: cfg datagen for anything involving CLDR identifiers please

components/datetime/src/provider/day_periods.rs
```diff
@@ -0,0 +1,263 @@
...
+            _ => None,
+        }
+    }
+}
+
+impl DayPeriodRules {
+    /// Looks up the day period for a given hour (0-24).
+    pub fn lookup(&self, hour: u8) -> DayPeriod {
+        // Make the lookup cyclic and handle hour 24 (as hour 0 of next day).
+        let hour = hour % 24;
```
robertbastian: modulo operations are comparably expensive (compared to the bit shifts that you are being done here). I would make it a precondition that hour is in 0..=24, our calling code will be able to guarantee that.

you could also use `icu::time::types::Hour` as the input, which is validated to be 0..=24, and I think is what we have internally already

components/datetime/src/provider/day_periods.rs
```diff
@@ -0,0 +1,263 @@
...
+
+        // GIGO: If presence is 0 (bad data), we return Morning1 as a safe fallback.
```
robertbastian, line 96: question: why is this bad data? I haven't seen an invariant that there's a minimum number of periods anywhere

should `presence` be `NonZeroU8` if this is an invariant? this is the niche

components/datetime/src/provider/day_periods.rs
```diff
@@ -0,0 +1,263 @@
...
+        // Since hour < 24, (hour + 1) <= 24, so 1u32 << (hour + 1) fits in u32 without overflow.
+        let mask = (1u32 << (hour + 1)) - 1;
+        // Count how many transitions occurred up to `hour`.
+        let transitions = (transitions_u32 & mask).count_ones() as usize;
+
+        // Assume first period (at hour 0) is the last present period in the sorted sequence.
+        // Adding `transitions` moves us forward in the sequence.
+        // We subtract 1 and modulo `count` to start from the last period (index count - 1)
+        // and wrap around correctly.
+        let which_period = (transitions + count - 1) % count;
```
robertbastian: using a modulo operation because we might be 1 outside the range is overkill. this can be a conditional addition, `+ (transition + count == 0) as usize`

components/datetime/src/provider/day_periods.rs
```diff
@@ -0,0 +1,263 @@
...
+            0,
+        ]);
+
+        // Create a mask for bits 0..=hour.
+        // For hour 0, mask is 0x1 (bit 0).
+        // For hour 23, mask is 0x00FF_FFFF (bits 0..23).
+        // Since hour < 24, (hour + 1) <= 24, so 1u32 << (hour + 1) fits in u32 without overflow.
+        let mask = (1u32 << (hour + 1)) - 1;
+        // Count how many transitions occurred up to `hour`.
+        let transitions = (transitions_u32 & mask).count_ones() as usize;
```
robertbastian: nit: I feel this could be more clearly expressed with count_one of a shifted number, without applying a mask. with the mask you're removing the top `32 - hour` bits, but instead you could just shift them out before counting. I think `(transition << (32 - hour)).count_ones()` works, or `transition.count_ones() - (transition >> hour).count_ones()`

components/datetime/src/provider/day_periods.rs
```diff
@@ -0,0 +1,263 @@
...
+        // Assume first period (at hour 0) is the last present period in the sorted sequence.
+        // Adding `transitions` moves us forward in the sequence.
+        // We subtract 1 and modulo `count` to start from the last period (index count - 1)
+        // and wrap around correctly.
+        let which_period = (transitions + count - 1) % count;
+
+        // Find the which_period-th set bit in presence (0-indexed).
+        let mut period_idx = 0u8;
+        let mut found_count = 0;
+        for i in 0..8 {
```
robertbastian: thought: I'd like to avoid this magic 8, but it's basically `core::mem::variant_count` which is unstable

components/datetime/src/provider/day_periods.rs
```diff
@@ -0,0 +1,263 @@
...
+            if (self.presence & (1 << i)) != 0 {
+                if found_count == which_period {
```
robertbastian: `which_period` sounds like we're still looking for it. this is the selected, or target period, right?

components/datetime/src/provider/day_periods.rs
robertbastian: issue: if you don't hit the break, period_idx stays 0, which is incorrect. I'd prefer if you wrote this as a `loop` where the `break` returns a value, and showi that it must terminate

components/datetime/src/provider/day_periods.rs
```diff
@@ -0,0 +1,263 @@
...
+            "evening1" => Some(DayPeriod::Evening1),
+            "evening2" => Some(DayPeriod::Evening2),
+            "night1" => Some(DayPeriod::Night1),
+            "night2" => Some(DayPeriod::Night2),
+            _ => None,
+        }
+    }
+
+    /// Converts a u8 index to a `DayPeriod` enum.
+    pub fn from_u8(v: u8) -> Option<Self> {
```
robertbastian: can be private?

components/datetime/src/provider/day_periods.rs
```diff
@@ -0,0 +1,263 @@
...
+                    break;
+                }
+                found_count += 1;
+            }
+        }
+
+        if let Some(period) = DayPeriod::from_u8(period_idx) {
+            period
+        } else {
+            // This should be unreachable by construction as which_period < count.
```
robertbastian: I don't think this is the correct justification. It's unreachable because `period_idx` gets assigned from `i`, which is less than the magic `8`

components/datetime/src/provider/day_periods.rs
```diff
@@ -0,0 +1,263 @@
...
+    fn test_lookup() {
+        // Create a dummy DayPeriodRules
+        // presence: Morning1 (0), Afternoon1 (2), Night1 (6)
+        // presence = (1<<0) | (1<<2) | (1<<6) = 1 | 4 | 64 = 69
+        // transitions: transitions at 6 (Morning1), 12 (Afternoon1), 18 (Night1)
+        // 6 is bit 6 of byte 0.
+        // 12 is bit 4 of byte 1.
+        // 18 is bit 2 of byte 2.
+
+        let mut transitions = [0u8; 3];
```
robertbastian, line 255: issue: the construction code should live in this file, behind the `datagen` feature, and you should test structs that were constructed using the constructor

components/datetime/src/provider/day_periods.rs
```diff
@@ -0,0 +1,263 @@
...
+        transitions[0] |= 1 << 6;
+        transitions[1] |= 1 << 4;
+        transitions[2] |= 1 << 2;
+
+        let rules = DayPeriodRules {
+            presence: 69,
```
robertbastian: if not using the constructor:

```suggestion
            presence: (1 << (DayPeriod::Morning as u8)) | (1 << (DayPeriod::Afternoon1 as u8)) | (1 << (DayPeriod::Night1 as u8)),
```

components/datetime/src/provider/day_periods.rs
```diff
@@ -0,0 +1,263 @@
...
+            transitions,
+        };
+
+        // Assume first period is Night1 (6) because it's the last night period.
```
robertbastian: why are we "assuming" this?

components/datetime/src/provider/day_periods.rs
```diff
@@ -0,0 +1,263 @@
...
+        let provider = crate::provider::Baked;
+        let rules: DataPayload<DayPeriodRulesV1> = provider
+            .load(DataRequest {
+                id: DataIdentifierBorrowed::for_locale(&icu_locale::langid!("en").into()),
+                ..Default::default()
+            })
+            .unwrap()
+            .payload;
+
+        assert_eq!(rules.get().lookup(0), DayPeriod::Morning1);
```
robertbastian, line 314: suggestion: might as well test all 24 hours

however, I think these really test the `icu_provider_source` parsing, and should live there. I'm also not convinced that we need to test three locales, they all test the same code, and hardcoding locale expectation has an update cost

you should add more tests like `test_lookup` that construct corner cases, like a struct with only one period

provider/source/data/debug/day/period/rules/v1/ar.json
robertbastian, line 1: suggestion, followup: these are not very useful for inspection, e.g. during a CLDR update. a custom serde human implementation could make these readable

provider/source/src/cldr_serde/day_periods.rs
```diff
@@ -0,0 +1,31 @@
...
+
+#[derive(Debug, Deserialize)]
+pub struct DayPeriodRule {
+    #[serde(rename = "_from")]
+    pub from: Option<String>,
+    #[serde(rename = "_before")]
+    pub before: Option<String>,
+    #[serde(rename = "_at")]
+    #[allow(dead_code)]
+    pub at: Option<String>,
```
robertbastian, line 29: issue: this is not dead code, you check it

provider/source/src/datetime/day_periods.rs
```diff
@@ -0,0 +1,268 @@
...
+            .core()
+            .read_and_parse("supplemental/dayPeriods.json")?;
+        Ok(day_periods
+            .supplemental
+            .day_period_rule_set
+            .0
+            .iter()
+            .filter_map(|(l, rules)| {
+                let langid: icu::locale::LanguageIdentifier = l.parse().unwrap();
+                let has_flexible = rules.keys().any(|k| DayPeriod::from_cldr_name(k).is_some());
```
robertbastian: suggestions: use `compute_day_periods`, which should exit early if there are no flexible periods, so that `iter_ids` goes through the same code path as `load`. then you can `unwrap` in load, because you called `check_req`

provider/source/src/datetime/day_periods.rs
robertbastian: I think fallback is fine, as long as the _names_ don't fall back. If `zh-SG: am, pm, noon, midnight` (I assume you meant to use an example that falls back to `zh`, `zh-Hant` does not), then it's fine to fall back to `zh`'s flexible day periods, as long as `zh-SG` doesn't end up using them because it doesn't have names for them. this would be cleaner if the names and rules were in a single struct

that said, I think this would be less risky if we store an empty value for `zh-SG`. the data struct could be `Option<DayPeriodRules>`, so that we keep the invariant that `DayPeriodRules` contains at least one period, and `lookup` is infallible (and we can give `DayPeriodRules` a niche, see below).

we should *probably* also cross-reference the day periods with the day period names and remove the whole thing if they don't match up

provider/source/src/datetime/day_periods.rs
```diff
@@ -0,0 +1,268 @@
...
+                period,
+                locale_str
+            );
+        }
+        if let Some(period_enum) = DayPeriod::from_cldr_name(period) {
+            let idx = period_enum as u8;
+            if let (Some(from), Some(before)) = (&rule.from, &rule.before) {
+                let start = parse_hour(from);
+                let end = parse_hour(before);
+                entries.push((idx, start, end));
```
robertbastian: suggestion: delay the `as u8` and store a `DayPeriod` in `entries`. the `DayPeriodRules` constructor should accept `DayPeriod`s, not `u8`s.

robertbastian: suggestion: `entries` should be a `BTreeMap` so you get automatic sorting, and you can detect duplicates and warn/error

provider/source/src/datetime/day_periods.rs
```diff
@@ -0,0 +1,268 @@
...
+                    None
+                }
+            })
+            .collect())
+    }
+}
+
+/// Parses a "HH:MM" time string and returns the hour as a u8.
+/// Logs a warning if the minute value is non-zero, as precision will be lost.
+fn parse_hour(s: &str) -> u8 {
```
robertbastian, line 113: style: these three are very scoped helper methods, nest them in their calling function for readability and so nobody adds 1000 lines of code in between. two of them will also move to the component crate as part of the constructor

provider/source/src/datetime/day_periods.rs
```diff
@@ -0,0 +1,268 @@
...
+    // Compute presence bitmask from entries.
+    let mut presence = 0u8;
+    for entry in &entries {
+        presence |= 1 << entry.0;
+    }
+
+    // Find the initial active period at hour 0.
+    // By definition, this is the last present period in the sorted sequence of present periods,
+    // which corresponds to the maximum period index present in entries.
+    let mut current_period = entries.iter().map(|e| e.0).max().unwrap();
```
robertbastian: style: declare mutable loop variables closer to the loop than immutable variables

provider/source/src/datetime/day_periods.rs
```diff
@@ -0,0 +1,268 @@
...
+
+            // If it still doesn't match, resync to the actual period.
+            if actual_period != current_period {
+                current_period = actual_period;
+            }
+        }
+    }
+
+    let bytes = transitions_u32.to_le_bytes();
+    let transitions = [bytes[0], bytes[1], bytes[2]];
```
robertbastian: I verified that this matches line 112 in components/datetime/src/provider/day_periods.rs, but this code should really be in the same file

provider/source/src/datetime/day_periods.rs
```diff
@@ -0,0 +1,268 @@
...
+        presence |= 1 << entry.0;
+    }
+
+    // Find the initial active period at hour 0.
+    // By definition, this is the last present period in the sorted sequence of present periods,
+    // which corresponds to the maximum period index present in entries.
+    let mut current_period = entries.iter().map(|e| e.0).max().unwrap();
+
+    // Determine the active period index (0-7) at each hour (0-23).
+    let hour_periods = compute_hour_periods(&entries);
```
robertbastian: this is a complicated temporary data structure that could be a map, or you just search `entries` for every hour. readability trumps performance here

provider/source/src/datetime/day_periods.rs
robertbastian: how about we store `start, end` as `start..end`, because it represents a range.

do you validate whether there are gaps anywhere?

components/datetime/src/provider/day_periods.rs
```diff
@@ -0,0 +1,263 @@
...
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+#[cfg_attr(feature = "datagen", derive(serde::Serialize, databake::Bake))]
+#[cfg_attr(feature = "datagen", databake(path = icu_datetime::provider::day_periods))]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize))]
+pub enum DayPeriod {
+    /// Morning 1
+    Morning1 = 0,
+    /// Morning 2
+    Morning2 = 1,
+    /// Afternoon 1
```
robertbastian: if that's all we can add for docs, just don't add docs

provider/source/src/datetime/day_periods.rs
```diff
@@ -0,0 +1,268 @@
...
+
+/// Computes the active period index (0-7) for each of the 24 hours.
+fn compute_hour_periods(entries: &[(u8, u8, u8)]) -> [u8; 24] {
+    let mut hour_periods = [0u8; 24];
+    for entry in entries {
+        let start = entry.1;
+        let end = entry.2;
+        let mut h = start;
+        loop {
+            hour_periods[h as usize] = entry.0;
```
robertbastian: this will do weird things if start > end, which I don't think you check anywhere

components/datetime/src/provider/day_periods.rs
lisamarielfabian: > would it make sense to have the night period have the lowest discriminant? days always start with night, so it'd be more "regular" if we didn't have to look at the last period to get the first period (but instead wrap around at the end)

Does it says it makes sense to have something that makes more sense about what's happening as to but in place something else looks like it's available from the contentity reliable it's me a reliability just to consistently adapt to the changes so just let me know what it's going to be the consistency before and then the need basis

components/datetime/src/provider/day_periods.rs
Manishearth: Datagen currently doesn't generate flex period rules if they're empty. That's the open question above.

It is GIGO to ever _ask_ for `lookup()` while presence is zero I think. We probably should fall back to ampm anyway; this is possible with mismatched data.

I'll make this return an Option.

components/datetime/src/provider/day_periods.rs
Manishearth: No, we want wraparound behavior because of night periods being possible at the beginning and the end. This isn't for overflow safety, we explicitly want that overflow behavior.

components/datetime/src/provider/day_periods.rs
Manishearth: Yeah, it's "whichth_period" really.

The problem is that period_index is unclear since the periods have intrinsic indices. Name suggestions welcome.

components/datetime/src/provider/day_periods.rs
Manishearth: > would it make sense to have the night period have the lowest discriminant? days always start with night, so it'd be more "regular" if we didn't have to look at the last period to get the first period (but instead wrap around at the end)

You have similar problems because sometimes night1/night2 are fully before midnight. THis complexity mostly just moves around. I'd rather keep it this way, as discussed in the design.

components/datetime/src/provider/day_periods.rs
Manishearth: Oh, this is a leftover comment

components/datetime/src/provider/day_periods.rs
robertbastian: No. Make this non-zero and use an Option as the data struct

components/datetime/src/provider/day_periods.rs
Manishearth: Then we have to allow missing_docs. That's fine i guess.

provider/source/src/datetime/day_periods.rs
Manishearth: I was going to but didn't end up doing it. Could do that.

components/datetime/src/provider/day_periods.rs
robertbastian: Do you actually have wrap around in both directions?

provider/source/src/datetime/day_periods.rs
Manishearth: This isn't designed this way for performance. The later bit code gets easier with this.

components/datetime/src/provider/day_periods.rs
robertbastian: `selected_period`, `target_period`

provider/source/src/datetime/day_periods.rs
Manishearth: Ah, I was trying to avoid double calling it. If we're okay with calling the more expensive thing in iter I'll do it.

provider/source/src/cldr_serde/day_periods.rs
Manishearth: Ah. Yes, it was dead code initially and then I fixed it.

components/datetime/src/provider/day_periods.rs
robertbastian: We have that for enum variants in quite a few places

components/datetime/src/provider/day_periods.rs
Manishearth: Why? What's the benefit?

robertbastian: this function doesn't have to return an option, you can resolve the absence of rules earlier, ideally along with the names

robertbastian: *and* you can represent the no-rules case as `None`, so you have a value for every locale

components/datetime/src/provider/day_periods.rs
Manishearth: We still want to wrap 24 around to zero, so either we use a modulo operation or a branch. a branch is worse

robertbastian: why? where would a 24 come from?

provider/source/src/datetime/day_periods.rs
Manishearth: Oh ,we can't, because in some cases `end < start`

Manishearth: We can still validate though

Manishearth: I added validation. I am a bit unsure about asserting though. Usually we warn when CLDR does something we don't like, but  I can't warn outside of provider/source without adding deps.

components/datetime/src/provider/day_periods.rs
Manishearth: I think this ties in to the earlier question about what we should be doing in the None case. I'd prefer this be figured out as a followup.

components/datetime/src/provider/day_periods.rs
Manishearth: What do you mean? We want to wrap around in one direction, the modulo operation is the easiest way to do that. A conditional add does not accomplish that.

Also I don't understand why you consider modulo operations overkill: they are not branchy, whereas most other ways of doing things are.

provider/source/src/datetime/day_periods.rs
```diff
@@ -0,0 +1,186 @@
...
+    for (period, rule) in rules {
+        if rule.at.is_some() {
+            assert!(
+                period == "noon" || period == "midnight",
+                "Found 'at' rule for non-noon/midnight period: {} in locale {}",
+                period,
+                locale_str
+            );
+        }
+        if let Some(period_enum) = DayPeriod::from_cldr_name(period) {
```
sffc, line 70: Issue: If the day period name is not found in the enum, it should be a warning, not a silent failure

provider/source/src/datetime/day_periods.rs
```diff
@@ -0,0 +1,186 @@
...
+            if let (Some(from), Some(before)) = (&rule.from, &rule.before) {
```
sffc, line 71: Issue: If these are None then it seems like it should be a warning, not a silent failure

components/datetime/src/provider/day_periods.rs
sffc: We need GIGO behavior here anyway, even if we do the thing @robertbastian suggested. (I am against making data structs have internal invariants for trivial things that don't meaningfully impact performance.)

components/datetime/src/provider/day_periods.rs
```diff
@@ -0,0 +1,361 @@
...
+        if self.presence == 0 {
+            debug_assert!(false, "presence must not be zero");
+            return DayPeriod::Morning1;
+        }
+
+        let count = self.presence.count_ones() as usize;
+
+        // Combine transitions bytes into a u32 for easier bit manipulation.
+        let transitions_u32 = u32::from_le_bytes([
+            self.transitions[0],
```
sffc, line 106: Question: Does indexing_slicing permit this? Is there no panic here?

You can destructure the array to avoid the panic if necessary.

provider/source/src/datetime/day_periods.rs
Manishearth: We assert now

provider/source/src/datetime/day_periods.rs
Manishearth: This now uses modulo to wrap around. start is allowed to be > end.

sffc (commented):
I didn't review the bit math in detail

components/datetime/src/provider/day_periods.rs
Manishearth: indexing_slicing understands fixed size arrays.

components/datetime/src/provider/day_periods.rs
Manishearth: Right. A lot of the code here (e.g. the modulo operations) help with gigo.

sffc (approved):

provider/source/src/datetime/day_periods.rs
Manishearth: https://github.com/unicode-org/icu4x/issues/7968

components/datetime/src/provider/day_periods.rs
Manishearth: https://github.com/unicode-org/icu4x/issues/7968

provider/source/data/debug/day/period/rules/v1/ar.json
Manishearth: https://github.com/unicode-org/icu4x/issues/7969

components/datetime/src/provider/day_periods.rs
Manishearth: https://github.com/unicode-org/icu4x/issues/7969

sffc (approved):

Manishearth:
Going to merge this and am happy to do followups later. In broad strokes it is clear we want something like this and the exact bit implementation and tests can be moved around. We have followup issues filed for some of this.

components/datetime/src/provider/day_periods.rs
robertbastian: why did you mark this as resolved?

provider/source/src/datetime/day_periods.rs
```diff
@@ -71,62 +68,21 @@ pub(crate) fn compute_day_periods(
...
     }
 
-    let bytes = transitions_u32.to_le_bytes();
-    let transitions = [bytes[0], bytes[1], bytes[2]];
-
-    Some(DayPeriodRules {
-        presence,
-        transitions,
-    })
+    DayPeriodRules::from_cldr_rules(&entries)
```
robertbastian, line 85: here you return `None` if there are data inconsistencies, but above you return `None` if there are no flexible day periods. in the caller these are handled the same, that's weird

components/datetime/src/provider/day_periods.rs
```diff
@@ -116,43 +109,135 @@ impl DayPeriodRules {
...
+                    break i;
                 }
+                // Found a period bit
                 found_count += 1;
             }
-        }
+            i += 1;
+            if i >= 8 {
+                // Unreachable: `target_period` is % count, which caps it to `count`,
+                // the number of set bits in `presence` (minus one), which must be at most 7
```
robertbastian, line 142: there is no "minus one". count can be 8



```rust
let count = self.presence.count_ones() as usize;
```

components/datetime/src/provider/day_periods.rs
Manishearth: I moved the construction code, no? Was this about tests? I have a followup filed about cleaning up tests.

Or is this comment on code that didn't get moved? I'll have to take a look on a computer when I'm back: I moved most of the construction code over and resolved comments about moving stuff

robertbastian: the point of the comment was that you shouldn't be manually constructing the encoded struct for your test, that's more work to maintain if we want to change the encoding, and it unnecessarily depends on internals. you added _another_ test that goes through the constructor, but this one should do that as well

Issue #7694, opened 2026-02-25
Should we have infallible Date::add()?


Manishearth:
See: https://github.com/unicode-org/icu4x/issues/3964

From https://github.com/unicode-org/icu4x/issues/3964#issuecomment-3889831325:

> Do not use `overflow: Constrain` as a hint for what to do on RD overflow. Open an issue in the Backlog milestone to collect feedback on whether we should have an infallible add function.
> 
> Do not add operators (`core::ops`) at this time. Open an issue in the Backlog milestone to collect feedback on whether the operators would be handy.

This is the issue for discussing that.

Basically, we currently have `Date::try_new_add[ed](&duration, &options)`. If we assume `Overflow: Constrain`, and saturate when adding past RD range, we can implement `add()` and maybe even `core::ops::Add`.

Let's see if people actually want that.

robertbastian:
A saturating `core::ops::Add` is probably unexpected, but then again, if you're reaching the saturation point, your dates don't have much meaning anymore. We could have panic-in-debug and saturate-in-release behaviour for `core::ops::Add`, similar to integers.

PR #7965, opened 2026-05-13
Use flexible day periods when they show up in our current set of fieldsets

Manishearth:
Based on https://github.com/unicode-org/icu4x/pull/7964. Can just be reviewed as a whole; I wanted that separate since it's the more complex data model change.

Progress on #487. Makes it no longer an ECMA402 problem; the remaining work is to add the `B` fieldset.

Mostly just implements our design.

I have reviewed this code, but less thoroughly compared to #7694. I am happy to wait for that to land and then go through this more carefully.

## Changelog

icu_datetime: Support flexible day periods for locales that use them in regular fieldsets.


provider/source/src/datetime/names.rs
```diff
@@ -189,10 +190,61 @@ fn weekday_convert(
...
+        }
+    }
+    false
+}
+
+/// Checks if the locale needs flexible day periods.
+/// A locale needs them if any of its supported calendars has standard time formats
+/// or available formats containing 'B' (flexible day periods) for non-B skeletons.
+fn needs_flexible_day_periods(datagen: &SourceDataProvider, locale: &DataLocale) -> bool {
+    const ALL_CALENDARS: &[DatagenCalendar] = &[
```
Manishearth, line 220: This isn't great. Day period names are stored in a calendar-agnostic way, and previously that was fine, but "does this day period show up in a non-B skeleton" changes by calendar. Specifically, `de` has `B` show up for an `h` skeleton in the Buddhist calendar. That's a bit weird but whatever.

This const should probably live on DatagenCalendar but I want to see what people think before cleaning it up.

provider/source/src/datetime/available_formats.rs
```diff
@@ -198,15 +198,15 @@ mod test {
...
+        "Bh", "Bhm", "Bhms", "EBhm", "EBhms",
     ];
 
     // NOTE: If you are moving this to the SUPPORTED section, make sure to remove the match
     //       on your symbol from impl From<fields::SymbolError> for SkeletonError
     //       and then regenerate baked data (`cargo make bakeddata components/datetime`)
     #[rustfmt::skip]
     const UNSUPPORTED_STRING_SKELETONS: &[&str] = &[
-        // TODO(#487) - Flexible day periods
-        "Bh", "Bhm", "Bhms", "EBhm", "EBhms",
```
Manishearth, line 209: These aren't fully supported yet, but the test fails now that B is supported. The test could be fixed to treat these as still fail-y.

components/datetime/src/provider/names.rs
```diff
@@ -527,3 +529,90 @@ pub use DatetimeNamesWeekdayV1 as WeekdayNamesV1;
...
+mod tests {
+    use super::*;
+
+    #[cfg(feature = "compiled_data")]
+    #[test]
+    fn test_dayperiod_names_compiled_data() {
+        let provider = crate::provider::Baked;
+
+        // Load names
+        let names_en: DataPayload<DayPeriodNamesV1> = provider
```
robertbastian, line 543: please don't test datagen results in component crates. these should be unit tests in `icu_provider_source` (_if_ you have to hardcode locale data for four locales in the first place)

provider/source/src/datetime/names.rs
```diff
@@ -189,10 +190,61 @@ fn weekday_convert(
...
+
+fn needs_flexible_day_periods_for_calendar(data: &ca::Dates) -> bool {
+    if time_format_has_b(&data.time_formats) {
+        return true;
+    }
+    if time_format_has_b(&data.time_skeletons) {
+        return true;
+    }
+    for (skeleton, pattern) in &data.datetime_formats.available_formats.0 {
+        if pattern.contains('B') && !skeleton.contains('B') {
```
robertbastian, line 209: please add a comment why we exclude patterns where the skeleton contains B. I don't understand

provider/source/src/datetime/names.rs
```diff
@@ -212,6 +264,18 @@ fn dayperiods_convert(
         periods.push(midnight)
     }
 
+    if needs_flexible_day_periods(datagen, locale) {
+        periods.resize(4, "");
+        let mut flex_periods = vec![""; 8];
+        for (period_name, name) in &day_periods.flexible {
+            if let Some(period_enum) = DayPeriod::from_cldr_name(period_name) {
+                let idx = period_enum as usize;
+                flex_periods[idx] = name;
```
robertbastian, line 273: nit: the intermediate name `idx` variable doesn't add anything

provider/source/src/datetime/names.rs
```diff
@@ -212,6 +264,18 @@ fn dayperiods_convert(
         periods.push(midnight)
     }
 
+    if needs_flexible_day_periods(datagen, locale) {
+        periods.resize(4, "");
+        let mut flex_periods = vec![""; 8];
+        for (period_name, name) in &day_periods.flexible {
```
robertbastian, line 270: right, so here you should check that the names match the rules

provider/source/src/debug_provider.rs
```diff
@@ -32,6 +32,11 @@ where
             Box::new(['\0'; 10])
         } else if type_id == TypeId::of::<DecimalSymbols>() {
             Box::new(DecimalSymbols::new_en_for_testing())
+        } else if type_id == TypeId::of::<icu::datetime::provider::day_periods::DayPeriodRules>() {
+            Box::new(icu::datetime::provider::day_periods::DayPeriodRules {
+                presence: 0,
```
robertbastian, line 37: `presence: 0` is garbage according to your `lookup` function

components/datetime/src/pattern/formatter.rs
```diff
@@ -717,4 +717,87 @@ mod tests {
...
+        // In German Buddhist calendar, "BBBB h 'Uhr'" formats B as Wide flexible day period and h as hour
+        let pattern: DateTimePattern = "BBBB h 'Uhr'".parse().unwrap();
+        let datetime = DateTime {
+            date: Date::try_new_iso(2023, 11, 17)
+                .unwrap()
+                .to_calendar(icu_calendar::cal::Buddhist),
+            time: Time::try_new(13, 0, 0, 0).unwrap(),
+        };
+        let formatted = names.with_pattern_unchecked(&pattern).format(&datetime);
+        assert_try_writeable_eq!(formatted, "nachmittags 1 Uhr", Ok(()));
```
robertbastian, line 801: thought: this is the weirdest data. first of all, you'd say "1 Uhr nachmittags", and second, this is universal and in no way tied to the Buddhist calendar

components/datetime/src/pattern/formatter.rs
```diff
@@ -717,4 +717,87 @@ mod tests {
...
+
+        let mut names: FixedCalendarDateTimeNames<Gregorian> =
+            FixedCalendarDateTimeNames::try_new(locale).unwrap();
+        names
+            .load_day_period_names(&crate::provider::Baked, DayPeriodNameLength::Wide)
+            .unwrap()
+            .load_day_period_rules(&crate::provider::Baked)
+            .unwrap();
+
+        let pattern: DateTimePattern = "BBBBh:mm".parse().unwrap();
```
robertbastian, line 768: why are you using custom patterns for this test? aren't these are used by some skeleton?

doing it this way doesn't test the name loading code

components/datetime/src/pattern/names.rs
```diff
@@ -3019,6 +3058,33 @@ impl<FSet: DateTimeNamesMarker> RawDateTimeNames<FSet> {
...
+        error_field: ErrorField,
+    ) -> Result<(), PatternLoadError>
+    where
+        P: BoundDataProvider<DayPeriodRulesV1> + ?Sized,
+    {
+        let locale = provider
+            .bound_marker()
+            .make_locale(prefs.locale_preferences);
+        // Normalize the request locale to language-only (suppressing script/region),
+        // because supplemental day period rules are strictly language-level in CLDR.
```
robertbastian, line 3074: issue: datagen needs to create data in a way that works with ICU4X's fallback behaviour. this is required by the data pipeline, e.g. filtering, and runtime hacks like this should not be done

components/datetime/src/pattern/names.rs
```diff
@@ -3926,14 +3996,31 @@ impl RawDateTimeNamesBorrowed<'_> {
...
             .dayperiod_names
             .get_with_variables(day_period_name_length)
             .ok_or(GetNameForDayPeriodError::NotLoaded)?;
         let option_value: Option<&str> = match (field_symbol, u8::from(hour), is_top_of_hour) {
+            (Flexible, hour, _) => {
+                let rules = self.dayperiod_rules.get_option();
+                // Look up the active DayPeriod enum value (0-7) chronologically for this hour.
+                let period_idx = rules.map(|r| r.lookup(hour) as usize);
+                // Flexible day period names are appended at indices 4..12 in `LinearNames`
+                // following standard day periods (am, pm, noon, midnight at 0..4).
```
robertbastian, line 4012: issue, blocking: this needs to be documented on the data struct, and ideally you'd move this logic to the data struct, i.e. `DayPeriodNames::get_flexible(DayPeriod)`

components/datetime/src/pattern/names.rs
```diff
@@ -3926,14 +3996,31 @@ impl RawDateTimeNamesBorrowed<'_> {
...
+                let name = period_idx.and_then(|idx| dayperiod_names.names.get(4 + idx));
+                // If rules are missing or names are unpopulated for this specific flexible period,
+                // we fall back to the locale's own standard AM/PM names.
+                name.or_else(|| {
+                    if hour < 12 {
```
robertbastian, line 4017: this seems to be a debug_assert case, especially if we verify this in datagen

components/datetime/src/pattern/names.rs
```diff
@@ -3926,14 +3996,31 @@ impl RawDateTimeNamesBorrowed<'_> {
...
             .ok_or(GetNameForDayPeriodError::InvalidFieldLength)?;
         let dayperiod_names = self
             .dayperiod_names
             .get_with_variables(day_period_name_length)
             .ok_or(GetNameForDayPeriodError::NotLoaded)?;
         let option_value: Option<&str> = match (field_symbol, u8::from(hour), is_top_of_hour) {
+            (Flexible, hour, _) => {
+                let rules = self.dayperiod_rules.get_option();
+                // Look up the active DayPeriod enum value (0-7) chronologically for this hour.
+                let period_idx = rules.map(|r| r.lookup(hour) as usize);
```
robertbastian, line 4010: don't throw away the strong `DayPeriod` type before you actually need it as an index (which is within a `.get()` call).

components/datetime/src/provider/day_periods.rs
```diff
@@ -36,6 +36,11 @@ icu_provider::data_marker!(
     /// `DayPeriodRulesV1` marker
     DayPeriodRulesV1,
     DayPeriodRules,
+    fallback_config = {
+        let mut config = icu_provider::fallback::LocaleFallbackConfig::default();
+        config.priority = icu_provider::fallback::LocaleFallbackPriority::Language;
```
robertbastian, line 41: issue: that's the default

provider/source/src/datetime/available_formats.rs
robertbastian: why aren't they fully supported now? we seem to support all fields?

Manishearth: They're supported as fields, but you still can't get these _skeletons_ since we don't have B fieldsets.

components/datetime/src/provider/names.rs
Manishearth: We test datagen results in `icu_datetime` all the time.

Maybe I shouldn't be testing it with data loading and instead e2e.

components/datetime/src/pattern/names.rs
Manishearth: This _is_ documented on the data struct. The data struct is LinearNames, this is documented there. LinearNames is used for many different things.

There is no DayPeriodNames.

sffc:
I should probably review this but I'm prioritizing this lower than other priorities since the dependent PR is more thoroughly deslopped and this PR depends on it.

provider/source/src/datetime/names.rs
Manishearth: We don't support B fieldsets yet. I'll add a comment

components/datetime/src/pattern/formatter.rs
Manishearth: Yeah I'm really unsure about this data and why it's only present for Buddhist.

components/datetime/src/pattern/names.rs
Manishearth: Huh, I thought I had fixed this already. I think I only partially fixed it then.

(This was a case of the agent picking a route that was incorrect, and then I undid a bunch of the changes, but there was some stuff left over. Like I said, I haven't fully reviewed this PR)

Manishearth:
Just rebased, I still need to go through this more.

components/datetime/src/provider/names.rs
robertbastian: we test end-to-end results, because they exercise runtime code in the crate. asserting that `icu_provider_source` produces a specific data struct should happen in `icu_provider_source`, because that makes updating CLDR data a lot easier. please trust me on this, as I always do this

components/datetime/src/pattern/names.rs
robertbastian: we can make a serde-transparent `DayPeriodNames` around `LinearNames`. I think there's enough complexity in this encoding to justify this

robertbastian: we have `am`, `pm`, `noon`, and `midnight` getters on `LinearNames` as well

components/datetime/src/provider/names.rs
Manishearth: Okay, understood.

There is one problem with this though:  anything you want to test in icu_provider_source needs to be a part of the relatively restricted set of testdata, yes? Or am I missing something? We have a lot of pressure to not bloat the list of locales in testdata.

Manishearth: Anyway, really what I plan to do here is replace this with an e2e test, this was mostly temporary.

components/datetime/src/pattern/formatter.rs
robertbastian: it's not actually, it's in `generic`: https://github.com/unicode-org/cldr/blob/1f4b115ec613020cca919d718904cef6e791baf0/common/main/de.xml#L2443

it should be reachable through `de-u-hc-h12` as it's in the pattern for `h`

robertbastian: the `"BBBB h 'Uhr'"` pattern that this test hardcodes is nowhere to be found in CLDR

sffc:
Is this PR still relevant or is it superseded by @robertbastian's other work?

Manishearth:
It is likely superseded but I need to have a look at that PR first

Manishearth:
#7971

PR #7966, opened 2026-05-13
Add data model for packed range patterns

Manishearth:
Progress on  #5448

This adds the data model for packed range patterns from [the design](https://docs.google.com/document/d/1ocgeiYiIJ7fUTr08qhC8CBF-I5AQqYYFmStPtlsORSw/edit?tab=t.0).

I can't datagen this without serde support (being worked on in https://github.com/unicode-org/icu4x/pull/8009), but the data model itself ought to be landable.

## Changelog

icu_datetime:
- Implement data structures for packed range patterns.

### DIFF
--- a/components/datetime/src/provider/range_patterns.rs
+++ b/components/datetime/src/provider/range_patterns.rs
@@ -1,0 +1,210 @@
+// This file is part of ICU4X. For terms of use, please see the file
+// called LICENSE at the top level of the ICU4X source tree
+// (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
+//! Data structures for packed range patterns.
+
+use crate::provider::packed_pattern::GenericPackedPatterns;
+use crate::provider::pattern::runtime::{Pattern, PatternULE};
+use icu_provider::prelude::*;
+use zerovec::VarZeroVec;
+
+/// The date fields that can have a greatest difference.
+///
+/// Ordered from smallest to largest field difference.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[repr(u8)]
+pub enum DateGreatestDifferenceField {
+    /// Day (`d`)
+    Day = 0,
+    /// Month (`M`)
+    Month = 1,
+    /// Year (`y`)
+    Year = 2,
+    /// Era (`G`)
+    Era = 3,
+}
+
+/// The time fields that can have a greatest difference.
+///
+/// Ordered from smallest to largest field difference.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[repr(u8)]
+pub enum TimeGreatestDifferenceField {
+    /// Minute (`m`)
+    Minute = 0,
+    /// Hour (`h`/`H`)
+    Hour = 1,
+    /// Flexible Day Period (`B`)
+    DayPeriodB = 2,
+    /// Day Period (`a`)
+    DayPeriodA = 3,
+}
+
+/// A bitset encoding which fields are present in the `GreatestDifference` pattern list.
+///
+/// The bitset maps greatest-difference field IDs to their presence in the patterns list,
+/// ordered from smallest to largest field difference.
+///
+/// Date and time fields reuse the same bits (0-3) because they are stored in separate
+/// range pattern collections (date vs time).
+///
+/// For date skeleta, the fields are:
+/// * Bit 0: Day (`d`)
+/// * Bit 1: Month (`M`)
+/// * Bit 2: Year (`y`)
+/// * Bit 3: Era (`G`)
+///
+/// For time skeleta, the fields are:
+/// * Bit 0: Minute (`m`)
+/// * Bit 1: Hour (`h`/`H`)
+/// * Bit 2: Flexible Day Period (`B`)
+/// * Bit 3: Day Period (`a`)
+///
+/// This sparse representation allows quickly finding the pattern corresponding to a
+/// differing field by counting the number of set bits before it.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[zerovec::make_ule(GreatestDifferenceHeaderULE)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+#[cfg_attr(feature = "datagen", derive(databake::Bake))]
+#[cfg_attr(feature = "datagen", databake(path = icu_datetime::provider::range_patterns))]
+#[zerovec::skip_derive(Ord)]
+pub struct GreatestDifferenceHeader(pub u8);
+
+impl GreatestDifferenceHeader {
+    /// Creates a new `GreatestDifferenceHeader` from a u8 value.
+    pub const fn new(val: u8) -> Self {
+        Self(val)
+    }
+
+    /// Returns whether the given date field is present in this header.
+    pub const fn is_date_field_present(self, field: DateGreatestDifferenceField) -> bool {
+        (self.0 & (1 << (field as u8))) != 0
+    }
+
+    /// Returns whether the given time field is present in this header.
+    pub const fn is_time_field_present(self, field: TimeGreatestDifferenceField) -> bool {
+        (self.0 & (1 << (field as u8))) != 0
+    }
+}
+
+/// A list of patterns that represent the greatest difference format for a given skeleton.
+#[derive(Debug, PartialEq, Eq, Clone, yoke::Yokeable, zerofrom::ZeroFrom)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize))]
+#[cfg_attr(feature = "datagen", derive(serde::Serialize))]
+#[cfg_attr(feature = "datagen", derive(databake::Bake))]
+#[cfg_attr(feature = "datagen", databake(path = icu_datetime::provider::range_patterns))]
+#[zerovec::make_varule(PatternsByGreatestDifferenceULE)]
+#[zerovec::derive(Debug)]
+#[zerovec::skip_derive(Ord)]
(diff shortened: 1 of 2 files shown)


components/datetime/src/provider/range_patterns.rs
```diff
@@ -0,0 +1,210 @@
...
+use crate::provider::pattern::runtime::{Pattern, PatternULE};
+use icu_provider::prelude::*;
+use zerovec::VarZeroVec;
+
+/// The date fields that can have a greatest difference.
+///
+/// Ordered from smallest to largest field difference.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[repr(u8)]
+pub enum DateGreatestDifferenceField {
```
Manishearth, line 17: design note: I decided to reuse bitfield indices for day and time fields (they don't mix).

This means that if CLDR adds more greatestDifference values in the future we are still all set.

components/datetime/src/provider/range_patterns.rs
```diff
@@ -0,0 +1,210 @@
...
+    pub header: GreatestDifferenceHeader,
+    /// The list of greatestDifference patterns.
+    #[cfg_attr(feature = "serde", serde(borrow))]
+    pub patterns: VarZeroVec<'data, PatternULE>,
+}
+
+/// Finds the smallest present bit index >= `requested` in `header` (limited to bits 0-3).
+/// Returns `Some((found_bit_index, pattern_index))` or `None`.
+fn resolve_fallback(header: u8, requested: u8) -> Option<(u8, usize)> {
+    for i in requested..4 {
```
sffc: 🤖: The magic number `4` here represents the number of fields. It might be better to use a constant or at least add a comment explaining why it is `4` (covering bits 0-3 for Day, Month, Year, Era or Minute, Hour, DayPeriodB, DayPeriodA).

@sffc: I do find it slightly confusing that GreatestDifferenceHeader has 8 bits but you check only 4 of them here

components/datetime/src/provider/range_patterns.rs
```diff
@@ -0,0 +1,210 @@
...
+///
+/// This sparse representation allows quickly finding the pattern corresponding to a
+/// differing field by counting the number of set bits before it.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[zerovec::make_ule(GreatestDifferenceHeaderULE)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
+#[cfg_attr(feature = "datagen", derive(databake::Bake))]
+#[cfg_attr(feature = "datagen", databake(path = icu_datetime::provider::range_patterns))]
+#[zerovec::skip_derive(Ord)]
+pub struct GreatestDifferenceHeader(pub u8);
```
sffc, line 82: 🤖: Is there a reason to make the inner `u8` public? If it is not needed by `zerovec` or other external crates directly, making it private would encapsulate the bitset representation better.

@sffc: it's because of databake, and all bit patterns are well-defined, so this is fine. however, it's slightly concerning that there are no bits for future extensions.

sffc (commented):
some initial comments, no time tonight for a full review

components/datetime/src/provider/range_patterns.rs
Manishearth: Yeah I was thinking about this. I keep waffling between designs where I use all eight bits (but then it's not expandable, and the math gets weirder) or I just use four and have space to grow.

I think I'll add a const.

components/datetime/src/provider/range_patterns.rs
Manishearth: We have bits for future extensions.

Manishearth: We use the first four bits, and the last four bits are currently unused. I don't think we need to care about that during validation, they will be ignored for now.

sffc: oh I didn't catch that the date bits and time bits are mutually exclusive.

components/datetime/src/provider/range_patterns.rs
```diff
@@ -0,0 +1,228 @@
...
+
+/// Finds the smallest present bit index >= `requested` in `header` up to `max_value`.
+/// Returns `Some((found_bit_index, pattern_index))` or `None`.
+fn resolve_fallback(header: u8, requested: u8, max_value: u8) -> Option<(u8, usize)> {
+    for i in requested..=max_value {
+        if (header & (1 << i)) != 0 {
+            let mask = (1 << i) - 1;
+            let pattern_index = (header & mask).count_ones() as usize;
+            return Some((i, pattern_index));
+        }
```
sffc, line 128: praise: clever bitmasking indexing. I'm 85% sure it is correct

components/datetime/src/provider/range_patterns.rs
```diff
@@ -0,0 +1,228 @@
...
+        let pattern2 = "y"
+            .parse::<reference::Pattern>()
+            .unwrap()
+            .to_runtime_pattern();
+
+        let patterns = vec![pattern1.clone(), pattern2.clone()];
+        let varzerovec = VarZeroVec::from(patterns.as_slice());
+
+        // Header with Day (bit 0) and Year (bit 2) present.
+        let header = GreatestDifferenceHeader(1 | 4);
```
sffc, line 249: Nit: I think this PR should include the constructor / builder, because otherwise in the PR where you use this, I need to re-learn the bits.

Manishearth:
Good call, and that makes #8014 even smaller.

components/datetime/src/provider/range_patterns.rs
```diff
@@ -160,6 +160,62 @@ impl<'data> PatternsByGreatestDifference<'data> {
...
+        for (bit, _) in parsed.iter() {
+            if *bit > 3 {
+                return Err("GreatestDifference bit index must be 0, 1, 2, or 3");
+            }
+        }
+        parsed.sort_by_key(|(bit, _)| *bit);
+
+        let mut header_val = 0u8;
+        let mut patterns = Vec::new();
+        for (bit, pattern) in parsed {
```
sffc: Suggestion (optional): deduplicate here for patterns that fall back to one another, such as hours to day periods

components/datetime/src/provider/range_patterns.rs
```diff
@@ -160,6 +160,62 @@ impl<'data> PatternsByGreatestDifference<'data> {
             .get(pattern_index)
             .map(<Pattern as zerofrom::ZeroFrom<PatternULE>>::zero_from)
     }
+
+    /// Construct from an iterator of (`bit_index`, pattern).
+    /// The `bit_index` must be <= 3.
+    /// The iterator will be sorted by `bit_index`.
+    #[cfg(feature = "datagen")]
+    fn try_from_unsorted_iter<I>(iter: I) -> Result<Self, &'static str>
```
sffc: Suggestion: why not require that they be sorted? LiteMap and BTreeMap can guarantee this

components/datetime/src/provider/range_patterns.rs
```diff
@@ -225,4 +281,53 @@ mod tests {
...
+            .to_runtime_pattern();
+        let pattern_y = "y"
+            .parse::<reference::Pattern>()
+            .unwrap()
+            .to_runtime_pattern();
+
+        // Valid date patterns (unsorted input)
+        let pgd = PatternsByGreatestDifference::try_from_date_patterns(vec![
+            (DateGreatestDifferenceField::Year, pattern_y.clone()),
+            (DateGreatestDifferenceField::Day, pattern_d.clone()),
```
sffc: Nit: Demonstrate that it works if you use ZeroFrom instead of Clone

sffc (approved):

components/datetime/src/provider/range_patterns.rs
Manishearth: Didn't end up doing this, I think CLDR data is quite clean here already. But worth considering.

components/datetime/src/provider/range_patterns.rs
```diff
@@ -199,22 +191,20 @@ impl<'data> PatternsByGreatestDifference<'data> {
...
     #[cfg(feature = "datagen")]
-    pub fn try_from_date_patterns<I>(iter: I) -> Result<Self, &'static str>
-    where
-        I: IntoIterator<Item = (DateGreatestDifferenceField, Pattern<'data>)>,
-    {
-        Self::try_from_unsorted_iter(iter.into_iter().map(|(f, p)| (f as u8, p)))
+    pub fn try_from_date_patterns(
+        map: alloc::collections::BTreeMap<DateGreatestDifferenceField, Pattern<'data>>,
+    ) -> Result<Self, &'static str> {
+        Self::try_from_sorted(map.into_iter().map(|(f, p)| (f as u8, p)).collect())
```
sffc: Issue: don't reallocate the BTreeMap.

components/datetime/src/provider/range_patterns.rs
```diff
@@ -161,31 +161,23 @@ impl<'data> PatternsByGreatestDifference<'data> {
...
+    /// Construct from a `BTreeMap` of (`bit_index`, pattern).
     /// The `bit_index` must be <= 3.
-    /// The iterator will be sorted by `bit_index`.
     #[cfg(feature = "datagen")]
-    fn try_from_unsorted_iter<I>(iter: I) -> Result<Self, &'static str>
-    where
-        I: IntoIterator<Item = (u8, Pattern<'data>)>,
-    {
+    fn try_from_sorted(
+        map: alloc::collections::BTreeMap<u8, Pattern<'data>>,
```
sffc: Issue: This should take a tuple iterator like before, and be named `try_from_unique_sorted`, and optionally debug assert the invariant.

sffc (approved):

PR #7967, opened 2026-05-13
Turn PackedPatterns into GenericPackedPatterns

Manishearth:
Part of  #5448

You can see this in use at https://github.com/unicode-org/icu4x/pull/7966

Depends on https://github.com/unicode-org/icu4x/pull/7999

I decided not to genericize the serde code for now. It doesn't seem worth it and I want to experiment with that in a separate step. But genericizing the builder/lookup code means that the whole "variant lookup" business is encapsulated.

Note that I have not yet experimented with this end-to-end. I think this is going to be a necessary change either way, though, and can be landed separately.

## Changelog: N/A


lisamarielfabian (approved):
Thank you

components/datetime/src/provider/packed_pattern.rs
```diff
@@ -368,29 +403,29 @@ impl PackedPatternsBuilder<'_> {
...
-    fn simplify(&mut self) {
-        if self.variant0.as_ref() == Some(&self.standard) {
-            self.variant0 = None;
-        }
-        if self.variant1.as_ref() == Some(&self.standard) {
-            self.variant1 = None;
-        }
+impl<'a> GenericPackedPatternsBuilder<PluralElements<Pattern<'a>>> {
+    /// Builds a packed pattern representation from the builder.
+    pub fn build(self) -> PackedPatterns<'static> {
```
Manishearth, line 433: Currently I didn't want to genericize this, we can wait until we have the actual Range type to figure out what bounds we want here.

components/datetime/src/provider/packed_pattern.rs
```diff
@@ -301,24 +304,56 @@ impl<'a> UnpackedPatterns<'a> {
...
+    /// returning a `GenericUnpackedPatterns<T>`.
+    pub(crate) fn build_unpacked(mut self) -> GenericUnpackedPatterns<T> {
         self.simplify();
 
         // Initialize the elements vector with the standard patterns.
         let mut elements = Vec::new();
         let mut has_explicit_medium = false;
         let mut has_explicit_short = false;
-        elements.push(self.standard.long.as_ref().map(Pattern::as_ref));
+        elements.push(self.standard.long.clone());
```
sffc: Question: why did you split build and build_unpacked?

You need the clone because of this. You didn't need it before.

(Possible "issue" if you don't have a good reason: please fix.)

sffc (commented):
Praise: quite a small diff.

One comment.

components/datetime/src/provider/packed_pattern.rs
Manishearth: Because UnpackedPatterns::build includes a lot of Plurals-specific stuff. See comment below; I didn't want to try genericizing that code until I had more of the Range type in place.

Manishearth: I'm going to try and genericize it now anyway since I have some time

Manishearth: Genericized. Still needs the clone though, because I still need the split to genericize it

Manishearth: One way to avoid the clone is to pre-map the elements to references and then require `T: Copy` for `build_unpacked`.

I think we don't need to optimize at that level but we can. Easy followup to explore, too.

Manishearth:
There's a datagen diff because the name of the struct changed. That's fine.

Manishearth:
Genericizing serde will make this diff larger but I want to do that separately.

components/datetime/src/provider/packed_pattern.rs
sffc: I made it use `ZeroFrom`. I pushed to your branch.

sffc (approved):

Issue #7969, opened 2026-05-13
Day period tests and test locations


Manishearth:
We should add more tests and decide where they go

https://github.com/unicode-org/icu4x/pull/7964#discussion_r3232881812

> suggestion: might as well test all 24 hours
> 
> however, I think these really test the `icu_provider_source` parsing, and should live there. I'm also not convinced that we need to test three locales, they all test the same code, and hardcoding locale expectation has an update cost
> 
> you should add more tests like `test_lookup` that construct corner cases, like a struct with only one period

https://github.com/unicode-org/icu4x/pull/7964#discussion_r3232888747

> suggestion, followup: these are not very useful for inspection, e.g. during a CLDR update. a custom serde human implementation could make these readable

PR #7970, opened 2026-05-15
Clean up after #7964

robertbastian:
Mainly addresses tests
* Splits tests between source data provider (locale-specific tests) and the component crate (behaviour-specific tests)
* Returns `Result` instead of panicking in constructor
* Tests edge cases like empty periods, single periods

#7969

## Changelog

N/A


provider/source/src/datetime/day_periods.rs
```diff
@@ -174,18 +175,104 @@ mod tests {
...
+                .into_iter()
+                .collect()
+            )
+            .unwrap()
+            .unwrap()
+        )
+    }
+
+    #[test]
+    fn test_data() {
```
robertbastian: it's unclear to me what the value of these tests is

Manishearth:
won't be able to do a proper review for another week. I am fine with Shane reviewing this though, if I have any issues I'll bring them up post merge. Thanks for helping!

robertbastian:
@sffc ?

components/datetime/src/provider/day_periods.rs
```diff
@@ -199,8 +197,14 @@ impl DayPeriodRules {
...
 
-        let hour_periods =
-            hour_periods.map(|p| p.expect("Gap detected in 24-hour day period coverage"));
+        for p in &hour_periods {
+            if p.is_none() {
+                return Err("gap between periods");
+            }
+        }
+
+        #[allow(clippy::unwrap_used)] // just checked
```
sffc, line 208: Nit: link to https://github.com/rust-lang/rust/issues/79711

```suggestion
        // TODO: change this to use array_try_map
        // <https://github.com/rust-lang/rust/issues/79711>
        #[allow(clippy::unwrap_used)] // just checked
```

components/datetime/src/provider/day_periods.rs
```diff
@@ -242,122 +246,86 @@ impl DayPeriodRules {
...
+            ]
+            .into_iter()
+            .collect(),
+        )
+        .is_err());
     }
 
-    #[cfg(feature = "compiled_data")]
     #[test]
-    fn test_compiled_data() {
```
sffc, line 303: Issue (optional if you offer justification): `test_compiled_data` is missing from the new code; you are testing the builder but not the end-to-end compiled data

robertbastian: As explained in the PR description, I've split this into tests that test behaviour, here, and tests that test datagen, in datagen. I've then removed the datagen tests because it was just asserting that e.g. `zh` has a specific CLDR value

sffc (approved):

Manishearth (commented):
thanks

Issue #7892, opened 2026-04-21
Remove ICU-defined properties


robertbastian:
The properties 
* `Case_Sensitive`
* `NFC_Inert`
* `NFD_Inert`
* `NFKC_Inert`
* `NFKD_Inert`
* `Segment_Starter`

are ICU-specific properties which ICU uses as part of its normalization, collation, or case mapping implementations. As they are not defined by Unicode, and we don't use these in our implementations, we should not expose them.

sffc:
- @Manishearth Are these exposed under icu_properties?
- @robertbastian Yes
- @sffc We only add properties to icu_properties when people asked for them
- @Manishearth Where was an initial set
- @sffc We should try to only include UCD properties
- @robertbastian Do you include UAX derived properties?
- @sffc Yes
- @robertbastian We should do those in UnicodeSet but not have extra data for them in icu_properties
- @sffc I'm okay marking these as deprecated, as long as we don't have users for them

Conclusion: OK to make these deprecated

sffc:
These are deprecated and can be removed in 3.0

PR #7973, opened 2026-05-18
Deprecate ICU-only properties

robertbastian:
#7892

Also cleaned up datagen to have different code paths for Unicode properties, ICU properties (deprecated), and compatibility properties (probably should be deprecated as well)

## Changelog

`icu_properties`:
* Deprecate some non-Unicode properties


sffc (approved):

sffc (approved):

sffc (approved):

Manishearth (commented):
lgtm

PR #7975, opened 2026-05-18
Address Askama 0.15 deprecation warnings in codegen templates

sffc:
Replaced uninitialized `let` declarations with `decl` in `datetime_formatter.rs.jinja` and `zoned_formatter.rs.jinja`. The warning message from Askama 0.15 suggested using `create`, but that resulted in an "unknown node" error. Using `decl` worked and is consistent with Askama documentation for declaring uninitialized variables.

## Changelog

N/A


sffc:
source: https://askama.rs/en/stable/upgrading.html?highlight=decl#from-askama-v015-to-askama-v016

sffc:
It's not helpful that the warning said to use "create" but the new syntax is actually "decl". But the AI agent figured it out.

robertbastian (approved):

Issue #7962, opened 2026-05-12
Ingest new segmentation rules from Unicode


sffc:
@eggrobin is contributing to new rule-based segmentation data to implement UAX 14 and 29. We should consume that data in ICU4X.

Bugs that the new implementation fixes:
* https://github.com/unicode-org/icu4x/issues/8321
* https://github.com/unicode-org/icu4x/issues/7218
* https://github.com/unicode-org/icu4x/issues/8243

robertbastian:
I wrote a doc

https://docs.google.com/document/d/11F4pJ96unMPhSiBi3JF2jyRpML2K231zqVd_huI66Qk/edit?tab=t.0

PR #7976, opened 2026-05-18
Data-driven neo segmenter tailorings

robertbastian:
Changes the sentence segmenter to use a (hardcoded) data-driven tailoring for Greek.

Also adds tailoring support to the line segmenter

#7962

## Changelog

N/A


sffc (approved):
Observation: the data was already here but it was code-generated around line 1250 of provider/source/src/segmenter/mod.rs, but now it is generated based on a data file, which will presumably make its way into CLDR.

PR #7977, opened 2026-05-18
Read explicit eot class from neo segmenter sources 

robertbastian:
This allows collapsing the eot (end of text) class with another class if that makes sense.

https://github.com/unicode-org/icu4x/issues/7962 

## Changelog

N/A


provider/source/data/segmenter/neo/GraphemeClusterBreakClasses.txt
```diff
@@ -1,4 +1,5 @@
 # Class name ; Class definition in UnicodeSet notation
+eot ; [{eot}]
```
eggrobin: GCB does not use eot at all (there is no eot in the transitions), so I will not be generating this.

provider/source/data/segmenter/neo/WordBreakClasses.txt
```diff
@@ -1,4 +1,5 @@
 # Class name ; Class definition in UnicodeSet notation
+eot ; [{eot}]
```
eggrobin: I think word also does not use eot.

provider/source/data/segmenter/neo/GraphemeClusterBreakClasses.txt
robertbastian: I think conceptually there should still be an EOT class generated at the end of text, even if there's no transition for it. Otherwise I need to start tracking whether a state machine needs EOT logic or not. But I think I can do this on our end

eggrobin (approved):

PR #7978, opened 2026-05-19
Use Maximal deduplication strategy for testdata

sffc:
This change switches the deduplication strategy in make_testdata.rs from RetainBaseLanguages to Maximal. This reduces the size of the generated testdata by removing redundant files that match their parent or root.

## Changelog

N/A


sffc:
This deletes 1682 json files and leaves 4945 json files.

This was motivated by adding zh-TW datetime testdata (I want it to be automatically excluded) but it seems like a good change to make. It makes PRs smaller.

robertbastian (approved):
I feel like there was a reason why we didn't do this years ago.

Manishearth (approved):
probably so the data is easier to inspect. I'm fine thinning this down

robertbastian (approved):

PR #7979, opened 2026-05-19
Replace `LinearNames` data struct by domain specific structs

robertbastian:
Split from #7971 

## Changelog

N/A


sffc (approved):
Praise: Thank you for splitting the refactoring to its own PR

Manishearth:
Hmm, I thought that merging together some of the duplicate structs was a benefit and done deliberately, but maybe the difference in data models is sufficient for this to be better.

sffc:
@Manishearth After @robertbastian's PRs, each of these has only one user, so my judgement was that it was reasonable to rename them. I had asked for it to at least be split into its own PR.

Manishearth:
This appears to have fixed a bug.

```rust
#[test]
fn test_sk_panic() {
    let field_set_with_options = fieldsets::YMDT::long();
    let datetime = DateTime {
        date: Date::try_new_iso(2020, 1, 7).unwrap(),
        time: Time::try_new(8, 25, 7, 0).unwrap(),
    };
    let dtf = DateTimeFormatter::try_new(locale!("sk").into(), field_set_with_options).unwrap();
    println!("Pattern: {:?}", dtf.format(&datetime).pattern());
    println!("Formatted: {}", dtf.format(&datetime));
}
```

<details><summary>AI analysis</summary>

### What is going wrong

In ICU4X 2.2.0, there is a divergence between the pattern items requested during data loading (`try_new`) and the pattern selected during formatting (`format`):

1. **During `try_new`**: To determine which names to load, `DateTimeFormatter` calls `pattern_items_for_data_loading()`. In 2.2.0, this method always inspected `PackedSkeletonVariant::Variant1` (the year-with-era skeleton variant).
2. **Slovak pattern mismatch**: For the Slovak (`sk`) locale, the `Variant1` date pattern for `YMD::long()` is `d. M. y G`, which uses a numeric month (`M`, i.e., `Month(Format)` length `One`).
3. Consequently, during `try_new`, `load_month_names` was only evaluated for numeric month length (`One`) and the wide month names (`MMMM`, `MonthNameLength::Wide`) were **not loaded**.
4. **During formatting**: When formatting a standard CE date (like `2020-01-07`), the year is unambiguous, so `select()` chooses `PackedSkeletonVariant::Standard` (the pattern without era).
5. For `sk`, the `Standard` long date pattern is `d. MMMM y`, which requires `MMMM` (wide month name in genitive case, e.g., `"januára"`).
6. When `FormattedDateTime::write_to_parts` attempts to write the `MMMM` field, it queries the loaded names for `MonthNameLength::Wide`, receives `NotLoaded`, and panics.

### Status on `main`
This bug does not occur on `main` because it was fixed by PR #7979 (`cf351445196094 Replace LinearNames data struct by domain specific structs`), which refactored domain name structures and pattern items loading so that required month name lengths (like `MMMM`) are correctly identified and loaded during initialization.

</details>

Manishearth:
Nope, the PR is https://github.com/unicode-org/icu4x/pull/7986. Sorry for the noise.

PR #7980, opened 2026-05-19
Enable additional CSS line break tests

robertbastian:
#7962 

## Changelog

N/A

### DIFF
--- a/components/segmenter/tests/css_line_break.rs
+++ b/components/segmenter/tests/css_line_break.rs
@@ -87,8 +87,7 @@ fn linebreak_strict() {
     strict("サ々サ", false, &["サ々", "サ"]);
 
     // from css/css-text/line-break/line-break-*-015a.xht
-    // XXX ID x IN in UAX14. But why?
-    strict("サ‥‥サ", false, &["サ‥‥", "サ"]);
+    strict("‥‥サ", false, &["‥‥", "サ"]);
 
     // from css/css-text/line-break/line-break-*-016a.xht
     strict("サ・サ", false, &["サ・", "サ"]);
@@ -121,7 +120,7 @@ fn linebreak_normal() {
     normal("サ々サ", true, &["サ々", "サ"]);
 
     // from css/css-text/line-break/line-break-*-015.xht
-    normal("サ‥‥サ", true, &["サ‥‥", "サ"]);
+    normal("‥‥サ", true, &["‥‥", "サ"]);
 
     // from css/css-text/line-break/line-break-*-016a.xht
     normal("サ・サ", true, &["サ・", "サ"]);
@@ -153,7 +152,7 @@ fn linebreak_loose() {
     loose("サ々サ", true, &["サ", "々", "サ"]);
 
     // from css/css-text/line-break/line-break-*-015.xht
-    loose("サ‥‥サ", true, &["サ", "‥", "‥", "サ"]);
+    loose("‥‥サ", true, &["‥", "‥", "サ"]);
 
     // from css/css-text/line-break/line-break-*-016a.xht
     loose("サ・サ", true, &["サ", "・", "サ"]);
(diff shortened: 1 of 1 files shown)


components/segmenter/tests/css_line_break.rs
```diff
@@ -133,10 +133,9 @@ fn linebreak_normal() {
     normal("サ€サ", true, &["サ", "€サ"]);
 
     // from css/css-text/i18n/unknown-lang/css-text-line-break-pr-normal.html
```
robertbastian, line 134: these characters are allowed to start a line per the test:
https://github.com/web-platform-tests/wpt/blob/748511b875003aae9baa1646b791b10a9e4a4029/css/css-text/i18n/unknown-lang/css-text-line-break-pr-normal.html#L63

components/segmenter/tests/css_line_break.rs
```diff
@@ -182,10 +181,10 @@ fn linebreak_loose() {
     loose("文︙文", false, &["文", "︙", "文"]);
 
     // from css/css-text/i18n/unknown-lang/css-text-line-break-pr-loose.html
```
robertbastian, line 182: these characters are allowed to start a line per the test:
https://github.com/web-platform-tests/wpt/blob/748511b875003aae9baa1646b791b10a9e4a4029/css/css-text/i18n/unknown-lang/css-text-line-break-pr-loose.html#L63

components/segmenter/tests/css_line_break.rs
```diff
@@ -97,7 +97,7 @@ fn linebreak_strict() {
     strict("サ°サ", false, &["サ°", "サ"]);
 
     // from css/css-text/line-break/line-break-*-018.xht
```
robertbastian, line 98: the euro sign is allowed to start a line per the test:
https://github.com/web-platform-tests/wpt/blob/748511b875003aae9baa1646b791b10a9e4a4029/css/css-text/line-break/line-break-normal-018.xht#L48-L54

components/segmenter/tests/css_line_break.rs
```diff
@@ -154,7 +153,7 @@ fn linebreak_loose() {
     loose("サ々サ", true, &["サ", "々", "サ"]);
 
     // from css/css-text/line-break/line-break-*-015.xht
```
robertbastian, line 154: the linked test only tests that there's a break between `‥` and `‥`. From reading the [spec](https://drafts.csswg.org/css-text-3/#line-break-property:~:text=breaks%20between%20inseparable%20characters), I'm unsure about the break between `サ` and `‥`:

> * The following breaks are [...] allowed in [loose](https://drafts.csswg.org/css-text-3/#valdef-line-break-loose):
>     * [...]
>     * breaks between inseparable characters (such as ‥ U+2025, … U+2026) i.e. characters from the Unicode line breaking class IN. [[UAX14]](https://drafts.csswg.org/css-text-3/#biblio-uax14)

If CSS wanted to allow a break between `サ` and `‥`, they would have written "breaks _before_ inseparable characters" instead. anyway, let's just test what the test tests

eggrobin: I agree that the spec does not say so, but another test does test that IN is allowed at the beginnning of a line: https://wpt.fyi/results/css/css-text/i18n/ja/css-text-line-break-ja-in-loose.html?label=experimental&label=master&aligned

And since it was committed by Koji, I assume that is what is actually meant: https://github.com/web-platform-tests/wpt/commits/97bbc7247a16231f4744a47a1d9b3d29633d5292/css/css-text/i18n/ja/css-text-line-break-ja-in-loose.html.

robertbastian: `css-text-line-break-zh-in-loose`, which is the same test as `css-text-line-break-ja-in-loose` is tested in line 188. it's just not _this_ test

eggrobin (approved):

PR #7981, opened 2026-05-19
Add CSS line break tailorings for neo segmenter

robertbastian:
#7962 

## Changelog

N/A


provider/source/src/segmenter/mod.rs
```diff
@@ -1117,7 +1123,7 @@ impl<'a> ParsedNfa<'a> {
         states: &'a str,
         transitions: &'a str,
     ) -> Result<Self, DataError> {
-        let mut eot_class = None;
+        let mut magic_classes = BTreeMap::new();
```
eggrobin, line 1126: Someday, and that day may never come, I will call upon you to use whatever terminology I came up with in the UAX14 text that describes the state machine. But until that day, accept this adjective as a gift.

eggrobin (approved):

provider/data/segmenter/fingerprints.csv
```diff
@@ -1,7 +1,14 @@
 segmenter/break/grapheme/cluster/v1, <singleton>, 10452B, 10346B, 648dd52b775f2e5d
 segmenter/break/grapheme/cluster/v2, <singleton>, 12418B, 12308B, 8f6898afca86f4f3
+segmenter/break/line/override/v2, <lookup>, 63B, 5 identifiers
+segmenter/break/line/override/v2, <total>, 5096B, 4800B, 5 unique payloads
+segmenter/break/line/override/v2, und/cj, 512B, 452B, a50d6ec49fc8301
+segmenter/break/line/override/v2, und/loose, 1104B, 1045B, a87f1a386b275465
```
sffc, line 6: Thought: making these be attributes seems fine… but it might get mess if we have something tailored by both locale and attribute. Let's figure that out when it happens.

provider/data/segmenter/fingerprints.csv
```diff
@@ -1,7 +1,14 @@
...
+segmenter/break/line/override/v2, und/loose_cj, 1544B, 1485B, 124e89ca35e86476
+segmenter/break/line/override/v2, und/normal, 920B, 861B, 9d8ed8209290d213
+segmenter/break/line/override/v2, und/normal_cj, 1016B, 957B, a1163c11890ad275
 segmenter/break/line/v1, <singleton>, 19860B, 19755B, 1abae346855fbab9
-segmenter/break/line/v2, <singleton>, 28756B, 28648B, c3a2cb7dd5b25587
+segmenter/break/line/v2, <singleton>, 35012B, 34905B, e4c91786e70c4825
```
sffc, line 11: Question: why the extra 7 kB on the base model?

sffc (approved):
It makes me smile that the CSS tailorings are data-driven.

provider/data/segmenter/fingerprints.csv
robertbastian: because the state machine needs to be extended to be tailorable

Manishearth (approved):

PR #7986, opened 2026-05-21
Correctly implement numeric month fields for calendars with leap months

robertbastian:
Using `M` or `MM` with the Hebrew or EastAsianTraditional calendars needs to load leap patterns. For EAT these are already being generated, but not loaded, for Hebrew we need to generate them.

## Changelog

`icu_datetime`: Correctly implement numeric month fields for calendars with leap months


components/datetime/src/pattern/names.rs
```diff
@@ -3725,7 +3725,24 @@ impl<FSet: DateTimeNamesMarker> RawDateTimeNames<FSet> {
                 }
 
                 // M..MM, L..LL
-                (FS::Month(_), One | Two) => numeric_field = Some(field),
+                (FS::Month(_), One | Two) => {
+                    if month_provider
```
robertbastian: this is hacky, is there a better way?

components/datetime/src/provider/names.rs
```diff
@@ -26,8 +26,8 @@ icu_provider::data_marker!(
     ///
     /// The full list is:
     /// - 3 is "abbreviated"
-    /// - 4 is "narrow"
-    /// - 5 is "wide"
+    /// - 4 is "wide"
+    /// - 5 is "narrow"
```
sffc: 😳

sffc (commented):
I think:

1. We _should definitely_ enforce a warning in datagen to prevent these from being constructed for the default Hebrew, Chinese, and Dangi patterns
2. We _could_ make `impl TryWriteable` write out an error part
3. We _shouldn't_ make this check significantly slow down the formatting hot path
4. We _probably shouldn't_ entirely prevent power users from making these patterns
5. We _could_ make `M` and `MM` use `MMMMM` behavior in lunisolar, but only after consulting with CLDR

components/datetime/src/pattern/names.rs
sffc: I would prefer to enforce this in datagen. Probably provider/source/src/datetime/semantic_skeletons.rs

robertbastian:
Actually, you've already built a mechanism to use leap patterns with numeric months, it's just not being used. Completely redid the PR

components/datetime/tests/simple_test.rs
```diff
@@ -394,3 +396,90 @@ fn test_flexible_dayperiod_formatting() {
...
+        FixedCalendarDateTimeFormatter::try_new(locale!("ar").into(), fieldsets::YMD::short())
+            .unwrap();
+
+    writeable::assert_writeable_eq!(
+        formatter.format(&Date::try_new_hebrew_v2(5771, Month::new(5), 23).unwrap()),
+        "23‏/5‏/5771 ص",
+    );
+    writeable::assert_writeable_eq!(
+        formatter.format(&Date::try_new_hebrew_v2(5771, Month::leap(5), 23).unwrap()),
+        "23‏/6a‏/5771 ص",
```
robertbastian: `23‏/6a‏/5771 ص`

provider/source/data/debug/datetime/patterns/date/chinese/v1/ym0/en.json
```diff
@@ -7,11 +7,12 @@
...
     0,
-    2
+    4
   ],
   "elements": [
     "MMMM r(U)",
     "MMM r",
-    "M/r"
+    "M/r",
+    "M r"
```
robertbastian, line 16: @sffc how exactly are the patterns selected, if this PR influences this algorithm? why was `M r` not included before, why is it now? it's almost the same as `M/r` after all

components/datetime/tests/simple_test.rs
```diff
@@ -394,3 +396,90 @@ fn test_flexible_dayperiod_formatting() {
...
+        "23\u{200f}/7\u{200f}/5771 ص",
+    );
+    writeable::assert_writeable_eq!(
+        formatter.format(&Date::try_new_hebrew_v2(5772, Month::new(6), 23).unwrap()),
+        "23\u{200f}/6\u{200f}/5772 ص",
+    );
+}
+
+#[test]
+fn test_adar_narrow() {
```
robertbastian, line 453: this is a test for #7728 and already passes on main. the other tests are fixed in this PR

sffc (commented):
This is good; it's a bit unfortunate to call `load_month_names` for every date formatter, but it's the right thing to do, and we can optimize it later if we really need to; for example, for calendars that don't need this, we can add an associated type on one of the scaffold traits.

I don't currently understand why the patterns data are changing. I would like to understand this.

provider/source/src/datetime/names.rs
```diff
@@ -387,8 +388,33 @@ fn months_convert(
             Context::Format,
             "numeric months only found for Context::Format"
         );
+        if calendar == DatagenCalendar::Hebrew {
+            return Ok(MonthNames::LeapNumericWithBase(
+                (&[
+                    // M05L should be 6a
```
sffc, line 392: Issue: Did you invent the 6a/6b thing or does it come from somewhere? Is it already the narrow display name? If so, should you copy it from there, instead of hard-coding the a/b here?

Follow-up OK

provider/source/src/datetime/names.rs
```diff
@@ -736,7 +762,7 @@ impl_symbols_datagen!(
 impl_symbols_datagen!(
     DatetimeNamesMonthBuddhistV1,
     DatagenCalendar::Buddhist,
-    NORMAL_MARKER_LENGTHS,
+    NUMERIC_MONTHS_MARKER_LENGTHS,
```
sffc: Suggestion: change the constants to `DAY_PERIODS_MARKER_LENGTHS` (which is the only remaining use of `NORMAL_MARKER_LENGTHS` and `MONTHS_MARKER_LENGTHS` (since all months should I think be using the same one now)

provider/source/data/debug/datetime/patterns/date/chinese/v1/ym0/en.json
sffc: It's not immediately obvious to me how this PR changes the date patterns being generated. Names, yes, of course, but patterns?

I observe that this is an override on only one specific pattern. Variant1 short: previously it was mapping to "MMM r" but now it maps to "M r"

For yM, Variant1 is intended for full year with era. The pattern should be reachable with fieldset YM and year style WithEra.

The pattern generation is pretty much all in semantic_skeletons.rs

sffc (approved):
Some questions I'd like to see answered, but this fixes bugs. We should merge and follow-up.

provider/source/src/datetime/names.rs
robertbastian: I already invented these for narrow display names. There are no CLDR names we could use, https://unicode-org.atlassian.net/browse/CLDR-19282

sffc (approved):

provider/source/data/debug/datetime/patterns/date/chinese/v1/ym0/en.json
robertbastian: For this case:
* the `standard` pattern is `M/r`
* the `variant0` pattern (full year) is `M/r`
* the `variant1` pattern (full year + era) is `MMM r`

With this PR, loading the last pattern into a `names` that has already loaded the previous patterns will return an error, because the month slot is already used by the length-1 names, so we can't load the length-3 names. The code then transforms the last pattern into `M r`.

I think this is stupid, it should just transform it into `M/r`. I don't really follow how the patterns are selected, is there a design doc or anything? The code has a TODO for https://github.com/unicode-org/icu4x/issues/4478, but that's closed...

sffc:
CAUTION: THIS POST IS 100% AI-GENERATED. IT IS INTENDED TO BE HELPFUL, BUT IT CONTAINS ERRORS.

---

Hi Robert,

I investigated why the datetime patterns data changes in this PR, even though the code changes seem to only impact names data. Here is what I found:

### Reproduction
1. **Baseline (`main` branch)**: Running the testdata generation task on `main` produces data where `datetime/patterns/date/chinese/v1/ym0/en.json` contains 3 elements: `"MMMM r(U)"`, `"MMM r"`, and `"M/r"`.
2. **This Branch**: Running the same testdata generation task on this branch reproduces the change, adding `"M r"` to the `elements` array.

### Trace of the Change
The change in patterns data is an indirect effect of the code changes in `components/datetime/src/pattern/names.rs` and how they interact with the data generation process in `provider/source/src/datetime/semantic_skeletons.rs`:

1. **Loading Numeric Month Names**: This PR modifies `RawDateTimeNames::load_for_pattern` to load month names even for numeric month fields (length 1 or 2). Previously, it did nothing for these fields.
2. **Conflict Detection in Datagen**: The data generation in `semantic_skeletons.rs` uses `FixedCalendarDateTimeNames::load_for_pattern` to validate patterns and detect conflicting fields (via `enforce_consistent_field_lengths`).
3. **New Conflicts/Behaviors**: Because it now tries to load names for numeric months, it can detect conflicts that were not detected before, or avoid panics if data was missing but now falls back to numeric.
4. **Fallback Pattern**: When `semantic_skeletons.rs` cannot find a good match in CLDR, it falls back to a last-resort pattern that joins fields with spaces. This fallback mechanism, combined with the new behavior, resulted in the addition of the `"M r"` pattern in the generated data for Chinese and Dangi calendars.

In summary, the runtime changes caused datagen's validation/fallback logic to produce different patterns.

Cited: This investigation was performed by Jetski, an AI assistant.

robertbastian:
> 4\. **Fallback Pattern**: When `semantic_skeletons.rs` cannot find a good match in CLDR, it falls back to a last-resort pattern that joins fields with spaces. This fallback mechanism, combined with the new behavior, resulted in the addition of the `"M r"` pattern in the generated data for Chinese and Dangi calendars.

This is absolutely not what happens. As stated in my analysis (which is actually mine), the starting pattern for `variant1` is `MMM r`, and the `MMM` gets replaced by `M` because `M` is already loaded. If it would just concatenate fields, we'd end up with `M y`.

Please verify AI output before posting it.

sffc:
Right, it isn't using the last-resort fallback, it is doing an alphabetic-to-numeric substitution.

It wants to find `GyM` but the closest match is `GyMMM`, which is `MMM r`, since there are no eras in the Chinese calendar. Then it replaces `MMM` with `M`. As far as I can tell, this is the CLDR specified behavior. Anything to change it would be a hack on top of the datetime pattern generator. The fix I think is for CLDR to populate data in `GyM`.

sffc:
> Please verify AI output before posting it.

The AI output was 90% correct and made progress on the investigation, so I thought it was worth posting. I am not happy that the post appeared to come from me. It would have been more clear if a bot account such as @gemini-code-assist had posted the comment, but I don't think I can do that.

robertbastian:
> The AI output was 90% correct and made progress on the investigation

It did not reveal anything that I hadn't posted in my comment 20 minutes earlier, and added false information.

sffc:
> > The AI output was 90% correct and made progress on the investigation
> 
> It did not reveal anything that I hadn't posted in my comment 20 minutes earlier, and added false information.

I did not see the comment 20 minutes earlier, or else I likely would not have posted the AI output.

Manishearth:
Seem slike this also fixed things for non-leap months

```rust
#[test]
fn test_sk_panic() {
    let field_set_with_options = fieldsets::YMDT::long();
    let datetime = DateTime {
        date: Date::try_new_iso(2020, 1, 7).unwrap(),
        time: Time::try_new(8, 25, 7, 0).unwrap(),
    };
    let dtf = DateTimeFormatter::try_new(locale!("sk").into(), field_set_with_options).unwrap();
    println!("Pattern: {:?}", dtf.format(&datetime).pattern());
    println!("Formatted: {}", dtf.format(&datetime));
}
```

robertbastian:
what was the bug?

Manishearth:
It panics because it can't find the names.

Not sure of details and I'll be logging off soon, just posting so others can verify. It appears to bisect here, I haven't fully tested this.

robertbastian:
I don't see any code in this PR that would explain this. This PR adds loading for `M` and `MM`, but the resolved pattern here uses `MMMM`. The error is `NamesNotLoaded(ErrorField(Field { symbol: Month(Format), length: Four }))`

However, there suspiciously *is* a data diff for the `sk` Gregorian YMD patterns:

```diff
diff --git a/provider/source/data/debug/datetime/patterns/date/gregorian/v1/ym0d/sk.json b/provider/source/data/debug/datetime/patterns/date/gregorian/v1/ym0d/sk.json
index d9d3b49189..00043429e0 100644
--- a/provider/source/data/debug/datetime/patterns/date/gregorian/v1/ym0d/sk.json
+++ b/provider/source/data/debug/datetime/patterns/date/gregorian/v1/ym0d/sk.json
@@ -5,12 +5,13 @@
     0,
     0,
     3,
-    3,
-    3
+    4,
+    4
   ],
   "elements": [
     "d. MMMM y",
     "d. M. y",
+    "d. MMMM. y GGG",
     "d. M. y GGG"
   ]
 }
```

I don't think this PR necessarily fixes this bug, it might merely hide it.

robertbastian:
Yeah, so this loads names for this pattern, which has `M`:

```
[components/datetime/src/raw/neo.rs:173:13] payload.get(options.length(), PackedSkeletonVariant::Variant1).items = ZeroVec([Field(Field { symbol: Day(DayOfMonth), length: One }), Literal('.'), Literal(' '), Field(Field { symbol: Month(Format), length: One }), Literal('.'), Literal(' '), Field(Field { symbol: Year(Calendar), length: One }), Literal(' '), Field(Field { symbol: Era, length: Three })])
```

but then later tries to format a pattern that has `MMMM`. This is really concerning. It's not fixed by this PR, the pattern data diff hides this by making the loading pattern have `MMMM` as well.

Why are there two different patterns? Why is the pattern datagen so brittle in the first place? @sffc

sffc:
> Why are there two different patterns?

The pattern payloads contain long/short/narrow in the data struct, as opposed to the names, which differentiate long/short/narrow in data marker attributes.

> I don't think this PR necessarily fixes this bug, it might merely hide it.

Any formatter that uses the long patterns will use _only_ the long patterns. However, the code assumes that the patterns for the same skeleton length contain the same month length, an assumption that seems to be violated by this locale. This PR fixes it by forbidding mixed numeric and alphabetic names in the same length.

sffc:
> Why is the pattern datagen so brittle in the first place?

When I wrote this code, I favored putting complexity in datagen as opposed to runtime, and making assumptions at runtime that datagen tries to enforce.

The assumption here is that (1) alphabetic months stay the same length when adding or removing the era field, and (2) if there is an alphabetic month, it is in the pattern that has an era field. The first assumption was being enforced by datagen, but not the second. This PR has the effect of fixing it.

I wish the CLDR data were more self-consistent. There are a number of CLDR issues tracking this, including https://unicode-org.atlassian.net/browse/CLDR-19298

sffc:
Here's the function in the code with the faulty assumption

https://github.com/unicode-org/icu4x/blob/9282b94d05c571da04e132fa5b34b4a2504f9ec4/components/datetime/src/raw/neo.rs#L171

```rust
    /// Borrows a pattern containing all of the fields that need to be loaded.
    #[inline]
    pub(crate) fn pattern_items_for_data_loading(
        &self,
        options: RawOptions,
    ) -> Option<impl Iterator<Item = PatternItem> + '_> {
        Some(
            payload
                .get(options.length(), PackedSkeletonVariant::Variant1)
                .items
                .iter(),
        )
    }
```

It assumes that `Variant1` (the pattern with the era field, which is supposedly the longest and most verbose) contains fields corresponding to all data that needs to be loaded. This assumption was apparently not being enforced in datagen, but it should have been.

robertbastian:
> This assumption was apparently not being enforced in datagen, but it should have been.

It's still not as far as I can tell. This PR definitely doesn't add enforcement.

robertbastian:
> When I wrote this code, I favored putting complexity in datagen as opposed to runtime, and making assumptions at runtime that datagen tries to enforce.

At the risk of repeating myself again: please _state these assumptions_ and please please _test them_.

Manishearth:
Filed https://github.com/unicode-org/icu4x/issues/8278 so I can get this out of my notification backlog.

sffc:
This PR appears to be responsible for a 47% increase in the `construct_and_format` benchmark for YMD

https://unicode-org.github.io/icu4x/benchmarks/perf/components/datetime/index.html

It only impacts the `construct_and_format` bench, not the `format_only` bench. Good we have both.

Manishearth:
Asked Gemini to investigate. 

Basically, this now unconditionally loads month names instead of only doing it when the pattern asks for it.

We could optimize by only doing it for calendars with leap months.

But before we decide to do something like that it would be good to get actual concrete numbers on this. "47% worse" can be a big deal or a nothingburger depending on what the base was; I perceive data loading to be very fast so I'd be surprised if these numbers were actually problematic.

<details>


* **PR Mission**: *"Correctly implement numeric month fields for calendars with leap months"* ([PR #7986](https://github.com/unicode-org/icu4x/pull/7986)).
* **Affected Benchmark**: `construct_and_format` on YMD fieldsets (both `any` and `fixed` calendars, e.g. [`ymd_short/any/construct_and_format/10_cases`](file:///usr/local/google/home/manishearth/dev/worktrees/icu4x-2/components/datetime/benches/datetime.rs#L175-L184) and [`ymd_short/fixed/construct_and_format/10_cases`](file:///usr/local/google/home/manishearth/dev/worktrees/icu4x-2/components/datetime/benches/datetime.rs#L215-L224)).

#### Root Cause
In [`components/datetime/src/pattern/names.rs`](file:///usr/local/google/home/manishearth/dev/worktrees/icu4x-2/components/datetime/src/pattern/names.rs#L3743-L3753):
* **Before PR #7986**: When resolving pattern fields in [`FixedCalendarDateTimeNames::include_for_pattern`](file:///usr/local/google/home/manishearth/dev/worktrees/icu4x-2/components/datetime/src/pattern/names.rs), numeric month fields (`M`, `MM`, `L`, `LL`) did not load any month name data:
  ```rust
  (FS::Month(_), One | Two) => numeric_field = Some(field),
  ```
  Constructing a fixed Gregorian YMD formatter (`construct_fixed_ymd_short`) only needed **1** provider data lookup (the date skeleton pattern `DatetimePatternsDateGregorianV1`).
* **In PR #7986**: To support leap month numeric formatting in Chinese and Hebrew (e.g. formatting leap month 5 as "閏5" or "6a"), PR #7986 changed this to unconditionally call [`load_month_names`](file:///usr/local/google/home/manishearth/dev/worktrees/icu4x-2/components/datetime/src/pattern/names.rs#L2988-L3011) with [`MonthNameLength::Numeric`](file:///usr/local/google/home/manishearth/dev/worktrees/icu4x-2/components/datetime/src/pattern/names.rs#L107):
  ```rust
  (FS::Month(m), One | Two) => {
      self.load_month_names(
          month_provider,
          prefs,
          MonthNameLength::from_field(m, field.length)
              .ok_or(PatternLoadError::UnsupportedLength(error_field))?,
          error_field,
      )?;
      numeric_field = Some(field)
  }
  ```
* **Why it regressed by 47%**:
  * For **all calendars** (including Gregorian, Buddhist, Indian, Persian, Roc, Ethiopian, Japanese, etc., which have no leap months), constructing a numeric-month formatter now performs an additional full data provider lookup (`DatetimeNamesMonth*V1`) through `ZeroTrie` and locale fallback, just to retrieve the dummy `MonthNames::Numeric` variant.
  * This doubled the number of data requests during construction (from 1 to 2), doubling construction overhead while leaving `format_only` unaffected.

#### Recommended Fix
* Calendars without leap months never require numeric month patterns.
* Add a capability constant/flag on [`CldrCalendar`](file:///usr/local/google/home/manishearth/dev/worktrees/icu4x-2/components/datetime/src/scaffold/calendar.rs#L33) (e.g. `const HAS_LEAP_MONTH_PATTERNS: bool = false;`).
* In [`FixedCalendarDateTimeNames::include_for_pattern`](file:///usr/local/google/home/manishearth/dev/worktrees/icu4x-2/components/datetime/src/pattern/names.rs#L3743-L3753), skip calling `load_month_names` for numeric lengths (`One | Two`) when `!C::HAS_LEAP_MONTH_PATTERNS`.
* Prune `MonthNames::Numeric` data generation for non-leap calendars in datagen, saving binary size as well.


</details>

PR #7989, opened 2026-05-21
Fix emoji collation

### LINKED ISSUE #7972: Emoji collation not working
I am trying to sort a list of emoji in an order that is intuitive to the end user (the same as in the [emoji-ordering.txt](https://unicode.org/emoji/charts/emoji-ordering.txt) data file). I believe this corresponds to the CLDR order with locale containing `-u-co-emoji`. For that, I am creating a `Collator` with the `und-u-co-emoji` locale, but that doesn't produce the result I expect.

Consider the following piece of code:

```rs
use icu_collator::Collator;
use icu_collator::options::{CollatorOptions, Strength};
use icu_locale::locale;
use std::cmp::Ordering;

fn main() {
    let mut options = CollatorOptions::default();
    options.strength = Some(Strength::Secondary);
    let collator = Collator::try_new(locale!("und-u-co-emoji").into(), options).unwrap();
    let grinning_face = "😀";
    let grinning_face_with_big_eyes = "😃";
    let red_circle = "🔴";
    // OK
    assert_eq!(collator.compare(grinning_face, grinning_face_with_big_eyes), Ordering::Less);
    // Panics
    assert_eq!(collator.compare(grinning_face, red_circle), Ordering::Less);
}
```

From the data at <https://github.com/unicode-org/cldr/blob/main/common/collation/root.xml>, I expect the code above to succeed. However, I get a panic at the second assertion.

I am sorry if this is not a bug and actually due to a lack of understanding on my end. In this case, I would appreciate any pointer that would help me solve my problem.

robertbastian:
Fixes #7972 

## Changelog

`icu_collator`:
* Fixed an issue where the emoji collation was not loading correctly


hsivonen (approved):
Thanks. Does `und-u-co-search` need the same fix?

robertbastian:
Maybe, we don't generate it. I've made the fix more general

hsivonen (approved):
Thanks!

PR #7994, opened 2026-05-27
Add note in LocaleFallbacker docs about normalization

### LINKED ISSUE #7993: icu_locale::fallback::LocaleFallbacker for "zh-Hans" does not contain "zh-Hans" in its output
Wrote the following test into https://source.chromium.org/chromium/chromium/src/+/main:base/i18n/internal/icu_bridge.rs
```
#[test]
    fn test_fallbacker_zh_hans() {
        let fallbacker = IcuFallbacker::new();
        let fallbacks = fallbacker.fallback_to_vec(b"zh-Hans");
        let fallback_strings: Vec<String> = fallbacks.iter().map(|f| f.to_string()).collect();
        // ICU4X fallback typically produces ["zh-Hans", "zh"] for zh-Hans.
        assert_eq!(fallback_strings, vec!["zh-Hans", "zh"]);
    }
```

This test fails. The only thing I am doing is breaking when I reach "und" locale. From the documentation for the fallbacker, I had the impression that the origin locale would always be output as the first locale in the fallback locales list.

The actual output in this case is just ["zh"].

Manishearth:
Fixes https://github.com/unicode-org/icu4x/issues/7993

I think a simple note and doctest is sufficient. Not sure if this is the best place for it.
## Changelog: N/A

### DIFF
--- a/components/locale/src/fallback/mod.rs
+++ b/components/locale/src/fallback/mod.rs
@@ -20,6 +20,9 @@ mod algorithms;
 /// algorithm. See *[the design doc]* for a detailed description and [#2243](
 /// https://github.com/unicode-org/icu4x/issues/2243) to track alignment with *UTS #35*.
 ///
+/// While the first element produced here is *usually* the input locale,
+/// the fallback iterator normalizes the input locale first, which may strip implied subtags.
+///
 /// If running fallback in a loop, use [`DataLocale::is_unknown()`] to break from the loop.
 ///
 /// # Examples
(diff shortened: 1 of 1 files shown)


danft:
nice :)

components/locale/src/fallback/mod.rs
```diff
@@ -51,6 +51,30 @@ mod algorithms;
 /// assert_eq!(fallback_iterator.get(), &locale!("und").into());
 /// ```
 ///
+/// # Normalization
```
sffc: Issue: The last heading should be `# Examples`

(I pushed a fix)

sffc (commented):
This is already documented on `LocaleFallbackerWithConfig::fallback_for`, but pulling it up into `LocaleFallbacker` is a good idea, and having an example is good, too

sffc (approved):

PR #7995, opened 2026-05-27
agent-driven: Refactor out enforce_consistent_field_length validation…

Manishearth:
Part of https://github.com/unicode-org/icu4x/issues/5488

Minor refactoring that seems useful for comprehending the packedpatterns code, discovered while working on range formatting (#7967, #7966). Probably will be useful as we genericize it, seemed separable from the other work so I split it out.

### DIFF
--- a/provider/source/src/datetime/semantic_skeletons.rs
+++ b/provider/source/src/datetime/semantic_skeletons.rs
@@ -62,6 +62,49 @@ impl<T> PatternsWithDistance<T> {
     }
 }
 
+/// Validates and resolves any conflicting field lengths inside a runtime pattern,
+/// normalizing them in-place to ensure consistency.
+///
+/// This checks the pattern for field conflicts (e.g., `EEE` vs `EEEE`) by loading it against
+/// a names provider. If a conflict is encountered, it invokes the logger callback and
+/// normalizes all conflicting fields to match the style of the first occurrence.
+fn enforce_consistent_field_length(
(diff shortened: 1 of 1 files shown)


provider/source/src/datetime/semantic_skeletons.rs
```diff
@@ -62,6 +62,50 @@ impl<T> PatternsWithDistance<T> {
     }
 }
 
+/// Validates and resolves any conflicting field lengths inside a runtime pattern,
+/// normalizing them in-place to ensure consistency.
+///
+/// This checks the pattern for field conflicts (e.g., `yyyy` vs `yy`) by loading it against
```
sffc: Issue: This comment is wrong I think. It only errors out if the fields load different data. `yyyy` and `yy` load the same data.

This is documented on PatternLoadError::ConflictingField:

```
    /// Fields conflict if they require the same type of data, for example the
    /// `EEE` and `EEEE` fields (short vs long weekday) conflict, or the `M`
    /// and `L` (format vs standalone month) conflict.
```

(since the purpose of this PR is to make the code more readable and understandable, I am making a blocking comment since an error in the docs does not move us toward that goal)

sffc (changes requested):

provider/source/src/datetime/semantic_skeletons.rs
Manishearth: Should have used E as an example, my bad

sffc (approved):

PR #7996, opened 2026-05-28
ScriptExtensions FFI should use the `Script` enum

robertbastian:
We currently use raw `u16`s, even though we have a `Script` type.

I'm not sure whether this is breaking for JS, and whether we care about the breakage for C++ and C, so I've only done this for Dart and Kotlin.

## Changelog

`icu_capi`:
* Dart, Kotlin: use the `Script` type on `ScriptExtension` APIs

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -367,13 +367,14 @@ jobs:
         rm dylib
 
     - name: Run `cargo make ci-job-test-dart` in fetch mode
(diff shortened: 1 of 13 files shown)


sffc (approved):

sffc (approved):
do we have a way to track that we turn the dart test back on again

robertbastian:
I do it when I do a Dart release

PR #7997, opened 2026-05-28
Make CCC a transparent newtype, deprecate `icu4c_value` conversions

robertbastian:
`CanonicalCombiningClass` is unique in the sense that the numeric values are define by Unicode, not by ICU4C. Currently they are only accessible through `to_icu4c_value`/`from_icu4c_value`, which are pretty much the only use cases of these methods (there are some internal ones that are easy to replace as well).

Instead, we should make the field public and allow clients to treat CCC as an integer.

## Changelog

`icu_properties`:
* Deprecate `to_icu4c_value`/`from_icu4c_value` on enumerated properties
* Expose `CanonicalCombiningClass`' integer field

### DIFF
--- a/components/properties/src/props.rs
+++ b/components/properties/src/props.rs
@@ -75,7 +75,7 @@ macro_rules! create_const_array {
                     $(
                         Self::$i => databake::quote!(icu_properties::props::$enum_ty::$i),
                     )*
-                    Self(v) => databake::quote!(icu_properties::props::$enum_ty::from_icu4c_value(#v)),
+                    Self(v) => databake::quote!(icu_properties::props::$enum_ty(#v)),
                 }
             }
         }
@@ -178,10 +178,12 @@ pub struct BidiClass(pub(crate) u8);
 
 impl BidiClass {
     /// Returns an ICU4C `UBidiClass` value.
+    #[deprecated]
     pub const fn to_icu4c_value(self) -> u8 {
         self.0
     }
     /// Constructor from an ICU4C `UBidiClass` value.
+    #[deprecated]
     pub const fn from_icu4c_value(value: u8) -> Self {
         Self(value)
     }
@@ -278,10 +280,12 @@ pub struct NumericType(pub(crate) u8);
 
 impl NumericType {
     /// Returns an ICU4C `UNumericType` value.
+    #[deprecated]
     pub const fn to_icu4c_value(self) -> u8 {
         self.0
     }
     /// Constructor from an ICU4C `UNumericType` value.
+    #[deprecated]
     pub const fn from_icu4c_value(value: u8) -> Self {
         Self(value)
     }
@@ -829,10 +833,12 @@ pub struct Script(pub(crate) u16);
 
 impl Script {
     /// Returns an ICU4C `UScriptCode` value.
+    #[deprecated]
     pub const fn to_icu4c_value(self) -> u16 {
         self.0
     }
     /// Constructor from an ICU4C `UScriptCode` value.
+    #[deprecated]
     pub const fn from_icu4c_value(value: u16) -> Self {
         Self(value)
     }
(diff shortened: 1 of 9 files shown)


provider/source/src/properties/script.rs
```diff
@@ -304,10 +304,6 @@ mod tests {
         assert!(swe.has_script32(0xFDF2, Script::Arabic)); // main Script value
         assert!(!swe.has_script32(0xFDF2, Script::Syriac));
         assert!(swe.has_script32(0xFDF2, Script::Thaana));
-
-        // The ICU4J comment for this test says:
-        // An unguarded implementation might go into an infinite loop.
-        assert!(!swe.has_script32(0x0640, Script::from_icu4c_value(0xAFFE)));
```
sffc, line 310: Issue: Unrelated test deletion

components/properties/src/props.rs
```diff
@@ -1575,14 +1590,16 @@ make_enumerated_property! {
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[allow(clippy::exhaustive_structs)] // newtype
 #[repr(transparent)]
-pub struct CanonicalCombiningClass(pub(crate) u8);
+pub struct CanonicalCombiningClass(pub u8);
```
sffc, line 1650: Question/Issue/Suggestion/Nit: should we give the field a better name?

sffc, line 1650: Question/Issue: Is there code that assumed an invariant about the value being valid, e.g. to cast it to an enum or something like that?

sffc (changes requested):

components/properties/src/props.rs
robertbastian: CCC _is_ a number, so I think the newtype pattern with a tuple struct applies here

robertbastian: Not in our code base. That would be incorrect anway: 

> The values in the Canonical_Combining_Class field in UnicodeData.txt are **numerical values** used in the Canonical Ordering Algorithm. Some of those numerical values also have explicit symbolic labels as property value aliases, to make their intended application more understandable.

https://www.unicode.org/reports/tr44/#Canonical_Combining_Class_Values, emphasis mine

Manishearth (approved):

components/properties/src/props.rs
Manishearth: I agree

components/properties/src/props.rs
```diff
@@ -1158,10 +1166,12 @@ pub struct EastAsianWidth(pub(crate) u8);
 
 impl EastAsianWidth {
     /// Returns an ICU4C `UEastAsianWidth` value.
+    #[deprecated]
```
Manishearth: issue: we should say why these are deprecated and what they are deprecated in favor of

Manishearth (dismissed):
Actually I'm not sure i'm on board with deprecating to/from integer conversions with no replacement. I think it's useful. Maybe it isn't that useful?

We've established that the utility of ICU4C interop is questionable, but I think integer conversions are more useful.

I'm in favor of the CCC changes.

robertbastian:
> I think it's useful. Maybe it isn't that useful?

We can link a ticket in the deprecation note where people can complain if they need them. Deprecated APIs are still _there_.

sffc:
If the discriminants are not going to be included in the [RSCD](https://github.com/unicode-org/properties/issues/546#issuecomment-4512837740), we should mark them deprecated and make an issue where clients can surface their use cases. If there are such clients, it will help us make a case to upstream the data file into the RSCD. If we leave them as-is, then we don't have a way to communicate with the clients that we need their input before potentially deleting them in 3.0.

sffc (approved):

eggrobin:
(Re RSCD, you can now link to the publicly available https://www.unicode.org/reports/tr44/tr44-37.html#Directory_Structure which introduces the term.)

PR #7999, opened 2026-05-29
Add generic ZeroFrom impls for PluralElements

sffc:
Adds a generic `ZeroFrom` implementation for `PluralElements` and `PluralElementsInner`. This allows mapping the lifetime of the inner type (e.g., from `PluralElements<Pattern<'a>>` to `PluralElements<Pattern<'b>>`).

Enables borrowing and zero-copy operations in builders (like `PackedPatternsBuilder`) that need to duplicate `PluralElements` data with a local lifetime, avoiding the need for `Clone`.

## Changelog

icu_plurals: Add generic `ZeroFrom` implementation for `PluralElements`
  * New trait implementations: `impl<'a, T, C> ZeroFrom<'a, PluralElements<C>> for PluralElements<T>`

### DIFF
--- a/components/plurals/src/lib.rs
+++ b/components/plurals/src/lib.rs
@@ -849,6 +849,45 @@ pub(crate) struct PluralElementsInner<T> {
     explicit_one: Option<T>,
 }
 
+impl<'a, T, C> zerofrom::ZeroFrom<'a, PluralElementsInner<C>> for PluralElementsInner<T>
+where
+    T: zerofrom::ZeroFrom<'a, C>,
+{
+    fn zero_from(other: &'a PluralElementsInner<C>) -> Self {
+        PluralElementsInner {
+            zero: other
+                .zero
+                .as_ref()
+                .map(|x| zerofrom::ZeroFrom::zero_from(x)),
+            one: other.one.as_ref().map(|x| zerofrom::ZeroFrom::zero_from(x)),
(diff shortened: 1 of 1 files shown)


components/plurals/src/lib.rs
```diff
@@ -849,6 +849,45 @@ pub(crate) struct PluralElementsInner<T> {
...
+        }
+    }
+}
+
+impl<'a, T, C> zerofrom::ZeroFrom<'a, PluralElements<C>> for PluralElements<T>
+where
+    T: zerofrom::ZeroFrom<'a, C>,
+{
+    fn zero_from(other: &'a PluralElements<C>) -> Self {
+        PluralElements(zerofrom::ZeroFrom::zero_from(&other.0))
```
robertbastian:
```suggestion
        other.as_ref().map(|p| zerofrom::ZeroFrom::zero_from(p))
```

robertbastian (commented):
do you actually need `PluralElementsInner`, or did you just add it as an implementation detail for `PluralElements`? if you need it, it should also use `map` (which you'll have to move from `PluralElements::map` to `PluralElementsInner::map`)

Manishearth (approved):

sffc:
good idea. I moved the as_ref and map impls to PluralElementsInner, which allows ZeroFrom to be written on top of them.

(the agent didn't understand this, I had to do it mostly manually)

Manishearth (approved):

PR #8000, opened 2026-05-29
Validate enumerated property defaults against UCD

robertbastian:
The UCD contains an entry of the shape

`# @missing: 0000..10FFFF; <name>`

for each property, which we treat as the default.

#4602 

## Changelog

N/A

### DIFF
--- a/provider/source/src/properties/enum_codepointtrie.rs
+++ b/provider/source/src/properties/enum_codepointtrie.rs
@@ -62,7 +62,10 @@ impl SourceDataProvider {
         let ucd_default = read_to_string
             .lines()
             .find_map(|l| {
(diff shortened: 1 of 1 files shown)


sffc (approved):

provider/source/src/properties/enum_codepointtrie.rs
```diff
@@ -63,7 +57,50 @@ impl SourceDataProvider {
             ),
         };
 
-        for line in self.unicode()?.read_to_string(&file)?.lines() {
+        let read_to_string = self.unicode()?.read_to_string(&file)?;
+
+        let ucd_default = read_to_string
```
Manishearth, line 62: issue (followup): this code is getting more and more complicated and is why I would like all of the stuff dealing directly with the UCD format to be moved to a separate file

Manishearth (approved):

PR #8001, opened 2026-05-29
Generate enumerated property constants from `PropertyDiscriminants.txt`

robertbastian:
Instead of maintaining the constants manually and then having unit tests against `PropertyDiscriminants.txt`, we should just generate them.

The first two commits reorder the already generated properties and values in `icu_capi`. As all values have explicit discriminants, this doesn't break semver (even though CSC complains, that's a false positive). It also simplifies some generated code as it makes these enums "contiguous". The third commit contains the meat of this PR, the fourth commit is optional, I think it makes sense to do this, if only to make the FFI enums contiguous.

## Changelog

N/A


sffc (approved):

Manishearth (approved):
always happy to see more of this mess being generated

PR #8002, opened 2026-05-29
Safety improvements for fixed_decimal

Manishearth:


sffc (commented):
Thought: I wish this fn in smallvec had a clear `# Safety` but I need to read the docs to figure out the condition.

Manishearth:
https://docs.rs/smallvec/latest/smallvec/struct.SmallVec.html#method.from_buf_and_len_unchecked (this function is just the const version of that one):


> Constructs a new SmallVec on the stack from an A without copying elements. Also sets the length. The user is responsible for ensuring that len <= A::size().

robertbastian (approved):

PR #8003, opened 2026-05-29
Safety improvements for resb

Manishearth:

### DIFF
--- a/utils/resb/src/binary/helpers.rs
+++ b/utils/resb/src/binary/helpers.rs
@@ -97,10 +97,18 @@ pub fn option_i32_tuple<'de, D: Deserializer<'de>>(
 
 /// Casts a slice of byte to a slice of T
 ///
+/// # Errors
+///
+/// Errors if `T` is a Zero-Sized Type (ZST), or if the length or alignment of the byte slice is incorrect.
+///
 /// # Safety
+///
 /// Alignment and length are checked, however the caller has to guarantee that the byte representation is valid for type T.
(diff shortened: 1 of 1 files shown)


utils/resb/src/binary/helpers.rs
```diff
@@ -98,8 +98,11 @@ pub fn option_i32_tuple<'de, D: Deserializer<'de>>(
 /// Casts a slice of byte to a slice of T
 ///
 /// # Safety
-/// Alignment and length are checked, however the caller has to guarantee that the byte representation is valid for type T.
+///
+/// 1. Type `T` must not be a Zero-Sized Type (ZST).
+/// 2. Alignment and length are checked, however the caller has to guarantee that the byte representation is valid for type T.
 pub unsafe fn cast_bytes_to_slice<T, E: Error>(bytes: &[u8]) -> Result<&[T], E> {
+    assert!(size_of::<T>() > 0, "ZSTs are not supported");
     if bytes.as_ptr().align_offset(align_of::<T>()) != 0 || bytes.len() % size_of::<T>() != 0 {
```
sffc: Observation: if `size_of::<T>() == 0` then we get NaN here and return an error. So, this doesn't harden security, but materially changes the function contract.

I think this is _fine_, but please include a changelog.

Manishearth: I'm going to change this to have an explicit check.

sffc (approved):
I was okay with the panic since it is statically analyzable but this is fine too

PR #8004, opened 2026-05-29
Safety improvements for tinystr

Manishearth:

### DIFF
--- a/utils/tinystr/src/int_ops.rs
+++ b/utils/tinystr/src/int_ops.rs
@@ -129,6 +129,10 @@ impl Aligned4 {
     pub const fn to_ascii_lowercase(&self) -> [AsciiByte; 4] {
         let word = self.0;
         let result = word | (((word + 0x3f3f_3f3f) & !(word + 0x2525_2525) & 0x8080_8080) >> 2);
+        // SAFETY: The bitwise arithmetic identifies uppercase ASCII bytes ('A'..='Z') in the word,
+        // shifts that match to produce 0x20 for those bytes, and ORs them to map to lowercase ('a'..='z').
+        // All other bytes are ORed with 0 and remain unchanged.
+        // Since all input bytes are valid ASCII (0..=127), this mapping preserves the ASCII range (0..=127) on all bytes.
         unsafe { AsciiByte::to_ascii_byte_array(&result.to_ne_bytes()) }
     }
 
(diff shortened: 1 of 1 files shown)


utils/tinystr/src/int_ops.rs
```diff
@@ -129,19 +129,25 @@ impl Aligned4 {
     pub const fn to_ascii_lowercase(&self) -> [AsciiByte; 4] {
         let word = self.0;
         let result = word | (((word + 0x3f3f_3f3f) & !(word + 0x2525_2525) & 0x8080_8080) >> 2);
+        // SAFETY: The input word consists of valid ASCII bytes (0..=127) since it originates from a validated TinyAsciiStr.
+        // The bitwise arithmetic preserves the ASCII range (0..=127) on each byte, satisfying the safety precondition.
```
sffc: this condition is hard to verify. Can we make it easier to verify?

This is a good catch. We lack safety comments here that we really should have.

Manishearth: Made an attempt. This is kind of verbose, which is fine,  I feel like there's a simpler proof tantalizingly close but I can't figure it out. I naïvely thought "& with 0x8080" would always work except not for the case wehre it is 0x808080 itself, and also we right shift by 2 which means low bits can spill into the next byte's high bits

Manishearth: Made it less verbose.

Verbose version: https://github.com/unicode-org/icu4x/compare/cc4f6167e79dc808274a61d8951187d2f9d37dc2..e0ea5e22158dea673a864124b04a7d55017d8c15

utils/tinystr/src/int_ops.rs
```diff
@@ -129,19 +129,32 @@ impl Aligned4 {
     pub const fn to_ascii_lowercase(&self) -> [AsciiByte; 4] {
         let word = self.0;
         let result = word | (((word + 0x3f3f_3f3f) & !(word + 0x2525_2525) & 0x8080_8080) >> 2);
+        // SAFETY: The bitwise arithmetic identifies uppercase ASCII bytes ('A'..='Z') in the word,
+        // shifts that match to produce 0x20 for those bytes, and ORs them to map to lowercase ('a'..='z').
+        // All other bytes are ORed with 0 and remain unchanged.
+        // Since all input bytes are valid ASCII (0..=127), this mapping preserves the ASCII range (0..=127) on all bytes.
```
sffc: Thought: It's a lot of words that doesn't really make it easier to reason about the bit operation.

To be convinced that this is safe, all you need to do is observe that `word` is ascii (by invariant), and it is being bitwise-ORed with a subset of `0x2020_2020`, which does not create non-ASCII bytes.

Maybe:

```rust
/// SAFETY: The existing word is ASCII, and it is bitwise-ORed with a subset of 0x2020_2020,
/// which keeps the word in the ASCII range (all bytes <= 127).
```

Manishearth: That's not quite correct, but I updated it with the correct version

utils/tinystr/src/int_ops.rs
```diff
@@ -260,6 +273,8 @@ impl Aligned8 {
                 & !(word + 0x2525_2525_2525_2525)
                 & 0x8080_8080_8080_8080)
                 >> 2);
+        /// SAFETY: The existing word is ASCII, and it is bitwise-ORed with a subset of 0x80_80_80_...,
```
sffc:
```suggestion
        /// SAFETY: The existing word is ASCII, and it is bitwise-ORed with a subset of 0x20_20_20_...,
```

sffc (commented):
Your safety comments for TinyStr4 and TinyStr8 are inconsistent with each other. I think there should be a standard safety comment for each of the 3 operations (lowercase, uppercase, titlecase) that is easy to reason about, and you should put the same comment in both TinyStr4 and TinyStr8.

The safety comment should focus on the important thing:

- The `word` in all cases is guaranteed ASCII by invariant
- Lowercase: ORs in 0x20 bits, which keeps ASCII as ASCII
- Uppercase: ANDs; doesn't actually matter which bits because ASCII cannot be ANDed to non-ASCII
- Titlecase: ORs 0x20 (safe) and then ANDs (safe)

Sorry for being nitpicky about this but it seems like the place to be nitpicky.

Manishearth:
Done. I had forgotten to update the others.

I went ahead and split out `mask` which makes it easier to talk about ANDing and ORing since otherwise there are many ANDs and ORs which makes your concise version of the comment harder to reason about, and I realized that was the actual issue were were circling around

utils/tinystr/src/int_ops.rs
sffc: no, it is ANDed with a subset of 0xCF (!0x20), but this doesn't matter, because AND can't introduce new bits

sffc (commented):
lowercase and titlecase look good. one more iteration on uppercase

utils/tinystr/src/int_ops.rs
Manishearth: Ah i fixed this for title but didn't for upper.

sffc (approved):
Praise: now I can clearly see that this code is safe without needing to understand how ASCII case mapping works.

Manishearth:
Bypassing network failures

PR #8008, opened 2026-05-29
Tuple layout fix for resb

### LINKED ISSUE #8007: resb/zoneinfo64 soundness: relies on tuple layout of (i32, i32)
https://github.com/unicode-org/icu4x/blob/b7c3aa954b8fb7fcf57957bf67d23509ae83023e/utils/resb/src/binary/helpers.rs#L82-L83

This isn't valid.

We really need to stop making assumptions about Rust tuple layouts where we don't need to.

This crate should be written to work with either `[i32; 2]` or `#[repr(C)] struct I32Pair`.

I didn't go ahead and fix this because this is public API on resb and affects zoneinfo64.

Marking as tech debt priority, safety issues are priority.

cc @sffc @robertbastian

Manishearth:
Fixes #8007 

## Changelog

resb: (breaking) APIs produce a typed `I32Pair` instead of `(i32, i32)` for maximum layout soundness
resb: (breaking) `cast_bytes_to_slice` replaced with type-specific cast functions

### DIFF
--- a/utils/resb/src/binary/helpers.rs
+++ b/utils/resb/src/binary/helpers.rs
@@ -75,26 +75,44 @@ pub fn option_u32<'de, D: Deserializer<'de>>(
 }
 
 /// A layout-guaranteed pair of 32-bit signed integers.
+///
+/// # Safety Layout Invariants
+///
+/// We apply `#[repr(C)]` to guarantee that this struct has a stable and defined layout matching the C ABI:
+/// 1. The fields are laid out sequentially in declaration order.
+/// 2. Since `i32` has size 4 and alignment 4, there is no padding between the first and second fields.
+/// 3. The total size of the struct is exactly 8 bytes, and the alignment matches the alignment of `i32` (4 bytes).
+///
+/// This ensures that casting a correctly aligned 4-byte raw byte slice of size `8 * N` to `&[I32Pair]` is fully sound and stable.
 #[repr(C)]
+#[allow(
+    clippy::exhaustive_structs,
+    reason = "Stable struct representing a specific C layout"
+)]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct I32Pair(pub i32, pub i32);
 
 /// TODO
 pub fn i32_tuple<'de, D: Deserializer<'de>>(deserializer: D) -> Result<&'de [I32Pair], D::Error> {
     let bytes = <&[u8]>::deserialize(deserializer)?;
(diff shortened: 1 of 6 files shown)


utils/resb/src/binary/helpers.rs
```diff
@@ -74,24 +74,27 @@ pub fn option_u32<'de, D: Deserializer<'de>>(
...
+
 /// TODO
-pub fn i32_tuple<'de, D: Deserializer<'de>>(
-    deserializer: D,
-) -> Result<&'de [(i32, i32)], D::Error> {
+pub fn i32_tuple<'de, D: Deserializer<'de>>(deserializer: D) -> Result<&'de [I32Pair], D::Error> {
     let bytes = <&[u8]>::deserialize(deserializer)?;
-    // Safety: all byte representations are valid (i32, i32)
+    // Safety: all byte representations are valid I32Pair
     unsafe { cast_bytes_to_slice(bytes) }
```
sffc: 🤖 This comment is written by an AI agent working with @sffc

Nice fix! Using `#[repr(C)] struct I32Pair` is much safer than assuming tuple layout.

The safety comment on line 91 is good, but could be slightly more explicit, e.g., 'Safety: all byte representations are valid for `i32`, and `I32Pair` has no padding and is `repr(C)`'.

sffc (commented):
I agree with the agent that if we are being more pedantic with safety comments, then you should be more explicit that the repr(C) makes it safe

Manishearth:
The safety comment it produced is more verbose than I would have, but essentially correct, and I won't begrudge it verbosity when it comes to safety.

sffc (changes requested):
It's confusing that `cast_bytes_to_slice` has its own safety requirement which is _almost_ complete except for needing Plain Old Data.

I think it would make this whole thing a lot more clear if you used zerovec or, actually, zerocopy (which deals with POD and knows how to do alignment checks).

Or, you could make a function that first casts from `&[u8]` to `&[u32]` (checking alignment only) and then another function that casts from `&[u32]` to `&[I32Pair]` (checking POD)?

Manishearth:
The latter seems fine, perhaps. I'd rather not add dependencies here, and resb's alignment requirements mean it doesn't fit existing zerocopy models.

Manishearth:
`cast_bytes_to_slice` was public API so these new typed helpers are public too. Not sure if they ought to be.

The `resb` crate needs a fair amount of polish...

sffc (approved):
Much easier to review now. thanks

PR #8009, opened 2026-05-29
Genericize serde impl of GenericPackedPatterns

Manishearth:
This works by moving a bunch of the custom serde logic over into PluralElements, so that the generic code in packed patterns just has to call it.

Note that this code only partially supports plurals. Some of the changes here bring it closer to full support.

To be reviewed commit by commit.

I tried to do this in a way that minimizes the complexity of the diff. First by moving PluralElements-specific serialization logic onto PluralElements, then genericizing from_packed, then moving over to genericized helpers that take in closures for converting between the human types.

I could have introduced a trait but I didn't want there to be a public API. This means in the end we still will have concrete `impl Deserialize for GenericPackedPatterns<Foo>` impls, but those concrete impls will be small.

## Changelog: N/A

The trait impl on PluralElements is a new trait impl, but it's a new trait impl on provider types.

### DIFF
--- a/components/datetime/src/provider/packed_pattern.rs
+++ b/components/datetime/src/provider/packed_pattern.rs
@@ -577,10 +577,69 @@ impl<'data> GenericPackedPatterns<'data, PluralElementsPackedULE<ZeroSlice<Patte
 mod _serde {
     use super::*;
     use crate::provider::pattern::reference;
-    use serde::Deserialize;
+    use serde::{Deserialize, Deserializer};
     #[cfg(feature = "datagen")]
-    use serde::Serialize;
+    use serde::{Serialize, Serializer};
     use zerovec::{ule::VarULE, VarZeroSlice, VarZeroVec};
+
+    pub(crate) trait PackedElementDeserialize<'de>: Sized {
+        fn deserialize_element<D>(deserializer: D) -> Result<Self, D::Error>
+        where
+            D: Deserializer<'de>;
+    }
+
+    #[cfg(feature = "datagen")]
+    pub(crate) trait PackedElementSerialize {
+        fn serialize_element<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+        where
+            S: Serializer;
+    }
+
+    #[derive(serde::Deserialize)]
+    #[cfg_attr(feature = "datagen", derive(serde::Serialize))]
+    #[serde(transparent)]
+    pub(crate) struct PackedElementWrap<T>(
+        #[cfg_attr(
+            feature = "serde",
+            serde(
+                deserialize_with = "PackedElementDeserialize::deserialize_element",
+                bound(deserialize = "T: PackedElementDeserialize<'de>")
+            )
+        )]
+        #[cfg_attr(
+            feature = "datagen",
+            serde(
+                serialize_with = "PackedElementSerialize::serialize_element",
+                bound(serialize = "T: PackedElementSerialize")
+            )
+        )]
+        pub T,
+    );
+
+    impl<'de, T> PackedElementDeserialize<'de> for PluralElements<T>
+    where
+        T: Deserialize<'de>,
+    {
+        fn deserialize_element<D>(deserializer: D) -> Result<Self, D::Error>
+        where
+            D: Deserializer<'de>,
+        {
+            icu_plurals::provider::deserialize_elements_flat_singleton(deserializer)
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


Manishearth:
Rebased. I would still prefer this be deprioritized over the other two PRs: I still think this could be simplified further. So it's ready for review, but if we wait longer I might be able to improve on it.

(I've spent quite a bit of time on arranging this code already but I think there's scope for more)

components/plurals/src/lib.rs
```diff
@@ -1135,3 +1152,48 @@ impl<T: PartialEq> PluralElements<T> {
...
+            }
+        } else {
+            let inner = PluralElementsInner::deserialize(deserializer)?;
+            Ok(PluralElements(inner))
+        }
+    }
+}
+
+#[cfg(feature = "datagen")]
+impl<T: serde::Serialize> serde::Serialize for PluralElements<T> {
```
sffc: Issue: we should not land `impl Serialize for PluralElements` because:

1. This is a public API change, even if it is feature = "datagen", and the serialization proposed here needs much more thought if it is to become the permastable format for this type
2. People needing this should use `PluralElementsPackedULE` which is much more efficient for serde storage. If we were to ever ship `impl Serialize for PluralElements`, it should probably use a `PluralElementsPackedULE` representation.

sffc (changes requested):
There's a lot in here but I'll post an initial comment that needs discussion

components/plurals/src/lib.rs
Manishearth: A simple thing I can do is just move this back into datetime as free functions.

Manishearth:
Argh, agent broke my VCS state. Give me a minute to undo this.

Manishearth:
Reverted changes.  817c431  is the commit you looked at.

components/datetime/src/provider/packed_pattern.rs
```diff
@@ -679,7 +749,7 @@ mod _serde {
                 }
             }
 
-            let elements = pack_fn(human.elements);
+            let elements = pack_fn(human.elements.into_iter().map(|w| w.0).collect());
```
Manishearth, line 752: it's not great that we have to do this, but this is also human deser so this type of thing isn't too bad

We _can_ avoid this allocation: instead of using PackedElementWrap we just use `T` here and use `deserialize_with` on the fields, but then we need to write manual `serialize_slice`/`deserilaize_slice`/vec impls. It's uglier.

Manishearth:
So it was not easy to move this into datetime as free functions: all of the serialization stuff uses the private PluralElementsInner helper type. We could make it public, but instead I ended up adding two helper functions to `icu_plurals::provider`, and then adding some serde wrapper scaffolding.

PluralElementsPackedULE does not work here, because its human readable serialization would include `[metadata, T]` with `FourBitMetadata`. We are constrained by a stable serde representation, and the serde representation stores singleton plural objects flat without any surrounding serde list.

sffc:
I'm still not clear on exactly your goal here. I want to try doing this. I wrote this code and have a better sense than an AI agent of how it should change.

This PR is huge and basically rewrites all the serde code. I do not understand _why_.

In other words, please state the impl that is missing that you think you need, and I will take a look and get that impl working for you.

Manishearth:
Discussed in person. Documenting for others: 

 - The goal is for GenericPackedPatterns<T> to have serde impls that work for both PluralElements and RangePatterns
 - The challenging bit here is that GenericPatternsBuilder::build_generic takes in a closure for the pack function (and as you can see in this CL, the serialization code needs an unpack function too). This needs to be threaded in to the serde impl. Our choices are to either have `trait GenericPackedItem {pack_fn, unpack_fn}` and `impl <T: GenericPackedItem> Serialize/Deserialize` (this is a publicly exposed new trait), or to have `impl Serialize/Deserialize for GenericPatternsBuilder<PluralElements or RangePatterns>` where the impl calls a helper `(de)serialize_helper()` that takes in `pack_fn` and `unpack_fn` as closures. I picked the latter to avoid adding a trait to our public `Serialize` impls.
 - Also the actual structure of this code is very much my doing and not the AI: I went through a bunch of different ways of doing this and settled on this. AI wrote the code, but the choice of how to do this is mine.

Manishearth:
Shane understands the challenges here and is going to try recreating my PR on his own since he understands this code better.

Manishearth:
#8024

PR #8013, opened 2026-05-29
Genericize skeleton-matching code in datagen

Manishearth:
Progress on https://github.com/unicode-org/icu4x/issues/5448

Simplifies datagen code so it can be easily reused for range patterns.

I figured out what to do here by first trying to write datagen code and noting the reusable code.

You can see this in use at https://github.com/unicode-org/icu4x/pull/8014

## Changelog: N/A

provider/datagen code only

### DIFF
--- a/provider/source/src/datetime/mod.rs
+++ b/provider/source/src/datetime/mod.rs
@@ -5,6 +5,11 @@
 use crate::cldr_serde;
 use crate::SourceDataProvider;
 use icu::calendar::AnyCalendarKind;
+use icu::datetime::options::Length;
+use icu::datetime::provider::fields::components;
+use icu::datetime::provider::packed_pattern::{
+    GenericLengthElements, GenericPackedPatternsBuilder,
+};
 use icu::datetime::provider::skeleton::reference::Skeleton;
 use icu::datetime::provider::skeleton::SkeletonError;
 use icu_provider::prelude::*;
@@ -276,6 +281,52 @@ where
         }
     }
     result
+}
+
+pub(crate) fn resolve_packed_patterns_builder<'data, T, F>(
+    provider: &'data SourceDataProvider,
+    locale: &DataLocale,
+    calendar: Option<DatagenCalendar>,
+    attributes: &DataMarkerAttributes,
+    to_components_bag: impl Fn(Length, &DataMarkerAttributes, &cldr_serde::ca::Dates) -> components::Bag,
+    mut match_fn: F,
+) -> Result<
+    (
+        GenericPackedPatternsBuilder<T>,
+        &'data cldr_serde::ca::Dates,
+    ),
+    DataError,
+>
+where
+    T: Clone + PartialEq,
+    F: FnMut(Length, &components::Bag, &cldr_serde::ca::Dates) -> T,
+{
+    let data = provider.get_dates_resource(locale, calendar)?;
+
+    let [long, medium, short] = [Length::Long, Length::Medium, Length::Short].map(|length| {
+        let components = to_components_bag(length, attributes, data);
+        match_fn(length, &components, data)
+    });
+
+    let builder = GenericPackedPatternsBuilder {
+        standard: GenericLengthElements {
+            long: long.clone(),
+            medium: medium.clone(),
+            short: short.clone(),
+        },
+        variant0: Some(GenericLengthElements {
+            long: long.clone(),
+            medium: medium.clone(),
+            short: short.clone(),
+        }),
+        variant1: Some(GenericLengthElements {
+            long: long.clone(),
+            medium: medium.clone(),
+            short: short.clone(),
+        }),
+    };
+
+    Ok((builder, data))
 }
 
 #[cfg(test)]
--- a/provider/source/src/datetime/semantic_skeletons.rs
+++ b/provider/source/src/datetime/semantic_skeletons.rs
@@ -24,41 +24,34 @@ mod tests;
 
 type VariantPatternsElement<'a> = PatternsWithDistance<PluralElements<runtime::Pattern<'a>>>;
 
-/// A set of patterns that may be used in the same formatter depending on input variations.
-#[derive(Debug)]
-struct VariantPatterns<'a> {
-    standard: VariantPatternsElement<'a>,
-    variant0: Option<VariantPatternsElement<'a>>,
-    variant1: Option<VariantPatternsElement<'a>>,
-}
-
-impl<'a> VariantPatterns<'a> {
-    pub fn iter_in_quality_order_mut(
-        &mut self,
-    ) -> impl Iterator<Item = &mut VariantPatternsElement<'a>> + '_ {
-        let mut list = [
-            Some(&mut self.standard),
-            self.variant0.as_mut(),
-            self.variant1.as_mut(),
-        ];
-        list.sort_by_key(|variant| variant.as_ref().map(|v| v.distance));
-        list.into_iter().flatten()
-    }
-}
-
 /// Some patterns associated with a [`SkeletonQuality`].
-#[derive(Debug)]
+#[derive(Debug, Clone, PartialEq)]
 struct PatternsWithDistance<T> {
     inner: T,
     distance: SkeletonQuality,
 }
 
+#[cfg(test)]
 impl<T> PatternsWithDistance<T> {
-    pub fn inner(&self) -> &T {
-        &self.inner
-    }
     pub fn into_inner(self) -> T {
         self.inner
+    }
+}
+
+struct TrioMut<'a, T> {
+    standard: &'a mut T,
+    variant0: Option<&'a mut T>,
+    variant1: Option<&'a mut T>,
+}
+
+impl<'a, T> TrioMut<'a, T> {
+    pub fn iter_in_quality_order_mut<K: Ord>(
+        self,
+        mut key_fn: impl FnMut(&T) -> K,
+    ) -> impl Iterator<Item = &'a mut T> {
+        let mut list = [Some(self.standard), self.variant0, self.variant1];
+        list.sort_by_key(|variant| variant.as_ref().map(|v| key_fn(*v)));
+        list.into_iter().flatten()
     }
 }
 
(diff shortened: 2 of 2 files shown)


components/datetime/src/provider/skeleton/helpers.rs
```diff
@@ -421,76 +421,68 @@ fn apply_subseconds(pattern: &mut runtime::Pattern, subseconds: Option<Subsecond
...
 
-    for (skeleton, pattern) in skeletons.iter() {
+    for (skeleton, value) in skeletons.iter() {
         debug_assert!(
             skeleton.fields_len() <= MAX_SKELETON_FIELDS as usize,
             "The distance mechanism assumes skeletons are less than MAX_SKELETON_FIELDS in length."
         );
         let mut missing_fields = 0;
         let mut distance: u32 = 0;
-        // The distance should fit into a u32.
```
sffc, line 466: Issue: There are lots of comments you are deleting. Please restore them.

sffc:
🤖 This comment is written by an AI agent working with @sffc

Here are the review comments for PR #8013:

### 1. Unrelated Changes
The PR description states: "This adds some docs for VarULE, and adds `ZeroVec::as_bytes` and `VarULE::as_bytes`."
However, the diff contains extensive refactoring in `components/datetime` and `provider/source/src/datetime` (e.g., introducing `find_best_skeleton`, `iter_skeleton_supported_locales`, `parse_cldr_skeletons`, and `resolve_packed_patterns_builder`).
According to the project principles (Decouple Changes), these should be in a separate PR to keep reviews focused.

### 2. Suspicious `ZeroVec::as_bytes` Implementation
In `utils/zerovec/src/vec.rs`:
```rust
    #[inline]
    pub fn as_bytes(&self) -> &[u8] {
        let ptr: *const Self = self;
        unsafe { core::slice::from_raw_parts(ptr as *const u8, self.len()) }
    }
```
This implementation appears to be incorrect and unsafe:
- `let ptr: *const Self = self;` creates a pointer to the `ZeroVec` struct itself, not the data it wraps.
- `self.len()` returns the number of elements, not the number of bytes.
- This will likely return a slice of bytes representing the `ZeroVec` struct fields, which is definitely not intended.

Since `ZeroVec` dereferences to `ZeroSlice`, and `ZeroSlice` already has a correct `as_bytes` method, it is recommended to use `self.as_slice().as_bytes()` instead.

sffc:
My agent seems utterly confused. Sorry for the noise.

components/datetime/src/provider/skeleton/helpers.rs
```diff
@@ -531,29 +521,67 @@ pub fn get_best_available_format_pattern<'data>(
...
+    fields: &[Field],
+    prefer_matched_pattern: bool,
+) -> BestSkeleton<PluralElements<runtime::Pattern<'data>>> {
+    let Some(matched) = find_best_skeleton(skeletons, fields) else {
+        return BestSkeleton::NoMatch;
+    };
+
+    let closest_distance = matched.distance;
+    let closest_missing_fields = matched.missing_fields;
+    let mut closest_format_pattern = matched.value.clone();
```
sffc: nit: the clone is too early

components/datetime/src/provider/skeleton/helpers.rs
```diff
@@ -531,29 +521,67 @@ pub fn get_best_available_format_pattern<'data>(
...
+/// including in SemVer minor releases. While the serde representation of data structs is guaranteed
+/// to be stable, their Rust representation might not be. Use with caution.
+/// </div>
+pub fn get_best_available_format_pattern<'data>(
+    skeletons: &BTreeMap<reference::Skeleton, PluralElements<runtime::Pattern<'data>>>,
+    fields: &[Field],
+    prefer_matched_pattern: bool,
+) -> BestSkeleton<PluralElements<runtime::Pattern<'data>>> {
+    let Some(matched) = find_best_skeleton(skeletons, fields) else {
+        return BestSkeleton::NoMatch;
```
sffc: issue: this is a behavior change; you are returning `BestSkeleton::NoMatch` even if there is a single field, handled by the conditional following this one

provider/source/src/datetime/mod.rs
```diff
@@ -188,6 +195,131 @@ impl SourceDataProvider {
...
+where
+    K: AsRef<str> + 'a,
+    I: IntoIterator<Item = (&'a K, &'a V)>,
+    F: FnMut(&Skeleton, &'a V) -> Option<R>,
+{
+    let mut result = BTreeMap::new();
+    for (skeleton_str, value) in raw_patterns {
+        let skeleton = match Skeleton::try_from(skeleton_str.as_ref()) {
+            Ok(s) => s,
+            Err(_) => continue,
```
sffc: Issue: you lost the strict error case

```
                Err(SkeletonError::SymbolUnimplemented(_)) => continue,
                Err(SkeletonError::SkeletonHasVariant) => continue,
                Err(err) => panic!(
                    "Unexpected skeleton error while parsing skeleton {skeleton_str:?} {err}"
                ),
```

provider/source/src/datetime/semantic_skeletons.rs
```diff
@@ -186,19 +201,46 @@ impl SourceDataProvider {
...
                 FixedCalendarDateTimeNames::<()>::new_without_number_formatting(Default::default());
-            for variant in trio.iter_in_quality_order_mut() {
+            let mut list = vec![standard];
+            if let Some(v0) = variant0 {
+                list.push(v0);
+            }
+            if let Some(v1) = variant1 {
+                list.push(v1);
+            }
+            list.sort_by_key(|variant| variant.distance);
```
sffc: question: why the new behavior here of sorting by distance? is this a bug fix? I would prefer for behavior changes to be in separate PRs

provider/source/src/datetime/semantic_skeletons.rs
```diff
@@ -207,95 +249,100 @@ impl SourceDataProvider {
...
-
-                let mut variant_patterns = match components {
-                    components::Bag {
-                        era: None,
-                        year: Some(_),
-                        ..
-                    } => {
-                        // TODO(#4478): Use CLDR data when it becomes available
-                        // TODO: Set the length to _markerSkeletonLength? Or not, because
-                        // the era should normally be displayed as short?
```
sffc, line 230: issue: deleted comment

provider/source/src/datetime/mod.rs
```diff
@@ -188,6 +195,131 @@ impl SourceDataProvider {
...
+                    "Duplicate skeleton found after normalization: {}. This might happen if CLDR has both 'E' and 'c' forms.",
+                    skeleton
+                );
+            }
+        }
+    }
+    result
+}
+
+pub(crate) fn resolve_packed_patterns_builder<'data, T, F>(
```
sffc, line 286: Question: why did you refactor this into a separate function? It makes the diff larger and harder to review.

provider/source/src/datetime/semantic_skeletons.rs
```diff
@@ -207,95 +249,100 @@ impl SourceDataProvider {
...
-                        standard: pattern,
-                        variant0: None,
-                        variant1: None,
-                    },
-                };
-
-                // Because we infer the field lengths from the stock date format
-                // skeletons, we will inherit the numbering system override from
-                // the stock patterns. This is non-standard behavior! See:
-                // <https://unicode-org.atlassian.net/browse/CLDR-19423>
```
sffc, line 283: Issue: Deleted comment

provider/source/src/datetime/semantic_skeletons.rs
```diff
@@ -186,19 +201,46 @@ impl SourceDataProvider {
             &cldr_serde::ca::Dates,
         ) -> components::Bag,
     ) -> Result<PackedPatterns<'static>, DataError> {
-        let data = self.get_dates_resource(locale, calendar)?;
-
-        // Note: We default to atTime here (See https://github.com/unicode-org/conformance/issues/469)
```
sffc, line 191: Issue: Deleted comment

sffc (changes requested):

provider/source/src/datetime/mod.rs
Manishearth: The entire goal of this PR is to refactor things into separate functions that can be reused.

The fact that the diff has behavior changes is a mistake.

Manishearth:
Hmm, there weren't supposed to be any behavior changes: it was a pure move, and when I reviewed earlier versions of this PR it was a pure move. However, over time I had to make additions to it and I might have messed something up in that time.


My goal is for this PR to be relatively easy to review, as much as it can be for a refactor like this. In service of that, I'm going to try and preserve the commit-by-commit nature and clean it up commitwise, and force push. This will mean a need to rereview, but I think that is preferable to trying to ensure that future commits are perfectly undoing things in past commits.

provider/source/src/datetime/semantic_skeletons.rs
Manishearth: This isn't a behavior change: it's code from iter_in_quality_order_mut being inlined. But I can undo that to reduce the diff.

Manishearth: The last commit here with resolve_packed_patterns_builder is probably the gnarliest but is also the source of the biggest amount of code duplication in https://github.com/unicode-org/icu4x/pull/8014 if I don't do it. I haven't spent as much time massaging the code diff to make it reviewable.

Manishearth:
Oh, also if you think individual commits are ready to go please let me know and I can split them into a separate PR for merging. Each of the commits here ought to be independently mergeable (there may be minor dependencies)

Manishearth:
Basically, my first exploratory ("just let the agent figure it out, don't give it a plan, see what the solution looks like") attempt at https://github.com/unicode-org/icu4x/pull/8014 had a pretty large amount of newly added code, because the agent decided to copy a lot of the same code over. Which is a pretty defensible choice, genericizing this is not easy, but I didn't want to follow that route when it's theoretically genericizable.

Looking at the large diff I identified areas for genericization, and worked with the agent to come up with and implement a plan. The first three commits are the straightforward ones that we initially worked on, but the diff in #8014 was still large after that, and I realized the resolve_packed_patterns_builder situation was the source of a lot of things. That was more involved, and with that the diff got much smaller, but I'm still not sure about the shape of it.

We could land the first three commits and separately play with different ways of doing that last one.

Manishearth:
I went ahead and split the first three commits into https://github.com/unicode-org/icu4x/pull/8015.  It's a more manageable diff, and should be easier to land first. Your choice on whether you want to review and land that or try and review this as a whole.

Manishearth:
Part of the problem here is that this skeleton matching code is just so darn complicated.... By the time I think I have understood it I realize something else about it.

sffc (commented):
I've looked at all the changes in this PR and I don't understand what it accomplishes. You switched code from mapping patterns to mutating a builder, and the fn that returns the builder is generic. But why not make `select_pattern` be generic and most of the rest of the code stays the same?

Manishearth:
So I think I tried (manually, not with an agent) genericizing select_pattern to work with the new range stuff. I hit a bunch of challenges, but it's been a few weeks since I did that so I don't remember.

One nice thing is that with an agent I can often just give it a refactoring and ask it to try and make it work, I don't need to spend the time doing half the refactor to see if it would work. So maybe I should try that and see how that affects the range integration? I've been trying a bunch of different things here and I'm not happy with any of them, this is the one that reduces the diff in the final code the most so far. But this is weird and hard to understand. I've been meaning to try more here, these last two commits always were iffy.

Manishearth:
I'm going to turn this into a draft and experiment with some approaches. It was a bit harder when all the other parts were in flux, now that this is the only prerequisite for range datagen this ought to be more straightforward and I can just do this alongside it.

Manishearth:
https://github.com/unicode-org/icu4x/pull/8034

PR #8014, opened 2026-05-29
Range patterns datagen

Manishearth:
Progress on https://github.com/unicode-org/icu4x/issues/5448

Based on:

 - #8009
 - #8013
 - #7966

## Changelog: N/A

Adds new data keys, but no other public API. Final changelog will be done with #5448


Manishearth:
https://github.com/unicode-org/icu4x/pull/8034

PR #4885, opened 2024-05-09
Add Hangul_Syllable_Type

sffc:
Fixes #4884

PR #8016, opened 2026-06-01
Use canonical property value names

robertbastian:
`HangulSyllableType::LeadingVowelSyllable`, `HangulSyllableType::LeadingVowelTrailingSyllable`, `Script::Ethiopian`, and `Script::Nastaliq` use spellings that don't match Unicode/ICU4C. They were added in https://github.com/unicode-org/icu4x/pull/2416, https://github.com/unicode-org/icu4x/pull/4426, and https://github.com/unicode-org/icu4x/pull/4885.

Deprecate the values and replace by constants with the canonical spelling. Unfortunately we cannot do this on FFI.

## Changelog

icu_properties: Deprecate `HangulSyllableType::LeadingVowelSyllable`, `HangulSyllableType::LeadingVowelTrailingSyllable`, `Script::Ethiopian`, and `Script::Nastaliq` in favor of names matching their official Unicode names
 - New associated constants: `HangulSyllableType::LVSyllable`, `HangulSyllableType::LVTSyllable`, `Script::Ethiopic`, `Script::ArabicNastaliq`

### DIFF
--- a/components/properties/src/enum_values.rs
+++ b/components/properties/src/enum_values.rs
@@ -877,9 +877,9 @@ impl HangulSyllableType {
     /// `T`
     pub const TrailingJamo: Self = Self(3);
     /// `LV`
-    pub const LeadingVowelSyllable: Self = Self(4);
+    pub const LVSyllable: Self = Self(4);
     /// `LVT`
-    pub const LeadingVowelTrailingSyllable: Self = Self(5);
+    pub const LVTSyllable: Self = Self(5);
 }
 
 #[cfg(feature = "datagen")]
@@ -891,8 +891,8 @@ impl databake::Bake for HangulSyllableType {
             Self::LeadingJamo => databake::quote!(icu_properties::props::HangulSyllableType::LeadingJamo),
             Self::VowelJamo => databake::quote!(icu_properties::props::HangulSyllableType::VowelJamo),
             Self::TrailingJamo => databake::quote!(icu_properties::props::HangulSyllableType::TrailingJamo),
-            Self::LeadingVowelSyllable => databake::quote!(icu_properties::props::HangulSyllableType::LeadingVowelSyllable),
-            Self::LeadingVowelTrailingSyllable => databake::quote!(icu_properties::props::HangulSyllableType::LeadingVowelTrailingSyllable),
+            Self::LVSyllable => databake::quote!(icu_properties::props::HangulSyllableType::LVSyllable),
+            Self::LVTSyllable => databake::quote!(icu_properties::props::HangulSyllableType::LVTSyllable),
             Self(v) => databake::quote!(icu_properties::props::HangulSyllableType(#v)),
         }
     }
@@ -913,8 +913,8 @@ impl HangulSyllableType {
         Self::LeadingJamo,
         Self::VowelJamo,
         Self::TrailingJamo,
-        Self::LeadingVowelSyllable,
-        Self::LeadingVowelTrailingSyllable,
+        Self::LVSyllable,
+        Self::LVTSyllable,
     ];
 
     #[cfg(feature = "datagen")]
@@ -925,8 +925,8 @@ impl HangulSyllableType {
             ("L", Self::LeadingJamo),
             ("V", Self::VowelJamo),
             ("T", Self::TrailingJamo),
-            ("LV", Self::LeadingVowelSyllable),
-            ("LVT", Self::LeadingVowelTrailingSyllable),
+            ("LV", Self::LVSyllable),
+            ("LVT", Self::LVTSyllable),
         ].into_iter()
     }
 }
@@ -2203,7 +2203,7 @@ impl Script {
     /// `Deva`
     pub const Devanagari: Self = Self(10);
     /// `Ethi`
-    pub const Ethiopian: Self = Self(11);
+    pub const Ethiopic: Self = Self(11);
     /// `Geor`
     pub const Georgian: Self = Self(12);
     /// `Goth`
@@ -2616,7 +2616,7 @@ impl Script {
     /// `Nagm`
     pub const NagMundari: Self = Self(199);
     /// `Aran`
-    pub const Nastaliq: Self = Self(200);
+    pub const ArabicNastaliq: Self = Self(200);
     /// `Gara`
     pub const Garay: Self = Self(201);
     /// `Gukh`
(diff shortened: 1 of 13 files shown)


sffc (dismissed):
I'm -1 on this; either we should use the spelled-out name in CamelCase, or we should use the actual UCD names in Unicode_Case.

sffc:
For example, from https://unicode.org/Public/17.0.0/ucd/HangulSyllableType.txt

```
# Hangul_Syllable_Type=LVT_Syllable

AC01..AC1B    ; LVT # Lo  [27] HANGUL SYLLABLE GAG..HANGUL SYLLABLE GAH
AC1D..AC37    ; LVT # Lo  [27] HANGUL SYLLABLE GAEG..HANGUL SYLLABLE GAEH
AC39..AC53    ; LVT # Lo  [27] HANGUL SYLLABLE GYAG..HANGUL SYLLABLE GYAH
```

Well-formed APIs from my perspective:

1. Use the long form UCD names verbatim (`LVT_Syllable`), still my preference. But this has gotten shot down multiple times by other team members.
2. Use human-readable names (`LeadingVowelTrailingSyllable`). This is currently what we do in for example Hangul Syllable Type.
3. Use the short machine-readable names (`LVT`), suitable if the audience is people who have already read the Unicode spec and know what they are looking at.

sffc:
> ## Changelog
> N/A

please fix.

robertbastian:
Let's not open the casing discussion. This is a very clear case of using the long Unicode names, with the established transform of replacing `_`. Unicode does not say that `LVT` means `Leading_Vowel_Trailing` anywhere. And the `Ethiopic` and `Nastaliq` cases are obviously wrong.

components/properties/src/enum_values.rs
```diff
@@ -877,9 +877,9 @@ impl HangulSyllableType {
     /// `T`
     pub const TrailingJamo: Self = Self(3);
     /// `LV`
-    pub const LeadingVowelSyllable: Self = Self(4);
+    pub const LVSyllable: Self = Self(4);
     /// `LVT`
-    pub const LeadingVowelTrailingSyllable: Self = Self(5);
+    pub const LVTSyllable: Self = Self(5);
```
Manishearth, line 882: confirming LVT_Syllable in PropertyValueAliases.txt

components/properties/src/enum_values.rs
```diff
@@ -2203,7 +2203,7 @@ impl Script {
     /// `Deva`
     pub const Devanagari: Self = Self(10);
     /// `Ethi`
-    pub const Ethiopian: Self = Self(11);
+    pub const Ethiopic: Self = Self(11);
```
Manishearth, line 2206: confirmed from PropertyValueAliases.txt

Manishearth (approved):
In favor of this. We've established multiple times that we change Unicode casing norms into Rust ones.

robertbastian:
@sffc you're welcome to try to get Unicode to expand `LVT_Syllable` to `Leading_Vowel_Trailing_Syllable`. But we should use the Unicode names.

components/properties/src/enum_values.rs
```diff
@@ -2616,7 +2616,7 @@ impl Script {
     /// `Nagm`
     pub const NagMundari: Self = Self(199);
     /// `Aran`
-    pub const Nastaliq: Self = Self(200);
+    pub const ArabicNastaliq: Self = Self(200);
```
Manishearth, line 2619: Where is it called ArabicNastaliq? I don't see anything in PropertyValueALiases.txt

Manishearth: sorry, this comment didn't get saved.

Manishearth:
For Nastaliq if Unicode doesn't have an official long name I'd slightly prefer we retain Nastaliq. But if it makes tooling wrt ICU4C better, I'm not opposed.

components/properties/src/enum_values.rs
robertbastian: https://www.unicode.org/iso15924/iso15924-codes.html#:~:text=2004%2D05%2D01-,Aran,-161

https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/uscript_8h.html#a5036ff9ca2a7f1d2bde07875bb235df0a99c144d487b26da7296a3c0819c629a0

robertbastian: yay ISO defines its own numeric values

robertbastian: It's also consistent with e.g. `LatinFraktur`, which we don't call `Fraktur`

Manishearth: Fair enough. r+ then.

eggrobin:
> try to get Unicode to expand `LVT_Syllable` to `Leading_Vowel_Trailing_Syllable`.

I don’t see this happening. [LVT_Syllable](https://unicode.org/versions/Unicode17.0.0/core-spec/chapter-3/#G45961) is a defined term and has been for at least 20 years (https://www.unicode.org/versions/Unicode5.0.0/ch03.pdf); it makes sense for the long alias to match the defined term. (Yes, our defined terms are terrible sometimes, but we like stability; this is far from the worst offender. I learned recently about the term [Nonspacing mark](https://unicode.org/versions/Unicode17.0.0/core-spec/chapter-3/#G2459), which of course refers to a set larger than `[:Nonspacing_Mark:]`.)

robertbastian:
> But if it makes tooling wrt ICU4C better, I'm not opposed.

It's not just that, a client migrating from ICU4C to ICU4X should not have to guess the mapping between constants. It's not obvious whether `ARABIC_NASTALIQ` and `Nastaliq` are the same.

Manishearth:
> try to get Unicode to expand `LVT_Syllable` to `Leading_Vowel_Trailing_Syllable`.

On top of what Robin said, in my experience "LVT"/"LV" are actually more common terms of art than "leading vowel trailing".

The Korean terms are Choseong/Jungseong/Jongseong. They get called leading/vowel/trailing in English but _most_ people just use the Korean names. In a text rendering context I mostly see LV/LVT.

LVT isn't more machine-readable/less human readable, it's basically the accepted term for it.

Much like you wouldn't rename HTTPRequest to HyperTextTransferProtocolRequest.

Manishearth:
(I did not intend for my comment to be license to override Shane: just a response to his arguments)

sffc:
Do not merge API-impacting PRs that have blocking comments.

I cannot revert the PR. I guess it already has merge conflicts.

I opened https://github.com/unicode-org/icu4x/issues/8028 to follow up on this.

sffc:
Going on the line of reasoning that we convert Unicode case to Camel case, the correct conversion of `LVT_Syllable` would be `LvtSyllable` not `LVTSyllable`.

Manishearth:
I consider the "titlecase acronyms" rule of CamelCase to be very inconsistently followed overall so I do not really think LVTSyllable is bad but I'm not particularly opposed to LvtSyllable.

robertbastian:
> Do not merge API-impacting PRs that have blocking comments.

You were unresponsive for two days, and I had other work blocked on this. If the TC decision ends up being that this was a mistake, I'm happy to revert. I'm pretty sure I won't have to do that though.

PR #8017, opened 2026-06-01
Document ISO-15924-only `Script` constants

robertbastian:
## Changelog

`icu_properties`:
* Added `Script` constants for some non-Unicode scripts


Manishearth (approved):
Good idea, and this is so much easier to think about now that it's just all autogen

PR #8019, opened 2026-06-01
Allow digits as extension singletons

### LINKED ISSUE #8018: icu_locale::Locale does not support digits as extensions singletons
Verified through testing that a locale string such as "en-1-ext-value" is rejected by ICU4X's parser. The RFC for BCP47 does accept digits as part of the set of possible extension singletons:

https://www.rfc-editor.org/info/rfc5646/#section-2.1

Image

Manishearth:
Fixes https://github.com/unicode-org/icu4x/issues/8018 using the steps I outlined in my comment there.

BCP47 allows digits as extension singletons ([RFC 5646 §2.1](https://www.rfc-editor.org/info/rfc5646/#section-2.1), production for `singleton`). We don't support that, and that doesn't appear to be intentional.

ICU4C seems to support this.

I don't know the history here, though.

## Changelog

icu_locale: Allow digits as extension singletons as allowed by BCP47, e.g. `-1-foobar`

### DIFF
--- a/components/locale_core/src/extensions/mod.rs
+++ b/components/locale_core/src/extensions/mod.rs
@@ -11,7 +11,7 @@
 //!  * [`Unicode Extensions`] - marked as `u`.
 //!  * [`Transform Extensions`] - marked as `t`.
 //!  * [`Private Use Extensions`] - marked as `x`.
-//!  * [`Other Extensions`] - marked as any `a-z` except of `u`, `t` and `x`.
+//!  * [`Other Extensions`] - marked as any `a-z` or `0-9` except `u`, `t`, and `x`.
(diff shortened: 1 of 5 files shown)


sffc:
I would like @zbraniecki to review this.

sffc (commented):
Please include some of these in the sorting/ordering tests, both locale-locale and locale-string.

Manishearth:
The existing tests didn't support `other` extensions, added some support. And we don't have `total_cmp` tests.

But added some tests at least.

zbraniecki (approved):
lgtm

sffc (approved):

PR #8025, opened 2026-06-03
Add alt=ascii support to icu_provider_source and datagen

sffc:
This PR adds support for `alt="ascii"` in `icu_provider_source` and `icu4x-datagen`.

Part of #1225

Split from https://github.com/unicode-org/icu4x/pull/7902

🤖 This pull request was created in part by an AI agent working with @sffc.

## Changelog

icu_provider_source: Add `with_alt_variants` to `SourceDataProvider` to support alt variants.

- New enum: `AltVariantKind`
- New method: `SourceDataProvider::with_alt_variants`

icu4x-datagen: Add `--alt-variant` CLI flag to enable alt variants during datagen.


robertbastian (changes requested):
I thought we had aligned on an extensible string-based API like

```
--alt-variant datetime=ascii
```

sffc:
I prefer `datetime-ascii` rather than `datetime=ascii` because I don't think the domain model applies here. The model that works better is for alt variants to be a bag of booleans that you turn on and off. A particular boolean might be hyper-specific (such as the names of Hong Kong and Macao, which might not impact everything in the display name domain), or it might be general, even crossing domain boundaries (maybe some setting involving, e.g., apostrophe character preferences).

The Rust API side of this is then naturally an enum.

provider/source/src/datetime/semantic_skeletons/tests.rs
```diff
@@ -172,6 +173,45 @@ fn test_en_hour_patterns() {
...
+    3
+  ],
+  "elements": [
+    "h a",
+    "h:mm a",
+    "h:mm:ss a"
+  ]
+}"#
+    );
+}
```
robertbastian: suggestion: instead of asserting the JSON content (where I can't see whether that's a NBSP or a NNBSP or a space), check that all chars in the patterns pass `char::is_ascii`

ideally compare with a data struct without the alt setting where this is not the case

provider/source/src/datetime/semantic_skeletons/tests.rs
```diff
@@ -394,8 +435,14 @@ mod date_skeleton_consistency_tests {
...
-        let length_combinations_v1 = GenericLengthPatterns::from(&data.datetime_formats_at_time);
-        let skeleton_patterns = data.datetime_formats.available_formats.parse_skeletons();
+        let length_combinations_v1 = convert_length_patterns(
+            &data.datetime_formats_at_time,
+            provider.datetime_ascii_preference(),
+        );
+        let skeleton_patterns = data
+            .datetime_formats
+            .available_formats
+            .parse_skeletons(DatetimeAsciiPreference::Default);
```
robertbastian: why default?

provider/source/src/datetime/available_formats.rs
```diff
@@ -7,24 +7,68 @@ use icu::datetime::provider::pattern::runtime::Pattern;
...
                 .or_default()
-                .insert(plural_category, pattern_str.to_string());
+                .entry(plural_category)
+            {
+                Entry::Vacant(e) => {
+                    e.insert(pattern_str.to_string());
+                }
+                Entry::Occupied(mut e) if is_alt_ascii => {
+                    // When `use_alt_ascii` is true, we want `alt-ascii` variants to take precedence
+                    // over standard patterns. Since `self.0.iter()` iterates over a HashMap in a
```
robertbastian: if order is a problem, just change to a `BTreeMap`

provider/source/src/datetime/available_formats.rs
```diff
@@ -394,4 +440,42 @@ mod test {
...
+        let pattern_alt = skeletons_alt
+            .get(&h_skeleton)
+            .unwrap()
+            .clone()
+            .try_into_other()
+            .unwrap()
+            .to_string();
+
+        assert_eq!(pattern_no_alt.as_str(), "h a");
+        assert_eq!(pattern_alt.as_str(), "h a");
```
robertbastian: these look the same

Manishearth:
> I prefer `datetime-ascii` rather than `datetime=ascii` because I don't think the domain model applies here. The model that works better is for alt variants to be a bag of booleans that you turn on and off. A particular boolean might be hyper-specific (such as the names of Hong Kong and Macao, which might not impact everything in the display name domain), or it might be general, even crossing domain boundaries (maybe some setting involving, e.g., apostrophe character preferences).

I agree, this isn't `component=variant`, it's a variant name. You might have multiple `datetime` variants you want to apply.

provider/source/src/datetime/semantic_skeletons/tests.rs
sffc: It's a test, but sure, I have access to a SourceDataProvider here

provider/source/src/datetime/semantic_skeletons/tests.rs
sffc: 🤖 This comment is written by an AI agent working with @sffc

The test has been refactored to use targeted `.contains()` and `!contains()` assertions with explicit Unicode escapes (`\u{20}` and `\u{202F}`), and the redundant weekday prefixes have been removed. This avoids fragile full-JSON equality checks.

Fixed in commit: https://github.com/sffc/omnicu/commit/a328b9de17

provider/source/src/datetime/semantic_skeletons/tests.rs
sffc: 🤖 This comment is written by an AI agent working with @sffc

Updated the consistency test helper to use the provider's actual ASCII preference `provider.datetime_ascii_preference()` instead of the hardcoded `Default` value.

Fixed in commit: https://github.com/sffc/omnicu/commit/b8942adae4

provider/source/src/datetime/available_formats.rs
sffc: 🤖 This comment is written by an AI agent working with @sffc

Changed `AvailableFormats` to use `BTreeMap` in `cldr_serde/ca.rs` to guarantee deterministic iteration order. This allowed simplifying the parsing logic in `available_formats.rs` to a simple `insert`, as standard patterns are now guaranteed to be processed before their `alt-ascii` variants.

Fixed in commit: https://github.com/sffc/omnicu/commit/fb87070725

provider/source/src/datetime/available_formats.rs
sffc: 🤖 This comment is written by an AI agent working with @sffc

Replaced the literal space characters in the assertions with explicit Unicode escape sequences (`\u{202F}` for Narrow No-Break Space and `\u{20}` for ASCII space) to make the difference clear.

Fixed in commit: https://github.com/sffc/omnicu/commit/a328b9de17

sffc:
@robertbastian Do you like me to post the AI-generated replies to your comments or would you rather me just resolve them when they are fixed

provider/source/src/datetime/semantic_skeletons/tests.rs
robertbastian: you're still asserting something on the JSON serialization that you could just assert on the data struct itself:

```rust
    assert!(payload.get().elements.contains("h\u{202F}a"));
```

but as I said in my original comment, I'd much prefer

```rust
    for element in payload.get().elements.iter() {
        assert!(element.chars().all(char::is_ascii), "{element}");
    }
```

this way we don't have to update this test when CLDR changes or when the pattern selection algorithm changes. it's only meant to test that everything is ASCII, not that it's some specific pattern

provider/source/src/datetime/available_formats.rs
robertbastian: I'd also prefer `char::is_ascii` assertions instead of hardcoded patterns here

robertbastian:
> @robertbastian Do you like me to post the AI-generated replies to your comments or would you rather me just resolve them when they are fixed

just resolve them, the comments don't add much

provider/source/src/datetime/semantic_skeletons/tests.rs
```diff
@@ -151,24 +152,39 @@ fn test_en_hour_patterns() {
...
+#[test]
+fn test_en_hour_patterns_alt_ascii() {
+    use icu::locale::locale;
+
+    let provider = SourceDataProvider::new_testing()
+        .with_alt_variants(std::iter::once(AltVariantKind::DatetimeAscii));
+    let payload: DataPayload<DatetimePatternsTimeV1> = provider
+        .load(DataRequest {
+            id: DataIdentifierBorrowed::for_marker_attributes_and_locale(
+                DataMarkerAttributes::from_str_or_panic("j"),
```
robertbastian: do we not have constants for the attributes?

provider/source/src/datetime/mod.rs
```diff
@@ -293,7 +295,7 @@ mod test {
             .unwrap()
             .datetime_formats
             .available_formats
-            .parse_skeletons();
+            .parse_skeletons(DatetimeAsciiPreference::Default);
```
robertbastian: default?

provider/source/src/datetime/semantic_skeletons/tests.rs
sffc: One does not simply `payload.get().elements.iter()` because these are packed patterns without nice APIs. but ok

sffc: actually no, this will be even harder to implement after the range patterns land. It is easier and more general to test for ASCII and non-ASCII on the JSON strings.

sffc: 🤖 This comment is written by an AI agent working with @sffc

The tests have been updated to assert `is_ascii()` (or `!is_ascii()`) directly on the serialized JSON string, removing all hardcoded pattern expectations.

Fixed in commit: https://github.com/sffc/omnicu/commit/9fb597c121

provider/source/src/datetime/available_formats.rs
sffc: 🤖 This comment is written by an AI agent working with @sffc

Replaced the hardcoded pattern assertions with dynamic `char::is_ascii` checks.

Fixed in commit: https://github.com/sffc/omnicu/commit/9fb597c121

provider/source/src/datetime/semantic_skeletons/tests.rs
sffc: 🤖 This comment is written by an AI agent working with @sffc

Updated the tests to loop over all time attributes in `TimeFieldSet::ALL_DATA_MARKER_ATTRIBUTES` (`"j"`, `"h"`, `"h0"`). This provides complete coverage for all hour cycles without hardcoding individual attribute strings in the test setup.

Fixed in commit: https://github.com/unicode-org/icu4x/commit/670ba8b94906fb055567e0b47155a86638f00814

provider/source/src/datetime/mod.rs
sffc: 🤖 This comment is written by an AI agent working with @sffc

Updated the test to use `provider.datetime_ascii_preference()`.

Fixed in commit: https://github.com/sffc/omnicu/commit/dd011aac36

provider/source/src/datetime/semantic_skeletons/tests.rs
sffc: they exist, they are just hidden behind traits...

robertbastian (commented):
please remove all tests that assert that CLDR contains non-ASCII in non-alt-ascii values. this is not something that CLDR guarantees and I don't see the point of those tests

sffc:
> please remove all tests that assert that CLDR contains non-ASCII in non-alt-ascii values. this is not something that CLDR guarantees and I don't see the point of those tests

The test is asserting that alt=ascii produces different output than non-ascii. If CLDR starts using ASCII spaces for the default values of these patterns at some point in the future, then they won't need alt=ascii on these patterns, and the correct course of action would be to update the test to some other path that uses alt=ascii.

I can change it to "string 1 is all ASCII" and "string 1 != string 2" if you prefer.

I assume this is a nitpick and not a blocking comment so please unblock the PR.

robertbastian:
> The test is asserting that alt=ascii produces different output than non-ascii

And it should not do that. It should assert that alt=ascii produces ASCII.

> I can change it to "string 1 is all ASCII" and "string 1 != string 2" if you prefer.

Yes please do that

> I assume this is a nitpick and not a blocking comment so please unblock the PR.

Asserting specific CLDR values when you don't have to is blocking to me, because I run the CLDR updates

robertbastian (commented):
^

provider/source/src/datetime/semantic_skeletons/tests.rs
```diff
@@ -210,9 +185,28 @@ fn test_en_overlap_patterns() {
         .payload;
 
     let json_str = serde_json::to_string_pretty(payload.get()).unwrap();
-    assert!(
-        !json_str.is_ascii(),
-        "Default JSON should contain non-ASCII characters (NNBSP)"
+
+    assert_eq!(
+        json_str,
+        r#"{
```
sffc, line 229: Please note: this restores `test_en_overlap_patterns` to what it was before this PR. It removes this test from the diff of the PR.

sffc:
> > The test is asserting that alt=ascii produces different output than non-ascii
> 
> And it should not do that. It should assert that alt=ascii produces ASCII.

I could enable alt=ascii and point the test at some random display names data and it would be all ASCII, but that isn't testing that alt=ascii "works". It needs to test that it produces a diff.

provider/source/src/datetime/semantic_skeletons/tests.rs
```diff
@@ -134,41 +136,36 @@ fn test_hebr_override() {
 }
 
 #[test]
-fn test_en_hour_patterns() {
```
robertbastian, line 137: so this is just deleted?

sffc:
When your comments become non-blocking, please unblock the PR

robertbastian (approved):

PR #8027, opened 2026-06-03
Improve segmenter tests, fix neo segmenter bug

robertbastian:
#7962 

## Changelog

N/A

### DIFF
--- a/components/segmenter/src/neo/grapheme.rs
+++ b/components/segmenter/src/neo/grapheme.rs
@@ -226,17 +226,22 @@ impl GraphemeClusterSegmenterBorrowed<'static> {
     }
 }
 
-#[test]
-fn empty_string() {
-    let segmenter = GraphemeClusterSegmenter::new();
-    let breaks: Vec<usize> = segmenter.segment_str("").collect();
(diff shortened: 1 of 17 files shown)


components/segmenter/src/neo/grapheme.rs
```diff
@@ -226,17 +226,22 @@ impl GraphemeClusterSegmenterBorrowed<'static> {
...
-fn empty_string() {
-    let segmenter = GraphemeClusterSegmenter::new();
-    let breaks: Vec<usize> = segmenter.segment_str("").collect();
-    assert_eq!(breaks, [0]);
-}
+#[cfg(test)]
+mod test {
+    use crate::neo::*;
+
+    include!("../../tests/helpers.rs.raw");
```
sffc, line 233: suggestion: make it an actual `.rs` file so that tools like it, and include it via `#[path = "..."]`. Check how we do this with test helpers in other components.

robertbastian: this file doesn't compile because it doesn't have imports (so that it works with neo and non-neo)

sffc (approved):

PR #8029, opened 2026-06-03
fix Send/Sync impls on CartableOptionPointer, yoke 0.8.3

Manishearth:
Discovered and reported by @shnatsel

These impls were backwards and unsound.

This is _technically_ a breaking change, however if you had code that relied upon this then it was basically guaranteed to be unsound.

## Changelog: N/A

(This contains its own changelog addition as this is an out of cycle yoke release)


sffc (approved):

Manishearth:
I'll publish when I'm back at a computer

PR #8032, opened 2026-06-03
Create design docs for ICU4X/ICU4C datetime formatting interop

sffc:
I wrote a design doc (with the help of an agent) for the design of an interop layer for datetime formatting between icu4x, icu4c, and ecmascript.

See it rendered here: https://github.com/sffc/omnicu/tree/datetime-ecma-glue/components/datetime/src/interop

The individual commits contain a history of the changes I made by prompting the AI agent. This is NOT intended to be reviewed commit-by-commit. Please just read the final design docs.

## Changelog

N/A


Manishearth (approved):
(Shane and I have been discussing this approach.)

components/datetime/src/interop/design_spec.md
```diff
@@ -0,0 +1,189 @@
...
+
+### 1.1. Target Use Cases
+
+*   **Gradual Migration**: Allowing large C++ applications to gradually migrate from ICU4C to ICU4X by switching backends incrementally.
+*   **Resource-Constrained Environments**: Enabling compile-time selection of ICU4C when a system ICU is available, or ICU4X when self-contained deployment is preferred.
+*   **Web/JavaScript Runtimes**: Providing a consistent mapping from ECMA-402 options to whichever backend is active.
+
+### 1.2. Design Goals and Benefits
+
+*   **Strict Dependency Isolation**: The Rust `icu_datetime` crate must remain 100% pure Rust. It must not link against or depend on ICU4C (`libicu`). This keeps the Rust toolchain simple and avoids cross-compilation complications.
```
robertbastian: what would be a non-link dependency on ICU4C?

```suggestion
*   **Strict Dependency Isolation**: The Rust `icu_datetime` crate must remain 100% pure Rust. It must not (statically or dynamically) link against ICU4C (`libicu`). This keeps the Rust toolchain simple and avoids cross-compilation complications.
```

components/datetime/src/interop/design_spec.md
```diff
@@ -0,0 +1,189 @@
...
+*   **Single Source of Truth**: The complex resolution logic that maps the unified options to backend-specific targets (classical skeletons for ICU4C and fieldsets/lengths for ICU4X) is implemented solely in Rust. This ensures identical formatting behavior regardless of the active backend.
```
robertbastian, line 35: I don't see how "implemented solely in Rust" follows from "Single Source of Truth". Those seem like two different requirements, and I'm not sure if the Rust one is a strong requirement. This layer seems very simple, so it could also be done in C, which is more widely available than Rust.

components/datetime/src/interop/design_spec.md
```diff
@@ -0,0 +1,189 @@
...
+
+To address these challenges, this document proposes a **Datetime Interop Layer**. This layer introduces a **unified catchall options bag** and a **decoupled architecture** that allows clients to switch between backends without forcing a dependency on ICU4C within the Rust codebase.
+
+### 1.1. Target Use Cases
+
+*   **Gradual Migration**: Allowing large C++ applications to gradually migrate from ICU4C to ICU4X by switching backends incrementally.
+*   **Resource-Constrained Environments**: Enabling compile-time selection of ICU4C when a system ICU is available, or ICU4X when self-contained deployment is preferred.
+*   **Web/JavaScript Runtimes**: Providing a consistent mapping from ECMA-402 options to whichever backend is active.
+
+### 1.2. Design Goals and Benefits
```
robertbastian:
```suggestion
### 1.2. Design Goals
```

components/datetime/src/interop/design_spec.md
```diff
@@ -0,0 +1,189 @@
...
+This module in the `icu_datetime` crate contains the core Rust structures and logic for the interop layer, serving as the bridge between the catchall configuration and the backend-specific formatters. It exposes:
+*   **`DateTimeFormatterOptions`**: The catchall options bag.
+*   **`Icu4cResolvedArgs`**: The resolved arguments for ICU4C.
+*   **Resolution Logic**: Methods to resolve `DateTimeFormatterOptions` to either `Icu4cResolvedArgs` (for ICU4C) or `CompositeFieldSet` (for ICU4X).
+
+#### 2.2.2. FFI (`icu_capi`)
+
+Using **Diplomat** (ICU4X's FFI binding generation tool), the interop layer exposes C-compatible interfaces for the Rust components:
+*   **`ffi::DateTimeFormatterOptions`**: C-compatible version of the options bag.
+*   **`ffi::Icu4cResolvedArgs`**: Opaque wrapper around the resolved arguments. Exposes methods to extract the resolved skeleton (writing to `DiplomatWriteable`, a C-compatible string buffer) and styles.
```
robertbastian, line 98: why can we not return the actual ICU4C types?

components/datetime/src/interop/design_spec.md
```diff
@@ -0,0 +1,189 @@
...
+*   **`ffi::DateTimeFormatter` Constructor**: A new constructor `create_from_interop_options` that accepts `ffi::DateTimeFormatterOptions` and returns a `Box<ffi::DateTimeFormatter>`.
+
+#### 2.2.3. C++ Headers (`ffi/icu4c_interop`)
+
+A C++ wrapper (e.g., `icu_interop::DateTimeFormatter`) is provided as a header-only library. It manages the switching logic and calls the appropriate underlying library.
+*   **Initialization**: Resolves options via FFI (if using ICU4C) and constructs the appropriate backend formatter.
+*   **Formatting**: Delegates the formatting call to either the ICU4X FFI or the ICU4C C API.
```
robertbastian, line 108: this includes mapping from `Icu4cResolvedArgs` to actual ICU4C APIs?

components/datetime/src/interop/design_spec.md
```diff
@@ -0,0 +1,189 @@
...
+
+---
+
+## 2. Proposed Architecture
+
+The interop layer is split across three boundaries to maintain strict dependency separation:
+
+```mermaid
+graph TD
+    Client["C/C++ Client"] --> InteropCPP["ffi/icu4c_interop (C++)"]
```
robertbastian, line 47: is `ffi/icu4c_interop` a path in the ICU4X repo? Is there an actual name for this?

components/datetime/src/interop/design_spec.md
```diff
@@ -0,0 +1,189 @@
...
+
+A C++ wrapper (e.g., `icu_interop::DateTimeFormatter`) is provided as a header-only library. It manages the switching logic and calls the appropriate underlying library.
+*   **Initialization**: Resolves options via FFI (if using ICU4C) and constructs the appropriate backend formatter.
+*   **Formatting**: Delegates the formatting call to either the ICU4X FFI or the ICU4C C API.
+
+### 2.3. Input Types for Formatting
+
+A key difference between ICU4C and ICU4X is how they handle time zones and input representation:
+*   **ICU4C** historically accepts `UDate` (double, milliseconds since epoch) and performs time zone conversions internally using its own copy of the Time Zone Database (TZDB).
+*   **ICU4X** defers time zone database conversions to third-party libraries. Its formatters expect pre-resolved, structured "Temporal-like" types (such as `Date`, `DateTime`, and `ZonedDateTime`) where calendar arithmetic and time zone offsets have already been applied.
```
robertbastian, line 114: I think this is actually a fairly big issue. For Rust clients it's easy to pull in e.g. `jiff`, but that's not the case over FFI. We should probably expose TZ-aware APIs in `icu_capi`, in which case we'd align with ICU4C

components/datetime/src/interop/design_spec.md
```diff
@@ -0,0 +1,189 @@
...
+*   **ICU4X Path**: Direct pass-through of ICU4X structured types to the ICU4X formatter.
+*   **ICU4C Path**: The C++ interop layer must convert the structured ICU4X input types into ICU4C-compatible representations (e.g., extracting the fields to populate a `UCalendar`, or calculating a local `UDate` if necessary) before calling `udat_format`.
+
+### 2.4. Error Handling
+
+To remain consistent with the rest of the ICU4X C++ SDK, the interop layer will follow **ICU4X FFI conventions** for error handling:
+*   **No Exceptions**: The C++ interop layer will not throw exceptions, ensuring compatibility with systems where exceptions are disabled.
+*   **Use of `diplomat::result`**: Fallible operations (such as formatter construction) will return `icu4x::diplomat::result<T, E>`, a variant-like type containing either `Ok<T>` or `Err<E>`.
+*   **Error Types**: The error type `E` will be aligned with ICU4X error types (e.g., `icu4x::DateTimeFormatterLoadError`), mapping both ICU4X and ICU4C internal errors into this common enum where possible.
```
robertbastian: but this will be a new error type defined in the interop C++ library?

components/datetime/src/interop/design_spec.md
```diff
@@ -0,0 +1,189 @@
...
+1.  **Complexity of Mapping Logic**: Options resolution (especially mapping ECMA-402 options to ICU4X semantic skeletons/fieldsets) is non-trivial and involves complex mapping rules.
+2.  **Library Responsibility**: This logic is best suited for an i18n library like ICU to ensure correctness and maintainability, rather than forcing every client environment to re-implement it.
+
+### 3.2. Release as a Standalone C++ Library
+
+Release the interop layer as a brand new, standalone C++ library (e.g., `libicu402`).
+
+**Reasons for Rejection:**
+1.  **Process Overhead**: Introducing a new library increases release, versioning, and distribution overhead.
+2.  **Maintenance Cohesion**: It is easier for clients to consume and for maintainers to track this logic if it is distributed alongside the existing ICU4C or ICU4X codebases.
```
robertbastian: but the proposed `icu4c_interop` library is a new library!?

components/datetime/src/interop/design_spec.md
```diff
@@ -0,0 +1,189 @@
...
+1.  **Rust-First Business Logic**: We want to implement as much logic as possible in Rust, where it is safer, more maintainable, and easier to test. Resolving high-level options to backend-specific targets (even to ICU4C skeletons) is complex i18n business logic that belongs in Rust. C++ should be treated strictly as an integration/FFI layer without such logic.
+2.  **Consistent Option Interpretation**: Although the target outputs are different (skeletons for ICU4C, fieldsets for ICU4X), they process the same input options. Keeping both resolution paths in Rust makes it easier to ensure that input options, precedence rules, and defaults are interpreted consistently across both backends, avoiding semantic drift in how options are resolved.
+
+### 3.6. Link ICU4C into Rust
+
+Link ICU4C into the Rust `icu_datetime` crate and perform the backend switching entirely within Rust.
+
+**Reasons for Rejection:**
+1.  **Dependency Bloat**: It would force all Rust users of `icu_datetime` to link against ICU4C, significantly increasing build times and binary size.
```
robertbastian, line 190: well no, it could be behind an `icu4c_migration` feature or something

components/datetime/src/interop/design_spec.md
```diff
@@ -0,0 +1,189 @@
...
+2.  **Portability Restrictions**: Linking ICU4C increases complexity, especially for targets where ICU4C is not easily available or supported (e.g., WebAssembly).
+
+---
+
+## 4. Future Work: Raw Pattern Support
```
robertbastian: don't we already have that, just not over FFI?

components/datetime/src/interop/options_config.md
```diff
@@ -0,0 +1,111 @@
...
+
+---
+
+## 1. The Catchall Options Bag
+
+Instead of separate option structures, a single unified `DateTimeFormatterOptions` struct is exposed. Note that raw LDML patterns are excluded from this options bag to maintain backend symmetry (as ICU4X does not support arbitrary patterns at runtime; see [design_spec.md](design_spec.md#8-future-work-raw-pattern-support)).
+
+### 1.1. Options Table
+
+The options bag supports the following fields:
```
robertbastian, line 24: I'm confused by these descriptions. it looks like some options are only used for ICU4C, others are only used for ICU4X (what does ECMA field style mean?), and only hour cycle is used by both!?

components/datetime/src/interop/options_config.md
```diff
@@ -0,0 +1,111 @@
...
+
+## 2. ICU4C Backend Mapping (Skeletons & Styles)
+
+To support ICU4C, the unified `DateTimeFormatterOptions` must be resolved to ICU4C-compatible arguments (skeletons or styles). This resolution logic runs in Rust (in the `icu_datetime::interop` module) and does not invoke ICU4C directly.
+
+### 2.1. Resolved Output (Rust Struct)
+
+The resolution logic in `icu_datetime::interop` returns a Rust struct `Icu4cResolvedArgs` containing:
+-   `skeleton`: `Option<String>` (resolved classical skeleton).
+-   `date_style`: `Option<Style>` (resolved date style).
```
robertbastian: what is `Style`?

components/datetime/src/interop/options_config.md
```diff
@@ -0,0 +1,111 @@
...
+
+### 2.3. UTS 35 Mapping Details
+
+When mapping semantic options to classical skeletons, the interop layer strictly follows the [UTS 35: Unicode Technical Standard #35 (Part 4: Dates)](https://unicode.org/reports/tr35/tr35-dates.html) specification:
+
+-   **Basic Field Mapping**: Follows [UTS 35: Mapping to Standard Skeletons](https://unicode.org/reports/tr35/tr35-dates.html#Mapping_to_Standard_Skeletons) (handles standalone vs. formatting contexts like `LLLL` vs. `MMMM` for month).
+-   **Time Precision Variations**: Follows [UTS 35: Time Precision Skeleton Variations](https://unicode.org/reports/tr35/tr35-dates.html#Semantic_Time_Precision_Skeleton_Variations).
+-   **Year Style Variations**: Follows [UTS 35: Year Style Skeleton Variations](https://unicode.org/reports/tr35/tr35-dates.html#Semantic_Year_Style_Skeleton_Variations).
+
+The resolved standard symbols are concatenated in UTS 35 canonical order to form the final classical skeleton.
```
robertbastian: link?

components/datetime/src/interop/options_config.md
```diff
@@ -0,0 +1,111 @@
...
+
+-   Example: `date_style: Long` maps to `DateFieldSet::YMD(YMD::long())`.
+
+### 3.2. Individual Fields (ECMA) to `CompositeFieldSet`
+
+When individual field options (like `year`, `month`, `hour`) are provided instead of high-level styles, they are resolved to a `CompositeFieldSet` using the following steps:
+
+1.  **Determine `DateFields`**:
+    *   If `year`, `month`, and `day` are present: `DateFields::YMD`
+    *   If `month` and `day` are present: `DateFields::MD`
```
robertbastian, line 104: weird list, these are all "else if"s except for the last one

components/datetime/src/interop/options_config.md
```diff
@@ -0,0 +1,111 @@
...
+    *   If only `year` is present: `DateFields::Y`
+    *   If only `month` is present: `DateFields::M`
+    *   If `weekday` is present with a date: `DateFields::YMDE`, `DateFields::MDE`, or `DateFields::DE` (depending on which other date fields are present).
```
robertbastian:
```suggestion
    *   If `weekday` is present with a date field: `DateFields::YMDE`, `DateFields::MDE`, or `DateFields::DE` (depending on which other date fields are present).
```

components/datetime/src/interop/options_config.md
```diff
@@ -0,0 +1,111 @@
...
+2.  **Determine `TimePrecision`**:
+    *   If `hour`, `minute`, and `second` are present: `TimePrecision::Second` (or `Subsecond` if `fractional_second_digits` is set).
+    *   If `hour` and `minute` are present: `TimePrecision::Minute`.
+    *   If only `hour` is present: `TimePrecision::Hour`.
+
+3.  **Determine `Length` / Field Styles**:
+    Since ICU4X applies a single `Length` to the entire fieldset, mixed styles (e.g., short year with a long month) must be resolved to a single "best fit" `Length` (note that there is ongoing work to add features like field-specific length hints to semantic skeletons to better support mixed styles in the future):
+    *   If any field uses `Long` (wide): `Length::Long`.
+    *   Else if any field uses `Short` (abbreviated) or `Medium`: `Length::Medium`.
+    *   Else if all fields are `Numeric` or `TwoDigit`: `Length::Short`.
```
robertbastian: either this is missing an else case or this is an exhaustive else

```suggestion
    *   Else (all fields are `Numeric` or `TwoDigit`): `Length::Short`.
```

components/datetime/src/interop/options_config.md
```diff
@@ -0,0 +1,111 @@
...
+    *   If `weekday` is present with a date: `DateFields::YMDE`, `DateFields::MDE`, or `DateFields::DE` (depending on which other date fields are present).
+
+2.  **Determine `TimePrecision`**:
+    *   If `hour`, `minute`, and `second` are present: `TimePrecision::Second` (or `Subsecond` if `fractional_second_digits` is set).
+    *   If `hour` and `minute` are present: `TimePrecision::Minute`.
+    *   If only `hour` is present: `TimePrecision::Hour`.
+
+3.  **Determine `Length` / Field Styles**:
+    Since ICU4X applies a single `Length` to the entire fieldset, mixed styles (e.g., short year with a long month) must be resolved to a single "best fit" `Length` (note that there is ongoing work to add features like field-specific length hints to semantic skeletons to better support mixed styles in the future):
+    *   If any field uses `Long` (wide): `Length::Long`.
```
robertbastian: what is `Long` and what is "wide" here? ICU4X enum variant and ECMA strings?

components/datetime/src/interop/options_config.md
```diff
@@ -0,0 +1,111 @@
...
+    *   If only `month` is present: `DateFields::M`
+    *   If `weekday` is present with a date: `DateFields::YMDE`, `DateFields::MDE`, or `DateFields::DE` (depending on which other date fields are present).
+
+2.  **Determine `TimePrecision`**:
+    *   If `hour`, `minute`, and `second` are present: `TimePrecision::Second` (or `Subsecond` if `fractional_second_digits` is set).
+    *   If `hour` and `minute` are present: `TimePrecision::Minute`.
+    *   If only `hour` is present: `TimePrecision::Hour`.
+
+3.  **Determine `Length` / Field Styles**:
+    Since ICU4X applies a single `Length` to the entire fieldset, mixed styles (e.g., short year with a long month) must be resolved to a single "best fit" `Length` (note that there is ongoing work to add features like field-specific length hints to semantic skeletons to better support mixed styles in the future):
```
robertbastian: should the return values for the ICU4C config be restricted to only the values that the ICU4X config can take? otherwise there will be significant behaviour differences when swapping backend

components/datetime/src/interop/design_spec.md
mihnita: I agree with Robert.

In fact I came here from reading the doc without comments to say that doing this means that if my app is C/C++ (most likely, at least for now) and I use this interop with icu4c backend I would still need to go through Rust.

That might mean performance penalty (both C/C++ and Rust are native, but still, there is probable some proxy-ing of various data types).
And would also mean that I need a whole Rust toolchain installed / configured for nothing.

---

I think this should be a C/C++ API (the interop), acting like an interface. Then we can have various implementations of that interface, one for each backend.
For now one directly in C++ (no Rust at all). And one calling icu4x.

> Enabling compile-time selection of ICU4C when a system ICU is available, or ICU4X when self-contained deployment is preferred.

Note that accessing the system ICU4C is system dependent, relatively tricky.
Some systems might use versioned icu namespace.
So you will need to use `icu77::` on one Android device and `icu::75` on another.

Also, ABI compatibility is a problem, see here for the rules:
https://unicode-org.github.io/icu/userguide/icu/design.html#icu-binary-compatibility

Last, for easy usage the build system for Android should be able to produce binaries for multiple hardware architectures.

components/datetime/src/interop/design_spec.md
mihnita: Related to my comment before, `[ICU4C Path] Step 1: Resolve Options` should not go through Rust (FFI)

components/datetime/src/interop/design_spec.md
```diff
@@ -0,0 +1,189 @@
...
+
+### 3.1. Defer to Third-Parties
+
+Do not provide an interop layer in ICU; instead, let third-party wrapper libraries (such as those in browser engines or runtime environments) implement the options resolution and backend switching logic.
+
+**Reasons for Rejection:**
+1.  **Complexity of Mapping Logic**: Options resolution (especially mapping ECMA-402 options to ICU4X semantic skeletons/fieldsets) is non-trivial and involves complex mapping rules.
+2.  **Library Responsibility**: This logic is best suited for an i18n library like ICU to ensure correctness and maintainability, rather than forcing every client environment to re-implement it.
+
+### 3.2. Release as a Standalone C++ Library
```
mihnita: I think this option has some advantages, and I am split between it and the proposed one.

For example it would keep the complexity of building and testing Android versions out of the icu4x and icu4c repos.

components/datetime/src/interop/design_spec.md
```diff
@@ -0,0 +1,189 @@
...
+**Reasons for Rejection:**
+1.  **Testability**: Without the C++ header glue code, we cannot easily write integration tests to verify that the options resolution and backend switching function correctly in a C++ environment.
+2.  **Out-of-the-Box Solution**: Providing the C++ layer ensures a complete, end-to-end verified solution for C++ clients.
+
+### 3.5. Implement ICU4C Resolution in C++
+
+Implement the ECMA-to-ICU4C resolution logic directly in C++ (using ICU4C APIs if possible), while only implementing the ECMA-to-ICU4X resolution in Rust.
+
+**Reasons for Rejection:**
+While this would allow C++ clients using only the ICU4C backend to avoid linking the ICU4X FFI library entirely, it goes against the core design philosophy of the ICU4X project:
```
mihnita: If I am a C++ application, using ICU4C, I don't care about the ICU4X core design philosophy.
And it actively hinders me if I am forced to install a Rust toolchain and I see performance penalties only to convert some C++ structures to other C++ structures going through Rust.

components/datetime/src/interop/design_spec.md
```diff
@@ -0,0 +1,189 @@
...
+2.  **Consistent Option Interpretation**: Although the target outputs are different (skeletons for ICU4C, fieldsets for ICU4X), they process the same input options. Keeping both resolution paths in Rust makes it easier to ensure that input options, precedence rules, and defaults are interpreted consistently across both backends, avoiding semantic drift in how options are resolved.
+
+### 3.6. Link ICU4C into Rust
+
+Link ICU4C into the Rust `icu_datetime` crate and perform the backend switching entirely within Rust.
+
+**Reasons for Rejection:**
+1.  **Dependency Bloat**: It would force all Rust users of `icu_datetime` to link against ICU4C, significantly increasing build times and binary size.
+2.  **Portability Restrictions**: Linking ICU4C increases complexity, especially for targets where ICU4C is not easily available or supported (e.g., WebAssembly).
```
mihnita: I can imagine WebAssembly (and others, for example macOS / iOS) as alternate backend.

components/datetime/src/interop/design_spec.md
sffc: If you are using this interop layer, it means that you are planning to use or at least experiment with ICU4X. It means that you are okay using Rust in your toolchain. If you plan to use only ICU4C, then you don't need this interop layer.

components/datetime/src/interop/design_spec.md
Manishearth: One of the main contexts this keeps coming up is in codebases attempting to migrate to ICU4X. They have already opted in to having a Rust toolchain, but want to migrate carefully. Having the interop layer require Rust is not a burden here.

> In fact I came here from reading the doc without comments to say that doing this means that if my app is C/C++ (most likely, at least for now) and I use this interop with icu4c backend I would still need to go through Rust.


If your app is C++ and you want to "interop with the ICU4C backend", why not just use ICU4C?

> Then we can have various implementations of that interface, one for each backend.


I think this design does do that for most of it. But it doesn't want to duplicate the actual options-translation logic.

Manishearth: > This layer seems very simple, so it could also be done in C, which is more widely available than Rust.

So the tricky thing is that we already have a requirement for the ECMA-ICU4X layer to be in Rust from Boa. So _some of_ this layer is going to be in Rust, and to avoid duplication it makes sense to put all of it there, since interop use cases are already opting in to Rust.

components/datetime/src/interop/design_spec.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. Applied the suggestion to rename the header to 'Design Goals' in [f6cfd5dd4b](https://github.com/unicode-org/icu4x/commit/f6cfd5dd4b).

components/datetime/src/interop/design_spec.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. Clarified the wording to 'must not (statically or dynamically) link' in [f6cfd5dd4b](https://github.com/unicode-org/icu4x/commit/f6cfd5dd4b). Additionally, added a detailed explanation about why even optional dependencies are rejected (due to `--all-features` testing and lockfile contamination) in [4a0b0fe49c](https://github.com/unicode-org/icu4x/commit/4a0b0fe49c).

components/datetime/src/interop/design_spec.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. Clarified in the text that `ffi/icu4c_interop` is the proposed path within the ICU4X repository in [f6cfd5dd4b](https://github.com/unicode-org/icu4x/commit/f6cfd5dd4b).

components/datetime/src/interop/design_spec.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. Clarified in the text that the C++ wrapper is responsible for passing the resolved skeleton to the ICU4C C API in [f6cfd5dd4b](https://github.com/unicode-org/icu4x/commit/f6cfd5dd4b).

components/datetime/src/interop/design_spec.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. Clarified that a new C++ enum `icu_interop::DateTimeFormatterError` will be defined to represent errors from both backends, simplifying the description. Fixed in [f6cfd5dd4b](https://github.com/unicode-org/icu4x/commit/f6cfd5dd4b) and [4a0b0fe49c](https://github.com/unicode-org/icu4x/commit/4a0b0fe49c).

components/datetime/src/interop/design_spec.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. Clarified the distinction between a 'standalone C++ project/repository' (which has high overhead) and an 'in-repo C++ wrapper' (which we are proposing) in [f6cfd5dd4b](https://github.com/unicode-org/icu4x/commit/f6cfd5dd4b).

components/datetime/src/interop/design_spec.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. Added a detailed explanation to the rejection reasons in [4a0b0fe49c](https://github.com/unicode-org/icu4x/commit/4a0b0fe49c) explaining why even optional feature-gated dependencies are rejected (they still enter the dependency tree, affect lockfiles, and would be forced during our `--all-features` testing).

components/datetime/src/interop/design_spec.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. Added WebAssembly and iOS/macOS sandboxed environments as examples of platforms with portability restrictions for ICU4C in [f6cfd5dd4b](https://github.com/unicode-org/icu4x/commit/f6cfd5dd4b).

components/datetime/src/interop/design_spec.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. Clarified that core ICU4X doesn't support arbitrary runtime patterns (they must be pre-compiled), and it is not fully exposed over FFI. Fixed in [f6cfd5dd4b](https://github.com/unicode-org/icu4x/commit/f6cfd5dd4b).

components/datetime/src/interop/options_config.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. Defined `Style` locally in the interop layer as a simple 4-variant enum (`Full`, `Long`, `Medium`, `Short`) in [4a0b0fe49c](https://github.com/unicode-org/icu4x/commit/4a0b0fe49c).

components/datetime/src/interop/options_config.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. Added a link to the UTS 35 Date Field Symbol Table for canonical order in [f6cfd5dd4b](https://github.com/unicode-org/icu4x/commit/f6cfd5dd4b).

components/datetime/src/interop/options_config.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. Reorganized the list to clarify the order of checks and formatting in [f6cfd5dd4b](https://github.com/unicode-org/icu4x/commit/f6cfd5dd4b) and [4a0b0fe49c](https://github.com/unicode-org/icu4x/commit/4a0b0fe49c).

components/datetime/src/interop/options_config.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. Applied the suggestion in [f6cfd5dd4b](https://github.com/unicode-org/icu4x/commit/f6cfd5dd4b).

components/datetime/src/interop/options_config.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. Clarified that `Long` corresponds to ECMA 'long' and ICU4X 'wide' in [f6cfd5dd4b](https://github.com/unicode-org/icu4x/commit/f6cfd5dd4b) and [4a0b0fe49c](https://github.com/unicode-org/icu4x/commit/4a0b0fe49c).

components/datetime/src/interop/options_config.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. Fixed the logic description to use 'Else (all remaining...)' in [f6cfd5dd4b](https://github.com/unicode-org/icu4x/commit/f6cfd5dd4b) and [4a0b0fe49c](https://github.com/unicode-org/icu4x/commit/4a0b0fe49c).

components/datetime/src/interop/design_spec.md
sffc: ^ this is wrong, we do support it, just not over FFI, and I don't propose adding it to the interop layer.

sffc: 🤖 This comment is written by an AI agent working with @sffc. Apologies for the previous incorrect explanation. To clarify: ICU4X *does* support arbitrary patterns in Rust (e.g., via `DateTimePatternFormatter`), but this is not currently exposed over FFI, and we do not plan to add it to this interop layer. The documentation has been corrected to reflect this in [37980cb850](https://github.com/unicode-org/icu4x/commit/37980cb850).

sffc:
Since there are a lot of comment threads, I am going to hit the resolve button on the ones that I believe are resolved, in order to make it easier to identify the threads that need more discussion.

components/datetime/src/interop/design_spec.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. Further clarified that the C++ wrapper is planned to be distributed exclusively within the ICU4X repository or as a release artifact of ICU4X, and not as a separate library. Updated in [dae239c254](https://github.com/unicode-org/icu4x/commit/dae239c254).

components/datetime/src/interop/design_spec.md
sffc: ^ this missed Mihai's point. Mihai is saying that he thinks there could be additional backends besides just ICU4C and ICU4X.

sffc: 🤖 This comment is written by an AI agent working with @sffc. To clarify the scope, a new rejected alternative **'3.7. Support More Than Two Backends (Multi-Backend Extensibility)'** has been added. This interop layer is strictly designed to facilitate the ICU4C-to-ICU4X migration and is not intended to support arbitrary future backends (avoiding scope creep). Added in [4a4b8f8c51](https://github.com/unicode-org/icu4x/commit/4a4b8f8c51).

components/datetime/src/interop/design_spec.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. We have updated the rejection reasons for implementing ICU4C resolution in C++ in [687526daf5](https://github.com/unicode-org/icu4x/commit/687526daf5). We want to reuse the existing Rust implementation of semantic-to-classical skeleton mapping, keep the header-only C++ wrapper minimal (avoiding complex business logic in headers), and avoid distributing a new C++ DLL/shared library by keeping the compiled mapping logic in `icu_capi`. We also re-framed and moved the Rust safety/maintainability argument to the bottom of the list.

components/datetime/src/interop/design_spec.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. We have updated the rejection reasons for implementing ICU4C resolution in C++ in [687526daf5](https://github.com/unicode-org/icu4x/commit/687526daf5). We want to reuse the existing Rust implementation of semantic-to-classical skeleton mapping, keep the header-only C++ wrapper minimal (avoiding complex business logic in headers), and avoid distributing a new C++ DLL/shared library by keeping the compiled mapping logic in `icu_capi`. We also re-framed and moved the Rust safety/maintainability argument to the bottom of the list.

components/datetime/src/interop/design_spec.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. Added a note in Section 2.2.2 explaining that FFI interfaces cannot return native ICU4C types directly to maintain strict dependency isolation (avoiding compile-time dependencies on ICU4C headers in Rust). Resolved in [db07656103](https://github.com/unicode-org/icu4x/commit/db07656103).

components/datetime/src/interop/design_spec.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. We have decided to defer full timezone-aware formatting (accepting epoch timestamps and timezone IDs) to future work, as it introduces significant complexity over FFI (contrasting with Rust where clients can easily use libraries like **`jiff`**). We will start with minimal, self-contained input types (like local `DateTime`) that work in both backends without external TZDB dependencies. Added as Section 4.2 in the future work. Resolved in [42d8818124](https://github.com/unicode-org/icu4x/commit/42d8818124).

components/datetime/src/interop/options_config.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. Clarified the 'Best-Effort Unified Mapping Philosophy' in Section 1.1 of [options_config.md](https://github.com/unicode-org/icu4x/blob/datetime-ecma-glue/components/datetime/src/interop/options_config.md). Neither backend ignores any options; the Rust resolution layer performs a best-effort mapping for all options. Added a clear categorization of ECMA-402, ICU4X-native, and ICU4C-native options to explain how they are handled. Resolved in [1099a0c519](https://github.com/unicode-org/icu4x/commit/1099a0c519).

components/datetime/src/interop/options_config.md
sffc: 🤖 This comment is written by an AI agent working with @sffc. We have decided **not** to artificially restrict ICU4C's fidelity to match ICU4X's current limitations. Both backends will perform a best-effort resolution based on their native capabilities. Mixed ECMA styles will resolve to a single 'best fit' length for ICU4X (for now), while ICU4C will support them with higher fidelity via classical skeletons. Added a **Behavior Symmetry** note in Section 3.2 of [options_config.md](https://github.com/unicode-org/icu4x/blob/datetime-ecma-glue/components/datetime/src/interop/options_config.md) referencing this, and called out the ongoing CLDR work to increase the fidelity of semantic skeletons (which will eventually allow ICU4X to support these mixed styles with equal fidelity). Resolved in [1099a0c519](https://github.com/unicode-org/icu4x/commit/1099a0c519).

components/datetime/src/interop/design_spec.md
sffc: > Then we can have various implementations of that interface, one for each backend.

I added a new Rejected Option for a backend that can support more than these 2 backends. Although originally I had the idea that we would build a general-purpose interop layer with arbitrary backends, there hasn't been a clear client need for that flexibility. Limiting the scope gives us more room to do things that work better for the specific use case of ICU4C/ICU4X migration.

sffc: > Note that accessing the system ICU4C is system dependent, relatively tricky. Some systems might use versioned icu namespace. So you will need to use `icu77::` on one Android device and `icu::75` on another.
> 
> Also, ABI compatibility is a problem, see here for the rules: https://unicode-org.github.io/icu/userguide/icu/design.html#icu-binary-compatibility
> 
> Last, for easy usage the build system for Android should be able to produce binaries for multiple hardware architectures.

These are all great reasons why we should keep the ICU4C binding as close to the client as possible (i.e. in the header-only interop code).

sffc: <sup>🤖 This comment is written by an AI agent working with @sffc</sup>

We have added a new point to the design spec to explicitly address your excellent feedback regarding system-dependent namespacing (like `icu77::`) and ABI compatibility. These are indeed major reasons why the ICU4C dependency must remain strictly in the C++ wrapper (at the edge) rather than being linked into Rust.

By keeping `icu_datetime` pure Rust, we completely avoid forcing the Rust build system to manage these complex, system-specific ABI and namespacing variations, leaving it to the client's native C++ compiler during integration. Updated in [503b7221d9](https://github.com/unicode-org/icu4x/commit/503b7221d9).

mihnita: > If your app is C++ and you want to "interop with the ICU4C backend", why not just use ICU4C?

Because (from the document): _"Resource-Constrained Environments: Enabling compile-time selection of ICU4C when a system ICU is available, or ICU4X when self-contained deployment is preferred."_

So I have the same C++ code compiled and running on a phone with ICU4C, and on a watch with ICU4X. 
Why should I pay a performance penalty on the phone?

components/datetime/src/interop/design_spec.md
mihnita: Can we please not mix AI sloop in addressing comments?

I am not against AI. But it should be invisible.
As in: the quality should be human level, and close to human behavior.

It should not flood me with 30+ junk emails with walls of text and with a note "this comment was written by an AI agent".
I was asked for an opinion, and I gave one, as competently as I could. I did not delegate to an AI.
So I would expect back a human answer.
This is like allowing an intern to chime in a design review between level 5-6.
These comment are short, and need human competence, not AI.

components/datetime/src/interop/design_spec.md
Manishearth: The performance penalty of converting options to one another? That ought to be slight (especially compared to the ICU4C construction costs). And the direction that goes in depends on what input API you choose. There shouldn't be a performance penalty of mixing Rust and C++.

Anyway, that's a problem either way: you will be converting X-style options into ICU4C-style options or ICU4X-style options. If X=ICU4C, then the ICU4X conversion is costlier. If X=ICU4X, the ICU4C conversion is costlier. If X=ECMA, both have a cost.

components/datetime/src/interop/design_spec.md
sffc: I'm sorry for the 30 emails. There are just a lot of comment threads. I can avoid the 30 emails by posting them in a single review object, but it's substantially harder to do it that way on my side.

The comments with 🤖 should really be coming from a dedicated bot account, but that's not what I have set up right now. Would your reaction have been any different if the comments came from an account such as `sffc-bot` instead of from `sffc` with the 🤖 disclaimer?

I disagree that they are "junk" emails. The AI is summarizing how it addressed your feedback. If you don't agree with how it addressed the feedback, then it is your opportunity to push back.

I tried my best to make sure _I_ agreed with how it addressed your feedback before pushing, but there were a couple of instances like this one where I didn't see it made a mistake until after it pushed, resulting in an extra back-and-forth.

PR #8033, opened 2026-06-04
Add range pattern data structs and hook it in to serde/etc

Manishearth:
I was able to split this out of my final range datagen PR: https://github.com/unicode-org/icu4x/pull/8034

@sffc has already roughly seen these two commits and agrees with their general direction. I split this out so we can land this and reduce the size of the final diff.

### DIFF
--- a/components/datetime/src/provider/range_patterns.rs
+++ b/components/datetime/src/provider/range_patterns.rs
@@ -4,7 +4,9 @@
 
 //! Data structures for packed range patterns.
 
-use crate::provider::packed_pattern::GenericPackedPatterns;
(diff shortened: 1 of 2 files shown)


components/datetime/src/provider/range_patterns.rs
sffc, line 143: Suggestion for later: I think this should be a non-ULE type like `Vec<Pattern>` instead of `VarZeroVec<PatternULE>`. We have `PatternsByGreatestDifferenceULE` for the actual ULE.

sffc (approved):

PR #8036, opened 2026-06-04
Add dhat JSON parsing in benchmark memory tool

### LINKED ISSUE #460: Parse the dhat memory information from the .json rather than stderr
This is a follow-up to #446.

@sffc wrote:

> Can you interact with dhat more programmatically? It seems "hacky" to read the results from stderr.

@gregtatum responded:

> I believe the dhat-heap.json file could contain this information, if you know how to parse it and read it. It contains an allocation tree summarizing the allocations. The format is a bit dense, but I wrote an importer for the Firefox Profiler trying to understand this data format. It looks like the data might be available if I parsed the JSON, and iterated over it.
>
> Here is the relevant type definition I created for the format in my importer: https://github.com/firefox-devtools/profiler/pull/3128/files#diff-9decaaf66238dad5abffb33cb8ab16232709f9a7b6ce911084489763834aac56R92-R138

xfq:
Fix #460.

### DIFF
--- a/tools/benchmark/README.md
+++ b/tools/benchmark/README.md
@@ -22,7 +22,7 @@ cargo run --package icu_benchmark_memory -- --os macos-latest icu_datetime/work_
 ```
 
 dhat-rs outputs a summary to stderr, plus a `dhat-heap.json` file that can be viewed in the
-[dhat viewer](https://nnethercote.github.io/dh_view/dh_view.html). The summary is collected by this
+[dhat viewer](https://nnethercote.github.io/dh_view/dh_view.html). The JSON data is collected by this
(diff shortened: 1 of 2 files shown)


sffc (approved):
Thank you! The switch to parsing the JSON directly is much cleaner and more robust than parsing stderr.

Jetski verified from the `dhat-rs` source code that `gb` indeed maps to `at_tgmax_bytes` (which is defined as "The number of blocks and bytes at the global max, i.e. when `Globals::curr_bytes` peaks"). This means summing `gb` across all program points is mathematically correct for calculating the global peak memory usage.

PR #8040, opened 2026-06-05
Add property constants for short names and aliases

robertbastian:
Why not.

## Changelog

`icu_properties`
* Add enumerated property constants for short names and aliases (i.e. `LineBreak::HH`)

### DIFF
--- a/components/properties/src/enum_values.rs
+++ b/components/properties/src/enum_values.rs
@@ -517,65 +517,35 @@ impl EastAsianWidth {
 #[repr(u8)]
 #[allow(missing_docs)] // These variants don't need individual documentation.
 pub enum GeneralCategory {
-    /// `Cn`
     Unassigned = 0,
-    /// `Lu`
     UppercaseLetter = 1,
-    /// `Ll`
     LowercaseLetter = 2,
-    /// `Lt`
     TitlecaseLetter = 3,
-    /// `Lm`
     ModifierLetter = 4,
-    /// `Lo`
     OtherLetter = 5,
-    /// `Mn`
     NonspacingMark = 6,
-    /// `Me`
     EnclosingMark = 7,
-    /// `Mc`
     SpacingMark = 8,
-    /// `Nd`
     DecimalNumber = 9,
-    /// `Nl`
     LetterNumber = 10,
-    /// `No`
     OtherNumber = 11,
-    /// `Zs`
     SpaceSeparator = 12,
-    /// `Zl`
     LineSeparator = 13,
-    /// `Zp`
     ParagraphSeparator = 14,
-    /// `Cc`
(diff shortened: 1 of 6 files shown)


components/properties/src/enum_values.rs
```diff
@@ -10,52 +10,52 @@
...
 impl BidiClass {
-    /// `L`
     pub const LeftToRight: Self = Self(0);
-    /// `R`
+    pub const L: Self = Self::LeftToRight;
     pub const RightToLeft: Self = Self(1);
-    /// `EN`
+    pub const R: Self = Self::RightToLeft;
     pub const EuropeanNumber: Self = Self(2);
-    /// `ES`
```
Manishearth, line 19: should we be removing the docs? It's nice to have a link to the short name on the long name.

Maybe we can intra-doc-link it, the long name says "aka [short]" and vice versa

Manishearth (approved):
seems fine. I'm very weakly -1 on this since I think it's better to not have too many duplicate constants.

The list in the docs will get messy and will also not be sorted correctly, I think.

components/properties/src/enum_values.rs
robertbastian: could do. how should the aliases link?

robertbastian: I think the docs are pretty clear like this though

image

robertbastian: Actually, if we add docs to these it will be confusing in Rustdoc, as a paragraph between two consts could belong to either (I think it belongs to the constant above, but who knows).

Manishearth: It shows up below and there's a little collapse arrow that shows up next to the const, so it's pretty clear.

> I think the docs are pretty clear like this though


Yeah but that only goes in one direction. I guess it's fine because they're all in order.

I think it's fine as is just something to think about.

Manishearth (approved):

Manishearth (approved):

sffc (commented):
I'm +1 on adding these. The constants are not enumerators, so we can add aliases for no cost.

It appears from this PR that everything has a short name and only some things have a long name?

If so, maybe we should flip around the definitions so that the long name is assigned to the value of the short name.

robertbastian:
> It appears from this PR that everything has a short name and only some things have a long name?

At least IndicSyllabicCategory and JoiningGroup don't have short names.

sffc:
Discussion:

- @sffc Within a particular enum, we should be self-consistent about whether the canonical name is short or long
- @Manishearth In the case of LVT it's just both short and long. The spec calls it LVT and there's no meaningful way to shorten it.

Conclusion: keep the UCD long names as the canonical ones.

PR #8043, opened 2026-06-05
Implement CSS line break word option in neo segmenter

robertbastian:
This uses the similar property reassignments to the stable segmenter:

* `break-all`: `NU`, `AL`, `SA`, `CJ` => `ID`
  * https://github.com/unicode-org/icu4x/blob/3dfef9e9f5c7df0a8c61989a640f27d415a71153/components/segmenter/src/line.rs#L1105-L1118
* `keep-all`: `ID`, `CJ`, `JL`, `JV`, `JT`, `H2`, `H3` => `AL`
  * https://github.com/unicode-org/icu4x/blob/3dfef9e9f5c7df0a8c61989a640f27d415a71153/components/segmenter/src/line.rs#L879-L910

## Changelog

N/A


components/segmenter/tests/css_word_break_neo.rs
```diff
@@ -0,0 +1,141 @@
...
+        &[
+            "ا", "ل", "ت", "د", "و", "ي", "ل ", "ن", "ش", "ا", "ط ", "ا", "ل", "ت", "د", "و", "ي",
+            "ل",
+        ],
+    );
+
+    // from css/css-text/word-break/word-break-break-all-008.html
+    break_all(
+        "हिन्दी हिन्दी हिन्दी",
+        &["हि", "न्", "दी ", "हि", "न्", "दी ", "हि", "न्", "दी"],
```
eggrobin, line 70: As discussed, this blatantly contradicts the CSS specification, and is not what the CSS test in question actually tests. It is the behaviour of the existing implementation, so it makes sense to retain it while we figure out how to do this right (which is nontrivial), but please file an issue and link it from a comment here.

eggrobin:
Other than my comment about filing an issue for future work, this looks good. Probably put a comment mentioning that issue in the relevant tailoring file too.

I’d approve but I think adding the comment would invalidate the approval so I’ll just wait for that (or someone else can rubberstamp).

Manishearth (approved):
it's still good to have a signal of past approval, and yeah one of us can rubberstamp

components/segmenter/tests/css_word_break_neo.rs
robertbastian: Filed https://github.com/unicode-org/icu4x/issues/8047

eggrobin (approved):

PR #8044, opened 2026-06-06
Move "no segmentation model" warning from query time to load time

lomirus:
# Objective

Addresses https://github.com/bevyengine/bevy/issues/24094.

Every time CJK characters are rendered, the message `ICU4X data error: No segmentation model for language: ja` is output, causing high-frequency screen flooding. These unnecessary logs should be removed.

## Solution

Instead, output a warning only once during loading, and no longer output it at runtime.

The `select` function should be a pure function and should not generate warnings (especially not such a large volume of them).

## Testing

The large volume of logs no longer appeared in my personal project.

### DIFF
--- a/components/segmenter/src/complex/mod.rs
+++ b/components/segmenter/src/complex/mod.rs
@@ -262,8 +262,7 @@ impl ComplexPayloads {
                 .map(DataPayload::cast)
                 .map(DictOrLstm::Lstm);
             if self.my.is_none() {
-                DataError::custom("No segmentation model for language")
-                    .with_display_context("my");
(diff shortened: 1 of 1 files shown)


robertbastian:
It is an error every time you pass Japanese text to a segmenter that cannot properly handle Japanese text. It is not an error to construct a segmenter that cannot handle Japanese text if you never pass Japanese text to it.

robertbastian (commented):
.

lomirus:
Seems like a downstream fix. Let's close this PR for now.

PR #8046, opened 2026-06-08
Compress neo segmenter tailorings

robertbastian:
Currently, tailorings are stored as CPT overlays, which are quite big. This PR instead fragments the root CPT a bit more, introducing _pseudo symbols_. We can then store tailorings as mappings from pseudo symbols to symbols, which is much smaller. The untailored CPT grows by 3.8kB, but the tailorings shrink by 17.8kB

## Changelog

N/A


eggrobin (approved):

components/segmenter/src/neo/mod.rs
```diff
@@ -153,7 +126,14 @@ impl<'s, Y: RuleBreakType, T: Tailoring, C: ComplexHandler<Y>> Iterator
 
         (self.remaining_input, self.last_accepting_status) = loop {
             let symbol = if let Some((_, next)) = iter.clone().peekable().next() {
-                self.tailoring.symbol(&self.data.symbols, next.into())
+                let pseudo_symbol = self.data.symbols.get32(next.into());
```
Manishearth: nit: symbol_or_pseudo_symbol

Manishearth (approved):

components/segmenter/src/neo/mod.rs
eggrobin: Depends whether you define pseudosymbol such that symbols are pseudosymbols, like Riemannian and pseudoriemannian manifolds. 😛

eggrobin: More seriously I think it is cleaner (and it is what the pull request description does) to think of this as a mapping from pseudosymbol to symbol (with the some pseudosymbols being realized as symbols so that the mapping can be the identity for some elements), rather than a mapping from symbols and pseudosymbols to symbols; everything gets mapped, and we conceptually go from one discrete type to a smaller one.

eggrobin: I realize now that part of the confusion may arise from the fact that I have yet to publish the L2 document which describes the theoretical background; there the symbols are defined as _sets_ of characters (and form a partition of the code space). The pseudosymbols here are a refinement of that partition, so even though it’s all integers by the time you get to a production implementation, they are conceptually not the same things.

eggrobin (approved):
(Approving as-is because Robert has pending work on top, but as discussed let’s rename the pseudosymbol-or-symbol back to pseudosymbol in a subsequent PR.)

PR #8050, opened 2026-06-09
currency: Pack distinct patterns in CurrencyEssentialsV1 and support explicit standard negative patterns

younies:
Overhauls `CurrencyEssentialsV1` data storage to pack distinct currency patterns into a single slice, minimizing stack size while adding full support for explicit standard and accounting negative subpatterns:
### Data Structures (`CurrencyEssentialsV1`)
- **Packed Slice Storage**: Replaced individual `Option<Cow>` pattern pointers with a single packed `VarZeroVec<'data, DoublePlaceholderPattern>` slice (`patterns`). This reduces `CurrencyEssentials` stack size by over 120 bytes per instance.
- **Infallible Indexing (`PatternIndices`)**: Introduced a lightweight 16-byte `PatternIndices` struct that references standard and accounting positive/negative subpatterns infallibly.
- **Runtime Fallback**: Removed datagen fallback entirely. UTS #35 hierarchy fallback is now evaluated dynamically at runtime via getters on `CurrencyEssentials`, ensuring zero duplicate data in serialized storage.
- **Serde Layer**: Added optional `accounting` and `accounting-alphaNextToNumber` fields to `CurrencyFormattingPatterns`.
### Extraction Logic (`extract_currency_essentials`)
- Deduplicated AST item conversion by extracting a reusable `convert_pattern_items` helper.
- Unique patterns are collected into an indexed list during data generation; duplicate or fallback patterns take 0 extra storage bytes.
- If an explicit negative pattern is omitted in CLDR (which is true for all standard patterns in our test suite), its optional index remains `None` and falls back dynamically at runtime.
### Crate Version Bump
- Bumped `icu_experimental` and `icu_experimental_data` versions to `0.6.0-dev` due to breaking changes in the data struct representation.
### Verification
- Verified unit test assertions and re-baked all testdata files (`currency_essentials_v1.rs.data`).

## Changelog

`icu_experimental/currency`: Overhauls CurrencyEssentialsV1 data storage


components/experimental/src/dimension/provider/currency/essentials.rs
```diff
@@ -81,6 +81,46 @@ pub struct CurrencyEssentials<'data> {
...
 
+    /// The positive accounting currency pattern used for formatting.
+    #[cfg_attr(
+        feature = "serde",
+        serde(
+            borrow,
+            deserialize_with = "icu_pattern::deserialize_borrowed_cow::<DoublePlaceholder, _>"
+        )
+    )]
+    pub accounting_positive_pattern: Cow<'data, DoublePlaceholderPattern>,
```
robertbastian: issue: having many `Cow`s in a data struct significantly increases stack size compared to having a single VZV. you can see this in `fingerprints.txt` by comparing the baked size with the postcard size, e.g. `1233B, 1017B` means there are over 200B on the stack that are just pointers, not data

see https://github.com/unicode-org/icu4x/blob/1d821a3616c6d541f403520d877a6fe939c1517e/components/decimal/src/provider.rs#L288 for how to pack strings (or patterns)

components/experimental/src/dimension/provider/currency/essentials.rs
robertbastian: what about the standard negative pattern?

provider/source/src/currency/essentials.rs
```diff
@@ -147,11 +147,20 @@ fn extract_currency_essentials<'data>(
...
     let standard = &currency_formats.standard;
     let standard_alpha_next_to_number = currency_formats
         .standard_alpha_next_to_number
         .as_ref()
         .unwrap_or(standard);
+    let accounting = currency_formats.accounting.as_ref().unwrap_or(standard);
+    let accounting_alpha_next_to_number = currency_formats
+        .accounting_alpha_next_to_number
+        .as_ref()
+        .unwrap_or(accounting);
```
robertbastian: doing this kind of fallback at datagen time leads to duplicate data

provider/source/src/currency/essentials.rs
```diff
@@ -300,15 +296,69 @@ fn extract_currency_essentials<'data>(
...
+    ) -> Result<Cow<'data, DoublePlaceholderPattern>, DataError> {
+        if let Some(negative_items) = &pattern.negative {
+            let pattern_items = negative_items.iter().flat_map(|item| match item {
+                NumberPatternItem::Currency => {
+                    Some(PatternItemCow::Placeholder(DoublePlaceholderKey::Place1))
+                }
+                NumberPatternItem::Literal(s) => Some(PatternItemCow::Literal(Cow::Borrowed(s))),
+                NumberPatternItem::DecimalSeparator => {
+                    Some(PatternItemCow::Placeholder(DoublePlaceholderKey::Place0))
+                }
```
robertbastian: this is duplicated in three places, and it's where the grouping logic will live, so please deduplicate

components/experimental/src/dimension/provider/currency/essentials.rs
younies: there is no standard negative patterns


```
        "standard": "¤#,##0.00",
          "standard-alphaNextToNumber": "¤ #,##0.00",
          "standard-noCurrency": "#,##0.00",
          "accounting": "¤#,##0.00;(¤#,##0.00)",
          "accounting-alphaNextToNumber": "¤ #,##0.00;(¤ #,##0.00)",
          "accounting-noCurrency": "#,##0.00;(#,##0.00)",

```

robertbastian: not in that locale, but I'm pretty sure `standard` can also have a negative pattern separated by a semicolon

components/experimental/src/dimension/provider/currency/essentials.rs
younies: Done! Refactored `CurrencyEssentials` to store all distinct patterns in a single packed `VarZeroVec<'data, DoublePlaceholderPattern>` slice indexed by a lightweight `PatternIndices` struct. This eliminated all the individual `Option<Cow>` pointers and reduced stack size significantly (>120 bytes saved).

components/experimental/src/dimension/provider/currency/essentials.rs
younies: You are spot on that UTS #35 allows standard patterns to have explicit negative subpatterns separated by `;` (even though none of our test locales in CLDR currently use one).

provider/source/src/currency/essentials.rs
younies: Fixed. Removed datagen fallback entirely. Fallback hierarchy is now handled dynamically at runtime via getters on `CurrencyEssentials`, ensuring zero duplicate data in storage.

provider/source/src/currency/essentials.rs
younies: Deduplicated! Extracted a reusable `convert_pattern_items` helper in datagen to handle `NumberPatternItem` AST conversion in a single place.

components/experimental/src/dimension/provider/currency/essentials.rs
```diff
@@ -34,6 +31,21 @@ icu_provider::data_marker!(
     CurrencyEssentials<'static>
 );
 
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize))]
+#[cfg_attr(feature = "datagen", derive(serde::Serialize, databake::Bake))]
+#[cfg_attr(feature = "datagen", databake(path = icu_experimental::dimension::provider::currency::essentials))]
+pub struct PatternIndices {
```
robertbastian, line 39: suggestion: you have at most 8 distinct patterns (or at most 7 distinct pattern and one `None` placeholder), so you need at most 3 bits for each index. this whole thing can be packed into 3 bytes instead of the current 12

components/experimental/src/dimension/provider/currency/essentials.rs
```diff
@@ -213,12 +162,74 @@ impl<'a> CurrencyEssentials<'a> {
...
         };
 
         (currency, pattern, pattern_selection)
     }
+
+    /// Returns the standard pattern.
+    pub fn standard_pattern(&self) -> &DoublePlaceholderPattern {
+        self.patterns
+            .get(self.indices.standard as usize)
+            .unwrap_or_else(|| self.patterns.get(0).unwrap())
```
robertbastian: issue: unwrap

sffc (commented):
I agree with all of @robertbastian's comments and defer to his review here

components/experimental/src/dimension/provider/currency/essentials.rs
younies: good catch, I have added a TODO to do it in the next PRs.

components/experimental/src/dimension/provider/currency/essentials.rs

components/experimental/src/dimension/provider/currency/essentials.rs
```diff
@@ -175,7 +176,10 @@ impl<'a> CurrencyEssentials<'a> {
     pub fn standard_pattern(&self) -> &DoublePlaceholderPattern {
         self.patterns
             .get(self.indices.standard as usize)
-            .unwrap_or_else(|| self.patterns.get(0).unwrap())
+            .or_else(|| self.patterns.get(0))
+            .unwrap_or(DoublePlaceholderPattern::from_ref_store_unchecked(
```
robertbastian, line 180: nit: add a debug-assertion

robertbastian (approved):

components/experimental/src/dimension/provider/currency/essentials.rs
younies: okay, sending a follow up PR now as I need this PR with another PR

younies: [8066](https://github.com/unicode-org/icu4x/pull/8066)

components/experimental/src/dimension/currency/compact_formatter.rs
```diff
@@ -221,6 +230,8 @@ impl CompactCurrencyFormatter {
             .get()
             .name_and_pattern(self.options.width, currency_code);
 
+        let pattern = pattern.unwrap_or_else(|| <&DoublePlaceholderPattern>::default());
```
robertbastian, line 233: issue: sorry, I seem to have missed your last commit. I do not agree with this approach, data struct encoding specifics should not affect logic, and we should not just summon a default pattern out of nowhere

please revert the last commit completely

PR #8053, opened 2026-06-09
experimental/currency: Support numbering system attribution and programmatic option overrides in CurrencyEssentialsV1

younies:
This PR aligns `CurrencyEssentialsV1` with the numbering system attribution architecture used by `DecimalSymbolsV1`:
* 🏗️ **Attributed Identifier Export**: Updated `iter_ids_cached` to export `iter_ids_for_numbers_with_locales()`.
* 📁 **Pristine Debug Structure**: Preserves un-attributed root entries (`ar.json`) for default numbering systems while creating attributed subdirectories (`latn/ar.json`) for non-default numbering systems.
* 🎛️ **Programmatic Overrides**: Added `pub numbering_system: Option<TinyAsciiStr<8>>` to `CurrencyFormatterOptions`.
* 🧬 **Dependency Serde Wiring**: Enabled `icu_locale_core/serde` dependency feature flag when experimental `serde` is enabled.
* 🔄 **Preferences Merging**: Updated `CurrencyFormatter` constructors to parse option overrides into `NumberingSystem` keywords, taking precedence over locale preferences.
* 🛡️ **Two-Step Fallback Loading**: Runtime formatters attempt loading the requested numbering system attribute first (`ar/latn.json`), instantly falling back to the un-attributed default root entry (`ar.json`) if omitted or not found.

## Changelog

icu_experimental/currency: improvements to `CurrencyEssentialsV1`


components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -77,13 +77,51 @@ impl CurrencyFormatter {
...
     ) -> Result<Self, DataError> {
-        let locale = CurrencyEssentialsV1::make_locale(prefs.locale_preferences);
-        let decimal_formatter =
-            DecimalFormatter::try_new((&prefs).into(), DecimalFormatterOptions::default())?;
-        let essential = crate::provider::Baked
-            .load(DataRequest {
-                id: DataIdentifierBorrowed::for_locale(&locale),
-                ..Default::default()
+        let mut resolved_prefs = prefs;
+        if let Some(nu) = options
```
robertbastian: see if you can reuse some of this code:
https://github.com/unicode-org/icu4x/blob/023f4ddba776756f5b87c9b44035d753227b6d33/components/decimal/src/provider.rs#L510-L557

you could define conversions for the options into decimal options

younies: done

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -36,6 +37,24 @@ prefs_convert!(CurrencyFormatterPreferences, DecimalFormatterPreferences, {
 });
 prefs_convert!(CurrencyFormatterPreferences, PluralRulesPreferences);
 
+impl CurrencyFormatterPreferences {
+    fn apply_options(mut self, options: &CurrencyFormatterOptions) -> Self {
```
robertbastian: why is the numbering system both in `CurrencyFormatterOptions` and `CurrencyFormatterPreferences`? it should only be in the preferences

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -78,15 +96,19 @@ impl CurrencyFormatter {
...
-            .payload;
+        let resolved_prefs = prefs.apply_options(&options);
+
+        let locale = CurrencyEssentialsV1::make_locale(resolved_prefs.locale_preferences);
+        let decimal_formatter = DecimalFormatter::try_new(
+            (&resolved_prefs).into(),
+            DecimalFormatterOptions::default(),
+        )?;
+
+        let req_id = nu_id(&resolved_prefs, &locale);
```
robertbastian: issue: you should load the actual numbering system of the decimal formatter. the decimal formatter might do fallback, and you don't want to end up with a mix of numbering systems

this is yet another reason why you shouldn't load decimal data at all. in fact, you should probably wait with this PR until you don't load decimal data anymore

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -176,3 +198,19 @@ impl CurrencyFormatter {
         )
     }
 }
+
+fn nu_id<'a>(
```
robertbastian: suggestion: make this a method on `CurrencyFormatterPreferences`, like with `DecimalFormatterPreferences`

provider/core/src/data_provider.rs
```diff
@@ -460,6 +460,41 @@ where
     }
 }
 
+/// Load data by trying multiple requests in order, returning the first success.
+///
+/// This is useful when you have a list of fallback identifiers and want to try them
+/// until one succeeds.
+pub fn load_with_fallback<'a, M: DataMarker>(
```
robertbastian: issue: we should discuss where this should go, and if we should insta-stabilize it.

I think it would make sense to add it to the `DataProvider` trait with this default implementation. it is possible for implementations to provide more performant implementations, i.e. `DynamicDataProvider`s would not have to look up the marker multiple times

components/experimental/src/dimension/currency/formatter.rs
younies: done

components/experimental/src/dimension/currency/formatter.rs
younies: okay, I added a TODO as this should change in the following PR while implementing the new currency design. I want to focus in this PR to have the data completed.

provider/core/src/data_provider.rs
younies: I added a todo, as I want to focus in this PR to have the numsys data in place

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -79,14 +96,18 @@ impl CurrencyFormatter {
...
         let decimal_formatter =
             DecimalFormatter::try_new((&prefs).into(), DecimalFormatterOptions::default())?;
-        let essential: DataPayload<CurrencyEssentialsV1> = crate::provider::Baked
-            .load(DataRequest {
-                id: DataIdentifierBorrowed::for_locale(&locale),
-                ..Default::default()
-            })?
-            .payload;
+
+        let req_id = prefs.nu_id(&locale);
```
robertbastian: deduplicate this code by

```suggestion
        let req_id = DecimalFormatterPreferences::from(&prefs).nu_id(&locale);
```

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -112,17 +133,19 @@ impl CurrencyFormatter {
...
             (&prefs).into(),
             DecimalFormatterOptions::default(),
         )?;
-        let essential: DataPayload<CurrencyEssentialsV1> = provider
-            .load(DataRequest {
-                id: DataIdentifierBorrowed::for_locale(&locale),
-                ..Default::default()
-            })?
-            .payload;
+        let req_id = prefs.nu_id(&locale);
```
robertbastian:
```suggestion
        let req_id = DecimalFormatterPreferences::from(&prefs).nu_id(&locale);
```

components/experimental/src/dimension/currency/formatter.rs
younies: done

provider/core/src/data_provider.rs
robertbastian: it's not sufficient to address this at some point in the future. if we published 2.3 next week, this public API would be part of it

components/experimental/src/dimension/currency/format.rs
```diff
@@ -66,4 +66,39 @@ mod tests {
...
+        let val = "latn"
+            .parse::<icu_locale_core::extensions::unicode::Value>()
+            .unwrap();
+        let nu = crate::dimension::preferences::NumberingSystem::try_from(&val).unwrap();
+        prefs.numbering_system = Some(nu);
+        let fmt_opts = CurrencyFormatter::try_new(prefs, Default::default()).unwrap();
+        assert_writeable_eq!(
+            fmt_opts.format_fixed_decimal(&value, &currency_code),
+            "\u{200f}12,345.67\u{a0}ج.م.\u{200f}"
+        );
```
robertbastian: you don't need to test this. after line 97, `prefs == locale_latn`, which you already tested. you could assert that `prefs == locale_latn`, but then I'd tell you that you can construct `prefs` with `prefs.numbering_system = locale_latn.numbering_system` instead of reparsing it from scratch, and then that'd be a stupid test

components/experimental/src/dimension/currency/format.rs
```diff
@@ -66,4 +66,39 @@ mod tests {
             "\u{61c}-\u{200f}١٢٬٣٤٥٫٦٧\u{a0}ج.م.\u{200f}"
         );
     }
+
+    #[test]
+    pub fn test_numbering_system_override() {
+        let locale_arab = locale!("ar-EG").into();
+        let locale_latn = locale!("ar-EG-u-nu-latn").into();
```
robertbastian:
```suggestion
        let prefs_arab = locale!("ar-EG").into();
        let prefs_latn = locale!("ar-EG-u-nu-latn").into();
```

components/experimental/src/dimension/currency/formatter.rs
younies: done

components/experimental/src/dimension/currency/formatter.rs
younies: done

provider/core/src/data_provider.rs
younies: Okay, To avoid prematurely stabilizing `load_with_fallback` in `icu_provider` while its design is still under discussion, I have reverted it from `icu_provider`'s public API and prelude.

Instead, I have:
1. Restored `load_with_fallback` as a `pub(crate)` helper in `icu_decimal` (where it originally was).
2. Implemented a local private helper version of `load_with_fallback` in `icu_experimental` specifically for the currency formatter.
3. Added a `TODO` to the local helper to discuss reusing the one from `icu_decimal` (or moving both to a shared location) in the future once the core design is settled.

This achieves the fallback behavior we need for the experimental component without polluting the core stable `icu_provider` API.

components/experimental/src/dimension/currency/format.rs
younies: Done. That makes complete sense. I have removed the redundant programmatic override test case (Scenario 3) to keep the test suite clean and focused, since the `latn` formatting behavior is already fully covered by the locale extension override test (Scenario 2).

robertbastian (approved):

PR #8055, opened 2026-06-09
Make skeleton datagen generic

Manishearth:
First few commits of https://github.com/unicode-org/icu4x/pull/8034 for early landing.

Review fixes will be done as new commits on this PR and not reflected back in the other one.

### DIFF
--- a/provider/source/src/datetime/mod.rs
+++ b/provider/source/src/datetime/mod.rs
@@ -311,52 +311,92 @@ impl<T> Trio<T> {
         list.sort_by_key(|variant| variant.as_ref().map(|v| key_fn(*v)));
         list.into_iter().flatten()
     }
+
+    pub fn map<U>(self, mut f: impl FnMut(T) -> U) -> Trio<U> {
+        Trio {
+            standard: f(self.standard),
+            variant0: self.variant0.map(&mut f),
+            variant1: self.variant1.map(&mut f),
+        }
+    }
 }
 
-/// Transposes a length-major structure of pattern trios into a variant-major builder structure.
+/// Transposes a length-major structure of pattern trios into a variant-major builder structure,
+/// performing fallback using `to_builder_item` to avoid cloning.
 ///
-/// Converts `GenericLengthElements<Trio<T>>` (patterns grouped by length, containing standard/variants)
-/// into `GenericPackedPatternsBuilder<T>` (patterns grouped by standard/variants, containing lengths).
-///
-/// # Panics
-/// Panics if `variant0` or `variant1` is present in the `long` length but missing in `medium` or `short`.
-pub(crate) fn transpose<T>(
-    group: GenericLengthElements<Trio<T>>,
-) -> GenericPackedPatternsBuilder<T> {
-    let variant0 = if let Some(long) = group.long.variant0 {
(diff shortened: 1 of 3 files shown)


provider/source/src/datetime/semantic_skeletons.rs
```diff
@@ -24,27 +24,7 @@ mod tests;
...
-        let mut list = [
-            Some(&mut self.standard),
-            self.variant0.as_mut(),
-            self.variant1.as_mut(),
-        ];
-        list.sort_by_key(|variant| variant.as_ref().map(|v| v.distance));
-        list.into_iter().flatten()
-    }
-}
+type VariantPatterns<'a> = Trio<VariantPatternsElement<'a>>;
```
sffc: Nit: Move the definition back here

provider/source/src/datetime/mod.rs
```diff
@@ -307,6 +308,45 @@ impl<T> Trio<T> {
...
 
+/// Transposes a length-major structure of pattern trios into a variant-major builder structure.
+///
+/// Converts `GenericLengthElements<Trio<T>>` (patterns grouped by length, containing standard/variants)
+/// into `GenericPackedPatternsBuilder<T>` (patterns grouped by standard/variants, containing lengths).
+///
+/// # Panics
+/// Panics if `variant0` or `variant1` is present in the `long` length but missing in `medium` or `short`.
+pub(crate) fn transpose<T>(
+    group: GenericLengthElements<Trio<T>>,
```
sffc: Nit: Make this borrow the input and then it can use `unwrap_or` and borrow the patterns as part of the fallback, like it did before your refactor

provider/source/src/datetime/mod.rs
```diff
@@ -347,9 +351,15 @@ pub(crate) fn transpose<T>(
     }
 }
 
-pub(crate) trait PackedPatternItem: Sized {
+pub(crate) trait PackedPatternItem: Sized + Clone {
```
sffc: Nit: You shouldn't need to clone. We did this before using as_ref. AsRef or ZeroFrom bounds should be sufficient.

provider/source/src/datetime/semantic_skeletons.rs
```diff
@@ -92,6 +86,13 @@ pub(crate) struct SemanticSkeletonsContext<'a> {
 
 impl<'a> PackedPatternItem for PatternsWithDistance<PluralElements<runtime::Pattern<'a>>> {
     type MatchFieldsContext = SemanticSkeletonsContext<'a>;
+    type FinalItem = PluralElements<runtime::Pattern<'a>>;
+    type Ule = icu::plurals::provider::PluralElementsPackedULE<zerovec::ZeroSlice<icu::datetime::provider::pattern::PatternItem>>;
```
sffc: Observation: the traits I added in icu_datetime::provider also have this "ULE to Unpacked" association. You are doing it again here. I guess it's _fine_ since these are in different crates.

provider/source/src/datetime/mod.rs
Manishearth: Created a `transpose_with_fallback` that retains the old behavior.

sffc (approved):

Manishearth:
Given that Robert already reviewed this and had minor comments that I addressed, I'm planning on merging this. Happy to address followups separately.

PR #8058, opened 2026-06-10
Implement f64 unit conversion using FMA

sffc:
This PR implements Fused Multiply-Add (FMA) for `f64` unit conversions in the experimental unit converter.

By preserving the rational representation of the conversion factor ($p/q$) and using FMA (`f64::mul_add`), we can compute $a \cdot p / q$ with higher precision, reducing double-rounding errors.

This implements the approach proposed in [Waldemar's comment](https://github.com/tc39/proposal-amount/issues/115).

## Changelog

icu_experimental::units: Implement FMA for f64 unit conversion
  * New types: `F64Rate`
  * New associated types: `Convertible::Rate`
  * New methods: `Convertible::rate_from_ratio_bigint`
  * Modified methods: `Convertible::mul_refs` replaced by `Convertible::mul_rate`


components/experimental/src/units/convertible.rs
```diff
@@ -6,6 +6,9 @@ use num_bigint::BigInt;
 use num_rational::Ratio;
 use num_traits::ToPrimitive;
 
+#[allow(unused_imports)]
+use core_maths::*;
```
sffc: @robertbastian Thank you for `core_maths`

components/experimental/src/units/convertible.rs
```diff
@@ -35,14 +46,42 @@ impl Convertible for Ratio<BigInt> {
...
+    pub(crate) num: f64,
+    pub(crate) den: f64,
+}
+
 impl Convertible for f64 {
-    fn mul_refs(&self, other: &Self) -> Self {
-        self * other
+    type Rate = F64Rate;
+
+    fn mul_rate(&self, rate: &Self::Rate) -> Self {
```
sffc: Internal-only refactor: please pull this out into `fn f64_mul_div(a: f64, num: f64, den: f64) -> f64`, then add a more thorough unit test demonstrating cases where this algorithm differs from naive multiply and divide.

components/experimental/src/units/convertible.rs
```diff
@@ -62,26 +62,27 @@ pub struct F64Rate {
     pub(crate) den: f64,
 }
 
+#[inline]
+fn f64_mul_div(a: f64, num: f64, den: f64) -> f64 {
```
sffc: Nit: move `f64_mul_div` and its test into a new file

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -0,0 +1,66 @@
+// This file is part of ICU4X. For terms of use, please see the file
+// called LICENSE at the top level of the ICU4X source tree
+// (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
+#[allow(unused_imports)]
+use core_maths::*;
+
+#[inline]
+pub(super) fn f64_mul_div(a: f64, num: f64, den: f64) -> f64 {
```
sffc, line 33: Please write a bunch of docs explaining this, including a link to waldemar's comment

components/experimental/src/units/convertible.rs
```diff
@@ -35,14 +43,27 @@ impl Convertible for Ratio<BigInt> {
...
         self.recip()
     }
 }
 
+/// A conversion rate for `f64` that preserves the rational representation.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct F64Rate {
+    pub(crate) num: f64,
+    pub(crate) den: f64,
+}
```
younies, line 77: Just to confirm, we store `num` and `den` as `f64` here so they can be fed directly into FMA (`f64::mul_add`) without type-casting overhead, right?

If so, it would be super helpful to document that rationale right here. Seeing a "rational representation" using floating-point types is a bit counter-intuitive at first glance, so a quick comment would clear up that confusion for future readers!

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -0,0 +1,84 @@
+// This file is part of ICU4X. For terms of use, please see the file
+// called LICENSE at the top level of the ICU4X source tree
+// (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
+#[allow(unused_imports)]
+use core_maths::*;
+
+/// Computes `a * num / den` with higher precision using Fused Multiply-Add (FMA).
```
younies, line 8: +1

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -0,0 +1,84 @@
...
+        // Naive: 0.010000000000000002
+        // FMA: 0.01
+        let val = 0.1;
+        let num = 1.0;
+        let den = 10.0;
+        let naive = val * (num / den);
+        let fma = f64_mul_div(val, num, den);
+        assert_ne!(naive, fma);
+        assert_eq!(fma, 0.01);
+        assert_eq!(naive, 0.010000000000000002);
```
younies: shall we use precision testing (with delta) ?

younies (approved):
I left some comments

components/experimental/src/units/convertible.rs
```diff
@@ -9,22 +9,30 @@ use num_traits::ToPrimitive;
 // TODO: add Mul & Add for references to avoid cloning.
 /// A trait for types that can be converted between two units.
 pub trait Convertible: Clone {
+    /// The type representing the conversion rate.
+    type Rate: Clone + core::fmt::Debug;
```
robertbastian: suggestions: `Ratio` or `Factor` are better terms (more mathematical) than `Rate`

I particularly like `Ratio`; `Ratio::<BigInt>::Ratio = Ratio<BigInt>` makes sense

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -0,0 +1,84 @@
...
+        // FMA: 0.02
+        let val = 0.1;
+        let num = 1.0;
+        let den = 5.0;
+        let naive = val * (num / den);
+        let fma = f64_mul_div(val, num, den);
+        assert_ne!(naive, fma);
+        assert_eq!(fma, 0.02);
+        assert_eq!(naive, 0.020000000000000004);
+    }
```
robertbastian, line 99: please add test cases for NaN and $\infty$. we probably want invariants on `RatioF64` that it's finite

components/experimental/src/units/convertible.rs
```diff
@@ -35,14 +43,27 @@ impl Convertible for Ratio<BigInt> {
...
+    }
+
     fn reciprocal(&self) -> Self {
         self.recip()
     }
 }
 
+/// A conversion rate for `f64` that preserves the rational representation.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct F64Rate {
```
robertbastian: I was hoping to say use `Ratio<f64>` for this, but that's not a valid type.

Call this `RatioF64` though to mirror `Ratio`

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -0,0 +1,84 @@
...
+///
+/// This uses the native FMA operation when available, falling back to the `libm` crate
+/// on platforms that don't support it.
+///
+/// If the FMA calculation results in an infinite value or NaN (e.g., due to overflow
+/// in intermediate steps), it falls back to the naive `a * (num / den)`.
+///
+/// For more details and the mathematical justification, see
+/// [this post by Waldemar Horwat](https://github.com/tc39/proposal-amount/issues/115).
+#[inline]
```
robertbastian:
```suggestion
#[inline(always)]
```

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -0,0 +1,84 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_f64_mul_div_precision() {
+        // Case 1: 5 grams to tonnes (1/1_000_000)
+        // Naive: 5.0 * (1.0 / 1_000_000.0) = 5.000000000000001e-6
+        // FMA: 5e-6
```
robertbastian: don't call this algorithm "FMA". it uses FMA in its implementation, but the operation is conceptually not a mutiply-add

```suggestion
        // f64_mul_div: 5e-6
```

components/experimental/src/units/convertible.rs
```diff
@@ -35,14 +43,27 @@ impl Convertible for Ratio<BigInt> {
...
+    pub(crate) num: f64,
```
robertbastian: nit: either type out `numerator` and `denominator` or just use a tuple struct

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -0,0 +1,84 @@
...
+/// [this post by Waldemar Horwat](https://github.com/tc39/proposal-amount/issues/115).
+#[inline]
+pub(super) fn f64_mul_div(a: f64, num: f64, den: f64) -> f64 {
+    let hi = a * num;
+    let err = a.mul_add(num, -hi);
+    let res = hi / den;
+    let rem = res.mul_add(-den, hi) + err;
+    let b = res + (rem / den);
+
+    if b.is_infinite() || b.is_nan() {
```
robertbastian:
```suggestion
    if !b.is_finite() {
```

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -0,0 +1,84 @@
...
+        // Fallback to simple multiplication if FMA fails due to overflow/NaN
+        a * (num / den)
+    } else {
+        b
+    }
```
robertbastian: the variable names are not very understandable. for example, `res` is some intermediate value, not the actual result. `hi` and `rem` are also pretty useless.

this could also use comments

```suggestion
    let double_rounded = a * num / den;

    // The multiplication error is the difference of the rounded multiplication
    // and the multiplication evaluated in full precision (FMA)
    let multiplication_error = a.mul_add(num, -(a * num));

    // The total error is the difference between the rounded multiplication+division
    // and the multiplication evaluated in full precision, plus the multiplication
    // error divided by the denominator, all evaluated in times-den-space. 
    let total_error = (double_rounded.mul_add(-den, a * num) + multiplication_error) / den;
    
    if total_error.is_finite() {
        double_rounded + total_error
    } else {
        a * (num / den)
    }
```

components/experimental/src/units/convertible.rs
```diff
@@ -53,6 +74,12 @@ impl Convertible for f64 {
         ratio.to_f64()
     }
 
+    fn rate_from_ratio_bigint(ratio: Ratio<BigInt>) -> Option<Self::Rate> {
+        let num = ratio.numer().to_f64()?;
+        let den = ratio.denom().to_f64()?;
+        Some(F64Rate { num, den })
```
robertbastian: issue: I don't like this being fallible. we should add an invariant on `conversion_factor` that it's always convertible into a `RatioF64`

components/experimental/src/units/f64_mul_div.rs
sffc: In this test, we are interested in the exact floating point results.

components/experimental/src/units/f64_mul_div.rs
sffc: I don't actually like this function being forced to be inline. This function has at least several system calls on modern hardware, and more when the `core_maths` polyfill is used. I would rather move the `#[inline]` up a level and leave this one to the compiler's discretion.

robertbastian: this function has a single call site, why would you not inline it.

I'd actually prefer if this didn't live in a separate file but in the trait impl.

components/experimental/src/units/convertible.rs
sffc: Acknowledged and agreed, but this is a pre-existing. The function above this one, `from_ratio_bigint`, has the same `Option` pattern. I would like to fix this in a follow-up.

components/experimental/src/units/f64_mul_div.rs
sffc: the agent originally put it in the trait impl, but I moved it to a separate file because interesting bits of business logic are best in their own file because of test locality, ease of referencing the file from elsewhere

components/experimental/src/units/convertible.rs
```diff
@@ -67,7 +67,9 @@ impl Convertible for f64 {
     type Ratio = RatioF64;
 
     fn mul_ratio(&self, ratio: &Self::Ratio) -> Self {
-        super::f64_mul_div::f64_mul_div(*self, ratio.numerator, ratio.denominator)
+        let result = super::f64_mul_div::f64_mul_div(*self, ratio.numerator, ratio.denominator);
+        debug_assert!(result.is_finite());
```
sffc: only assert this if the input was finite.

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -78,4 +78,21 @@ mod tests {
         assert_eq!(fma, 0.02);
         assert_eq!(naive, 0.020000000000000004);
     }
+
+    #[test]
+    fn test_f64_mul_div_nan_infinity() {
```
sffc: also test divide by zero and 0/0

components/experimental/src/units/f64_mul_div.rs
sffc: @robertbastian As a heuristic for what to inline, I prefer to inline one-liners, and fns like this that have actually good names and actually good scope are the exact ones to not force-inline. The compiler will almost certainly inline it in release builds if it has one call site, but in debug builds, it is actually _helpful_ to not inline this, because then you can see for example how much binary size this fn is contributing.

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -81,4 +77,33 @@ mod tests {
...
+
+    #[test]
+    fn test_f64_mul_div_extreme_cases() {
+        // If we pass NaN, it should return NaN
+        assert!(f64_mul_div(f64::NAN, 1.0, 2.0).is_nan());
+        assert!(f64_mul_div(1.0, f64::NAN, 2.0).is_nan());
+        assert!(f64_mul_div(1.0, 1.0, f64::NAN).is_nan());
+
+        // If we pass Infinity, it might return NaN due to intermediate Inf - Inf
+        assert!(f64_mul_div(f64::INFINITY, 1.0, 2.0).is_nan());
```
robertbastian, line 115: well this is incorrect. $\infty * a / b$ should be $\infty$ (for positive $a/b$, negative for negative).

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -81,4 +77,33 @@ mod tests {
...
+        assert!(f64_mul_div(1.0, f64::INFINITY, 2.0).is_nan());
```
robertbastian, line 116: same here

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -81,4 +77,33 @@ mod tests {
...
+        assert!(f64_mul_div(1.0, 1.0, f64::INFINITY).is_nan());
```
robertbastian, line 117: $a * b / \infty$ should be $0$

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -81,4 +77,33 @@ mod tests {
...
+        assert!(f64_mul_div(1.0, 1.0, 0.0).is_nan());
+        assert!(f64_mul_div(0.0, 1.0, 0.0).is_nan());
+
+        // 0/0 (indeterminate forms)
+        assert!(f64_mul_div(1.0, 0.0, 0.0).is_nan());
+        assert!(f64_mul_div(0.0, 0.0, 0.0).is_nan());
+
+        // Intermediate overflow (known limitation: FMA fails/returns NaN due to
+        // intermediate overflow in a * num, even though the naive math is finite)
+        assert!(f64_mul_div(1e300, 1e300, 1e300).is_nan());
```
robertbastian, line 135: you had code that handled this limitation, why did you remove that?

components/experimental/src/units/converter.rs
```diff
@@ -114,8 +114,8 @@ pub(crate) struct ProportionalConverter<N>
 where
     N: Convertible,
 {
-    /// The conversion rate between the input and output units.
-    pub(crate) conversion_rate: N::Rate,
+    /// The conversion ratio between the input and output units.
+    pub(crate) conversion_ratio: N::Ratio,
```
robertbastian: less keen on renaming this field to conversion ration. the *type* should be ratio, because it's conceptually `Ratio<f64>`, i.e. the ratio between two floats.

the *value* should not be called ratio. if I didn't know what this was, I'd interpret a field called ratio as a divisor, i.e. the 100 in liters-per-100-kilometers

also, you describe this as "the conversion ratio between input and output units", but it's actually the inverse of that: $output / input =  (input * conversionratio) / input  = conversionratio$.

this should be called conversion *factor* because you multiply it with the value, not divide the value by it.

rate was also fine though if you don't like factor

components/experimental/src/units/convertible.rs
```diff
@@ -9,50 +9,97 @@ use num_traits::ToPrimitive;
...
     fn reciprocal(&self) -> Self {
         self.recip()
     }
 }
 
+/// A conversion ratio for `f64` that preserves the rational representation.
+///
+/// We store the `numerator` and `denominator` as `f64` (instead of integers)
+/// so they can be fed directly into the precision-improving division algorithm
+/// (which uses `f64::mul_add`).
```
robertbastian: please add a comment that numerator and denominator are always integers, so this is an exact representation of a `Ratio<BigInt>`. realising this really helped me understand this

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -0,0 +1,109 @@
...
+/// This is particularly useful for unit conversion where the conversion rate is represented
+/// as a rational fraction `num / den`, and we want to compute `a * (num / den)` more accurately
+/// than the naive calculation.
+///
+/// This uses the native FMA operation when available, falling back to the `libm` crate
+/// on platforms that don't support it.
+///
+/// For more details and the mathematical justification, see
+/// [this post by Waldemar Horwat](https://github.com/tc39/proposal-amount/issues/115).
+pub(super) fn f64_mul_div(a: f64, num: f64, den: f64) -> f64 {
```
robertbastian, line 33: suggestion: `impl Mul<f64> for RatioF64`?

components/experimental/src/units/convertible.rs
```diff
@@ -9,50 +9,97 @@ use num_traits::ToPrimitive;
...
+    pub(crate) denominator: f64,
+}
+
 impl Convertible for f64 {
-    fn mul_refs(&self, other: &Self) -> Self {
-        self * other
+    type Ratio = RatioF64;
+
+    #[inline]
+    fn mul_ratio(&self, ratio: &Self::Ratio) -> Self {
```
robertbastian: suggestion: this could take `ratio: &Ratio<BigInt>`, and you do the logic from `ratio_from_ratio_bigint` here. the output of that function is only ever used as the input of this one (modulo intermediate storage), and the whole integerness of `RatioF64` becomes a lot clearer if this code is merged

storing `Ratio<BigInt>` in `ProportionalConverter` also removes the `Covertible` generic from `UnitsCoverter`, instead its `convert` method becomes generic. I think that makes much more sense, so that the same converter can be used with all input types

components/experimental/src/units/f64_mul_div.rs
sffc: I prefer a named function. It has only the one call site right now as you've observed. If we get a lot more call sites then maybe we can add an `impl Mul`.

robertbastian (changes requested):
I tried to compromise here, but I think the changes this makes to the public `Convertible` trait are going in the wrong direction

components/experimental/src/units/convertible.rs
sffc: As I commented in https://github.com/unicode-org/icu4x/pull/8073, I don't think we should be using BigInts at all in the float path.

components/experimental/src/units/f64_mul_div.rs
robertbastian: > in debug builds, it is actually helpful to not inline this, because then you can see for example how much binary size this fn is contributing.

please stop coming up with ridiculous arguments for why you want to ignore my review comments. build size of debug builds is not a metric

> fns like this that have actually good names and actually good scope are the exact ones to not force-inline

the compiler doesn't care about good names. this just makes it harder for the review to review this function. for example, it's not clear where the function is used, that the `num` and `den` arguments are always integers, etc

components/experimental/src/units/converter.rs
sffc: `factor` is fine. done.

components/experimental/src/units/f64_mul_div.rs
sffc: I hypothesize that it is enough of an edge case thing that I would rather land without the extra branch, and we can add the branch later if we have an actual bug to fix.

I considered making it a debug panic.

components/experimental/src/units/f64_mul_div.rs
Manishearth: FWIW `inline(always)` has been an antipattern in Rust for a long time. It makes sense when micro-optimizing, BUT you should try to just use `inline` unless you really have measured everything within an inch of its life and the performance is relevant.

LLVM is smart, and LLVM plus `inline` will basically always inline anyway.

A thing people have observed is that `inline(always)` tends to actually slow builds down at times. I'm not sure why, and this might be a bug that is since fixed, but back in the day it appeared to be an LLVM-side issue.

Many comments in [this thread](https://www.reddit.com/r/rust/comments/vxuz9r/inlining_functions_that_will_only_ever_be_called/) suggesting people avoid `inline(always)`.

components/experimental/src/units/f64_mul_div.rs
sffc: 8ebdfc8

components/experimental/src/units/f64_mul_div.rs
sffc: 8ebdfc8

components/experimental/src/units/f64_mul_div.rs
sffc: 8ebdfc8

components/experimental/src/units/convertible.rs
sffc: done.

sffc:
> I tried to compromise here, but I think the changes this makes to the public `Convertible` trait are going in the wrong direction

This PR implements the new FMA logic in a way that is consistent with the current trait structure of this module.

What this PR does in terms of the trait:

1. Adds a new associated type to the existing trait
2. Adds one new trait function
3. Updates existing trait functions to use the new associated type

It has no impact on the userland call site. Sure, it makes the code complexity of this crate slightly higher. But, _the mechanism for doing the change in this PR is consistent with the pre-existing design of the traits and types._

The only reason to block this PR is if you believe that the new associated type and trait function do not justify the benefit of FMA. _I think_ it's a _slam dunk_ because FMA significantly improves the accuracy _and_ aligns us with ECMA. But, you are welcome to disagree.

If you believe FMA is a good change, please unblock this PR. Then, we can discuss a new trait structure in #8073. In that PR, you are welcome to cite the new associated type as evidence that the trait is too complex.

components/experimental/src/units/f64_mul_div.rs
sffc: > please stop coming up with ridiculous arguments for why you want to ignore my review comments.

Please stop calling my arguments ridiculous or made up. It is both the wrong culture and also simply not true.

I am honestly concerned about potential code size implications of this[^1], and it is easier to analyze named functions in ELF and Wasm files than trait functions[^2].

The inline attribute added here at first was AI slop; I never intended it to be inline.

[^1]: not all platforms have FMA, and the libm polyfill is probably pretty big, and even on platforms with FMA, I don't know the impact on instruction count. I care about this because a V8 tech lead said I should care about it. 

[^2]: based on my experience using tools like Ida Pro and Twiggy.

Manishearth: I disagree with Robert's framing but i am, like him, confused why "code size in debug" is a metric we care about. We haven't historically cared about that, and I struggle to see how it's useful.

robertbastian: As far as I know, code properties we care about are
* readability
* performance (in release)
* code size (in release)

Outlining this into its own function probably does not impact code size or performance, because as a tiny function with a single call site, the compiler will very likely inline it. It does, however, impact readability, because the whole context that this is being called from a single call site with only integer-valued arguments, is lost.

components/experimental/src/units/convertible.rs
```diff
@@ -81,10 +83,14 @@ impl Convertible for f64 {
     fn ratio_from_ratio_bigint(ratio: Ratio<BigInt>) -> Option<Self::Ratio> {
         let numerator = ratio.numer().to_f64()?;
         let denominator = ratio.denom().to_f64()?;
-        Some(RatioF64 {
-            numerator,
-            denominator,
-        })
+        if numerator.is_finite() && denominator.is_finite() {
+            Some(RatioF64 {
```
robertbastian: issue: the invariant needs to be stated on the type, otherwise it gets lost

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -33,14 +30,10 @@ pub(super) fn f64_mul_div(a: f64, num: f64, den: f64) -> f64 {
...
+    // error divided by the denominator, all evaluated in times-den-space.
     let total_error = (double_rounded.mul_add(-den, a * num) + multiplication_error) / den;
-    
-    if total_error.is_finite() {
-        double_rounded + total_error
-    } else {
-        a * (num / den)
-    }
+
+    double_rounded + total_error
```
robertbastian, line 49: I don't think `num.is_finite() && den.is_finite()` implies that `total_error.is_finite()`.

components/experimental/src/units/convertible.rs
robertbastian: why is this not `#[inline]` then?

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -0,0 +1,124 @@
...
+        assert_eq!(naive, 0.010000000000000002);
+
+        // Case 3: 0.1 * (1.0 / 5.0)
+        // Naive: 0.020000000000000004
+        // f64_mul_div: 0.02
+        let val = 0.1;
+        let num = 1.0;
+        let den = 5.0;
+        let naive = val * (num / den);
+        let fma = f64_mul_div(val, num, den);
```
robertbastian:
```suggestion
        let f64_mul_div = f64_mul_div(val, num, den);
```

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -0,0 +1,124 @@
...
+    #[test]
+    fn test_f64_mul_div_precision() {
+        // Case 1: 5 grams to tonnes (1/1_000_000)
+        // Naive: 5.0 * (1.0 / 1_000_000.0) = 5.000000000000001e-6
+        // f64_mul_div: 5e-6
+        let val = 5.0;
+        let num = 1.0;
+        let den = 1_000_000.0;
+        let naive = val * (num / den);
+        let fma = f64_mul_div(val, num, den);
```
robertbastian:
```suggestion
        let f64_mul_div = f64_mul_div(val, num, den);
```

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -0,0 +1,124 @@
...
+        assert_eq!(naive, 0.0000049999999999999996);
+
+        // Case 2: 0.1 * (1.0 / 10.0)
+        // Naive: 0.010000000000000002
+        // f64_mul_div: 0.01
+        let val = 0.1;
+        let num = 1.0;
+        let den = 10.0;
+        let naive = val * (num / den);
+        let fma = f64_mul_div(val, num, den);
```
robertbastian:
```suggestion
        let f64_mul_div = f64_mul_div(val, num, den);
```

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -0,0 +1,124 @@
...
+
+        // Division by zero
+        assert!(f64_mul_div(1.0, 1.0, 0.0).is_nan());
+        assert!(f64_mul_div(0.0, 1.0, 0.0).is_nan());
+
+        // 0/0 (indeterminate forms)
+        assert!(f64_mul_div(1.0, 0.0, 0.0).is_nan());
+        assert!(f64_mul_div(0.0, 0.0, 0.0).is_nan());
+
+        // Intermediate overflow (known limitation: FMA fails/returns NaN due to
```
robertbastian:
```suggestion
        // Intermediate overflow (known limitation: f64_mul_div fails/returns NaN due to
```

components/experimental/src/units/f64_mul_div.rs
robertbastian: I don't think it's sensible to return `NaN` instead of infinity. I don't want to make this trade off just to avoid a branch, without knowing how much that branch even costs (there are divisions in this code which might outweigh any branch pipeline stalls anyway).

In any case, this should be _documented on the API_.

We should anyway document the floating point semantics we're implementing here (on `impl Convertible for f64`). Our current semantics are that the conversion factor is evaluated in full precision (`BigInt`), if we cannot convert that to `f64` we *fail*, and then the multiplication by the conversion factor introduces two more roundings (which this PR changes to one, but it still fails if the `RatioF64` is not exact). The addition of the offset also introduces one more rounding. I believe you want to change the calculation of the conversion factor to use floating point roundings as well instead of being fallible?

components/experimental/tests/units/units_test.rs
```diff
@@ -290,3 +290,40 @@ fn test_unparsable_units() {
...
     });
 }
+
+#[test]
+fn test_fma_precision() {
+    let factory = ConverterFactory::new();
+
+    // Case 1: 5 g to tonnes (factor 1/1000000)
+    // Today: 0.0000049999999999999996
+    // Expected with FMA: 0.000005 (exactly representable double closest to 5e-6)
```
robertbastian:
```suggestion
    // Expected with f64_mul_div: 0.000005 (exactly representable double closest to 5e-6)
```

components/experimental/tests/units/units_test.rs
```diff
@@ -290,3 +290,40 @@ fn test_unparsable_units() {
...
+    let g = MeasureUnit::try_from_str("gram").unwrap();
+    let tonne = MeasureUnit::try_from_str("tonne").unwrap();
+    let converter = factory.converter::<f64>(&g, &tonne).unwrap();
+    let result = converter.convert(&5.0);
+    assert_eq!(result, 5e-6);
+
+    // Case 2: 825 g to kg (factor 1/1000)
+    // Today: 0.8250000000000001
+    // Expected with FMA: 0.825 (exactly representable double closest to 0.825)
```
robertbastian:
```suggestion
    // Expected with f64_mul_div: 0.825 (exactly representable double closest to 0.825)
```

components/experimental/src/units/convertible.rs
```diff
@@ -9,50 +9,100 @@ use num_traits::ToPrimitive;
...
     fn reciprocal(&self) -> Self {
         self.recip()
     }
 }
 
+/// A conversion ratio for `f64` that preserves the rational representation.
+///
+/// We store the `numerator` and `denominator` as `f64` (instead of integers)
+/// so they can be fed directly into the precision-improving division algorithm
+/// (which uses `f64::mul_add`).
```
robertbastian:
```suggestion
/// (`f64_mul_div`).
```

components/experimental/tests/units/units_test.rs
```diff
@@ -290,3 +290,40 @@ fn test_unparsable_units() {
         );
     });
 }
+
+#[test]
+fn test_fma_precision() {
```
robertbastian:
```suggestion
fn test_f64_precision() {
```

components/experimental/tests/units/units_test.rs
```diff
@@ -290,3 +290,40 @@ fn test_unparsable_units() {
...
+    let factory = ConverterFactory::new();
+
+    // Case 1: 5 g to tonnes (factor 1/1000000)
+    // Today: 0.0000049999999999999996
```
robertbastian: "Today: " will not age well

components/experimental/src/units/f64_mul_div.rs
sffc: Back when I was more routinely looking at code size issues in the early days of ICU4X, I would build with parameters that apply most optimizations but don't inline as many functions. If there is a single function with a whole lot of code, it is easier to spot than if the whole call tree is inlined.

My use of the term "debug build" in this context was an elision of "a build setting which is not quite release, retaining certain debug artifacts like function names and less inlining, while still being useful as a proxy metric for the code size of the fully optimized build." I'm surprised that wasn't obvious and I needed to type it out.

---

Putting this code in its own file improves readability because (1) it keeps the tests more localized and (2) it explicitly highlights that this code is interesting / nontrivial / non-obvious business logic, rather than just trait impl plumbing.

robertbastian (commented):
I think you should test the `f64` implementation against the `Ratio<BigInt>` implementation, which is the optimal `f64` implementation with a single rounding

it probably makes sense to have a lot more test cases, maybe randomly generated

components/experimental/src/units/convertible.rs
sffc: The type `RatioF64` already says

```
/// The numerator and denominator are always integers, so this is an exact
/// representation of a `Ratio<BigInt>`.
```

Did you want more, like the `is_finite` being documented? I think all integers are finite so it seems implied to me but I'm happy to reiterate it.

components/experimental/src/units/convertible.rs
sffc: Because it's more than 1 line, but I guess this is safe to inline; I don't see anything here of value that would get lost in code size analysis

sffc:
I opened https://github.com/unicode-org/icu4x/issues/8094 to investigate whether we can get fahrenheit celsius exact too.

components/experimental/src/units/convertible.rs
```diff
@@ -12,12 +12,12 @@ pub trait Convertible: Clone {
     type Addend: core::fmt::Debug + Clone;
     type Result: core::fmt::Debug;
 
-    /// Computes `self * factor + addend`.
-    fn mul_add(self, factor: &Self::Factor, addend: &Self::Addend) -> Self::Result;
-
     /// Computes `self * factor`
     fn mul(self, factor: &Self::Factor) -> Self::Result;
 
+    /// Computes `self * factor + addend`.
```
robertbastian, line 15: unnecessary diff

components/experimental/src/units/convertible.rs
```diff
@@ -33,57 +33,92 @@ impl Convertible for &'_ Ratio<BigInt> {
     type Addend = Ratio<BigInt>;
     type Result = Ratio<BigInt>;
 
-    // Exact
+    #[inline]
```
robertbastian, line 37: issue (everywhere in this file): retain previous comments regarding accuracy and infallible unwraps

components/experimental/src/units/convertible.rs
```diff
@@ -33,57 +33,92 @@ impl Convertible for &'_ Ratio<BigInt> {
...
+/// conversion is performed in a single rounded operation using a precision-improving
+/// FMA-based algorithm (`f64_mul_div`).
+///
+/// However, because the algorithm tracks errors using floating-point math, it has
+/// the following limitations:
+/// - **Non-finite inputs/results:** If the input value, the ratio components, or the
+///   intermediate product `value * numerator` is non-finite (NaN, Infinity, or overflow),
+///   the conversion will return `NaN`. This is a known limitation where the FMA error-tracking
+///   math encounters indeterminate forms (like `Inf - Inf`), even if naive math would
+///   have returned `Infinity` or `0.0` (tracked in [unicode-org/icu4x#8092](https://github.com/unicode-org/icu4x/issues/8092)).
```
robertbastian: why are these semantics listed on this type? it's behaviour of `f64_mul_div`

what I asked to add to the type is an invariant that both fields are non-`NaN`, as this is an invariant that is established in `factor_from_ratio_bigint` and used in `mul`, and it needs to be tracked across these functions

components/experimental/src/units/convertible.rs
```diff
@@ -33,57 +33,92 @@ impl Convertible for &'_ Ratio<BigInt> {
...
     fn mul_add(self, factor: &Self::Factor, addend: &Self::Addend) -> Self::Result {
-        self * factor + addend
+        super::f64_mul_div::f64_mul_div(self, factor.numerator, factor.denominator) + addend
     }
 
-    // TODO: reduce error
+    #[inline]
     fn reciprocal_mul(self, factor: &Self::Factor) -> Self::Result {
-        1.0 / (self * factor)
+        1.0 / super::f64_mul_div::f64_mul_div(self, factor.numerator, factor.denominator)
```
robertbastian, line 101: issue: I'm not convinced this is more accurate than
`factor.denominator / (self * factor.nominator)`

robertbastian (commented):
did you see #8088?

components/experimental/src/units/convertible.rs
sffc: Sorry about this one. I think it was from the messy merge so I didn't catch it first.

components/experimental/src/units/convertible.rs
sffc: Also added comments in the places where the invariant is established and used.

sffc:
> did you see #8088?

Acknowledged. #8088 makes different choices, such as not making a new file and function for this. Since we are already deep in those discussions here, I figured we should finish here instead of moving over there.

components/experimental/tests/units/units_test.rs
```diff
@@ -291,3 +291,99 @@ fn test_unparsable_units() {
...
+        ("gram", "tonne", 5.0, true),
+        ("gram", "tonne", 0.123456789, true),
+        ("gram", "kilogram", 825.0, true),
+        ("inch", "centimeter", 5.0, true),
+        ("mile-per-hour", "kilometer-per-hour", 120.0, true),
+        // Offset conversions
+        ("fahrenheit", "celsius", 32.0, true), // 32 F = 0 C (exact)
+        ("fahrenheit", "celsius", 212.0, true), // 212 F = 100 C (exact)
+        ("fahrenheit", "celsius", -40.0, true), // -40 F = -40 C (exact)
+        ("fahrenheit", "celsius", 0.123456789, false), // 1 unit off (verified)
```
robertbastian: verified?

components/experimental/tests/units/units_test.rs
```diff
@@ -291,3 +291,99 @@ fn test_unparsable_units() {
...
+        let converter_f64 = factory.converter::<f64>(&input_unit, &output_unit).unwrap();
+        let converter_bigint = factory
+            .converter::<&Ratio<BigInt>>(&input_unit, &output_unit)
+            .unwrap();
+
+        // 1. Compute using f64 (FMA)
+        let actual = converter_f64.convert(val);
+
+        // 2. Compute using Ratio<BigInt> (exact + single rounding)
+        let val_ratio = Ratio::from_float(val).expect("Failed to convert float to Ratio");
```
robertbastian, line 331: issue: this adds more roundings to the exact ratio code path, as `0.123456789` is not exact in floats. you should set up the test cases with integers

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -0,0 +1,137 @@
...
+/// [this post by Waldemar Horwat](https://github.com/tc39/proposal-amount/issues/115).
+///
+/// # Limitations
+/// This algorithm does not handle non-finite inputs (NaN, Infinity) or intermediate
+/// overflows (e.g., `a * num > f64::MAX`) gracefully. In these cases, the error-tracking
+/// math will naturally encounter indeterminate forms (like `Inf - Inf` or `0 * Inf`) and
+/// return `NaN`, even if the naive calculation would have returned a sensible value
+/// (like `Infinity` or `0.0`). We accept this limitation to keep the function extremely
+/// simple, fast, and branchless, relying on callers to enforce finiteness invariants
+/// (which we do in `RatioF64` and `Convertible::mul_ratio`) (tracked in
```
robertbastian:
```suggestion
/// (which we do in `RatioF64` and `Convertible::mul`) (tracked in
```

components/experimental/src/units/f64_mul_div.rs
robertbastian: I really think figuring this out (especially if the claim that we need to avoid the branch holds water) is an important part of this PR and shouldn't be chucked on the backlog

components/experimental/tests/units/units_test.rs
```diff
@@ -291,3 +291,99 @@ fn test_unparsable_units() {
         );
     });
 }
+
+#[test]
+fn test_f64_precision() {
```
robertbastian: does this test anything that the other test doesn't?

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -0,0 +1,137 @@
...
+#[allow(unused_imports)]
+use core_maths::*;
+
+/// Computes `a * num / den` with higher precision using Fused Multiply-Add (FMA).
+///
+/// This function implements an algorithm that reduces rounding errors in floating-point
+/// division by utilizing `f64::mul_add` to compute the remainder of the multiplication
+/// and incorporate it into the division.
+///
+/// This is particularly useful for unit conversion where the conversion rate is represented
```
robertbastian: yes this is particularly useful for the single place that uses it, in the crate where it's defined

components/experimental/src/units/f64_mul_div.rs
```diff
@@ -0,0 +1,137 @@
...
+// (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
+
+#[allow(unused_imports)]
+use core_maths::*;
+
+/// Computes `a * num / den` with higher precision using Fused Multiply-Add (FMA).
+///
+/// This function implements an algorithm that reduces rounding errors in floating-point
+/// division by utilizing `f64::mul_add` to compute the remainder of the multiplication
+/// and incorporate it into the division.
```
robertbastian:
```suggestion
/// division by utilizing `f64::mul_add` to compute the errors of the multiplication
/// and division.
```

components/experimental/src/units/f64_mul_div.rs
sffc: I prefer to start with a minimalistic solution, and fix it once we have a need to fix it.

I consider the exact nature of the fallback behavior of intermediate overflows to be an edge case we will probably eventually need to address, but one that should wait for a clearer use case and not be rushed or forced, since there are multiple ways we could go about doing it. For example, if multiplying by 5/3 causes intermediate overflow, should we do something like reordering the multiply and divide? Or introducing an intermediate factor like 2/3 then 5/2? Or fall back to BigInt? Or give the user multiple options to choose from? This needs use cases and discussion that I'm not in a position to provide at this moment.

sffc:
The f64_mul_div should really be its own crate, complete with toggles to tune the behavior, maybe some feature flags for platform support, and probably with more fns like f64_div_div and f64_mul_div_add.

Making it a standalone file with its own test suite is a good first step.

components/experimental/tests/units/units_test.rs
```diff
@@ -303,7 +303,7 @@ fn test_f64_vs_bigint_precision() {
     let test_cases = [
         // Proportional conversions (always exact)
         ("gram", "tonne", 5.0, true),
-        ("gram", "tonne", 0.123456789, true),
+        ("gram", "tonne", 0.125, true),
```
robertbastian, line 306: the correctness of this test now relies on these floats being exact, which is neither documented anywhere nor easy to determine for future test cases. why not declare the test values as exact integer fractions?

components/experimental/src/units/f64_mul_div.rs
robertbastian: We currently have a minimalistic solution, which is correct according to very simple semantics. This PR is correct only according to very complicated semantics that are basically "you have to read the whole implementation". I consider this a regression.

I'm also still waiting for you to write out the semantics, as I have demonstrated in #8088

components/experimental/src/units/convertible.rs
```diff
@@ -34,54 +34,88 @@ impl Convertible for &'_ Ratio<BigInt> {
...
+        // Relies on the invariant: factor.numerator and factor.denominator are finite.
+        1.0 / super::f64_mul_div::f64_mul_div(self, factor.numerator, factor.denominator)
     }
 
+    #[inline]
     fn factor_from_ratio_bigint(factor: Ratio<BigInt>) -> Self::Factor {
-        // Ratio::<BigInt>::to_f64 is infallible
-        factor.to_f64().unwrap_or(f64::NAN)
+        // Ratio::<BigInt>::to_f64 is infallible. Establishes the invariant that
+        // numerator and denominator are finite integers.
```
robertbastian, line 107: issue: this is not actually true, see the implementation: https://docs.rs/num-bigint/latest/src/num_bigint/biguint/convert.rs.html#382-391

robertbastian (commented):
> The f64_mul_div should really be its own crate, complete with toggles to tune the behavior, maybe some feature flags for platform support, and probably with more fns like f64_div_div and f64_mul_div_add.

> Making it a standalone file with its own test suite is a good first step.

No it should not, in the same way that binary search with toggles shouldn't be in its own crate. It's a very simple algorithm that needs to be adapted to specific circumstances, where analysis in the general case is going to be prohibitively complicated and useless.

components/experimental/src/units/convertible.rs
robertbastian: I'm frustrated that it took like 4 rounds of feedback for you and your bot to state the invariant here

sffc:
>  It's a very simple algorithm

Multiple very smart people were not familiar with fused multiply add being applicable to a situation with multiply and divide. Just because something is "simple" does not mean it shouldn't be in its own module or crate.

sffc:
I'm not interested in finishing this PR because I don't have the time right now to define and defend a robust algorithm. I thought it was better to implement the new algorithm that improves the 99% case and open follow-ups to better define the 1% case, which is what I was trying to do here, but it's clear we don't have consensus on that approach.

I opened https://github.com/unicode-org/icu4x/issues/8116 to track future work on this.

You can keep working on #8088 if you want.

sffc:
Backed up as https://github.com/sffc/omnicu/tree/archive/fma-unit-converter

Manishearth:
N.b. PRs have their git history permanently available under `pull/#/head` (e.g.: https://github.com/unicode-org/icu4x/tree/pull/8058/head) so you don't have to archive branches.

components/experimental/src/units/f64_mul_div.rs
sffc: This PR makes the 99% case better by removing an intermediate float rounding. I think it improves the code base to land this. But, I don't think my approach of "step in the right direction, figure out detailed edge case handling later" is compatible with your requirement of "write out and defend the semantics in this PR".

Issue #8060, opened 2026-06-10
Consider optimizing range patterns


Manishearth:
Our current design for range patterns is to store them in PatternsByGreatestDifference, as a pattern.

However, most range patterns have a fair amount of duplication: A range pattern will typically look something like `A - A B" where A and B are pattern strings like `y/M/d`. We could recognize this structure and store them in an efficient way so the duplication doesn't actually happen.

I'm a little skeptical though: patterns are efficiently packed in a VarULE representation, and these strings are small. Adding more structure to this type will probably make it bigger in a way that doesn't work out. We might be able to make it work by using the pattern metadata and basically hijacking PatternULE to give it more interpretation in the case of range patterns (e.g. "if you see an X field, that just means you repeat the last pattern string").

> I can see two axes:
> 
> * all patterns (seem to) share the same glue (" - ")
> * each range pattern often, but not always, repeats the non-range pattern

_Originally posted by @robertbastian in https://github.com/unicode-org/icu4x/pull/8034#discussion_r3387075771_

Manishearth:
Threw Gemini at some ideas to see what it thought and if it could fill in the blanks.

A promising idea is to use the header to track the type of range pattern. We have symmetric range patterns (`A - A`) and asymmetric range patterns `A - B`. Note that `M d - M d y` is perfectly okay to model as an asymmetric pattern for this purpose.

The header tells us if a greatestDifferenceValue is symmetric or asymmetric, and that's how we know whether we should read two patterns or read one pattern twice.

A very nice property here is that we then _always_ store patterns that have no duplicate fields. This will simplify the actual range formatting impl, the original plan would have required us to complicate "format this pattern" code in datetime. We should probably verify that this property is true.

One small note is that this is using the glue pattern. It is possible that the glue isn't always the same (it ought to be, though? we can check). We have two bits per greatest difference field, so we can have the states be "empty, symmetric, asymmetric, NoGlue", and for NoGlue we store the patterns as e.g. `M d -` and `M d y`.


A second optimization I realized after looking at the data is that the data has lots of spots where we have multiple identical greatestDifference entries. We should just ... not do that.


<details>

# Range Pattern Optimization Approaches

This document outlines two separate, orthogonal optimizations to reduce the storage size of datetime range patterns in ICU4X (`PatternsByGreatestDifference`).

These optimizations can be applied independently or together for maximum impact.

---

## Optimization 1: Structural Decomposition & Glue Sharing

[...]

PR #8061, opened 2026-06-10
Deduplicate identical interval format patterns where possible

Manishearth:
This is the smaller optimization from https://github.com/unicode-org/icu4x/issues/8060

We have a number of cases where the data just looks like this:

```

						<intervalFormatItem id="yMd">
							<greatestDifference id="d">U/MM/dd～U/MM/dd</greatestDifference>
							<greatestDifference id="M">U/MM/dd～U/MM/dd</greatestDifference>
							<greatestDifference id="y">U/MM/dd～U/MM/dd</greatestDifference>
						</intervalFormatItem>
```

This is silly, we should just store this as a single pattern with `greatestDfiference=d`.

## Changelog: N/A

internal, new feature


Manishearth:
image

Manishearth:
One thing we should definitely do here is document exactly how the deduplication works both here and on the type itself.

There already is an implicit fallback algorithm which this relies on, but it would be nice for it to be documented.

components/datetime/src/provider/range_patterns.rs
```diff
@@ -125,6 +125,18 @@ impl GreatestDifferenceHeader {
...
+///
+/// ### Fallback Behavior
+/// At runtime, if a pattern for a specific greatest difference field is requested but not present
+/// in this struct, the resolver will fall back to the next larger present field (e.g., Day -> Month
+/// -> Year -> Era).
+///
+/// ### Datagen Deduplication
+/// To minimize data size, datagen leverages this fallback behavior to omit redundant patterns.
+/// If a smaller field (e.g., Day) has the exact same pattern as a larger field (e.g., Month),
+/// the smaller field is marked as absent in the header and its pattern is omitted. At runtime,
```
robertbastian, line 137: this contradicts the PR description which says

> This is silly, we should just store this as a single pattern with `greatestDifference=d`.

so which direction does the deduplication actually go?

robertbastian (commented):
please generate testdata

components/datetime/src/provider/range_patterns.rs
Manishearth: PR body is wrong, it should fall back to Y. gD=Y is the most general case

components/datetime/src/provider/range_patterns.rs
```diff
@@ -187,58 +199,86 @@ impl<'data> PatternsByGreatestDifference<'data> {
...
+    ///
     /// The `bit_index` must be <= 3.
+    ///
+    /// This function automatically sorts the inputs by `bit_index` and performs
+    /// fallback-based deduplication: if a pattern for a smaller field is identical
+    /// to the pattern for the next larger present field, the smaller field's pattern
+    /// is omitted, as it will naturally fall back to the larger field at runtime.
     #[cfg(feature = "datagen")]
-    pub fn try_from_sorted<I>(iter: I) -> Result<Self, &'static str>
+    pub fn try_from_patterns<I>(iter: I) -> Result<Self, &'static str>
```
sffc, line 211: Vent: we previously discussed tradeoffs and decided to use a sorted iterator (with a BTreeMap frontend). Changing this function's contract to take a potentially-unsorted iterator is orthogonal to the bug this PR is fixing. I do not disagree with moving the sorting inside this fn, but I would really prefer you to decouple refactors, _especially_ previously-discussed ones, from functional changes.

sffc (approved):

PR #8063, opened 2026-06-11
Store range sub-patterns instead of patterns

### LINKED ISSUE #8060: Consider optimizing range patterns

Manishearth:
Fixes #8060

Progress on #5448

This is the other optimization from https://github.com/unicode-org/icu4x/issues/8060

The idea is to sometimes store PatternULEs for the left and right halves of a range pattern (connected by glue), and use the existing metadata to indicate how that works. We have three modes:

 - None: No pattern for this greatest difference mode, fall back to smaller difference
 - Symmetric: A single pattern is stored, and it renders as `[pat] [glue] [pat]`
 - FullRange: Single pattern

A previous version of this PR had an Asymmetric mode that worked as `[pat1] [glue] [pat2]` and a NoGlue one that didn't use the same glue.


Manishearth:
image

components/datetime/src/provider/range_patterns.rs
```diff
@@ -77,28 +77,96 @@ impl TimeGreatestDifferenceField {
...
+    Symmetric(Pattern<'a>),
+    /// Asymmetric pattern: `[left] [glue] [right]`. (2 patterns)
+    ///
+    /// The pattern is formed by formatting the start pattern, then the shared
+    /// fallback glue, then the end pattern.
+    ///
+    /// **Example**: `y M d – M d` (English Month range, e.g., `2020 Nov 5 – Dec 6`)
+    /// with fallback glue ` – `.
+    /// Decomposes to `Asymmetric { left: y M d, right: M d }`.
+    Asymmetric {
```
sffc: Observation: Symmetric is a special case of Asymmetric where the patterns are the same, so technically you don't need these to be separate variants. Actually, the whole enum is representable as `(Pattern<'a>, Pattern<'a>, bool)` where the bool says whether the glue pattern is needed.

sffc (commented):
This is good. I'm not sure `NoGlue` and `Asymmetric` save us much; doing the splitting at runtime is not too expensive _I think_. If we merge them into a `Ranged` with a runtime split then we can do direct indexing instead of the summing thing, and the data is more readable and mutatable.

components/datetime/src/provider/range_patterns.rs
sffc: In the range version of the intermediate (inside of `FormattedDateTime`, internal type `DateTimeZonePatternDataBorrowed`), you could store:

```rust
(PatternBorrowed<'a>, Option<DoublePlaceholderPatternBorrowed<'a>>)
```

and then do the splitting if the glue is None in `impl Writeable`, or you could store:

```rust
(PatternBorrowed<'a>, PatternBorrowed<'a>, DoublePlaceholderPatternBorrowed<'a>)
```

where you split ahead of time and use a no-op glue pattern `"{0}{1}"` if necessary.

components/datetime/src/provider/range_patterns.rs
```diff
@@ -155,62 +223,79 @@ pub struct PatternsByGreatestDifference<'data> {
...
+    /// GIGO: If the `resolved_field_idx` is `Absent`, it returns a default symmetric pattern
+    /// to avoid panicking at runtime.
+    fn get_pattern_internal<'a>(&'a self, resolved_field_idx: u8) -> RangePatternInfo<'a> {
+        let state = self.header.get_state(resolved_field_idx);
+        let start_idx = self.header.get_pattern_index(resolved_field_idx);
+
+        let get_pat = |offset| {
+            self.patterns
+                .get(start_idx + offset)
+                .map(<Pattern as zerofrom::ZeroFrom<PatternULE>>::zero_from)
```
sffc, line 250: praise: thanks for not cloning

provider/source/src/datetime/range_patterns.rs
```diff
@@ -133,12 +135,86 @@ fn skeleton_has_time_fields(skeleton: &Skeleton) -> bool {
     })
 }
 
+/// Extracts the glue string from a CLDR fallback pattern (e.g. "{0} – {1}").
+///
+/// Returns `None` if the pattern is invalid or if `{0}` and `{1}` are not at the ends
+/// of the string (e.g. if there is a prefix or suffix).
+fn extract_glue(fallback_str: &str) -> Option<&str> {
+    let pos0 = fallback_str.find("{0}")?;
+    let pos1 = fallback_str.find("{1}")?;
```
sffc: Issue: I really would prefer to avoid parsing things out of strings. I have found from experience that agents really like doing operations on strings. But we have DoublePlaceholderPattern that has a structured representation of this, and it handles the parsing, so we should use it.

provider/source/src/datetime/range_patterns.rs
```diff
@@ -133,12 +135,86 @@ fn skeleton_has_time_fields(skeleton: &Skeleton) -> bool {
...
+        Some(&fallback_str[pos1 + 3..pos0])
+    }
+}
+
+/// Decomposes a CLDR range pattern into sub-patterns based on the fallback glue.
+///
+/// If the pattern is symmetric (i.e. it can be split by the fallback glue and both sides
+/// are identical), it returns `RangePatternInfo::Symmetric`. Otherwise, it returns
+/// `RangePatternInfo::FullRange` containing the entire pattern.
+fn decompose_pattern(
```
sffc, line 176: Suggestion: change this to take a `DoublePlaceholderPattern` instead of `Option<&[PatternItem]>` and then do operations based on it. You are just trying to see if `pattern_str` can be decomposed using the glue pattern. I don't think you should be parsing it into `PattermItem`s at all in this step.

sffc (changes requested):

Manishearth:
Got it to work. I'm assuming that DoublePlaceholderPattern will always parse the intervening glue literals as a single `PatternItem::Literal`, whihc I think is a reasonable assumption and avoids intermediate allocs.

components/datetime/src/provider/range_patterns.rs
```diff
@@ -235,48 +320,64 @@ impl<'data> PatternsByGreatestDifference<'data> {
...
+            let keep = if let Some((_, next_info)) = it.peek() {
+                &info != next_info
             } else {
                 true // Always keep the last element
             };
 
             if keep {
-                header_val |= 1 << bit;
-                minimized_patterns.push(pattern);
+                minimized_infos.push((bit, info));
```
sffc: Suggestion: You don't need the two vectors any more, since they are the same length. You can do the mapping all in the first loop.

provider/source/src/datetime/range_patterns.rs
```diff
@@ -133,12 +136,98 @@ fn skeleton_has_time_fields(skeleton: &Skeleton) -> bool {
...
+    pattern_str: &str,
+    fallback_pattern: Option<&DoublePlaceholderPattern>,
+) -> RangePatternInfo<'static> {
```
sffc, line 179: I think we can clean up this fn more. I'm okay with a follow-up.

Manishearth: Did some cleanup.

sffc (approved):

PR #8066, opened 2026-06-11
experimental/currency: Implement runtime fallback for missing patterns and remove constructor validation

younies:
This PR enables support for incomplete or custom currency data by implementing runtime fallback for missing patterns in `CurrencyEssentials` and removing strict constructor validation in formatters.

Previously, we attempted to make positive patterns optional in `PatternIndices`, but reverted that approach to avoid footprint overhead in standard data. Instead, we keep positive patterns as non-optional `u8` indices, resolving fallbacks at datagen time for standard data, while introducing a runtime fallback safety net in the getters to support custom data.

## Changelog

* **Runtime Fallback in `CurrencyEssentials`**:
  * Positive pattern getters now return non-optional `&DoublePlaceholderPattern`. If an index is out of bounds (indicating a missing pattern in custom data), it falls back according to the hierarchy, with `FALLBACK_PATTERN` (`"{1}{0}"`) as the ultimate fallback.
  * Established positive pattern fallback hierarchies:
    * `standard_alpha_next_to_number_positive` ➔ `standard`
    * `accounting_positive` ➔ `standard`
    * `accounting_alpha_next_to_number_positive` ➔ `standard_alpha_next_to_number` ➔ `standard`
  * Negative pattern getters return `Option<&DoublePlaceholderPattern>`, with `accounting_negative` falling back to `standard_negative` if missing (`None`).
* **Relaxed Formatter Validation**:
  * Removed standard pattern validation from `CurrencyFormatter` and `CompactCurrencyFormatter` constructors. Formatters now initialize successfully even if the standard pattern is missing, relying on runtime fallback during formatting.
* **Documentation & Cleanups**:
  * Documented fallback hierarchies in `CurrencyEssentials` getters.
  * Added comments in datagen explaining datagen-time fallback resolution.
* **Data**:
  * Regenerated experimental baked data.

TAG=agy
CONV=de8f6008-79a8-45a5-9a6f-3332bb99904f


components/experimental/src/dimension/provider/currency/essentials.rs
```diff
@@ -179,9 +179,14 @@ impl<'a> CurrencyEssentials<'a> {
 
     /// Returns the standard pattern.
     pub fn standard_pattern(&self) -> Option<&DoublePlaceholderPattern> {
+        debug_assert!(
+            (self.indices.standard as usize) < self.patterns.len(),
+            "Standard pattern index {} is out of bounds for patterns of length {}",
+            self.indices.standard,
+            self.patterns.len()
+        );
```
younies: @robertbastian : https://github.com/unicode-org/icu4x/pull/8050/changes/BASE..36013c38afcdb0e14dbca1c66a9e515400bd8b5d#r3396052964

robertbastian: please see my comment on the other PR

add by @younies https://github.com/unicode-org/icu4x/pull/8050#discussion_r3396868963

robertbastian (changes requested):
no, for several reasons:

* an `Option<u8>` takes more space than a `u8`
* it shifts work that can be done at datagen time to runtime

components/experimental/src/dimension/provider/currency/essentials.rs

younies:
I have made all pattern indices `u8` except for the negative ones. At runtime, we need to know if a negative pattern does not exist so we can fall back to the decimal formatter.

Regarding the struct size, we will address that when we add all the patterns and implement the `ULE`.

components/experimental/src/dimension/provider/currency/essentials.rs
```diff
@@ -133,6 +141,14 @@ pub struct CurrencyPatternConfig {
     pub narrow_placeholder_value: Option<PlaceholderValue>,
 }
 
+// Fallback pattern: "{1}{0}" (currency followed by number, e.g. "$10")
+//
+// Even though the baked data handles the fallback at data generation time,
+// we have the fallback here for users feeding their own data without
+// handling the fallback logic in their data generation.
```
robertbastian, line 148: You should document the invariant that the standard pattern always exists, and phrase any case where it doesn't as a condition that we don't support.

```suggestion
// we have the fallback here in case the data struct is corrupted.
```

components/experimental/src/dimension/provider/currency/essentials.rs
```diff
@@ -178,10 +190,20 @@ impl<'a> CurrencyEssentials<'a> {
...
+        debug_assert!(
+            (self.indices.standard as usize) < self.patterns.len(),
+            "Standard pattern index {} is out of bounds for patterns of length {}",
+            self.indices.standard,
+            self.patterns.len()
+        );
         self.patterns
             .get(self.indices.standard as usize)
-            .or_else(|| self.patterns.get(0))
+            .unwrap_or(FALLBACK_PATTERN)
```
robertbastian:
```suggestion
        self.patterns
            .get(self.indices.standard as usize)
            .unwrap_or_else(|| {
                debug_assert!(
                    false,
                    "Standard pattern index {} is out of bounds for patterns of length {}",
                    self.indices.standard,
                    self.patterns.len()
                );
                // GIGO
                FALLBACK_PATTERN
            })
```

utils/zoneinfo64/src/lib.rs
```diff
@@ -787,6 +787,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(feature = "chrono")]
```
robertbastian: don't make unrelated changes.

components/experimental/src/dimension/provider/currency/essentials.rs
```diff
@@ -178,10 +190,20 @@ impl<'a> CurrencyEssentials<'a> {
     }
 
     /// Returns the standard pattern.
-    pub fn standard_pattern(&self) -> Option<&DoublePlaceholderPattern> {
+    ///
+    /// Even though the baked data handles the fallback at data generation time,
+    /// we have the fallback here for users feeding their own data without
+    /// handling the fallback logic in their data generation.
```
robertbastian: there is no fallback here

components/experimental/src/dimension/provider/currency/essentials.rs
```diff
@@ -191,14 +213,24 @@ impl<'a> CurrencyEssentials<'a> {
...
+    /// `standard_alpha_next_to_number` -> `standard`
+    ///
+    /// Even though the baked data handles the fallback at data generation time,
+    /// we have the fallback here for users feeding their own data without
+    /// handling the fallback logic in their data generation.
+    pub fn standard_alpha_next_to_number_pattern(&self) -> &DoublePlaceholderPattern {
         self.patterns
             .get(self.indices.standard_alpha_next_to_number as usize)
-            .or_else(|| self.standard_pattern())
+            .unwrap_or_else(|| self.standard_pattern())
```
robertbastian, line 230:
```suggestion
            .unwrap_or_else(|| {
                debug_assert!(
                    false,
                    "Standard-alpha-next-to-number pattern index {} is out of bounds for patterns of length {}",
                    self.indices.standard,
                    self.patterns.len()
                );
                // GIGO
                FALLBACK_PATTERN
            })
```

components/experimental/src/dimension/provider/currency/essentials.rs
```diff
@@ -191,14 +213,24 @@ impl<'a> CurrencyEssentials<'a> {
...
-    /// Returns the `standard_alpha_next_to_number` pattern, falling back to `standard_pattern` if not present.
-    pub fn standard_alpha_next_to_number_pattern(&self) -> Option<&DoublePlaceholderPattern> {
+    /// Returns the `standard_alpha_next_to_number` pattern.
+    ///
+    /// Fallback hierarchy:
+    /// `standard_alpha_next_to_number` -> `standard`
+    ///
+    /// Even though the baked data handles the fallback at data generation time,
+    /// we have the fallback here for users feeding their own data without
+    /// handling the fallback logic in their data generation.
```
robertbastian, line 226: don't document like this. we should treat the fallback as a datagen concern, here we just assume we have a value for each type

```suggestion
```

components/experimental/src/dimension/provider/currency/essentials.rs
```diff
@@ -191,14 +213,24 @@ impl<'a> CurrencyEssentials<'a> {
...
+    pub fn standard_alpha_next_to_number_pattern(&self) -> &DoublePlaceholderPattern {
         self.patterns
             .get(self.indices.standard_alpha_next_to_number as usize)
-            .or_else(|| self.standard_pattern())
+            .unwrap_or_else(|| self.standard_pattern())
     }
 
     /// Returns the `standard_alpha_next_to_number` negative pattern if specified, falling back to standard negative.
+    ///
+    /// Fallback hierarchy:
```
robertbastian, line 235: issue: handle the negative fallbacks at datagen time as well. they can still be `Option<u8>`

robertbastian (changes requested):

components/experimental/src/dimension/provider/currency/essentials.rs
younies: what will happen if a user fed the data without handling the fallback in their datagen?

robertbastian: if they break a documented invariant on the data struct, that's their problem

components/experimental/src/dimension/provider/currency/essentials.rs

components/experimental/src/dimension/provider/currency/essentials.rs
younies: I like the original comment

components/experimental/src/dimension/provider/currency/essentials.rs

components/experimental/src/dimension/provider/currency/essentials.rs
younies: I think it is valuable to keep these comments here to document the runtime behavior for **custom data**. While it's true that for standard data the fallback is resolved at datagen time, `CurrencyEssentials` is designed to support users feeding their own custom data. For those users, the runtime fallback hierarchy (e.g., `accounting` falling back to `standard`) is active and important to understand. Keeping this documented in the getters makes the API behavior clear and explains the design of the `unwrap_or_else` fallback code.

components/experimental/src/dimension/provider/currency/essentials.rs
younies: I think it is better to keep the fallback to `self.standard_pattern()` here. 

If `standard_alpha_next_to_number` is missing in custom data, falling back to the localized `standard_pattern` (which is still locale-sensitive) is much better and more accurate than immediately falling back to the hardcoded `FALLBACK_PATTERN` (`"{1}{0}"`). 

Furthermore, since `standard_pattern()` itself already has the GIGO safety net (returning `FALLBACK_PATTERN` and debug-asserting if it is also out of bounds), falling back to it allows us to centralize this error-handling. This keeps the code DRY and avoids duplicating the GIGO boilerplate in every single getter, while still guaranteeing safety.

components/experimental/src/dimension/provider/currency/essentials.rs
younies: We need to keep the negative fallbacks as `Option` and resolve them at runtime because of how the formatter handles negative numbers.

During formatting, we must determine if a negative currency pattern is available (either explicitly defined or via fallback). If one exists (is `Some`), we format the absolute value of the number and interpolate it into that negative currency pattern. 

If no negative currency pattern is available (is `None`), we want to fall back to letting the decimal formatter handle the negative sign using the decimal negative patterns. 

Resolving this at runtime allows the formatter to dynamically check the existence of the pattern and choose the correct formatting path (currency-specific negative pattern vs. decimal-formatter sign handling).

utils/zoneinfo64/src/lib.rs
younies: done

components/experimental/src/dimension/provider/currency/essentials.rs
robertbastian: the original comment suggests that this is something that users are allowed to do, and we need to work around it. they should not be allowed to do that and we shouldn't have to work around it

components/experimental/src/dimension/provider/currency/essentials.rs
robertbastian: ok but you should still debug-assert, otherwise you'll never notice if your data is corrupted

components/experimental/src/dimension/provider/currency/essentials.rs
robertbastian: > If no negative currency pattern is available (is `None`), we want to fall back to letting the decimal formatter handle the negative sign using the decimal negative patterns.

this is not how your code currently behaves. all your negative patters fall back to `standard_negative`, which is a fallback that can be resolved at datagen time. if you mean to have a fallback like 

```
`standard_alpha_next_to_number_negative` -> `standard_alpha_next_to_number` but with a minus
```

you need to change the code.

younies: Just to clarify, these negative patterns or accounting are not actually used in the formatter code yet. This PR is focused on adding the patterns to the datagen and setting up the provider structures.

I agree that the runtime fallback in `CurrencyEssentials` should eventually align with this design (returning `None` to let the `DecimalFormatter` handle it). I have added a `TODO` in both `CurrencyFormatter` and `CompactCurrencyFormatter` to apply the negative or accounting patterns with handling for missing negative patterns when we hook them up at runtime.

For now, since it is unused at runtime, this shouldn't block the PR, and we can refine the exact fallback behavior when we implement the formatting side of it.

younies:
Okay, it seems that in this PR we are going back and forth between two different implementation approaches: vertical and horizontal. 

I am going to close this PR and instead add the data vertically in separate, smaller PRs.

@sffc FYI.

PR #8067, opened 2026-06-11
currency: Move experimental currency formatting to legacy

younies:
## Changelog

This PR moves the existing experimental currency formatting implementation to a `legacy` subdirectory/module. This is in preparation for a new implementation based on the design at https://hackmd.io/@younies/number_formatter_4x.

Changes:
- Moved all files (except `mod.rs`) from `components/experimental/src/dimension/currency` to a new `legacy` subdirectory.
- Added a TODO comment to each moved file indicating it is legacy code to be removed.
- Refactored `name_and_pattern` out of `essentials.rs` into `legacy/mod.rs` to ensure that data provider structs do not depend on legacy code.
- Updated imports and module structure accordingly.

### DIFF
--- a/components/experimental/src/dimension/currency/legacy/compact_format.rs
+++ b/components/experimental/src/dimension/currency/legacy/compact_format.rs
@@ -2,60 +2,78 @@
 // called LICENSE at the top level of the ICU4X source tree
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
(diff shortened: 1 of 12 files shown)


robertbastian (commented):
I'm not sure this is necessary, I believe the current code can be transformed into the new design fairly easily without rewriting it from scratch.

younies:
okay

PR #8069, opened 2026-06-12
Range formatting: Fix distance calculation for single fields

Manishearth:
We discovered we weren't applying the early "distance too high" calculation, leading to e.g. EMd patterns showing up for E:

https://github.com/unicode-org/icu4x/blob/d4c6ddf23239aa03a0c9cf914cc9d9646a254243/provider/source/data/debug/datetime/patterns/range/date/gregorian/v1/e/en.json

Not sure if there's logic for non-single-field skeleta that also needs to be copied; this was the prominent fix. There might be more things to fix here

Part of #5448


provider/source/src/datetime/range_patterns.rs
```diff
@@ -244,7 +246,16 @@ fn match_range_skeleton<'a, 'data>(
...
 ) -> Option<(&'a Skeleton, &'a PatternsByGreatestDifference<'data>)> {
-    find_best_skeleton(skeletons, fields).map(|m| (m.skeleton, m.value))
+    let matched = find_best_skeleton(skeletons, fields)?;
+
+    // A single field was requested and the best pattern either includes extra fields
+    // or can't be adjusted to match (e.g. text vs numeric). We reject the match
+    // so that we fall back to the glue pattern.
+    if is_bad_match_for_single_field(fields, matched.distance) {
+        return None;
+    }
```
sffc, line 258: Nit (optional): leave a TODO to eventually implement the rest of the skeleton matching resolution algorithm.

sffc (approved):
What you're doing here is basically taking the first line of `get_best_available_format_pattern` and copying it into the range pattern call site. This is fine to fix the bug at hand, but eventually I think we will need to genericize the rest of `get_best_available_format_pattern`, at which point this workaround will be obsolete.

provider/source/src/datetime/range_patterns.rs
Manishearth: And link to https://github.com/unicode-org/icu4x/issues/8070

sffc (approved):

Manishearth:
The CLA bot is broken more and more, and it's frustrating that you can't click on the CI job to see what's wrong. I wonder if there's a force wake-up button somewhere.

sffc:
clas-assistant.io uses a github webhook, and they probably get a very high QPS, and there isn't really anyone paid to maintain it or make it a good service.

PR #8071, opened 2026-06-12
Implement data infrastructure for single displaynames

sffc:
This pull request implements the data infrastructure (markers and datagen) for the new `single` displaynames module, bringing it up to the new specification.

Specifically, this PR:
1.  **Renames/Adds Markers**: Renames the existing "Long" markers (which actually represented the default/medium width) to "Medium" to align with the spec. Adds a new `LocaleNamesLanguageLongV1` marker that correctly loads the long variant (with `-alt-long` suffix) from CLDR, and removes the unused `LocaleNamesVariantShortV1` marker.
2.  **Fixes Menu Fallback**: Updates the `LocaleNamesLanguageMenuMediumV1` datagen to support fallback to `alt-menu` (using `as_str()` to resolve type ambiguity). If the split `menu-core` and `menu-extension` keys are missing in CLDR, it now falls back to using the `alt-menu` value as the `core` and an empty `extension`.
3.  **Introduces Essentials Marker**: Defines the `LocaleNamesEssentialsV1` marker and `LocaleNamesEssentials` struct to hold locale display patterns (`localePattern` and `localeSeparator`) using `DoublePlaceholderPattern`.
4.  **Updates Datagen & Registry**: Updates all datagen readers, tests, and the central registry to support these new markers.
5.  **Regenerates Data**: Regenerates both the source test data and the baked data for the `experimental` component to include all these changes.
6.  **Aligns Existing Implementation**: Updates the existing `singular.rs` implementation (`RegionDisplayNameOwned` and `ScriptDisplayNameOwned`) to use the renamed `Medium` markers, which correctly represents the default width under the new specification.

## Changelog

N/A (data only)


sffc (commented):
🤖 This review was generated by an AI agent working with @sffc.

### CI Status
* ❌ **Failing**: `Build and Test/fmt` check failed. Please run `cargo fmt` to fix.
*  **Passing**: All other 32 checks passed.

### Review Summary
This pull request successfully implements the data infrastructure for the new `single` displaynames module, aligning it with the new specification. The changes are clean, consistent, and follow ICU4X best practices.

#### Detailed Findings:
1. **Zero-Copy Essentials Marker**: The new `LocaleNamesEssentialsV1` marker and `LocaleNamesEssentials` struct correctly use `VarZeroCow<'data, DoublePlaceholderPattern>` for both `locale_pattern` and `locale_separator`. Parsing happens during the datagen phase, keeping the runtime allocation-free.
2. **Spec-Compliant Renames**: Renaming the default markers from `Long` to `Medium` (e.g., `LocaleNamesRegionMediumV1`) correctly reflects the default width in the new specification.
3. **Robust Menu Fallback**: The updates to `LocaleNamesLanguageMenuMediumV1` datagen to support fallback to `alt-menu` when split `menu-core` and `menu-extension` keys are missing is robust and well-tested.
4. **Documentation**: The `README.md` was correctly updated.

Looks good! Once `cargo fmt` is run to fix the CI, this is ready to go.

components/experimental/src/displaynames/provider.rs
```diff
@@ -175,6 +176,22 @@ pub struct MenuNameParts<'data> {
...
+#[cfg_attr(feature = "serde", derive(serde::Deserialize))]
+#[cfg_attr(feature = "datagen", derive(serde::Serialize, databake::Bake))]
+#[cfg_attr(feature = "datagen", databake(path = icu_experimental::displaynames::provider))]
+pub struct LocaleNamesEssentials<'data> {
+    /// The pattern used to combine the base language name with qualifiers (e.g., `"{0} ({1})"`).
+    #[cfg_attr(feature = "serde", serde(borrow))]
+    pub locale_pattern: VarZeroCow<'data, DoublePlaceholderPattern>,
+    /// The separator used to join multiple qualifiers (e.g., `"{0}, {1}"`).
+    #[cfg_attr(feature = "serde", serde(borrow))]
+    pub locale_separator: VarZeroCow<'data, DoublePlaceholderPattern>,
```
sffc, line 190: I could go either way on these being two VarZeroCow or a VarZeroVec of length 2.

With a VZV, we could more easily add other patterns like `"localeKeyTypePattern": "{0}: {1}"` when we need them, but we maybe could instead put that in a separate data marker for locale name formatting.

sffc:
Human commentary: I'm happy with how the agent worked on this PR. I found I did not need to do much extra prompting. I posted one design question where I'm split on whether the AI agent made the right choice or not (I would have had the same question).

I asked the agent to split datagen into separate commits, and it did.

components/experimental/src/displaynames/provider.rs
```diff
@@ -175,10 +176,26 @@ pub struct MenuNameParts<'data> {
...
+}
+
+icu_provider::data_struct!(LocaleNamesEssentials<'_>, #[cfg(feature = "datagen")]);
```
robertbastian, line 193: shouldn't this use the MaybeAsVarULE encoding thing?

robertbastian (commented):
> ## Changelog

> icu_experimental: Implement data infrastructure for single displaynames
>  * New types: `LocaleNamesEssentialsV1`, > `LocaleNamesLanguageMediumV1`, `LocaleNamesLanguageMenuMediumV1`, > `LocaleNamesRegionMediumV1`, `LocaleNamesScriptMediumV1`, > `LocaleNamesVariantMediumV1`
>  * Removed types: `LocaleNamesLanguageMenuLongV1`, `LocaleNamesRegionLongV1`, `LocaleNamesScriptLongV1`, `LocaleNamesVariantLongV1`, `LocaleNamesVariantShortV1`
>  * Modified types: `LocaleNamesLanguageLongV1` (now loads `-alt-long` variant instead of default)

I don't think we want to list provider changes in the changelog?

components/experimental/src/displaynames/provider.rs
sffc: hmm, probably not necessary for `LocaleNamesEssentials` since it doesn't use attributes and there are only a few versions of the payload. But maybe we should use it more even for payloads that don't _need_ it?

components/experimental/src/displaynames/provider.rs
Manishearth: probably doesn't matter. if expanding seems likely then we should use a vzv.

components/experimental/src/displaynames/provider.rs
```diff
@@ -175,10 +176,26 @@ pub struct MenuNameParts<'data> {
 
 icu_provider::data_struct!(VariantDisplayNames<'_>, #[cfg(feature = "datagen")]);
 
+/// [`LocaleNamesEssentials`] provides the formatting patterns used to combine subtags.
```
Manishearth, line 179: suggestion (nb): LocaleNamesGlueEssentials?

Manishearth (approved):
architecture seems normal and good

components/experimental/src/displaynames/provider.rs
sffc: 🤖 This comment is written by an AI agent working with @sffc.

We chose `LocaleNamesEssentialsV1` to be consistent with several other "Essentials" markers in the codebase, such as:
*   `CurrencyEssentialsV1` (in `icu_experimental::dimension::provider::currency`)
*   `PercentEssentialsV1` (in `icu_experimental::dimension::provider::percent`)
*   `UnitsEssentialsV1` (in `icu_experimental::dimension::provider::units`)
*   `tz::EssentialsV1` (in `icu_datetime`)

In all these cases, "Essentials" is used to group the core, essential formatting data (like patterns or symbols) for a specific domain that are required for the basic functioning of the formatter. In our case, it contains `localePattern` and `localeSeparator`, which are the essential patterns needed to combine subtags.

Adding "Glue" might make it inconsistent with these existing markers, as we don't have `CurrencyGlueEssentials` or `PercentGlueEssentials` even though they serve a similar purpose of "gluing" values together.

components/experimental/src/displaynames/provider.rs
robertbastian: yeah I guess it's a speed/size tradeoff

PR #8072, opened 2026-06-12
Return concrete types from LiteMap iter methods

### LINKED ISSUE #8054: litemap: Consider returning concrete types on `LiteMap::{iter,iter_mut,keys,values}`
## What I wanted

I needed to `.clone()` the return value of `LiteMap::iter` (which is much cheaper than cloning the whole map).

## What happened

```rust
let my_map = LiteMap::new();
let my_iter = my_map.iter();
let iter_clone = my_iter.clone();
// ^^^ Compilation error!
```

## Current workaround

It's still possible to get an iterator that is clonable as follow:

```rust
let my_map = LiteMap::new();
let my_iter = (&my_map).into_iter();
let iter_clone = my_iter.clone();
// ^_^ works fine!
```

## Possible solution

Instead of returning `impl DoubleEndedIterator<Item = _>`, methods like `LiteMap::iter` could return the same type as `<&'_ LiteMap as IntoIterator>::IntoIter` (that would be `<S as StoreIterable<'a, K, V>>::KeyValueIter`)

## More context

I was using the `serde_iter` crate to serialize through a view type my base struct. But for it to work, I needed my iterator to be `Clone`. I expected the return type of  `.iter()` to be `Clone`, since that's how the std lib works.

I was specifically using `serde_iter` to avoid having to create a new owned struct and cloning all my values (the actual use-case `Bar` is much larger).

The view pattern was as follow:

<details>

<summary>Code listing</summary>

```rust
struct Bar { some: i32, field: i32 }
struct Foo { bars: Vec<BarId> }
struct Foos {
  name: String,
  foos: LiteMap<FooId, Foo>,
  bars: LiteMap<BarId, Bar>,
}

#[derive(Serialize)]
struct ViewFoo<'a, I: IntoIterator<Item = (&'a BarId, &'a Bar)> + Clone> {
  #[serde(with = "serde_iter::map")]
  bars: I,
}

#[derive(Serialize)]
struct ViewFoos<'a, V: Serialize, I: IntoIterator<Item = (&'a FooId, V)> + Clone> {
  name: &'a str
  #[serde(with = "serde_iter::map")]
  foos: I,
}

impl Serialize for Foos {
    // I want to serialize to: '{ "name": "the_name", foos: {
    //   "foo1": { "bars": {
    //     "bar1": { "some": 1, "field": 32 },
    //     "bar2": { "some": 2, "field": 42 }
    //   }},
    //   "foo2": { "bars": {
    //     "bar3": { "some": 3, "field": 52 },
    //     "bar4": { "some": 4, "field": 62 }
    //   }}
    // }}'
    fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
      let ser_foos = ViewFoos {
        name: &self.name,
        foos: self.foos.iter().map(|(id, foo)| {
          (id, ViewFoo { bars: foo.bars.iter().map(|id| (id, &self.bars[id])) })
        })
      };
      ser_foos.serialize(serializer)
    }
}
```

</details>

Since the types get quite long (because we have those generic structs that themselves contain the iterator types, which may become quite long) the error message wasn't very helpful in pointing out from where the error came. (in this case, it is `self.foos.iter()` and `foo.bars.iter()`) and it took me a bit of time to find a fix.

nicopap:
Return concrete types from LiteMap iter methods

impl Trait is limited to a single trait. But we'd like to expose
iterators that implement more than DoubleEndedIterator (eg: ExactSize,
Fused). This is why we return a concrete type now.

This closes #8054.

## Considerations

* No Copy derive, because that's against API guidelines for Iterators
* Manual impl of `size_hint`, to be able to use ExactSizeIterator
  without panicking.
* Manual impl of `nth` and `fold`, as other iterator methods tend to be
  implemented in term of those, and forwarding them to the inner
  iterator **might** avoid performance penalty of the default impl.

## CLA status

* Signed

## Changelog

litemap: Make return types of `LiteMap` iter methods (`iter`, `iter_mut`, `values`, `keys`) concrete:
  * New types: `ValuesIter`, `KeysIter`, implements relevant traits, returned by `values` and `keys` respectively.
  * Changed: The return types of the aforementioned methods go from an `impl DoubleEndedIterator` to concrete types such as `S::KeyValueIter`, `S::KeyValueIterMut` or the aforementioned new types.

### DIFF
--- a/utils/litemap/src/map.rs
+++ b/utils/litemap/src/map.rs
@@ -10,7 +10,7 @@ use alloc::vec::Vec;
 use core::borrow::Borrow;
 use core::cmp::Ordering;
 use core::fmt::Debug;
-use core::iter::FromIterator;
+use core::iter::{FromIterator, FusedIterator};
 use core::marker::PhantomData;
 use core::mem;
 use core::ops::{Index, IndexMut, Range};
@@ -901,7 +901,7 @@ where
     S: StoreIterable<'a, K, V>,
 {
     /// Produce an ordered iterator over key-value pairs
-    pub fn iter(&'a self) -> impl DoubleEndedIterator<Item = (&'a K, &'a V)> {
+    pub fn iter(&'a self) -> S::KeyValueIter {
         self.values.lm_iter()
     }
 
@@ -918,13 +918,17 @@ where
     }
 
     /// Produce an ordered iterator over keys
-    pub fn keys(&'a self) -> impl DoubleEndedIterator<Item = &'a K> {
-        self.values.lm_iter().map(|val| val.0)
+    pub fn keys(&'a self) -> KeysIterWrap<S::KeyValueIter> {
+        KeysIterWrap {
+            iter: self.values.lm_iter(),
+        }
     }
 
     /// Produce an iterator over values, ordered by their keys
-    pub fn values(&'a self) -> impl DoubleEndedIterator<Item = &'a V> {
-        self.values.lm_iter().map(|val| val.1)
+    pub fn values(&'a self) -> ValuesIterWrap<S::KeyValueIter> {
+        ValuesIterWrap {
+            iter: self.values.lm_iter(),
+        }
     }
 }
 
@@ -933,7 +937,7 @@ where
     S: StoreIterableMut<'a, K, V>,
 {
     /// Produce an ordered mutable iterator over key-value pairs
-    pub fn iter_mut(&'a mut self) -> impl DoubleEndedIterator<Item = (&'a K, &'a mut V)> {
+    pub fn iter_mut(&'a mut self) -> S::KeyValueIterMut {
         self.values.lm_iter_mut()
     }
 }
@@ -1216,6 +1220,88 @@ impl_const_get_with_index_for_integer!(i32);
 impl_const_get_with_index_for_integer!(i64);
 impl_const_get_with_index_for_integer!(i128);
 impl_const_get_with_index_for_integer!(isize);
+
+/// An [`Iterator`] adapter over the keys of a [`StoreIterable::KeyValueIter`].
+///
+/// This `struct` is created by the [`LiteMap::keys`] method See its documentation for more.
+#[derive(Debug, Clone)]
+pub struct KeysIterWrap<I> {
+    iter: I,
+}
+
+/// An [`Iterator`] adapter over the values of a [`StoreIterable::KeyValueIter`].
+///
+/// This `struct` is created by the [`LiteMap::values`] method. See its documentation for more.
+#[derive(Debug, Clone)]
+pub struct ValuesIterWrap<I> {
+    iter: I,
+}
+
+impl<K, V, I: Iterator<Item = (K, V)>> Iterator for KeysIterWrap<I> {
+    type Item = K;
+
+    #[inline]
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        self.iter.size_hint()
+    }
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
+        self.iter.next().map(|val| val.0)
+    }
+    #[inline]
+    fn fold<B, F>(self, init: B, f: F) -> B
+    where
+        Self: Sized,
+        F: FnMut(B, Self::Item) -> B,
+    {
+        self.iter.map(|val| val.0).fold(init, f)
+    }
+    #[inline]
+    fn nth(&mut self, n: usize) -> Option<Self::Item> {
+        self.iter.nth(n).map(|val| val.0)
+    }
+}
+impl<K, V, I: ExactSizeIterator<Item = (K, V)>> ExactSizeIterator for KeysIterWrap<I> {}
+impl<K, V, I: FusedIterator<Item = (K, V)>> FusedIterator for KeysIterWrap<I> {}
+impl<K, V, I: DoubleEndedIterator<Item = (K, V)>> DoubleEndedIterator for KeysIterWrap<I> {
+    #[inline]
+    fn next_back(&mut self) -> Option<Self::Item> {
+        self.iter.next_back().map(|val| val.0)
+    }
+}
+
+impl<K, V, I: Iterator<Item = (K, V)>> Iterator for ValuesIterWrap<I> {
+    type Item = V;
+
+    #[inline]
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        self.iter.size_hint()
+    }
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
+        self.iter.next().map(|val| val.1)
+    }
+    #[inline]
+    fn fold<B, F>(self, init: B, f: F) -> B
+    where
+        Self: Sized,
+        F: FnMut(B, Self::Item) -> B,
+    {
+        self.iter.map(|val| val.1).fold(init, f)
+    }
+    #[inline]
+    fn nth(&mut self, n: usize) -> Option<Self::Item> {
+        self.iter.nth(n).map(|val| val.1)
+    }
+}
+impl<K, V, I: ExactSizeIterator<Item = (K, V)>> ExactSizeIterator for ValuesIterWrap<I> {}
+impl<K, V, I: FusedIterator<Item = (K, V)>> FusedIterator for ValuesIterWrap<I> {}
+impl<K, V, I: DoubleEndedIterator<Item = (K, V)>> DoubleEndedIterator for ValuesIterWrap<I> {
+    #[inline]
+    fn next_back(&mut self) -> Option<Self::Item> {
+        self.iter.next_back().map(|val| val.1)
+    }
+}
 
 /// An entry in a `LiteMap`, which may be either occupied or vacant.
 #[allow(clippy::exhaustive_enums)]


Manishearth (approved):
This is a good idea. Generally we should do this more to make it easier to  compose our iterators.

Waiting on CLA.

nicopap:
I'll keep up to date with the CLA application progression. I just submitted it to management.

utils/litemap/src/map.rs
```diff
@@ -1217,6 +1221,106 @@ impl_const_get_with_index_for_integer!(i64);
...
+}
+
+pub struct ValuesIter<'a, K, V, S: StoreIterable<'a, K, V>> {
+    iter: MapKv<'a, S::KeyValueIter, K, V, V>,
+}
+
+impl<'a, K, V, S: StoreIterable<'a, K, V>> Iterator for KeysIter<'a, K, V, S> {
+    type Item = &'a K;
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
```
sffc, line 1244: Nit: make the functions that delegate be `#[inline]`

utils/litemap/src/map.rs
```diff
@@ -901,7 +901,7 @@ where
     S: StoreIterable<'a, K, V>,
 {
     /// Produce an ordered iterator over key-value pairs
-    pub fn iter(&'a self) -> impl DoubleEndedIterator<Item = (&'a K, &'a V)> {
+    pub fn iter(&'a self) -> S::KeyValueIter {
```
sffc, line 904: Thought: This is fine since `S::KeyValueIter` is already public and already used in a bunch of bounds.

An alternative would be to add a third iterator wrapper for this.

sffc (approved):
This looks about right. I don't like how Rust doesn't have a good way to automatically delegate the `Iterator` trait. You picked 4 fns here, which seem like good ones, but you could have picked more or fewer; there isn't a great way to pick which ones to leave as default.

nicopap:
I have the CLA documents signed, I need to scan them and send them to the unicode org email still.

I will implement the suggested changes to the PR, and rebase.

utils/litemap/src/map.rs
```diff
@@ -1217,6 +1221,106 @@ impl_const_get_with_index_for_integer!(i64);
 impl_const_get_with_index_for_integer!(i128);
 impl_const_get_with_index_for_integer!(isize);
 
+type MapKv<'a, I, K, V, Out> = core::iter::Map<I, fn((&'a K, &'a V)) -> &'a Out>;
```
nicopap: This is the biggest worry here for me, I'm not sure how the compiler handles the function pointer. In practice, it will always be the closure used in `LiteMap::values` and `LiteMap::keys` passed to `iter().map(_)`, so I don't know if the compiler is smart enough to directly inline the closure.

The alternative is described in the PR description, which might make the exposed type quite difficult to read.

nicopap:
* I added the inline attributes,
* I submitted the signed corporate CLA to the email mentioned in www.unicode.org.
* I rebased to follow main.

utils/litemap/src/map.rs
sffc: It's probably fine. But, since you're exporting a new type anyway, a cleaner solution would be to remove the `core::iter::Map` and just do the mapping in the fns you are implementing. The map fn is only a field access.

```rust
pub struct KeysIter<'a, K, V, S: StoreIterable<'a, K, V>> {
    iter: S::KeyValueIter
}

impl<'a, K, V, S: StoreIterable<'a, K, V>> Iterator for KeysIter<'a, K, V, S> {
    type Item = &'a K;

    #[inline]
    fn size_hint(&self) -> (usize, Option<usize>) {
        // don't need to map here
        self.iter.size_hint()
    }
    #[inline]
    fn next(&mut self) -> Option<Self::Item> {
        // access the .0 of the value using Option::map
        self.iter.next().map(|val| &val.0)
    }
    #[inline]
    fn fold<B, F>(self, init: B, f: F) -> B
    where
        Self: Sized,
        F: FnMut(B, Self::Item) -> B,
    {
        // here we can apply a map fn
        self.iter.map(|val| val.0).fold(init, f)
    }
    #[inline]
    fn nth(&mut self, n: usize) -> Option<Self::Item> {
        self.iter.nth(n).map(|val| &val.0)
    }
}
```

nicopap: Good point, with this, it's also possible to make `KeysIter` and `ValuesIter` generic over the iterator itself rather than `S: StoreIterable` which allows to use `derive` for the `Debug` and `Clone` impls.

nicopap:
Applying the suggestion from @sffc wrt to mapping inside the methods rather than before creating the iterator (I should have thought about that) made the code much cleaner, since `KeysIter` and `ValuesIter` are now simply iterator adaptor to get a key/value from an iterator over tuples. We can even use `derive` for `Clone` and `Debug`.

Looking at the code for implementing `StoreIterable` in `vec_impl` and `slice_impl`, I feel like it might be possible to apply the same trick there to avoid the function pointer (although it's def not worth changing the code, as I don't know it's meaningful to avoid it) 

CLA ETA: Email sent yesterday, waiting for the reply.

nicopap:
My employer appears to now be on the list of Unicode-approved contributors (heliotis AG) https://api.unicode.org/cla/corporate

sffc:
> My employer appears to now be on the list of Unicode-approved contributors (heliotis AG) https://api.unicode.org/cla/corporate

Great! Please complete the cla-assistant.io, and enter your organization on the form to record that you are covered under a corporate CLA.

nicopap:
I think everything should be ready at this point.

utils/litemap/src/map.rs
```diff
@@ -1217,6 +1221,84 @@ impl_const_get_with_index_for_integer!(i64);
 impl_const_get_with_index_for_integer!(i128);
 impl_const_get_with_index_for_integer!(isize);
 
+/// An [`Iterator`] adapter over the 0th element of a 2-tuple iterator.
+#[derive(Debug, Clone)]
+pub struct KeysIter<I> {
```
Manishearth: Actually, can this be `LiteMapKeysIter` and have it take `S` as a parameter?

Then it can just be "iterator over the keys of a litemap". The fact that it encapsulates an "internal" iterator should be an implementation detail.

Same for the other one.

Manishearth (commented):
Looking mostly good, but now that I've had a second look I realized this can be better.

utils/litemap/src/map.rs
nicopap: This would require to specify the `S: StoreIterable` bounds at the `struct` level so that we internally can store `S::KeyValueIter`, and add the `'a, K, V` type parameters, ending up with

```rust
#[derive(Debug, Clone)]
pub struct LiteMapKeysIter<'a, K, V, S: StoreIterable<'a, K, V>> {
    iter: S::KeyValueIter,
}
```

The iterator needs to keep an internal state distinct from `S` that mutates through the `next` iterator method. Currently, that internal state is `S::KeyValueIter`.

Then I'd have to manually impl the `Debug` and `Clone` (because the `#[derive]` ones add bounds on type parameters that are not needed), and the trait bounds on the blanket impls for `ExactSizeIterator` and `FusedIterator` start to look as follow:

```rust
impl<'a, K, V, I, S> ExactSizeIterator for LiteMapKeysIter<'a, K, V, S>
where
    I: ExactSizeIterator<Item = (&'a K, &'a V)>,
    S: StoreIterable<'a, K, V, KeyValueIter = I>,
{
}
```

While currently they look like this:

```rust
impl<K, V, I: ExactSizeIterator<Item = (K, V)>> ExactSizeIterator for KeysIter<I> {}
```

I think keeping the way it is currently reduces considerably complexity, it's easier to deduce which traits are implemented by just looking at the type signature, and we have only 1 type parameter instead of 3 + lifetime parameter.

Plus, as mentioned by @sffc, `S::KeyValueIter` is already exposed publicly in the `StoreIterable` trait, and returned by the `LiteMap::iter` function, so there is not much benefit in hiding in the first place.

----

**Edit:** Also, checking for how possible the change is, I looked at the first iteration of this PR, and I found a bug in it, that was obfuscated and difficult to notice precisely because of the complexity of the bounds on the blanket impls. (the `S::KeyValueIter` wasn't bound to `ExactSizeIterator` in the `ExactSizeIterator impl for KeysIter`) (see [this old revision](https://github.com/unicode-org/icu4x/commit/0a0097778128ca66649a91deb19e577d0a70812c#diff-4f939354a38d472c28346086f005d9f3dbb881573a7d511e578bfff50c89672aR1254))

Manishearth: Hmmm, makes sense. Can we call it KeysIterWrap and ValuesIterWrap then?

Manishearth: And expand the documentation to say "iterator over keys over litemap" and explain what the `I` parameter is.

nicopap: sounds good, I'll do this tomorrow.

nicopap: I took inspiration from the `std::iter` structs documentation. I think the documentation makes more sense now.

Manishearth (approved):

nicopap:
Thank you for the time and energy you put in this review, and holding my hand for the CLA signing.

PR #8073, opened 2026-06-12
`UnitsConverter`: accept inputs by value, return correct error type

robertbastian:
~Currently we have `UnitsConverter<N: Convertible>`. This is because we convert the conversion factors to the `N` type during construction, which both loses precision and means we cannot use the same converter for multiple `N`.~

~This PR moves the conversion into the `convert` call, and into the `Convertible` trait, which now defines the three operations we need for proportional, reciprocal, and offset converters. For each type, such as `f64`, those operations can be holistically improved.~

It also changes the `UnitsConverter` constructor to return `InvalidConversionError` instead of just an `Option`, and changes the bounds on `convert` so that `f64` can be passed by value.

## Changelog

`icu_experimental/units`
* `Convertibles`s are now passed by value, and for `Ratio<BigInt>` the `Convertible` impl is now on the reference
* The `Convertibles` trait was completely overhauled to allow for more accurate calculations

### DIFF
--- a/components/experimental/src/units/converter_factory.rs
+++ b/components/experimental/src/units/converter_factory.rs
@@ -191,13 +191,14 @@ impl ConverterFactory {
                     .get()
                     .conversion_info_by_unit_id(single_unit.unit_id);
 
-                debug_assert!(conversion_info.is_some(), "Failed to get convert info");
-
                 match conversion_info {
                     Some(items) => {
                         insert_base_units(items.basic_units(), single_unit.power as i16, sign, map)
                     }
-                    None => return Err(InvalidConversionError),
+                    None => {
+                        debug_assert!(false, "Failed to get convert info");
+                        return Err(InvalidConversionError);
+                    }
                 }
             }
 
@@ -284,18 +285,20 @@ impl ConverterFactory {
         &self,
         input_unit: &MeasureUnit,
         output_unit: &MeasureUnit,
-    ) -> Option<UnitsConverter<T>> {
-        let is_reciprocal = match self.is_reciprocal(input_unit, output_unit) {
-            Ok(is_reciprocal) => is_reciprocal,
-            Err(InvalidConversionError) => return None,
-        };
+    ) -> Result<UnitsConverter<T>, InvalidConversionError> {
+        let is_reciprocal = self.is_reciprocal(input_unit, output_unit)?;
+
+        // is_reciprocal checked that all simple units can be loaded, so we can ignore load errors after this.
 
         // Determine the sign of the powers of the units from root to unit2.
         let root_to_unit2_direction_sign = if is_reciprocal { 1 } else { -1 };
 
         let mut conversion_rate = IcuRatio::one();
         for &input_single_unit in input_unit.single_units.as_slice() {
-            conversion_rate *= Self::compute_conversion_term(self, input_single_unit, 1)?;
+            let Some(t) = Self::compute_conversion_term(self, input_single_unit, 1) else {
+                continue;
+            };
+            conversion_rate *= t;
         }
 
         if input_unit.constant_denominator() != 0 {
@@ -303,11 +306,14 @@ impl ConverterFactory {
         }
 
         for &output_single_unit in output_unit.single_units.as_slice() {
-            conversion_rate *= Self::compute_conversion_term(
+            let Some(t) = Self::compute_conversion_term(
                 self,
                 output_single_unit,
                 root_to_unit2_direction_sign,
-            )?;
+            ) else {
+                continue;
+            };
+            conversion_rate *= t;
         }
 
         if output_unit.constant_denominator() != 0 {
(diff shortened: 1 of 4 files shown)


Manishearth (commented):
without going through the actual code in depth this seems like a good idea and the right way to do it

robertbastian:
are you going to go through the code?

components/experimental/src/units/convertible.rs
```diff
@@ -6,54 +6,57 @@ use num_bigint::BigInt;
...
-        self * other
-    }
+    type Result = Self;
 
-    fn add_refs(&self, other: &Self) -> Self {
-        self + other
+    // TODO: reduce error
+    #[allow(clippy::unwrap_used)] // TODO
+    fn mul_ratio_bigint(&self, factor: &Ratio<BigInt>) -> Self::Result {
+        self * factor.to_f64().unwrap()
```
robertbastian: @sffc your optimisation would go here, but we should also try to optimise the other operations, which have one additional floating point operation that introduces error

```suggestion
        let Some(num) = factor.numer().to_f64() else { return f64::NAN };
        let Some(den) = factor.denom().to_f64() else { return f64::NAN };

        let double_rounded = self * num / den;

        // The multiplication error is the difference of the rounded multiplication
        // and the multiplication evaluated in full precision (FMA)
        let multiplication_error = self.mul_add(num, -(self * num));

        // The total error is the difference between the rounded multiplication+division
        // and the multiplication evaluated in full precision, plus the multiplication
        // error divided by the denominator, all evaluated in times-den-space.
        let total_error = (double_rounded.mul_add(-den, self * num) + multiplication_error) / den;

        double_rounded + total_error
```

Manishearth:
> are you going to go through the code?

I could: my review was not explicitly requested and Younies and Shane have been primarily reviewing this component. I'm happy to do a thorough review if you'd like, but I'm usually not doing units reviews.

sffc (dismissed):
The issue I have with this approach is I think it ties us in even more to the BigInt data model. I approved the BigInt landing initially as experimental code, but I think we should be able to support unit conversions completely with floats, storing a float ratio (and deserializing to a float ratio!) instead of a BigInt ratio. So I think this PR is in the wrong direction.

robertbastian:
> I approved the BigInt landing initially as experimental code

This is the same BigInt code that we had before, and it's still experimental, so please approve anyway.

> storing a float ratio (and deserializing to a float ratio!) instead of a BigInt ratio

sure this gets us further away from completely rewriting this, and losing accuracy on the way. not sure if that's a goal

sffc (commented):
I approve PRs that go in the right direction, but this goes in the wrong direction.

sffc:
At some point along the way, we need to convert from the serialized integer ratio type to the runtime type.

Currently, this happens when going from ConverterFactory -> UnitsConverter.

This PR makes it happen when calling the actual convert function.

If you are doing a lot of conversions in a loop, it seems like we should do the number type conversion earlier rather than later.

components/experimental/src/units/converter_factory.rs
```diff
@@ -286,30 +287,35 @@ impl ConverterFactory {
...
+        // is_reciprocal checked that all simple units can be loaded, so we can ignore load errors after this.
 
         // Determine the sign of the powers of the units from root to unit2.
         let root_to_unit2_direction_sign = if is_reciprocal { 1 } else { -1 };
 
         let mut conversion_rate = IcuRatio::one();
         for &input_single_unit in input_unit.single_units.as_slice() {
-            conversion_rate *= Self::compute_conversion_term(self, input_single_unit, 1)?;
+            let Some(t) = Self::compute_conversion_term(self, input_single_unit, 1) else {
+                continue;
```
sffc, line 299: Issue: I think we shouldn't just ignore conversion terms, right? We should fail, right? If there is a reason to skip them silently, please add a comment explaining why

sffc (commented):
I gave my feedback on commit 1 already. Commit 2 seems fine but I have a comment on it. Commit 3 is an internal refactor on which I have no opinion.

components/experimental/src/units/converter_factory.rs
robertbastian: `// is_reciprocal checked that all simple units can be loaded, so we can ignore load errors after this.`

younies (commented):
Let's discuss this PR in the ICU4X meeting. Making the conversion rate `Convertible` is a valid approach for a couple of reasons:

* First, it avoids extra processing overhead at runtime, since it only happens once during initialization.
* Second, while there might be a tiny bit of extra precision loss, this is expected behavior. Ultimately, the user is already requesting a type where a small amount of loss is acceptable.

However, this implementation also allows different input types to use the same converter, and I'm not entirely sure what the real-world use case is for this. Do we have specific scenarios where users actually need this kind of flexibility?

robertbastian:
reorganized commits^

components/experimental/src/units/convertible.rs
```diff
@@ -6,54 +6,84 @@ use num_bigint::BigInt;
...
+    // Exact
+    fn reciprocal_mul_ratio_bigint(self, factor: &Self::Factor) -> Self::Result {
+        (self * factor).recip()
     }
 
-    fn from_ratio_bigint(ratio: Ratio<BigInt>) -> Self {
-        ratio
+    fn factor_from_ratio_bigint(factor: Ratio<BigInt>) -> Self::Factor {
+        factor
     }
```
sffc, line 53: Thought: I think we probably will want to change this to read from a downgraded type (a "PotentialRatioBigInt", maybe encoded as a `&str`). This PR puts us in the correct direction.

sffc (commented):
Commits 1-3 are lgtm.

components/experimental/src/units/convertible.rs
robertbastian: if we already want to lose precision when determining the conversion factor, sure.

we need to agree on floating point semantics before making any such changes.

Manishearth (approved):

Manishearth (approved):

components/experimental/src/units/convertible.rs
sffc: I meant, not changing this to a BigInt eagerly, doing it lazily instead. Even assuming we want an exact factor, we can cancel out factors and do other things without ever reifying a BigInt.

robertbastian: storing a list of integers ratios will probably allocate more space than storing a single `Ratio<BigInt>`. cancelling out factors is what multiplication on `Ratio<BigInt>` does, `Ratio<BigInt>` *is* the compact representation of a product of integer ratios.

PR #7891, opened 2026-04-21
Move collation invariants to data structs

robertbastian:
## Changelog

N/A

PR #8075, opened 2026-06-12
release/2.1: Fix off-by-one errors in CollationSpecialPrimariesValidated

sffc:
Depends on #8078

On release/2.2: #8081
On main: #8080

***

This PR fixes a critical off-by-one bug in `icu_collator` 2.1.x that causes a panic when using `AlternateHandling::Shifted` with `MaxVariable::Currency`. 

This bug was found by the Unicode conformance testing project (credit to @sven-oly).

### Cause
When converting `CollationSpecialPrimaries` to `CollationSpecialPrimariesValidated`, the `last_primaries` vector was incorrectly truncated to `MaxVariable::Currency as usize` (3) instead of `MaxVariable::Currency as usize + 1` (4). This resulted in `last_primaries` lacking the 4th element (index 3, for `Currency`), causing a panic (`unwrap` on `None`) during comparison when `last_primary_for_group` was called.
Additionally, `compressible_bytes` extraction was misaligned by one index and failed the length check, causing it to always fall back to hardcoded defaults.

### Fix
Corrected the off-by-one errors in both `Collator::try_new_unstable` and `CollatorBorrowed::try_new` in `comparison.rs` by adding 1 to `MaxVariable::Currency as usize` where appropriate.

~Also bumped `icu_collator` to `2.1.2`.~ --> this will be done in another PR

### Context on main branch
On the main branch, this was refactored. The intermediate `CollationSpecialPrimariesValidated` struct was removed, and `CollationSpecialPrimaries` is used directly. The bug is naturally avoided there because the deserialization logic correctly uses `MaxVariable::Currency as usize + 1` to split the data:
See: [provider.rs:L635 on main](https://github.com/unicode-org/icu4x/blob/d4c6ddf23239aa03a0c9cf914cc9d9646a254243/components/collator/src/provider.rs#L635)

### icu_collator (2.1.2)
- Fix panic when using `AlternateHandling::Shifted` with `MaxVariable::Currency` (off-by-one in special primaries validation).

### DIFF
--- a/components/collator/src/comparison.rs
+++ b/components/collator/src/comparison.rs
@@ -39,6 +39,7 @@ use crate::provider::CollationMetadataV1;
 use crate::provider::CollationReordering;
 use crate::provider::CollationReorderingV1;
 use crate::provider::CollationRootV1;
+use crate::provider::CollationSpecialPrimaries;
 use crate::provider::CollationSpecialPrimariesV1;
 use crate::provider::CollationSpecialPrimariesValidated;
 use crate::provider::CollationTailoringV1;
@@ -663,12 +664,12 @@ impl Collator {
         }
         let special_primaries = special_primaries.map_project(|csp, _| {
             let compressible_bytes = (csp.last_primaries.len()
-                == MaxVariable::Currency as usize + 16)
+                == CollationSpecialPrimaries::TOTAL_LEN_WITH_COMPRESSIBLE_BYTES)
                 .then(|| {
                     csp.last_primaries
                         .as_maybe_borrowed()?
                         .as_ule_slice()
-                        .get((MaxVariable::Currency as usize)..)?
+                        .get(MaxVariable::VARIANT_COUNT..)?
                         .try_into()
                         .ok()
                 })
@@ -678,7 +679,7 @@ impl Collator {
                 );
 
             CollationSpecialPrimariesValidated {
-                last_primaries: csp.last_primaries.truncated(MaxVariable::Currency as usize),
+                last_primaries: csp.last_primaries.truncated(MaxVariable::VARIANT_COUNT),
                 numeric_primary: csp.numeric_primary,
                 compressible_bytes,
             }
@@ -786,7 +787,7 @@ impl CollatorBorrowed<'static> {
                         .last_primaries
                         .as_slice()
                         .as_ule_slice()
-                        .split_at(MaxVariable::Currency as usize)
+                        .split_at(MaxVariable::VARIANT_COUNT)
                         .0,
                 )
                 .as_zerovec(),
@@ -798,8 +799,8 @@ impl CollatorBorrowed<'static> {
                             .last_primaries
                             .as_slice()
                             .as_ule_slice();
-                    if C.len() == MaxVariable::Currency as usize + 16 {
-                        let i = MaxVariable::Currency as usize;
+                    if C.len() == CollationSpecialPrimaries::TOTAL_LEN_WITH_COMPRESSIBLE_BYTES {
+                        let i = MaxVariable::VARIANT_COUNT;
                         #[allow(clippy::indexing_slicing)] // protected, const
                         &[
                             C[i],
--- a/components/collator/tests/tests.rs
+++ b/components/collator/tests/tests.rs
@@ -2128,3 +2128,20 @@ fn test_latin1_utf16_fi() {
 // TODO: Test that nn and nb are aliases for no
 
 // TODO: Consider testing ff-Adlm for supplementary-plane tailoring, including contractions
+
+#[test]
+// Regression test for a bug where setting `AlternateHandling::Shifted` and
+// `MaxVariable::Currency` caused a panic during collation.
+// The bug was due to an off-by-one error when validating special primaries,
+// which truncated the `last_primaries` vector and removed the `Currency` element.
+// This test verifies that the collator can be successfully constructed with these
+// options and that it correctly compares an empty string and a space as equal
+// (since space is shifted and ignored at the default Tertiary strength).
+fn test_shifted_max_variable_currency() {
+    let prefs = CollatorPreferences::default();
+    let mut options = CollatorOptions::default();
+    options.alternate_handling = Some(AlternateHandling::Shifted);
+    options.max_variable = Some(MaxVariable::Currency);
+    let collator = Collator::try_new(prefs, options).unwrap();
+    assert_eq!(collator.compare("", " "), Ordering::Equal);
+}
--- a/components/collator/src/options.rs
+++ b/components/collator/src/options.rs
@@ -208,6 +208,15 @@ pub enum MaxVariable {
     Currency = 3,
 }
 
+impl MaxVariable {
+    /// The number of variants in `MaxVariable` (Space, Punctuation, Symbol, Currency),
+    /// which correspond to the "real" special primaries.
+    ///
+    /// `variant_count` isn't stable yet:
+    /// <https://github.com/rust-lang/rust/issues/73662>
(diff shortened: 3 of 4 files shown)


components/collator/tests/tests.rs
```diff
@@ -2128,3 +2128,20 @@ fn test_latin1_utf16_fi() {
...
 // TODO: Consider testing ff-Adlm for supplementary-plane tailoring, including contractions
+
+#[test]
+fn test_repro_user_bug() {
+    let provider = TestingProvider;
+    let payload: DataPayload<CollationSpecialPrimariesV1> =
+        provider.load(Default::default()).unwrap().payload;
+    let csp = payload.get();
+    println!("csp.last_primaries.len() = {}", csp.last_primaries.len());
+    println!("csp.last_primaries = {:?}", csp.last_primaries.to_vec());
```
sffc: Instead of println, please assert the lengths of the things

components/collator/src/comparison.rs
```diff
@@ -663,12 +663,12 @@ impl Collator {
         }
         let special_primaries = special_primaries.map_project(|csp, _| {
             let compressible_bytes = (csp.last_primaries.len()
-                == MaxVariable::Currency as usize + 16)
+                == MaxVariable::Currency as usize + 17)
```
Manishearth: While we're here we should add a comment fixing the magic numbers

components/collator/src/comparison.rs
```diff
@@ -663,12 +663,12 @@ impl Collator {
...
                 .then(|| {
                     csp.last_primaries
                         .as_maybe_borrowed()?
                         .as_ule_slice()
-                        .get((MaxVariable::Currency as usize)..)?
+                        .get((MaxVariable::Currency as usize + 1)..)?
```
Manishearth: and just explaining the +1 here

components/collator/src/comparison.rs
```diff
@@ -786,7 +788,7 @@ impl CollatorBorrowed<'static> {
                         .last_primaries
                         .as_slice()
                         .as_ule_slice()
-                        .split_at(MaxVariable::Currency as usize)
+                        .split_at(MaxVariable::Currency as usize + 1)
```
Manishearth: actually, should these be static methods on MaxVariable? after_currency and something different for the 17 one?

components/collator/tests/tests.rs
```diff
@@ -2128,3 +2128,19 @@ fn test_latin1_utf16_fi() {
 // TODO: Test that nn and nb are aliases for no
 
 // TODO: Consider testing ff-Adlm for supplementary-plane tailoring, including contractions
+
+#[test]
+fn test_repro_user_bug() {
```
Manishearth: please explain this test

Manishearth (commented):
It's kind of hard to understand what actually broke here, but that's because I don't understand this part of the collation api

Manishearth:
@sffc Also I believe our patch release stance is to merge to main and then copy a patch over ,yes?

sffc:
> @sffc Also I believe our patch release stance is to merge to main and then copy a patch over ,yes?

The bug is in 2.1. It is fixed on main due to other refactorings, perhaps #7891 

I should check 2.2

Manishearth:
Ah, okay

tools/make/tidy.toml
robertbastian, line 1: also here, why is this needed? if this is needed it should be a separate PR on the release branch ("Fix CI") and not part of this commit, which might be cherry-picked, patched, etc.

tools/make/bakeddata/src/main.rs
```diff
@@ -21,7 +21,11 @@ const REPO_VERSION: &str = "version.workspace = true";
 const COMPONENTS: &[(&str, &[DataMarkerInfo], &str)] = &[
     ("calendar", icu::calendar::provider::MARKERS, REPO_VERSION),
     ("casemap", icu::casemap::provider::MARKERS, REPO_VERSION),
-    ("collator", icu::collator::provider::MARKERS, REPO_VERSION),
+    (
+        "collator",
+        icu::collator::provider::MARKERS,
+        "version = \"2.1.2\"",
```
robertbastian: there's no data diff, why are you bumping this?

components/collator/tests/tests.rs
```diff
@@ -2128,3 +2128,19 @@ fn test_latin1_utf16_fi() {
...
+    let provider = TestingProvider;
+    let payload: DataPayload<CollationSpecialPrimariesV1> =
+        provider.load(Default::default()).unwrap().payload;
+    let csp = payload.get();
+    assert!(csp.last_primaries.len() > MaxVariable::Currency as usize);
```
robertbastian: issue: some condition on the data struct is not a "user bug". this seems to be testing implementation details, why does the constructor call below not suffice as a test?

components/experimental/src/duration/validated_options.rs
robertbastian, line 1: we have a pinned toolchain, a pinned clippy version, a pinned MSRV, etc., so why does this unrelated code need to change?

robertbastian (changes requested):
> ### Context on 2.2 (main)
> In ICU4X 2.2 (main branch), this was refactored. 

Main is not 2.2, main is 2.3. This bug exists on 2.2, in fact users that depend on `version = "2"` will be on 2.2, so this should be fixed first and foremost on the 2.2 release branch, and then backported to other releases if needed (2.1, 2.0). This PR only fixes users that depend on `version = "~2.1"`, which, while the conformance project does that, is a small minority of users.

robertbastian:
> > @sffc Also I believe our patch release stance is to merge to main and then copy a patch over ,yes?
> 
> The bug is in 2.1. It is fixed on main due to other refactorings, perhaps #7891
> 
> I should check 2.2

It would still be good to have the test on main

tools/make/tidy.toml
sffc: #8078

components/experimental/src/duration/validated_options.rs
sffc: Somehow this got patched into the 2.1 branch without the docs. fixing in #8078

components/collator/tests/tests.rs
sffc: Checking the root cause seems harmless, but since you labeled this "issue", I assume you consider this blocking, so I will remove this extra assertion.

tools/make/bakeddata/src/main.rs
sffc: good catch! I missed this. Fixed.

sffc:
The force-push was to make this PR sit cleanly on top of the CI-fixing commits which are moved into #8078

sffc:
There are now 4 PRs:

- The CI fixes from here are split into https://github.com/unicode-org/icu4x/pull/8078
- The fix is ported to release/2.2: https://github.com/unicode-org/icu4x/pull/8081
- The test is added to main: https://github.com/unicode-org/icu4x/pull/8080
- And this PR is the fix on release/2.1

Manishearth (approved):

robertbastian (commented):
I don't particularly like the architecture of the constants, but as this is not main I don't care. I have left comments on #8083.

> ~Also bumped `icu_collator` to `2.1.2`.~ --> this will be done in another PR

This should be done in the same PR, doing this in a different PR is extra work for downstream users if they want to patch this fix.

sffc:
> This should be done in the same PR, doing this in a different PR is extra work for downstream users if they want to patch this fix.

I thought we decoupled fixes from the Cargo.toml diff, since people who vendor don't particularly care about the Cargo.toml diff, and it makes it harder to apply the same diff across multiple versions (like 2.1 and 2.2)?

sffc:
I'll update this after https://github.com/unicode-org/icu4x/pull/8089 lands.

robertbastian (approved):

PR #8077, opened 2026-06-15
Move all UCD parsing code into one file, share some code with helpers

### LINKED ISSUE #7947: Collect UCD-format parsing code in one file
Followup from https://github.com/unicode-org/icu4x/pull/7904#discussion_r3148832021 , https://github.com/unicode-org/icu4x/pull/7904#pullrequestreview-4193035179

The UCD format is not too complex, but it has some nuances, e.g. around `@missing`. As a maintainer of the unicode-rs crates, updating all of the code around UCD parsing when things change has been a bit of a slog, and often hard to do consistently. Markus has alluded to further complexities around these files, though from my conversation with him I think aside from `@missing` we don't need to worry about most of that.

From a maintenance perspective, having all of the code in one file would be nicer, so we can ensure things behave similarly, and if things change (or we find a bug) it's easy to update everyone.

I'm not advocating for consolidating them all into a single "parse UCD format" function. Mostly just seeing if all of the for loops that look like this

https://github.com/unicode-org/icu4x/blob/ad0697a918bca1d9867011b015a0e869e8714aff/provider/source/src/properties/bin_cp_set.rs#L202-L218

are in a separate file. It's fine if that file ends up with 5 functions with very similar bodies but different signatures.

@robertbastian was opposed to doing this in a way that introduces intermediate structures, which seems fine. We should try and consolidate similar functions if possible but I care more about the code being in one place rather than making an abstraction that lets us DRY.

Manishearth:
Fixes https://github.com/unicode-org/icu4x/issues/7947

This is an optional refactor. I think it's a good idea; it makes it easy for us to deal with parsing issues as we find them. This came up during the review for #7904 and Robert wasn't interested in doing it so I filed a followup.

I deliberately chose to not genericize `@missing` parsing, but we should consider looking into that. I'm not sure if it would be a net positive.

Having all these in one place cleans up the code significantly IMO.

Review commit by commit: the first commits are clean code moves (they also add lines of documentation), and the last commit adds and uses some nicer helpers.


sffc (approved):
seems ok, I am neutral on this

Manishearth:
In that case I will wait for Robert's opinion since this is his code and he has been actively working on it.

provider/source/src/properties/ucd_helper.rs
```diff
@@ -0,0 +1,474 @@
...
+    ///
+    /// It reads the file, strips comments (lines starting with `#` or anything after `#`),
+    /// skips empty lines, and passes a mutable reference to an iterator over the trimmed,
+    /// semicolon-separated fields to the callback.
+    ///
+    /// **Important:** Eagerly strips all comments, so it **cannot** be used for parsing
+    /// `@missing` rules which are semantically significant comments starting with `# @missing`.
+    fn parse_ucd_lines<F>(&self, file: &str, mut action: F) -> Result<(), DataError>
+    where
+        F: for<'a> FnMut(&mut dyn Iterator<Item = &'a str>) -> Result<(), DataError>,
```
robertbastian, line 54: issue: JS vibes. why not return an iterator of iterators? or at least let the callback return a value and this function an iterator of values, then you can do things like `find` without juggling mutable state, use for-loops, etc

provider/source/src/properties/ucd_helper.rs
```diff
@@ -0,0 +1,474 @@
...
+    /// Validates that the given property name matches its expected short name.
+    ///
+    /// This helper reads `PropertyValueAliases.txt` to verify the association.
+    pub(super) fn validate_property_name(
+        &self,
+        name: &str,
+        short_name: &str,
+    ) -> Result<(), DataError> {
+        let mut sn = None;
+        self.parse_ucd_lines("ucd/PropertyAliases.txt", |fields| {
```
robertbastian, line 76: issue: this function was a lot more readable before. you now have three short name variables and mutable state because `parse_ucd_lines` does not return anything

provider/source/src/properties/ucd_helper.rs
```diff
@@ -0,0 +1,474 @@
...
+            Ok(())
+        })?;
+
+        let inv_list = inv_list.build();
+
+        Ok(CodePointInversionListAndStringList::try_from(
+            inv_list,
+            VarZeroVec::from(&strings.into_iter().collect::<Vec<_>>()),
+        )
+        .expect("invariants upheld"))
```
robertbastian, line 194: nit: might as well unwrap if you don't state what the invariants are

provider/source/src/properties/ucd_helper.rs
```diff
@@ -0,0 +1,474 @@
...
+            paired_brackets.insert(cp, typ);
+            Ok(())
+        })?;
+        Ok(paired_brackets)
+    }
+
+    /// Parses `ScriptExtensions.txt` to create script sets and a lookup map for code points.
+    ///
+    /// Returns a tuple containing:
+    /// 1. A list of unique script sets (`Vec<Vec<Script>>`), where each set is sorted.
```
robertbastian, line 245: this representation is tied to tightly to the data struct, this code really should live in script.rs

robertbastian (commented):
I'm not sure if I like this. Yes, maybe abstracting out `parse_range` and `parse_ucd_lines` is useful (although I'm not happy with `parse_ucd_line`), but moving basically all interesting logic into a single huge file and having the organised files do nothing does not improve readability imho. Also, with the way `# @missing` is handled, this actually doesn't simplify much code.

provider/source/src/properties/ucd_helper.rs
Manishearth: Ah, this was a deliberate choice to try and keep the diff smaller and keep the code working the same.

Manishearth:
> but moving basically all interesting logic into a single huge file and having the organised files do nothing does not improve readability imho

This was my primary impetus, sharing code is secondary. I found it harder to read when all of this code had a bunch of parsing stuff stuck inside it.

provider/source/src/properties/ucd_helper.rs
Manishearth: Yeah. I agree

Manishearth:
https://github.com/unicode-org/icu4x/pull/8115

PR #8078, opened 2026-06-15
Fix pre-existing CI failures on release/2.1 branch (experimental docs & tidy)

sffc:
This PR fixes pre-existing CI failures on the `release/2.1` branch by addressing two independent issues:

### 1. missing docs and unresolved link in `icu_experimental`
*   **The Problem**: A recent commit on the `release/2.1` branch ([#7616](https://github.com/unicode-org/icu4x/pull/7616)) re-exported `ValidatedDurationFormatterOptions`, making it part of the public API. This suddenly subjected it to strict documentation checks, exposing two issues:
    1.  A missing doc comment on the `pub fn validate` function (triggering `missing_docs` lint).
    2.  A broken intra-doc link `[DurationFormatter](DurationFormatter)` that could not be resolved in that scope.
*   **The Impact**: Because CI runs with `-D warnings` (deny warnings), these two issues caused the **`clippy`**, **`doc`**, and **`msrv`** jobs to fail.
*   **The Fix**: Added the missing doc comment to `ValidatedDurationFormatterOptions::validate` and corrected the intra-doc link to use the proper path (`crate::duration::DurationFormatter`), which unblocks all three jobs.

### 2. cargo-rdme installation failure in `tidy`
*   **The Problem**: The `tidy` job in CI automatically installs `cargo-rdme`. Recently, a new version of `cargo-rdme` (or its dependencies) was released on crates.io that requires Rust 1.91. However, the `release/2.1` branch is pinned to Rust 1.90. This caused the automatic installation to fail in CI.
*   **The Impact**: This caused the **`tidy`** job to fail.
*   **The Fix**: Updated the installation command in `tidy.toml` to use `cargo install cargo-rdme --locked`. This forces Cargo to respect the locked, compatible dependency versions that still support Rust 1.90, unblocking the `tidy` job.

***

Both of these fixes are required to make the CI green for any PR targeted to `release/2.1`. They are combined here because checking them in separately would require each one to not pass CI.

## Changelog

N/A

### DIFF
--- a/tools/make/tidy.toml
+++ b/tools/make/tidy.toml
@@ -133,9 +133,22 @@ end
 echo "License header check complete"
 '''
 
+[tasks.install-cargo-rdme]
+description = "Install cargo-rdme"
+private = true
+script_runner = "@duckscript"
+script = '''
+cargo_list = exec cargo --list
+if not contains ${cargo_list.stdout} "rdme"
+    echo "Installing cargo-rdme..."
+    exec cargo install cargo-rdme --locked
+end
+'''
+
 [tasks.generate-readmes]
 description = "Automatically generate README.md for each component."
-install_crate = { crate_name = "cargo-rdme", test_arg = ["--help"] }
+install_crate = false
+dependencies = ["install-cargo-rdme"]
 category = "ICU4X Development"
 script_runner = "@duckscript"
 script = '''
--- a/components/experimental/src/duration/validated_options.rs
+++ b/components/experimental/src/duration/validated_options.rs
@@ -4,7 +4,7 @@
 
 use crate::duration::options::*;
 
-/// Validated options for [DurationFormatter](DurationFormatter).
+/// Validated options for [`DurationFormatter`](crate::duration::DurationFormatter).
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct ValidatedDurationFormatterOptions {
     /// The style that will be applied to units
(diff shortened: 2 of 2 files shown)


Manishearth (approved):

tools/make/tidy.toml
```diff
@@ -133,9 +133,22 @@ end
...
+if not contains ${cargo_list.stdout} "rdme"
+    echo "Installing cargo-rdme..."
+    exec cargo install cargo-rdme --locked
+end
+'''
+
 [tasks.generate-readmes]
 description = "Automatically generate README.md for each component."
-install_crate = { crate_name = "cargo-rdme", test_arg = ["--help"] }
+install_crate = false
```
robertbastian, line 150: I would have preferred if we changed this to install `cargo-rdme` using the stable toolchain instead of the repo toolchain. we gain nothing from installing it using the pinned toolchain

also, this should probably be changed on main then, so we have it going forward?

robertbastian (commented):
I'm confused why https://github.com/unicode-org/icu4x/pull/7616 didn't trigger CI but this did

tools/make/tidy.toml
sffc: I don't really want to change it on main. I would like to keep using `install_crate`.

robertbastian: what I'm saying is that we can probably tell `install_crate` to use the stable toolchain. you've fixed the 2.1 branch, but this issue is eventually going to happen on all release branches, so we should fix it on main for all future branches

Issue #7825, opened 2026-03-30
Implement full display names algorithm


sffc:
This issue tracks the full display name algorithm implementation on top of the optimized data (#3260), which is designed for granular data slicing and supports more efficient loading where only a single name lookup is required (e.g., "Current Location: France").

While the `multi` submodule provides broader coverage, the `single` submodule is currently in active development and has a more limited feature set. We are seeking early adopters to test the current implementation and provide feedback on real-world utility.

See also https://github.com/unicode-org/icu4x/issues/7824 for evaluating the `single` vs `multi` API style.

### Implementation Status

The `single` submodule currently supports a subset of the display name types available in `multi`.

| Feature | Supported in `single`? | Type Name (Example) |
| :--- | :---: | :--- |
| **Regions** | ✅ | `RegionDisplayName` |
| **Scripts** | ✅ | `ScriptDisplayName` |
| **Languages** | ❌ | *Not yet implemented* |
| **Variants** | ❌ | *Not yet implemented* |
| **Locales** | ❌ | *Not yet implemented* |

sffc:
Language identifiers are done. I'll split out Locale identifiers since ECMA-402 doesn't need them.

sffc:
Splitting locale display names into https://github.com/unicode-org/icu4x/issues/8352.

Issue #7824, opened 2026-03-30
Feedback Request for `icu4x` DisplayNames Experimental Module


sffc:
We have introduced an experimental `displaynames` module in ICU4X and are seeking feedback on its API design. We are currently evaluating two submodules: **`multi`** and **`single`**. 

The primary distinction is how data is loaded and accessed. We want to determine which approach better suits real-world performance requirements and developer workflows.

Some technical discussion of the tradeoffs: https://github.com/unicode-org/icu4x/issues/3260#issuecomment-3881644475

### Comparison of Submodules

| Submodule | Type Naming (Example) | Data Loading Pattern |
| :--- | :--- | :--- |
| **`multi`** | `RegionDisplayNames` | One data load for multiple lookups (e.g., populating a list or dropdown menu). |
| **`single`** | `RegionDisplayName` | One data load for a single name (e.g., "Current Location: France"). |

### Key Feedback Areas
We are specifically looking for input on the following:

* **Real-World Use Cases:** Which pattern (batch vs. granular) occurs more frequently in your applications?
* **Performance:** Is the overhead of loading broader data structures in `multi` a concern, or is reducing the number of load calls more critical?
* **Unified vs. Split:** Should we maintain both patterns, or is one clearly superior for most internationalization tasks?
* **Same Application:** If you have use cases for both single and multi, do they occur in the same application?

### How to Help
Please comment on this issue with your specific use case or a code snippet showing how you would expect to use these APIs.

Manishearth:
(Prior architecture ideas are [here](https://github.com/unicode-org/icu4x/issues/3260#issuecomment-3881644475))

I was thinking more about my preferences here and I realized that while ICU4X generally eschews "duplicate data" (data stored in different, non-deduplicatable formats, that gets loaded based on usage), it might be something where we allow ourselves to do it a _little bit_ for some cases. So we'd just export `multi` and `single` permanently, and users use the one they need, with a comment on whole-program usage.


That made me realize that in that world we can still, under the hood, not have duplicate data. What we'd do is have a `multi` feature (or a `single` and `multi` feature), and when `multi` is enabled, under the hood `single::DisplayName` will load `multi` data.

This will work just fine for baked data. For postcard and `try_new_unstable` we may need to have some fallbacking, or something else, but I think there is a decent design space for those.

sffc:
[...]

PR #8082, opened 2026-06-16
Implement single language display names

sffc:
Depends on #8085

Implements the single display names component for Languages, bringing them up to the planned specification. This PR lands the core implementation of:

- `LanguageDisplayNameOwned` & `LanguageDisplayName`: Supports standard and dialect formatting (dialect resolution is fully implemented). Currently limited to the default `Medium` width.

The `QualifiersWriteable` helper makes it all zero-copy.

This does not implement menu names. That will be in an upcoming PR.

Addresses #7824 and #7825.

## Open Questions

- **Constructor Argument Order**: In `LanguageDisplayNameOwned::try_new(prefs, locale_id, options)`, we have placed `options` last. This is because `options` behaves like a trailing optional bag (similar to varargs in other languages). Does this match the preferred style for ICU4X constructors, or should `options` be placed elsewhere?
- **Writeable on Owned Types**: Currently, owned types like `ScriptDisplayNameOwned`, `RegionDisplayNameOwned`, `VariantDisplayNameOwned`, and `LanguageDisplayNameOwned` implement `Writeable` (by forwarding to their borrowed counterparts via `.as_borrowed()`). Should owned types implement `Writeable` directly, or should users be forced to call `.as_borrowed()` to format them? Keeping `Writeable` on owned types is convenient but increases API surface. We should decide if we want to keep this pattern or deprecate it.
- **Fallback Behavior in Single Formatters**: Currently, single formatters (`Script`, `Region`, `Variant`, `Language`) fail-fast in the constructor (`try_new` returns `Err(DataError)`) if the specific subtag data is missing from the provider (e.g., `xx` or an untranslated language). Should we redesign them to support fallback to the code (similar to the multi formatter) by storing the input identifier and making the payloads optional? Or is the current fail-fast behavior preferred for single formatters? If we choose fallback, we will need to update the existing `Region` and `Script` formatters in a future PR to maintain consistency.

## Changelog

`icu_experimental/displaynames`:
  - Added `displaynames::single::LanguageDisplayNameOwned` and `LanguageDisplayName` for formatting language display names.


sffc:
This PR is reviewable commit-by-commit, but each commit is split into its own PR.

components/experimental/tests/displaynames/tests.rs
```diff
@@ -95,11 +115,21 @@ fn test_concatenate() {
             let result = result.into_owned();
             assert_eq!(result.capacity(), result.len());
         }
+
+        // Test our new single formatter implementation (only for cases that are in the data, i.e. not "xx")
```
robertbastian: please also test `xx`, if only to demonstrate how a client would do the fallback themselves

components/experimental/tests/displaynames/tests.rs
```diff
@@ -110,12 +140,12 @@ fn test_single_language_display_name() {
...
     let options: DisplayNamesOptions = Default::default();
 
-    // This should format "zh-Hant-HK" to "Traditional Chinese (Hong Kong)" in "en-001"
+    // This should format "zh-Hant-HK" to "Traditional Chinese (Hong Kong SAR China)" in "en-001"
     let lang_id = langid!("zh-Hant-HK");
     let lang_name = LanguageDisplayNameOwned::try_new(locale.into(), lang_id, options)
         .expect("Data should load successfully");
 
-    assert_writeable_eq!(lang_name, "Traditional Chinese (Hong Kong)");
+    assert_writeable_eq!(lang_name, "Traditional Chinese (Hong Kong SAR China)");
```
robertbastian: why not test this as part of `test_concatenate`?

components/experimental/src/displaynames/single/language.rs
```diff
@@ -0,0 +1,389 @@
...
+///     langid!("fr-CA"),
+///     options,
+/// )
+/// .expect("Data should load successfully");
+///
+/// assert_writeable_eq!(display_name, "Canadian French");
+/// ```
+#[allow(dead_code)]
+#[derive(Debug)]
+pub struct LanguageDisplayNameOwned {
```
robertbastian: issue: `Language` is the subtag, this formats `LanguageIdentifier`s, so it should be called `LanguageIdentifierDisplayNameOwned`.

the `multi` implementation take `&Locale`, and we do want to format unicode extensions at some point

components/experimental/src/displaynames/single/language.rs
```diff
@@ -0,0 +1,389 @@
...
+            try_new_unstable,
+            Self
+        ]
+    );
+
+    #[doc = icu_provider::gen_buffer_unstable_docs!(UNSTABLE, Self::try_new)]
+    pub fn try_new_unstable<D>(
+        provider: &D,
+        prefs: DisplayNamesPreferences,
+        locale_id: icu_locale::LanguageIdentifier,
```
robertbastian: `langid`?

components/experimental/src/displaynames/single/language.rs
```diff
@@ -0,0 +1,389 @@
...
+    ) -> Result<Self, DataError>
+    where
+        D: DataProvider<LocaleNamesLanguageMediumV1>
+            + DataProvider<LocaleNamesScriptMediumV1>
+            + DataProvider<LocaleNamesRegionMediumV1>
+            + DataProvider<LocaleNamesVariantMediumV1>
+            + DataProvider<LocaleNamesEssentialsV1>
+            + ?Sized,
+    {
+        let locale = LocaleNamesLanguageMediumV1::make_locale(prefs.locale_preferences);
```
robertbastian: please use better names for the two locales, like display locale and subject locale or something

components/experimental/src/displaynames/single/language.rs
```diff
@@ -0,0 +1,389 @@
...
+            }
+        }
+
+        // Format the collected qualifiers.
+        // If we have 1 qualifier, we just write it.
+        // If we have multiple, we join them pair-wise using the locale separator pattern.
+        // For example, if we have [Q0, Q1, Q2], we format it as:
+        //   interpolate(interpolate(Q0, Q1), Q2)
+        // This avoids any allocation and matches the TR35 spec for list formatting of qualifiers.
+        match len {
```
robertbastian: issue: this reimplements list formatting. one thing that we realised in list formatting is that we don't need to use `Pattern`s, because for example ` {0}, {1}` is an illegal list joiner pattern, as there is text before the first placeholder, which makes the operation non-associative. once we added the invariant that a join pattern needs to start with `{0}` and end with `{1}`, the `write_to` code could be implemented as a loop over the items, without an intermediate collection

please apply those learnings here.

components/experimental/src/displaynames/single/language.rs
```diff
@@ -0,0 +1,389 @@
...
+    region: Option<&'a str>,
+    variants: &'a [DataPayload<LocaleNamesVariantMediumV1>],
+    separator: &'a DoublePlaceholderPattern,
+}
+
+impl<'a> writeable::Writeable for QualifiersWriteable<'a> {
+    fn write_to<W: core::fmt::Write + ?Sized>(&self, sink: &mut W) -> core::fmt::Result {
+        // Collect all non-consumed qualifiers (script, region, variants) into a stack array.
+        // We have at most 4 qualifiers (1 script, 1 region, and up to 2 variants in practice,
+        // though BCP-47 allows more, CLDR data rarely has more than 2 variants for display).
```
robertbastian: yeah but what if it does? this code will just truncate

Manishearth:
> Constructor Argument Order: In LanguageDisplayNameOwned::try_new(prefs, locale_id, options), we have placed options last. This is because options behaves like a trailing optional bag (similar to varargs in other languages). Does this match the preferred style for ICU4X constructors, or should options be placed elsewhere?


This is the normal style

> Writeable on Owned Types: Currently, owned types like ScriptDisplayNameOwned, RegionDisplayNameOwned, VariantDisplayNameOwned, and LanguageDisplayNameOwned implement Writeable (by forwarding to their borrowed counterparts via .as_borrowed()). Should owned types implement Writeable directly, or should users be forced to call .as_borrowed() to format them? Keeping Writeable on owned types is convenient but increases API surface. We should decide if we want to keep this pattern or deprecate it.


No harm in having multiple `Writeable` impls. Hopefully it optimizes codesize wise

> Fallback Behavior in Single Formatters: Currently, single formatters (Script, Region, Variant, Language) fail-fast in the constructor (try_new returns Err(DataError)) if the specific subtag data is missing from the provider (e.g., xx or an untranslated language). Should we redesign them to support fallback to the code (similar to the multi formatter) by storing the input identifier and making the payloads optional? Or is the current fail-fast behavior preferred for single formatters? If we choose fallback, we will need to update the existing Region and Script formatters in a future PR to maintain consistency.


Say more about what this would look like? I'm not sure I understand "storing the input identifier and making the payloads optional"

components/experimental/src/displaynames/single/language.rs
```diff
@@ -0,0 +1,389 @@
...
+        if options.language_display == LanguageDisplay::Dialect {
+            // 1a. Try 3-subtag combination (if both script and region are present)
+            if let (Some(script), Some(region)) = (locale_id.script, locale_id.region) {
+                let temp_id = icu_locale::LanguageIdentifier {
+                    language: locale_id.language,
+                    script: Some(script),
+                    region: Some(region),
+                    variants: Default::default(),
+                };
+                let attr = temp_id.to_string();
```
robertbastian: suggestion: use a stack string for this. this has a maximum length

you can even reuse the same stack string for all attributes

components/experimental/src/displaynames/single/language.rs
```diff
@@ -0,0 +1,389 @@
...
+                let mut metadata = DataRequestMetadata::default();
+                metadata.silent = true;
+                if let Ok(response) = provider.load(DataRequest { id, metadata }) {
+                    language_payload = Some(response.payload);
+                    script_consumed = true;
+                    region_consumed = true;
+                }
+            }
+
+            // 1b. Try language + script (if script is present and not consumed)
```
robertbastian: you don't check that script is not consumed

components/experimental/src/displaynames/single/language.rs
```diff
@@ -0,0 +1,389 @@
...
+                let id = DataIdentifierBorrowed::for_marker_attributes_and_locale(
+                    DataMarkerAttributes::try_from_str(&attr)
+                        .map_err(|_| DataError::custom("Invalid dialect attr"))?,
+                    &locale,
+                );
+                let mut metadata = DataRequestMetadata::default();
+                metadata.silent = true;
+                if let Ok(response) = provider.load(DataRequest { id, metadata }) {
+                    language_payload = Some(response.payload);
+                    script_consumed = true;
```
robertbastian: suggestion: instead of keeping track of consumed subtags, remove them from the langid

components/experimental/src/displaynames/single/language.rs
```diff
@@ -0,0 +1,389 @@
...
+                    region: Some(region),
+                    variants: Default::default(),
+                };
+                let attr = temp_id.to_string();
+                let id = DataIdentifierBorrowed::for_marker_attributes_and_locale(
+                    DataMarkerAttributes::try_from_str(&attr)
+                        .map_err(|_| DataError::custom("Invalid dialect attr"))?,
+                    &locale,
+                );
+                let mut metadata = DataRequestMetadata::default();
```
robertbastian, line 125: thought: we should really add a `DataRequestMetadata::silent()` constructor

components/experimental/src/displaynames/single/language.rs
```diff
@@ -0,0 +1,389 @@
...
+                metadata.silent = true;
+                if let Ok(response) = provider.load(DataRequest { id, metadata }) {
```
robertbastian: issue: this discards all `DataError`s. use `allow_identifier_not_found`

components/experimental/src/displaynames/single/language.rs
```diff
@@ -0,0 +1,389 @@
...
+                        ),
+                        ..Default::default()
+                    })?
+                    .payload
+            }
+        };
+
+        // Step 2: Load script name (if present in locale_id and not consumed)
+        let script_payload = if let Some(script) = locale_id.script {
+            if !script_consumed {
```
robertbastian: this can be collapsed into a single if, but you shouldn't have `script_consumed` in the first place

components/experimental/src/displaynames/single/language.rs
```diff
@@ -0,0 +1,389 @@
...
+                }
+            } else {
+                DataPayloadOr::none()
+            }
+        } else {
+            DataPayloadOr::none()
+        };
+
+        // Step 4: Load variant names (if present in locale_id)
+        let mut variant_payloads = Vec::new();
```
robertbastian: shortvec?

components/experimental/src/displaynames/single/language.rs
```diff
@@ -0,0 +1,389 @@
...
+            let id = DataIdentifierBorrowed::for_marker_attributes_and_locale(
+                DataMarkerAttributes::try_from_str(variant.as_str())
+                    .map_err(|_| DataError::custom("Invalid variant"))?,
+                &locale,
+            );
+            let mut metadata = DataRequestMetadata::default();
+            metadata.silent = true;
+            match provider.load(DataRequest { id, metadata }) {
+                Ok(response) => variant_payloads.push(response.payload),
+                Err(DataError {
```
robertbastian: `allow_identifier_not_found`

components/experimental/src/displaynames/single/language.rs
```diff
@@ -0,0 +1,389 @@
...
+            None => {
+                provider
+                    .load(DataRequest {
+                        id: DataIdentifierBorrowed::for_marker_attributes_and_locale(
+                            DataMarkerAttributes::try_from_str(locale_id.language.as_str())
+                                .map_err(|_| DataError::custom("Invalid language"))?,
+                            &locale,
+                        ),
+                        ..Default::default()
+                    })?
```
robertbastian, line 162: please add a TODO that this is the location where we'd fall back to the code

components/experimental/src/displaynames/single/language.rs
```diff
@@ -0,0 +1,389 @@
...
+                &locale,
+            );
+            let mut metadata = DataRequestMetadata::default();
+            metadata.silent = true;
+            match provider.load(DataRequest { id, metadata }) {
+                Ok(response) => variant_payloads.push(response.payload),
+                Err(DataError {
+                    kind: DataErrorKind::IdentifierNotFound,
+                    ..
+                }) => {}
```
robertbastian: issue: you just drop unknown variants, they should fall back to the code

robertbastian (changes requested):
A missing name should not be a `DataError`, it should fall back to the code

components/experimental/src/displaynames/single/language.rs
sffc: When designing this, I settled on `Language` because:

- A language display name formatter that formats only the language subtag is not well-defined in CLDR. We always read the other subtags to pick the correct dialect name.
- This formatter does what most regular people expect for a type named `LanguageDisplayName`.
- If we did implement a formatter with language-subtag-only behavior, calling it `LanguageDisplayName` would be misleading, because it isn't the thing most people want. It should be called something like `LanguageSubtagDisplayName`.
    - Note: in icu_locale_core, we have a `Language` type which is a subtag, but it is scoped inside the `subtag` module to emphasize what it is and is not.
- [ECMA-402 calls this "language".](https://tc39.es/ecma402/#sec-Intl.DisplayNames)

In the [design doc](https://github.com/unicode-org/icu4x/blob/main/components/experimental/src/displaynames/single/README.md#formatters--constructors), I listed `LanguageDisplayName` as being the thing that formats a `LanguageIdentifier`.

I would like to stick with the design doc for this PR and do this bikeshed in a follow-up discussion.

components/experimental/tests/displaynames/tests.rs
sffc: It's an end-to-end test we are now enabling. Would you prefer to delete the whole test (and merge it in with one of the other tests)?

components/experimental/src/displaynames/single/language.rs
sffc: Oops, this is terrible. I should have caught this from AI output but I didn't. Thank you.

components/experimental/src/displaynames/single/language.rs
robertbastian: > * A language display name formatter that formats only the language subtag is not well-defined in CLDR. We always read the other subtags to pick the correct dialect name.

> * If we did implement a formatter with language-subtag-only behavior, calling it `LanguageDisplayName` would be misleading, because it isn't the thing most people want. It should be called something like `LanguageSubtagDisplayName`.
>     * Note: in icu_locale_core, we have a `Language` type which is a subtag, but it is scoped inside the `subtag` module to emphasize what it is and is not.

The naming of a potential language subtag formatter shouldn't impact the name here.

> This formatter does what most regular people expect for a type named `LanguageDisplayName`.

Both `LanguageIdentifier` and `Locale` are what most regular people expect a "language" to be. But we should be exact with terminology.

> - [ECMA-402 calls this "language".](https://tc39.es/ecma402/#sec-Intl.DisplayNames)

Not an argument, ECMA-402 has a lot of differing naming decisions

> In the design doc, I listed LanguageDisplayName as being the thing that formats a LanguageIdentifier.

I was never given this doc for approval, we could have had the discussion there, so we're having it now.

components/experimental/src/displaynames/single/language.rs
```diff
@@ -84,79 +85,83 @@ impl LanguageDisplayNameOwned {
...
-                    script: Some(script),
-                    region: Some(region),
-                    variants: Default::default(),
-                };
-                let attr = temp_id.to_string();
+            if let (Some(script), Some(region)) = (subject.script, subject.region) {
+                let lang_str = subject.language.to_tinystr();
+                let script_str = script.to_tinystr();
+                let region_str = region.to_tinystr();
+                let hyphen = tinystr::tinystr!(1, "-");
```
robertbastian: make this a const

components/experimental/src/displaynames/single/language.rs
```diff
@@ -84,79 +85,83 @@ impl LanguageDisplayNameOwned {
...
+                let attr1: tinystr::TinyAsciiStr<16> = lang_str.concat(hyphen);
+                let attr2: tinystr::TinyAsciiStr<16> = attr1.concat(script_str);
+                let attr3: tinystr::TinyAsciiStr<16> = attr2.concat(hyphen);
+                let attr: tinystr::TinyAsciiStr<16> = attr3.concat(region_str);
```
robertbastian: if things are hard to name, don't name them

```suggestion
                let attr = lang_str.concat::<16>(hyphen).concat::<16>(script_str).concat::<16>(hyphen).concat::<16>(region_str);

File: components/experimental/src/displaynames/single/language.rs [hunk 20]
```diff
@@ -84,79 +85,83 @@ impl LanguageDisplayNameOwned {
...
                 let id = DataIdentifierBorrowed::for_marker_attributes_and_locale(
-                    DataMarkerAttributes::try_from_str(&attr)
+                    DataMarkerAttributes::try_from_str(attr.as_str())
                         .map_err(|_| DataError::custom("Invalid dialect attr"))?,
```
robertbastian: issue: this code should be in an infallible `LocaleNamesLanguageMediumV1::make_attributes(Language, Script, Region)`

components/experimental/src/displaynames/single/language.rs
```diff
@@ -165,99 +170,101 @@ impl LanguageDisplayNameOwned {
         let language_payload = match language_payload {
             Some(payload) => payload,
             None => {
+                // TODO(#8100): Fall back to the code instead of failing with DataError if the language name is not found
                 provider
                     .load(DataRequest {
                         id: DataIdentifierBorrowed::for_marker_attributes_and_locale(
-                            DataMarkerAttributes::try_from_str(locale_id.language.as_str())
+                            DataMarkerAttributes::try_from_str(subject.language.as_str())
```
robertbastian: `make_attributes`

components/experimental/src/displaynames/single/language.rs
robertbastian: the "location where we'd fall back to the code" needs to be created with `allow_identifier_not_found` first, like you do for variants below

components/experimental/src/displaynames/single/language.rs
```diff
@@ -165,99 +170,101 @@ impl LanguageDisplayNameOwned {
...
+            let id = DataIdentifierBorrowed::for_marker_attributes_and_locale(
+                DataMarkerAttributes::try_from_str(region.as_str())
+                    .map_err(|_| DataError::custom("Invalid region"))?,
+                &formatting_locale,
+            );
+            let mut metadata = DataRequestMetadata::default();
+            metadata.silent = true;
+            match provider.load(DataRequest { id, metadata }) {
+                Ok(response) => DataPayloadOr::from_payload(response.payload),
+                Err(DataError {
```
robertbastian: `allow_identifier_not_found`

components/experimental/src/displaynames/single/language.rs
```diff
@@ -165,99 +170,101 @@ impl LanguageDisplayNameOwned {
...
+                .load(DataRequest { id, metadata })
+                .allow_identifier_not_found()?
+            {
+                loaded_variants.push(response.payload);
+            } else {
+                // TODO(#8100): Fall back to the code instead of dropping it if the variant name is not found
             }
         }
 
+        let variant_payloads = if loaded_variants.len() == 1 {
```
robertbastian: at this point you have already allocated a vec, there's little point in deallocating it now

components/experimental/src/displaynames/single/language.rs
```diff
@@ -294,81 +312,80 @@ impl writeable::Writeable for LanguageDisplayNameOwned {
 
 writeable::impl_display_with_writeable!(LanguageDisplayNameOwned);
 
+/// Borrowed variants representation to avoid heap allocation.
+#[derive(Debug, Clone, Copy)]
+pub enum BorrowedVariants<'a> {
+    None,
+    One(&'a str),
+    Slice(&'a [DataPayload<LocaleNamesVariantMediumV1>]),
```
robertbastian, line 270: these are two different payloads, maybe be explicit about that. `One(&'a str)` borrows from data, which is `'static` in the compiled data case, but `Slice(&'a ...)` borrows from the owned name

components/experimental/src/displaynames/single/language.rs
robertbastian: I feel like @sffc would have definitely done this, and @sffc-bot is producing extra work for me

components/experimental/tests/displaynames/tests.rs
robertbastian: in #8100 you state that you don't even want to implement fallback, which is exactly why I'm asking you to demonstrate here how a client would do it

components/experimental/tests/displaynames/tests.rs
robertbastian: yes

I don't agree with the concept of having disabled tests in code anyway

components/experimental/tests/displaynames/tests.rs
```diff
@@ -110,12 +140,12 @@ fn test_single_language_display_name() {
...
 }
 
 #[cfg(any())]
```
robertbastian: nit: disable tests with `#[ignore]`

components/experimental/tests/displaynames/tests.rs
sffc: This test is testing LanguageDisplayName. I have no good way for a client to perform their own fallback other than falling back to the whole BCP-47 string.

I prefer to keep PRs small and incremental but I can implement it in this PR if you insist.

components/experimental/src/displaynames/single/language.rs
sffc: > The naming of a potential language subtag formatter shouldn't impact the name here.

I raised that point to say that there is nothing else competing for the name `LanguageDisplayName` and nothing else that I would expect a user to assume when they see a type named `LanguageDisplayName`.

> Not an argument, ECMA-402 has a lot of differing naming decisions

ECMA-402 naming decisions are always an input; we just often have other inputs that are stronger.

---

In an effort to unblock progress, I'll ask @sffc-bot to change this to LanguageIdentifierDisplayName (including in README.md), erring on the side of being more precise, and open a follow-up issue to bikeshed.

components/experimental/src/displaynames/single/language.rs
sffc: you're right. I'll walk through the code in more detail (and maybe rewrite parts of it) before sending it for review again. sorry

components/experimental/src/displaynames/single/language.rs
sffc: omg, why did the ai fail at refactoring this when it did all the others? It is usually is good at refactors. I'm trusting it less and less. :man_facepalming:

components/experimental/src/displaynames/single/language.rs
sffc: @sffc-bot you did a very bad job at this change. Your second lifetime is meaningless because it is always bound to the same lifetime as the first. I will fix it for you.

components/experimental/tests/displaynames/tests.rs
sffc: Actually I think the right way to do this is to implement TryWriteable, with error parts over the fallback strings. Do you agree, and will you let me do this in a follow-up?

sffc: I started implementing this. It should work but it needs some missing TryWriteable impls in icu_pattern. I'll make a separate PR for those. In the mean time I think I would like to merge this PR without the TryWriteable.

Here's my branch: https://github.com/sffc/omnicu/tree/dname-trywriteable

components/experimental/tests/displaynames/tests.rs
sffc: fixed in 6fc2b9f

sffc:
ok I think I got it all, let's merge this. I'll keep working on TryWriteable in separate PRs.

components/experimental/src/displaynames/single/language.rs
```diff
@@ -0,0 +1,362 @@
...
+    RegionDisplayNameOwned, ScriptDisplayNameOwned, VariantDisplayNameOwned,
+};
+use crate::displaynames::{DisplayNamesOptions, DisplayNamesPreferences, LanguageDisplay};
+use alloc::vec::Vec;
+use icu_pattern::DoublePlaceholderPattern;
+use icu_provider::DataPayloadOr;
+use icu_provider::prelude::*;
+use tinystr::TinyAsciiStr;
+
+/// A localized display name for a language, owned version.
```
robertbastian: nit: 
```suggestion
/// A localized display name for a language identifier, owned version.
```

components/experimental/src/displaynames/single/language.rs
```diff
@@ -0,0 +1,362 @@
...
+                }
+            }
+        }
+
+        // If the language name is not loaded yet, try loading it from the language subtag alone.
+        // TODO(#8100): Fall back to the code instead of failing with DataError if the language name is not found
+        let language_payload = match language_payload {
+            Some(payload) => payload,
+            None => {
+                let mut buffer = TinyAsciiStr::EMPTY;
```
robertbastian, line 148: nit: reuse the previous buffer

components/experimental/src/displaynames/single/language.rs
```diff
@@ -0,0 +1,362 @@
...
+        // present subtags (script, region, variants) will be formatted as qualifiers.
+
+        let mut language_payload = None;
+
+        // Only try dialect if requested (which is the default)
+        if options.language_display == LanguageDisplay::Dialect {
+            for (language, script, region) in [
+                (subject.language, Some(subject.script), Some(subject.region)),
+                (subject.language, Some(subject.script), None),
+                (subject.language, None, Some(subject.region)),
```
robertbastian, line 98: thought: you could have `(subject.language, None, None)` here as the fourth case, and `continue` the first ones (in your existing matches) if `Dialect` is disabled. then you only have the attribute construction and data loading code once

components/experimental/src/displaynames/single/language.rs
robertbastian: I meant what the bot did

components/experimental/src/displaynames/provider.rs
```diff
@@ -281,57 +281,88 @@ icu_provider::data_marker!(
...
+                let script_str = script.to_tinystr();
+                lang_str.concat::<1, 16>(HYPHEN).concat::<4, 16>(script_str)
+            }
+            (None, Some(region)) => {
+                let region_str = region.to_tinystr();
+                lang_str.concat::<1, 16>(HYPHEN).concat::<3, 16>(region_str)
+            }
+            (None, None) => lang_str.resize::<16>(),
+        };
+        // This won't panic because the buffer contains subtag strings, which are valid attributes
```
robertbastian:
```suggestion
        // This is infallible (will not panic) because validated `Language`, `Script`, 
        // `Region`, and hyphens are guaranteed to conform to `DataMarkerAttributes` syntax.
```

robertbastian (commented):
I would like to see tests for all the error cases
* language not found
* script not found
* region not found
* variant not found
* more than two variants

the tests can test the implementation as is (with TODOs everywhere), but should demonstrate that the code errors or truncates.

components/experimental/src/displaynames/single/language.rs
sffc: I tried this but it made the code quite a bit harder to follow because both the setup and the error handling of the two cases are different.

components/experimental/src/displaynames/single/language.rs
sffc: I disagree with preemptively splitting the lifetime. If we add a pure compiled data constructor for this, we need to redesign BorrowedVariants anyway, maybe adding a new variant that doesn't use a lifetime, and then `'static` will work fine. In the current code, one lifetime is the correct number.

components/experimental/src/displaynames/single/language.rs
sffc: done

components/experimental/src/displaynames/single/language.rs
sffc: I would rather keep this localized. There is no allocation happening.

sffc:
I added the extra tests. I don't see anything else open that's blockworthy / can't be fixed in a follow-up.

The previous review should have also been an approval because we don't typically block on more exhaustive test coverage (we had tests for the positive cases), but I'm happy to add the tests.

sffc:
tidy says:

```
Automatically generating components/calendar/README.md 
error: failed to transform intralinks: rust toolchain not installed: nightly
```

robertbastian (approved):
> The previous review should have also been an approval because we don't typically block on more exhaustive test coverage (we had tests for the positive cases), but I'm happy to add the tests.

If I want to see test cases for `DataError`s being thrown left and right because of incomplete code, it's reasonable to ask for test cases demonstrating this behaviour

Manishearth (approved):
Splitting the writeable into a subwriteable is a nice design.

PR #8083, opened 2026-06-16
Replace magic numbers with constants in CollationSpecialPrimaries deserialization

sffc:
See #8075

Defines `NUM_PRIMARIES` and `COMPRESSIBLE_BYTES_LEN` constants on `CollationSpecialPrimaries` and uses them in the custom `Deserialize` implementation to avoid magic numbers, making the deserialization logic self-documenting.

This addresses reviewer feedback regarding magic numbers in the collation data handling.

## Changelog

N/A (Replace magic numbers with constants in `CollationSpecialPrimaries` deserialization.)

### DIFF
--- a/components/collator/src/provider.rs
+++ b/components/collator/src/provider.rs
@@ -611,6 +611,15 @@ pub struct CollationSpecialPrimaries<'data> {
 }
 
 #[cfg(feature = "serde")]
+impl CollationSpecialPrimaries<'_> {
+    /// The number of real special primaries (Space, Punctuation, Symbol, Currency).
+    pub(crate) const NUM_PRIMARIES: usize = MaxVariable::Currency as usize + 1; // 4
+
+    /// The length of the compressible bytes array (256 bits packed in 16 u16s).
+    pub(crate) const COMPRESSIBLE_BYTES_LEN: usize = 16;
+}
+
+#[cfg(feature = "serde")]
 impl<'de> serde::Deserialize<'de> for CollationSpecialPrimaries<'de> {
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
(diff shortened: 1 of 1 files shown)


sffc:
This seems like a good change because it makes the code on main a little bit more parallel with the code on 2.1/2.2

Manishearth (approved):

Manishearth (approved):

components/collator/src/provider.rs
```diff
@@ -632,15 +641,15 @@ impl<'de> serde::Deserialize<'de> for CollationSpecialPrimaries<'de> {
             .as_ule_slice()
             // `variant_count` isn't stable yet:
             // https://github.com/rust-lang/rust/issues/73662
-            .split_at_checked(MaxVariable::Currency as usize + 1)
+            .split_at_checked(CollationSpecialPrimaries::NUM_PRIMARIES)
```
robertbastian, line 644: issue: the "`variant_count` isn't stable yet" comment should be moved to the constant. it would also make more sense for the constant to live on `MaxVariable`, and be called something like `VARIANT_COUNT`, or `fn variant_count()`.

components/collator/src/provider.rs
```diff
@@ -610,6 +610,15 @@ pub struct CollationSpecialPrimaries<'data> {
     pub compressible_bytes: ZeroVec<'data, u16>,
 }
 
+#[cfg(feature = "serde")]
+impl CollationSpecialPrimaries<'_> {
+    /// The number of real special primaries (Space, Punctuation, Symbol, Currency).
+    pub(crate) const NUM_PRIMARIES: usize = MaxVariable::Currency as usize + 1; // 4
```
robertbastian, line 616: I'd be happier with this way of variant counting if this lived next to the enum

components/collator/src/provider.rs
```diff
@@ -610,6 +610,15 @@ pub struct CollationSpecialPrimaries<'data> {
...
+
+    /// The length of the compressible bytes array (256 bits packed in 16 u16s).
+    pub(crate) const COMPRESSIBLE_BYTES_LEN: usize = 16;
```
robertbastian, line 619:
```suggestion
    /// The length of the compressible bytes array (256 bits packed in `u16`s).
    pub(crate) const COMPRESSIBLE_BYTES_LEN: usize = 256 / u16::BITS;
```

robertbastian: that said, the `/ u16::BITS` could also be moved to the use site, with this just being `256`, the number of bytes that are compressible, without worrying about how they are packed

PR #8084, opened 2026-06-16
Fix variant casing in display names attributes

sffc:
This PR implements support for key and attribute transforms in the `impl_displaynames_v1!` and `impl_displaynames_iter_v1!` macros, and uses it to normalize variant keys to lowercase in the `LocaleNamesVariantMediumV1` provider.

This is a standalone PR containing the data casing fixes split out from the main DisplayNames Phase 2 implementation.

### Key Changes
*   **Macro Overloads**: Overloaded `impl_displaynames_v1!` and `impl_displaynames_iter_v1!` in `provider/source/src/displaynames/mod.rs` to support optional loader and iterator key/attribute transforms.
*   **Double-Check Validation**: Implemented a double-check validation pattern in the iterator macro to construct the `Box<DataMarkerAttributes>` safely without cloning in the success path.
*   **Variant Normalization**: Refactored the `LocaleNamesVariantMediumV1` provider in `provider/source/src/displaynames/variant.rs` to use these transforms, normalizing the CLDR uppercase keys (e.g., `POSIX`) to BCP-47 lowercase (e.g., `posix`).
*   **Testdata & Baked Data**: Regenerated the JSON testdata (renaming `POSIX/` to `posix/` under `data/debug/`) and the baked Rust data to reflect the new lowercase variant key.

Addresses #7824 and #7825.

🤖 This pull request was created by an AI agent working with @sffc.

Changes not in changelog:

`icu_provider_source`:
    *   Updated `LocaleNamesVariantMediumV1` datagen to normalize variant keys to lowercase and perform case-insensitive lookups.

## Changelog

N/A (datagen only)

PR #8085, opened 2026-06-16
Implement VariantDisplayName and refactor single module

sffc:
Depends on #8084

This PR implements the single formatter for variant display names (`VariantDisplayName` and `VariantDisplayNameOwned`) and refactors the `single` module structure.

I chose to do the refactor in this PR because adding the third type made the module worth splitting. If there is an ounce of controversy about this I am happy to revert the module splitting and move it to another PR.

This PR is stacked and depends on the data fixes in #8084.

### Key Changes
*   **Variant API**: Implemented `VariantDisplayName` and `VariantDisplayNameOwned` in `components/experimental/src/displaynames/single/variant.rs` supporting the `Medium` width.
*   **Module Restructuring**: Refactored the `single` module to use the **`single/mod.rs`** structure, moving all single formatter files inside the `single/` directory.
*   **Split Region & Script**: Split `singular.rs` into `single/region.rs` and `single/script.rs` to keep the directory symmetric and clean, eliminating `singular.rs` entirely.
*   **Shared Helpers**: Moved generic loader helper functions (`try_new_unstable` and `try_new_short_unstable`) into `single/mod.rs` as `pub(crate)` and updated their error messages to be generic (`"Invalid subtag"`).
*   **Isolated Scope**: Only `Variant`, `Region`, and `Script` are re-exported from `single/mod.rs`. `Language` display names are excluded and will be introduced in a subsequent stacked PR.

Addresses #7824 and #7825.

## Changelog

icu_experimental/displaynames: New types `VariantDisplayName` and `VariantDisplayNameOwned`


Manishearth (approved):
Fetched locally and reviewed.

For a hand-done PR I would be fine with this, but for something agent-driven I think I would have preferred the "key changes" to each be a separate commit.

It worked out since I was reviewing locally anyway, so I wasn't benefiting much from GitHub's review, and needed to do local comparisons _anyway_.

One thing I noted: the variant code doesn't have as extensive ctor docs as the script/region code. This is fine, just a note.

components/experimental/src/displaynames/single/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+) -> Result<DataPayload<M>, DataError>
+where
+    M: DataMarker<DataStruct = VarZeroCow<'static, str>>,
+    D: DataProvider<M> + ?Sized,
+{
+    let locale = M::make_locale(prefs.locale_preferences);
+    let payload = provider
+        .load(DataRequest {
+            id: DataIdentifierBorrowed::for_marker_attributes_and_locale(
+                DataMarkerAttributes::try_from_str(attr)
```
robertbastian: issue: if you pass `DataMarkerAttributes` instead of `&str` to this function, you can infallibly create the attributes at the call sites, because there you have the information that they conform to the naming rules.

the error you return here otoh is either unreachable, or not client-actionable

I would also suggest creating `LocaleNamesRegionMediumV1::make_attributes(Region)` functions, so that you don't duplicate, and potentially diverge, the subtag -> attribute code between runtime and datagen.

components/experimental/src/displaynames/single/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+//! Currently, this module has limited support. It supports regions, scripts,
+//! and variants, but support for languages and locales is currently missing.
+//! More features are on their way.
+//!
+//! If you have any feedback, please let us know at
+//! <https://github.com/unicode-org/icu4x/issues/7825>.
+//!
+//! See [the parent module](mod@super) for a comparison of single and multi.
+
+mod region;
```
robertbastian, line 21: suggestion: it seems to me that the code in `region`/`script`/`variant` is pretty much identical and could be macro generated.

components/experimental/src/displaynames/single/mod.rs
sffc: This is a good suggestion, and I would appreciate in the future if you would label it as "issue (pre-existing)" to emphasize that it is not a regression introduced by this pull request. It is minor so I will fix it here instead of opening a follow-up.

components/experimental/src/displaynames/single/mod.rs
sffc: I had previously pulled out `try_new_unstable` and `try_new_short_unstable` to reduce the amount of duplicated code. I'll look for more opportunities, short of shoving everything into a macro.

components/experimental/src/displaynames/provider.rs
```diff
@@ -279,3 +280,43 @@ icu_provider::data_marker!(
     "locale/names/essentials/v1",
     LocaleNamesEssentials<'static>
 );
+
+impl LocaleNamesRegionMediumV1 {
+    /// Helper to create data marker attributes from a region.
+    #[inline]
+    pub fn make_attributes(region: &Region) -> &DataMarkerAttributes {
```
sffc: Since these are in the `provider` module, I think they don't need to be explicitly tagged as unstable?

EDIT: since we don't currently use these out-of-crate, I'll mark them pub(crate)

Manishearth (approved):

components/experimental/src/displaynames/provider.rs
```diff
@@ -283,6 +283,9 @@ icu_provider::data_marker!(
 
 impl LocaleNamesRegionMediumV1 {
     /// Helper to create data marker attributes from a region.
+    ///
+    /// This is infallible (will not panic) because a validated `Region` is guaranteed to
```
robertbastian, line 287: this should not be a doc comment, it should be a code comment on the call to `from_str_or_panic

components/experimental/src/displaynames/single/mod.rs
robertbastian: it's not preexisting though. if you don't deduplicate code between the subtags, I will consider it new code

components/experimental/src/displaynames/provider.rs
sffc: this is nitpicky but I don't disagree, ok

components/experimental/src/displaynames/single/mod.rs
sffc: This fn existed before, shared between Region and Script. This PR moves it to a new file and uses it also in Variant.

components/experimental/src/displaynames/provider.rs
sffc: Fixing this in the next PR

https://github.com/unicode-org/icu4x/pull/8082/commits/673af438e37613ec07001a811822ef51c6d8ee56

PR #8088, opened 2026-06-16
More accurate `f64` unit conversions

robertbastian:
Document `f64` rounding semantics, and improve accuracy. Tested against the `Ratio<BigInt>` code path.

#8058 

## Changelog

icu_experimental:
units:
    * document `f64` semantics and improve accuracy

### DIFF
--- a/components/experimental/src/units/convertible.rs
+++ b/components/experimental/src/units/convertible.rs
@@ -57,33 +57,136 @@ impl Convertible for &'_ Ratio<BigInt> {
     }
 }
 
+#[derive(Debug, Clone, Copy)]
+#[doc(hidden)]
+pub struct RatioF64 {
+    numerator: f64,
+    denominator: f64,
+}
+
 impl Convertible for f64 {
-    type Factor = f64;
+    type Factor = RatioF64;
     type Addend = f64;
     type Result = f64;
 
-    // TODO: reduce error
+    // This code performs up to 4 rounding operations:
+    // * factor numerator and denominator from BigInt to f64 (factor_from_ratio_bigint)
+    //   * this will be exact in most cases, but may be inexact for conversions involving
+    //     large SI prefixes
+    // * a single rounding for the multiplication and division
+    //   * if the intermediate error is not representable, it falls back to double-rounding
+    //     the multiplication and division
     fn mul(self, factor: &Self::Factor) -> Self::Result {
-        self * factor
+        let double_rounded = self * factor.numerator / factor.denominator;
+        // The multiplication error is the difference of the rounded multiplication
+        // and the multiplication evaluated in full precision (FMA)
+        let multiplication_error = self.mul_add(factor.numerator, -(self * factor.numerator));
+        // The total error is the difference between the rounded multiplication+division
+        // and the multiplication evaluated in full precision, plus the multiplication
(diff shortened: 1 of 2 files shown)


components/experimental/src/units/convertible.rs
```diff
@@ -57,33 +57,136 @@ impl Convertible for &'_ Ratio<BigInt> {
...
+}
+
 impl Convertible for f64 {
-    type Factor = f64;
+    type Factor = RatioF64;
     type Addend = f64;
     type Result = f64;
 
-    // TODO: reduce error
+    // This code performs up to 4 rounding operations:
```
eggrobin, line 72: I am not sure how you are counting things, but I have either five or six roundings involved in the final error (depending on whether you count the last one), namely:
* 1,2: the two in the evaluation of `double_rounded`, 
* ~~3: the one in the FMA of total_error (contrary to `multiplication_error` which is the classic exact TwoProduct, I see no reason that $fl(\mathtt{self} \times \mathtt{factor.numerator})-\mathtt{double\textunderscore rounded}\times \mathtt{factor.denominator}$  should be representable, there is no such thing as an exact TwoQuotient, so the FMA rounds, albeit once)~~ EDIT: Nevermind, that one is probably exact too, this is something like the low half of TwoProduct(d, dr).
* 4: the one from the sum of the result of that FMA and `multiplication_error`
* 5: the one from the division in `total_error`.

Of course I may well have screwed up the error analysis.

But more importantly, describing it that way is short-changing the computation, because in contrast with Convertible::mul_add below that is built on top of this one, and really just straightforwardly adds another roundoff to the pile, these five roundings contribute to the error on the error: they are _squared_ roundoffs. Obviously the last rounding does give you one roundoff (this is never going to be better than correctly rounded!), but the relative error bound I have on the result is $u+5u^2(1+u)^4$, where $u$ is the unit roundoff.
$u=2^{-53}$ in binary64 rounded to nearest (half the ULP of 1 above), so the error bound is a smidge above $(1+5\times 2^{-53})u$: this is not quite correctly rounded but it often will be, and it certainly is faithfully rounded.

components/experimental/src/units/convertible.rs
```diff
@@ -57,33 +57,136 @@ impl Convertible for &'_ Ratio<BigInt> {
...
+    // * factor numerator and denominator from BigInt to f64 (factor_from_ratio_bigint)
+    //   * this will be exact in most cases, but may be inexact for conversions involving
+    //     large SI prefixes
+    // * a single rounding for the multiplication and division
+    //   * if the intermediate error is not representable, it falls back to double-rounding
+    //     the multiplication and division
     fn mul(self, factor: &Self::Factor) -> Self::Result {
-        self * factor
+        let double_rounded = self * factor.numerator / factor.denominator;
```
eggrobin, line 80: `double_rounded` is an unfortunate term, [double rounding](https://en.wikipedia.org/wiki/Rounding#Double_rounding) being a term of art and not what is happening here.

sffc (commented):
Please move the impl to its own file, and pull any useful tests from #8058.

robertbastian:
I closed this PR last week, I'm not planning on working on this.

sffc:
ok, let's just continue discussion in https://github.com/unicode-org/icu4x/issues/8116.

PR #8090, opened 2026-06-16
Datagen: Use structured keys with alt and menu variants for display names

sffc:
This PR refactors the `icu_provider_source` crate to migrate CLDR Serde display names to a new structured key type `WithAlt<T>`.

This replaces fragile string parsing at the provider layer with type-safe deserialization at the Serde layer, and introduces clean support for `menu` variants.

### Key Changes
1.  **Structured Key (`WithAlt<T>`)**: Defined a new key type that wraps a subtag `T` and carries optional `alt` and `menu` fields.
2.  **Smart Deserialization**: Implemented a custom `Deserialize` for `WithAlt<T>` that uses `split_once` to safely strip and extract `"-menu-"` and `"-alt-"` suffixes before parsing the subtag `T`.
3.  **Typed Keys**: Migrated all display name Serde maps (languages, regions, scripts, variants) to use `WithAlt<T>` with their respective typed subtags (`LanguageIdentifier`, `Region`, `Script`, `Variant`) instead of raw `String`s.
4.  **Infallible Providers**: Simplified the provider `TryFrom` implementations to infallible `From` implementations since all validation and parsing now happen during deserialization.
5.  **Macro Updates**: Preserved and updated the `impl_displaynames_v1!`, `impl_displaynames_iter_v1!`, and `impl_displaynames_menu_v1!` macros to be generic over the subtag type `T`. Removed the fully qualified path from the macros and imported `WithAlt` at the call sites.
6.  **Timezone Provider Update**: Updated `time_zones/convert.rs` to use the new structured keys for `territories.json`, replacing string suffix stripping with clean field checks.

## Changelog

N/A

### DIFF
--- a/provider/source/src/cldr_serde/displaynames/mod.rs
+++ b/provider/source/src/cldr_serde/displaynames/mod.rs
@@ -12,13 +12,13 @@ use core::str::FromStr;
 use serde::{Deserialize, Deserializer};
 
 #[derive(Hash, Eq, PartialEq, Clone, Debug)]
-pub(crate) struct SubtagWithOptionalAltVariant<T> {
+pub(crate) struct ModifiedSubtag<T> {
     pub(crate) subtag: T,
     pub(crate) alt_variant: Option<String>,
     pub(crate) menu_variant: Option<String>,
 }
 
-impl<'de, T> Deserialize<'de> for SubtagWithOptionalAltVariant<T>
+impl<'de, T> Deserialize<'de> for ModifiedSubtag<T>
 where
     T: FromStr,
     T::Err: core::fmt::Display,
@@ -34,7 +34,7 @@ where
             T: FromStr,
             T::Err: core::fmt::Display,
         {
-            type Value = SubtagWithOptionalAltVariant<T>;
+            type Value = ModifiedSubtag<T>;
 
             fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
                 formatter.write_str("a string with optional -alt- or -menu- suffix")
@@ -48,7 +48,7 @@ where
                     let (subtag_str, menu_str) = v.split_at(index);
                     let menu_variant = menu_str.strip_prefix("-menu-").unwrap().to_string();
                     let subtag = T::from_str(subtag_str).map_err(E::custom)?;
-                    Ok(SubtagWithOptionalAltVariant {
+                    Ok(ModifiedSubtag {
                         subtag,
                         alt_variant: None,
                         menu_variant: Some(menu_variant),
@@ -57,14 +57,14 @@ where
                     let (subtag_str, alt_str) = v.split_at(index);
                     let alt_variant = alt_str.strip_prefix("-alt-").unwrap().to_string();
                     let subtag = T::from_str(subtag_str).map_err(E::custom)?;
-                    Ok(SubtagWithOptionalAltVariant {
+                    Ok(ModifiedSubtag {
                         subtag,
                         alt_variant: Some(alt_variant),
                         menu_variant: None,
                     })
                 } else {
                     let subtag = T::from_str(v).map_err(E::custom)?;
-                    Ok(SubtagWithOptionalAltVariant {
+                    Ok(ModifiedSubtag {
                         subtag,
                         alt_variant: None,
                         menu_variant: None,
--- a/provider/source/src/cldr_serde/displaynames/mod.rs
+++ b/provider/source/src/cldr_serde/displaynames/mod.rs
@@ -12,12 +12,61 @@ const ALT_SEPARATOR: &str = "-alt-";
 const MENU_SEPARATOR: &str = "-menu-";
 use core::str::FromStr;
 use serde::{Deserialize, Deserializer};
+#[derive(Hash, Eq, PartialEq, Clone, Copy, Debug)]
+pub(crate) enum Alt {
+    Short,
+    Long,
+    Variant,
+    StandAlone,
+    Official,
+    Secondary,
+    Biot,
+    Chagos,
+    Menu,
+    Unknown,
+}
+
+impl FromStr for Alt {
+    type Err = ();
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "short" => Ok(Alt::Short),
+            "long" => Ok(Alt::Long),
+            "variant" => Ok(Alt::Variant),
+            "stand-alone" => Ok(Alt::StandAlone),
+            "official" => Ok(Alt::Official),
+            "secondary" => Ok(Alt::Secondary),
+            "biot" => Ok(Alt::Biot),
+            "chagos" => Ok(Alt::Chagos),
+            "menu" => Ok(Alt::Menu),
+            _ => Err(()),
+        }
+    }
+}
+
+#[derive(Hash, Eq, PartialEq, Clone, Copy, Debug)]
+pub(crate) enum Menu {
+    Core,
+    Extension,
+    Unknown,
+}
+
+impl FromStr for Menu {
+    type Err = ();
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "core" => Ok(Menu::Core),
+            "extension" => Ok(Menu::Extension),
+            _ => Err(()),
+        }
+    }
+}
 
 #[derive(Hash, Eq, PartialEq, Clone, Debug)]
 pub(crate) struct WithAlt<T> {
     pub(crate) subtag: T,
-    pub(crate) alt: Option<std::borrow::Cow<'static, str>>,
-    pub(crate) menu: Option<std::borrow::Cow<'static, str>>,
+    pub(crate) alt: Option<Alt>,
+    pub(crate) menu: Option<Menu>,
 }
 
 impl<'de, T> Deserialize<'de> for WithAlt<T>
@@ -51,13 +100,13 @@ where
                     Ok(WithAlt {
                         subtag,
                         alt: None,
-                        menu: Some(std::borrow::Cow::Owned(menu_str.to_string())),
+                        menu: Some(Menu::from_str(menu_str).unwrap_or(Menu::Unknown)),
                     })
                 } else if let Some((subtag_str, alt_str)) = v.split_once(ALT_SEPARATOR) {
                     let subtag = T::from_str(subtag_str).map_err(E::custom)?;
                     Ok(WithAlt {
                         subtag,
-                        alt: Some(std::borrow::Cow::Owned(alt_str.to_string())),
+                        alt: Some(Alt::from_str(alt_str).unwrap_or(Alt::Unknown)),
                         menu: None,
                     })
                 } else {
--- a/provider/source/src/cldr_serde/displaynames/mod.rs
+++ b/provider/source/src/cldr_serde/displaynames/mod.rs
@@ -8,7 +8,8 @@ pub(crate) mod region;
 pub(crate) mod script;
 pub(crate) mod variant;
 
-use crate::displaynames::{ALT_SEPARATOR, MENU_SEPARATOR};
+const ALT_SEPARATOR: &str = "-alt-";
+const MENU_SEPARATOR: &str = "-menu-";
 use core::str::FromStr;
 use serde::{Deserialize, Deserializer};
 
--- a/provider/source/src/cldr_serde/displaynames/mod.rs
+++ b/provider/source/src/cldr_serde/displaynames/mod.rs
@@ -16,8 +16,8 @@ use serde::{Deserialize, Deserializer};
 #[derive(Hash, Eq, PartialEq, Clone, Debug)]
 pub(crate) struct WithAlt<T> {
     pub(crate) subtag: T,
-    pub(crate) alt: Option<String>,
-    pub(crate) menu: Option<String>,
+    pub(crate) alt: Option<std::borrow::Cow<'static, str>>,
+    pub(crate) menu: Option<std::borrow::Cow<'static, str>>,
 }
 
 impl<'de, T> Deserialize<'de> for WithAlt<T>
@@ -51,13 +51,13 @@ where
                     Ok(WithAlt {
                         subtag,
                         alt: None,
-                        menu: Some(menu_str.to_string()),
+                        menu: Some(std::borrow::Cow::Owned(menu_str.to_string())),
                     })
                 } else if let Some((subtag_str, alt_str)) = v.split_once(ALT_SEPARATOR) {
                     let subtag = T::from_str(subtag_str).map_err(E::custom)?;
                     Ok(WithAlt {
                         subtag,
-                        alt: Some(alt_str.to_string()),
+                        alt: Some(std::borrow::Cow::Owned(alt_str.to_string())),
                         menu: None,
                     })
                 } else {
--- a/provider/source/src/cldr_serde/displaynames/mod.rs
+++ b/provider/source/src/cldr_serde/displaynames/mod.rs
@@ -19,9 +19,13 @@ pub(crate) enum Alt {
     Variant,
     StandAlone,
     Official,
+    /// Secondary name variant, used in languages and scripts.
     Secondary,
+    /// Abbreviation for territory code `IO` (British Indian Ocean Territory).
     Biot,
+    /// Alternate name for territory code `IO` (British Indian Ocean Territory) mapping to "Chagos Archipelago".
     Chagos,
+    /// "menu" variant, which is being replaced by menu=core|extension, but is still in CLDR.
     Menu,
     Unknown,
 }
--- a/provider/source/src/cldr_serde/displaynames/mod.rs
+++ b/provider/source/src/cldr_serde/displaynames/mod.rs
@@ -10,6 +10,8 @@ pub(crate) mod variant;
 
 const ALT_SEPARATOR: &str = "-alt-";
 const MENU_SEPARATOR: &str = "-menu-";
+use core::fmt::{self, Display};
+use core::marker::PhantomData;
 use core::str::FromStr;
 use serde::{Deserialize, Deserializer};
 #[derive(Hash, Eq, PartialEq, Clone, Copy, Debug)]
@@ -76,22 +78,22 @@ pub(crate) struct WithAlt<T> {
 impl<'de, T> Deserialize<'de> for WithAlt<T>
 where
     T: FromStr,
-    T::Err: core::fmt::Display,
+    T::Err: Display,
 {
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
         D: Deserializer<'de>,
     {
-        struct Visitor<T>(std::marker::PhantomData<T>);
+        struct Visitor<T>(PhantomData<T>);
 
         impl<'de, T> serde::de::Visitor<'de> for Visitor<T>
         where
             T: FromStr,
-            T::Err: core::fmt::Display,
+            T::Err: Display,
         {
             type Value = WithAlt<T>;
 
-            fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
                 formatter.write_str("a string with optional -alt- or -menu- suffix")
             }
 
@@ -138,6 +140,6 @@ where
             }
         }
 
-        deserializer.deserialize_str(Visitor(std::marker::PhantomData))
+        deserializer.deserialize_str(Visitor(PhantomData))
     }
 }
--- a/provider/source/src/cldr_serde/displaynames/mod.rs
+++ b/provider/source/src/cldr_serde/displaynames/mod.rs
@@ -14,6 +14,7 @@ use core::str::FromStr;
 use serde::{Deserialize, Deserializer};
 #[derive(Hash, Eq, PartialEq, Clone, Copy, Debug)]
 pub(crate) enum Alt {
+    Unknown,
     Short,
     Long,
     Variant,
@@ -27,7 +28,6 @@ pub(crate) enum Alt {
     Chagos,
     /// "menu" variant, which is being replaced by menu=core|extension, but is still in CLDR.
     Menu,
-    Unknown,
 }
 
 impl FromStr for Alt {
@@ -50,9 +50,9 @@ impl FromStr for Alt {
 
 #[derive(Hash, Eq, PartialEq, Clone, Copy, Debug)]
 pub(crate) enum Menu {
+    Unknown,
     Core,
     Extension,
-    Unknown,
 }
 
 impl FromStr for Menu {
@@ -101,16 +101,30 @@ where
             {
                 if let Some((subtag_str, menu_str)) = v.split_once(MENU_SEPARATOR) {
                     let subtag = T::from_str(subtag_str).map_err(E::custom)?;
+                    let menu = match Menu::from_str(menu_str) {
+                        Ok(m) => m,
+                        Err(_) => {
+                            log::warn!("Unknown menu variant: {}", menu_str);
+                            Menu::Unknown
+                        }
+                    };
                     Ok(WithAlt {
                         subtag,
                         alt: None,
-                        menu: Some(Menu::from_str(menu_str).unwrap_or(Menu::Unknown)),
+                        menu: Some(menu),
                     })
                 } else if let Some((subtag_str, alt_str)) = v.split_once(ALT_SEPARATOR) {
                     let subtag = T::from_str(subtag_str).map_err(E::custom)?;
+                    let alt = match Alt::from_str(alt_str) {
+                        Ok(a) => a,
+                        Err(_) => {
+                            log::warn!("Unknown alt variant: {}", alt_str);
+                            Alt::Unknown
+                        }
+                    };
                     Ok(WithAlt {
                         subtag,
-                        alt: Some(Alt::from_str(alt_str).unwrap_or(Alt::Unknown)),
+                        alt: Some(alt),
                         menu: None,
                     })
                 } else {
--- a/provider/source/src/cldr_serde/displaynames/mod.rs
+++ b/provider/source/src/cldr_serde/displaynames/mod.rs
@@ -7,3 +7,72 @@ pub(crate) mod locale_display_pattern;
 pub(crate) mod region;
 pub(crate) mod script;
 pub(crate) mod variant;
+
+use core::str::FromStr;
+use serde::{Deserialize, Deserializer};
+
+#[derive(Hash, Eq, PartialEq, Clone, Debug)]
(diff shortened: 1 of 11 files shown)


provider/source/src/cldr_serde/displaynames/variant.rs
```diff
@@ -22,3 +23,14 @@ pub(crate) struct LangDisplayNames {
...
 pub(crate) type Resource = super::super::LocaleResource<LangDisplayNames>;
+
+fn deserialize_normalized_keys<'de, D>(deserializer: D) -> Result<HashMap<String, String>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let map = HashMap::<String, String>::deserialize(deserializer)?;
+    Ok(map
+        .into_iter()
+        .map(|(k, v)| (k.to_ascii_lowercase(), v))
```
sffc: 🤖 This comment is written by an AI agent working with @sffc.

We cannot simply use `Variant::try_from_str` here because some keys in CLDR might contain alt variant suffixes (e.g., `-alt-variant`), which are not valid BCP-47 variants as a whole and would fail to parse, causing deserialization to fail.

If we wanted to use `Variant` for normalization, we would need to split the key, normalize the variant part, and reconstruct it. Here is how we could handle it:

```rust
fn normalize_variant_key(k: &str) -> String {
    if let Some((var, alt)) = k.split_once("-alt-") {
        if let Ok(variant) = Variant::try_from_str(var) {
            format!("{}-alt-{}", variant.as_str(), alt)
        } else {
            k.to_ascii_lowercase()
        }
    } else {
        if let Ok(variant) = Variant::try_from_str(k) {
            variant.as_str().to_string()
        } else {
            k.to_ascii_lowercase()
        }
    }
}
```

Since `to_ascii_lowercase()` is a safe fallback and is sufficient for all valid variants (which are just ASCII alphanumeric), we are sticking with it for simplicity.

robertbastian: `Variants` should work though? `POSIX-alt-foo` parses as a `Variants` and normalizes to `posix-alt-foo`

robertbastian (commented):
.

provider/source/src/cldr_serde/displaynames/variant.rs
sffc: I think it's a coincidence that `POSIX-alt-foo` parses as a `Variants`. The `-alt-foo` could be `-alt-x` (too short) or `-alt-abbreviated` (too long).

Given these options:

1. Apply `to_ascii_lowercase()` to the whole string (current PR)
2. Parse the first subtag as `Variant` and apply `to_ascii_lowercase()` to the rest of the string (the suggestion above in this thread, with tweaks)
3. Parse the whole string as `Variants`

my preference order is: 1 ~> 2 ~>> 3

Wdyt?

Manishearth:
Approach seems better, will let Robert review since it's his feedback

provider/source/src/cldr_serde/displaynames/variant.rs
robertbastian: it should be 2. we also need to error if it doesn't parse as a variant due to length, or because it contains non-ascii characters. you probably even want to store `(Variant, String)` as the key, if you want to use the `Variant -> DataMarkerAttributes` constructor later

robertbastian (commented):
.

provider/source/src/displaynames/mod.rs
```diff
@@ -8,54 +8,17 @@ pub(crate) mod region;
 pub(crate) mod script;
 pub(crate) mod variant;
 
-pub(crate) const ALT_SUBSTRING: &str = "-alt-";
-pub(crate) const ALT_SHORT_SUBSTRING: &str = "-alt-short";
-pub(crate) const ALT_LONG_SUBSTRING: &str = "-alt-long";
```
sffc, line 13: Restore the centralized constants. You might need to change their names and values, but they should stay centralized.

provider/source/src/displaynames/mod.rs
```diff
@@ -157,34 +143,26 @@ macro_rules! impl_displaynames_menu_v1 {
                 let mut result = HashSet::new();
                 let displaynames = self.cldr()?.displaynames();
                 for locale in displaynames.list_locales()?.filter(|locale| {
-                    // The directory might exist without the file
```
sffc, line 160: Don't delete comments, here and elsewhere

provider/source/src/displaynames/mod.rs
```diff
@@ -118,10 +87,23 @@ macro_rules! impl_displaynames_menu_v1 {
...
-                let mut key_extension = key_core.clone();
-                key_core.push_str($crate::displaynames::MENU_CORE_SUBSTRING);
-                key_extension.push_str($crate::displaynames::MENU_EXTENSION_SUBSTRING);
+                let subtag =
+                    <$subtag_ty as core::str::FromStr>::from_str(req.id.marker_attributes.as_str())
+                        .map_err(|_| {
+                            DataError::custom("failed to parse subtag").with_req($marker::INFO, req)
+                        })?;
+
+                let key_core = $crate::cldr_serde::displaynames::SubtagWithOptionalAltVariant {
```
sffc: nit: import the types at the call sites of this macro instead of fully qualifying them

provider/source/src/displaynames/language.rs
```diff
@@ -140,59 +147,52 @@ impl From<&cldr_serde::displaynames::language::Resource> for LanguageDisplayName
...
-        let mut short_names = BTreeMap::new();
-        let mut long_names = BTreeMap::new();
-        let mut menu_names = BTreeMap::new();
+        let mut names = ZeroMap::new();
+        let mut short_names = ZeroMap::new();
+        let mut long_names = ZeroMap::new();
+        let mut menu_names = ZeroMap::new();
         for (key, value) in other.main.value.localedisplaynames.languages.iter() {
-            if key.contains("-menu-") {
-                // TODO: handle -menu-core and -menu-extension
```
sffc, line 149: another comment, please don't delete

provider/source/src/cldr_serde/displaynames/mod.rs
```diff
@@ -7,3 +7,72 @@ pub(crate) mod locale_display_pattern;
 pub(crate) mod region;
 pub(crate) mod script;
 pub(crate) mod variant;
+
+use core::str::FromStr;
+use serde::{Deserialize, Deserializer};
+
+#[derive(Hash, Eq, PartialEq, Clone, Debug)]
+pub(crate) struct SubtagWithOptionalAltVariant<T> {
```
sffc: suggestion: feel free to pick a better name for this type

sffc (commented):
Please update the PR description and title with the new scope

provider/source/src/cldr_serde/displaynames/variant.rs
sffc: Context: I tried this for Variant but it was easier to do it for all of the display names data. I think it makes the code a lot cleaner, especially the parts where we have to match on alt variants.

provider/source/src/cldr_serde/displaynames/mod.rs
```diff
@@ -7,3 +7,73 @@ pub(crate) mod locale_display_pattern;
 pub(crate) mod region;
 pub(crate) mod script;
 pub(crate) mod variant;
+
+use crate::displaynames::{ALT_SEPARATOR, MENU_SEPARATOR};
+use core::str::FromStr;
+use serde::{Deserialize, Deserializer};
+
+#[derive(Hash, Eq, PartialEq, Clone, Debug)]
+pub(crate) struct ModifiedSubtag<T> {
```
robertbastian: not a great name, modified in what way? `WithAlt`?

provider/source/src/cldr_serde/displaynames/mod.rs
```diff
@@ -7,3 +7,73 @@ pub(crate) mod locale_display_pattern;
...
+                formatter.write_str("a string with optional -alt- or -menu- suffix")
+            }
+
+            fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+            where
+                E: serde::de::Error,
+            {
+                if let Some(index) = v.rfind(MENU_SEPARATOR) {
+                    let (subtag_str, menu_str) = v.split_at(index);
+                    let menu_variant = menu_str.strip_prefix(MENU_SEPARATOR).unwrap().to_string();
```
robertbastian:
```suggestion
                if let Some((subtag_str, menu_str)) = v.split_once(MENU_SEPARATOR) {
                    let menu_variant = menu_str.to_string();
```

provider/source/src/cldr_serde/displaynames/mod.rs
```diff
@@ -7,3 +7,73 @@ pub(crate) mod locale_display_pattern;
...
 pub(crate) mod variant;
+
+use crate::displaynames::{ALT_SEPARATOR, MENU_SEPARATOR};
+use core::str::FromStr;
+use serde::{Deserialize, Deserializer};
+
+#[derive(Hash, Eq, PartialEq, Clone, Debug)]
+pub(crate) struct ModifiedSubtag<T> {
+    pub(crate) subtag: T,
+    pub(crate) alt_variant: Option<String>,
```
robertbastian: `alt_variant` sounds like it's a value, but it's a key

```suggestion
    pub(crate) alt_key: Option<String>,
```

provider/source/src/displaynames/language.rs
```diff
@@ -89,27 +92,35 @@ impl From<&cldr_serde::displaynames::language::Resource> for LanguageDisplayName
...
+
+            let langid = &key.subtag;
+            if langid.script.is_some() || langid.region.is_some() || !langid.variants.is_empty() {
+                continue;
+            }
+            let lang = langid.language;
+
+            match key.alt_variant.as_deref() {
+                Some(ALT_SHORT) => {
                     short_names.insert(lang.to_tinystr(), value.as_ref());
```
robertbastian: nit: while you're here

```suggestion
                    short_names.insert(lang, value.as_ref());
```

and then do the `.to_tinystr()` with the `.to_unvalidated()` later

provider/source/src/displaynames/language.rs
```diff
@@ -140,59 +151,54 @@ impl From<&cldr_serde::displaynames::language::Resource> for LanguageDisplayName
 
 impl From<&cldr_serde::displaynames::language::Resource> for LocaleDisplayNames<'static> {
     fn from(other: &cldr_serde::displaynames::language::Resource) -> Self {
-        let mut names = BTreeMap::new();
-        let mut short_names = BTreeMap::new();
-        let mut long_names = BTreeMap::new();
-        let mut menu_names = BTreeMap::new();
+        let mut names = ZeroMap::new();
```
robertbastian: issue: this is pretty much the same code as `impl From<&Resource> for LanguageDisplayNames<'static>`, but written very differently

provider/source/src/cldr_serde/displaynames/mod.rs
sffc: it's "long", "short", "variant". Those aren't really keys

sffc:
I'm clocked out for the evening. Did only a cursory look at the latest AI commits.

provider/source/src/cldr_serde/displaynames/mod.rs
robertbastian: well they're definitely not values

sffc: (I changed the field name to simply `alt`)

robertbastian (commented):
I'll review this when it passes CI

provider/source/src/displaynames/language.rs
```diff
@@ -61,138 +61,146 @@ crate::displaynames::impl_displaynames_v1!(
...
+            if langid.script.is_some() || langid.region.is_some() || !langid.variants.is_empty() {
+                None
+            } else {
+                let lang = langid.language;
+                if lang.as_str() == val {
+                    None
+                } else {
+                    Some(lang)
+                }
+            }
```
robertbastian:
```suggestion
            if langid.script.is_some() || langid.region.is_some() || !langid.variants.is_empty() || langid.language.as_str() == val {
                None
            } else {
                Some(langid.language)
            }
```

robertbastian:
```suggestion
            (langid.script.is_none() && langid.region.is_none() && langid.variants.is_empty() && langid.language != val).then_some(langid.language)
```

provider/source/src/displaynames/language.rs
robertbastian: this has not been restored

provider/source/src/displaynames/language.rs
```diff
@@ -61,138 +61,146 @@ crate::displaynames::impl_displaynames_v1!(
...
+        });
+
+        let to_zero_map = |map: BTreeMap<icu::locale::subtags::Language, &str>| {
+            map.into_iter()
+                .map(|(k, v)| (k.to_tinystr().to_unvalidated(), v))
+                .collect()
+        };
+
         Self {
-            // Old CLDR versions may contain trivial entries, so filter
```
robertbastian, line 116: also preserve this comment

provider/source/src/displaynames/language.rs
```diff
@@ -61,138 +61,146 @@ crate::displaynames::impl_displaynames_v1!(
...
+    let mut names = BTreeMap::new();
+    let mut short_names = BTreeMap::new();
+    let mut long_names = BTreeMap::new();
+    let mut menu_names = BTreeMap::new();
+    for (key, value) in resource.main.value.localedisplaynames.languages.iter() {
+        if key.menu.is_some() {
+            continue;
+        }
+        let val_str = value.as_str();
+        if let Some(k) = filter_project(&key.subtag, val_str) {
```
robertbastian: if you're passing `val_str` just so you can filter ("Old CLDR versions may contain trivial entries"), is that still necessary?

in any case, I think the filter should happen outside of `filter_project`


```suggestion
        if let Some(k) = filter_project(&key.subtag) && k != val_str {
```

provider/source/src/displaynames/language.rs
```diff
@@ -61,138 +61,146 @@ crate::displaynames::impl_displaynames_v1!(
...
+    }
+    ExtractedNames {
+        names,
+        short_names,
+        long_names,
+        menu_names,
+    }
+}
+
+impl From<&cldr_serde::displaynames::language::Resource> for LanguageDisplayNames<'static> {
```
robertbastian, line 88: these three impls differ only in the `extract_names` `filter_project`, can you move them into the `impl_displaynames_v1` macro?

provider/source/src/displaynames/region.rs
```diff
@@ -83,10 +86,9 @@ impl TryFrom<&cldr_serde::displaynames::region::Resource> for RegionDisplayNames
                 .filter(|&(k, v)| k != v)
```
robertbastian: nit: here the filter still happens later

robertbastian (commented):
this is all messier than necessary. imho this is adding tech debt because it handles language and region/script/variant pretty differently, but we have todos for region/script/variant to handle alt variants as well, so they should really be handled the same.

sffc:
I'll do a more thorough human pass before bothering @robertbastian again.

provider/source/src/displaynames/language.rs
```diff
@@ -61,138 +61,103 @@ crate::displaynames::impl_displaynames_v1!(
...
+            &other.main.value.localedisplaynames.languages,
+            &[ALT_VARIANT, ALT_SECONDARY, ALT_OFFICIAL],
+            "language",
+            |langid, val| {
+                if langid.script.is_none() && langid.region.is_none() && langid.variants.is_empty()
+                {
+                    None
+                } else {
+                    let locale_str = langid.to_string();
+                    if locale_str == val {
```
sffc: use writeable cmp_bytes

provider/source/src/cldr_serde/displaynames/mod.rs
```diff
@@ -7,3 +7,70 @@ pub(crate) mod locale_display_pattern;
...
+const MENU_SEPARATOR: &str = "-menu-";
+use core::str::FromStr;
+use serde::{Deserialize, Deserializer};
+
+#[derive(Hash, Eq, PartialEq, Clone, Debug)]
+pub(crate) struct WithAlt<T> {
+    pub(crate) subtag: T,
+    pub(crate) alt: Option<String>,
+    pub(crate) menu: Option<String>,
+}
```
sffc, line 76: hmm, make this contain Cow<str> instead of String so that when we do lookups we don't need to allocate

provider/source/src/displaynames/language.rs
sffc: Nevermind; we need the String in the next step. We eventually need it as a String because it becomes the key of a ZeroMap.

provider/source/src/displaynames/language.rs
sffc: I restored it again

provider/source/src/displaynames/language.rs
sffc: done

sffc (commented):
ok I'm reasonably happy with this. It is a bigger diff than I usually want, but it's all in datagen, and it's kind-of what happens when I change the types of things used in so many places.

provider/source/src/displaynames/language.rs
sffc: It's already been simplified a lot from what was there before. I don't really want to move the zeromap-based structs into `impl_displaynames_v1`, which generates the attributes-based structs.

provider/source/src/displaynames/region.rs
sffc: looks like @sffc-bot fixed this

provider/source/src/displaynames/language.rs
sffc: done

provider/source/src/displaynames/language.rs
sffc: I find `bool::then_some` to be harder to read than the if statement

sffc:
> this is all messier than necessary. imho this is adding tech debt because it handles language and region/script/variant pretty differently, but we have todos for region/script/variant to handle alt variants as well, so they should really be handled the same.

Yeah, so, the main deliverable of this PR is changing the cldr_serde structs to be nicely typed instead of Strings. This touches a lot of places in the code.

On top of this, based on [this comment](https://github.com/unicode-org/icu4x/pull/8090#discussion_r3441059483), the agent decided to refactor more logic out of the five zeromap struct impls into a helper function. This seems like a good thing, and something I would expect you would quite like, since it deduplicates code. I'm surprised by your reaction that the change to refactor those impls into the helper fn adds tech debt.

Regardless, the zeromap-based structs will hopefully be sunset, so I don't have a very big appetite to continue to refactor their impls.

sffc:
I think this PR has made the impls converge in such a way that it is easier to see what is unique about each one, and most of the logic lives in the helper fn `extract_names_for_zeromap_struct`.

```rust
impl From<&cldr_serde::displaynames::language::Resource> for LanguageDisplayNames<'static> {
    fn from(other: &cldr_serde::displaynames::language::Resource) -> Self {
        let extracted = extract_names_for_zeromap_struct(
            &other.main.value.localedisplaynames.languages,
            &[ALT_VARIANT, ALT_SECONDARY, ALT_OFFICIAL],
            "language",
            |langid| {
                if langid.script.is_some() || langid.region.is_some() || !langid.variants.is_empty()
                {
                    None
                } else {
                    Some(langid.language.to_tinystr())
                }
            },
        );

        let to_zero_map = |map: BTreeMap<TinyAsciiStr<3>, &str>| {
            map.into_iter()
                .map(|(k, v)| (k.to_unvalidated(), v))
                .collect()
        };

        Self {
            names: to_zero_map(extracted.names),
            short_names: to_zero_map(extracted.short_names),
            long_names: to_zero_map(extracted.long_names),
            menu_names: to_zero_map(extracted.menu_names),
        }
    }
}

impl From<&cldr_serde::displaynames::region::Resource> for RegionDisplayNames<'static> {
    fn from(other: &cldr_serde::displaynames::region::Resource) -> Self {
        let extracted = extract_names_for_zeromap_struct(
            &other.main.value.localedisplaynames.regions,
            &[ALT_VARIANT, ALT_CHAGOS, ALT_BIOT],
            "region",
            |region| Some(region.to_tinystr()),
        );

        let to_zero_map = |map: BTreeMap<tinystr::TinyAsciiStr<3>, &str>| {
            map.into_iter()
                .map(|(k, v)| (k.to_unvalidated(), v))
                .collect()
        };

        Self {
            names: to_zero_map(extracted.names),
            short_names: to_zero_map(extracted.short_names),
        }
    }
}
```

provider/source/src/cldr_serde/displaynames/mod.rs
```diff
@@ -16,8 +16,8 @@ use serde::{Deserialize, Deserializer};
 #[derive(Hash, Eq, PartialEq, Clone, Debug)]
 pub(crate) struct WithAlt<T> {
     pub(crate) subtag: T,
-    pub(crate) alt: Option<String>,
-    pub(crate) menu: Option<String>,
+    pub(crate) alt: Option<std::borrow::Cow<'static, str>>,
```
robertbastian: why are you not importing `Cow`? you use it four times fully qualified in this file, and more in other files

provider/source/src/cldr_serde/displaynames/mod.rs
```diff
@@ -51,13 +51,13 @@ where
                     Ok(WithAlt {
                         subtag,
                         alt: None,
-                        menu: Some(menu_str.to_string()),
+                        menu: Some(std::borrow::Cow::Owned(menu_str.to_string())),
```
robertbastian: nit: leave a comment why this doesn't borrow (wrong lifetime)

provider/source/src/displaynames/language.rs
```diff
@@ -61,138 +62,96 @@ crate::displaynames::impl_displaynames_v1!(
...
+            },
+        );
+
+        let to_zero_map = |map: BTreeMap<String, &str>| {
+            let mut zero_map = ZeroMap::new();
+            for (k, v) in map.into_iter() {
+                zero_map.insert(PotentialUtf8::from_str(&k), v);
             }
-        }
+            zero_map
```
robertbastian: why not iter+map? you still do that everywhere else

provider/source/src/displaynames/mod.rs
```diff
@@ -8,54 +8,108 @@ pub(crate) mod region;
...
+}
 
-// TODO: ALT_MENU_SUBSTRING should be dead. Remove when possible.
-pub(crate) const ALT_MENU_SUBSTRING: &str = "-alt-menu";
+/// Process locale display names from a `cldr_serde` struct to `BTreeMaps`
+/// ready to be converted to `ZeroMaps`.
+///
+/// This fn is used by the zeromap-based structs, not the `LocaleNames`
+/// attributes-based structs.
+pub(crate) fn extract_names_for_zeromap_struct<'a, T, K, F>(
```
robertbastian, line 25: that the result of this is later converted into zeromaps is completely irrelevant for this function

provider/source/src/cldr_serde/displaynames/mod.rs
```diff
@@ -7,3 +7,70 @@ pub(crate) mod locale_display_pattern;
...
+
+const ALT_SEPARATOR: &str = "-alt-";
+const MENU_SEPARATOR: &str = "-menu-";
+use core::str::FromStr;
+use serde::{Deserialize, Deserializer};
+
+#[derive(Hash, Eq, PartialEq, Clone, Debug)]
+pub(crate) struct WithAlt<T> {
+    pub(crate) subtag: T,
+    pub(crate) alt: Option<std::borrow::Cow<'static, str>>,
```
robertbastian: suggestion: make an `Alt` enum that contains the variants we recognise, instead of having string constants. then you can discard unknown alts during parsing.

provider/source/src/time_zones/convert.rs
```diff
@@ -143,21 +143,20 @@ impl SourceDataProvider {
...
-                .chain(regions.iter().filter_map(|(region, value)| {
-                    Some((
-                        icu::locale::subtags::Region::try_from_str(
-                            region.strip_suffix("-alt-short")?,
-                        )
-                        .ok()?,
-                        value.as_str(),
-                    ))
+                .chain(regions.iter().filter_map(|(key, value)| {
+                    if key.alt.as_deref() == Some("short") && key.menu.is_none() {
```
robertbastian: nit: use the constant

robertbastian (approved):
I still don't think this is good, but I also don't think this is worth more of my time. I wonder if I spent more time on this than you.

sffc:
> I still don't think this is good, but I also don't think this is worth more of my time. I wonder if I spent more time on this than you.

I spent about 3-4 hours on this. Add another 45-60 minutes today.

provider/source/src/cldr_serde/displaynames/mod.rs
sffc: Yeah I keep chasing down these fully qualified type names, but didn't get to this one. I'll teach the agent to catch this when reviewing itself.

provider/source/src/displaynames/mod.rs
sffc: I wanted to emphasize that this fn is only reachable from the old zeromap structs, which I want to delete at some point. I could have called the fn `extract_names_for_old_struct` but old/new in code goes stale.

provider/source/src/time_zones/convert.rs
sffc: I don't know why I didn't catch this one.

provider/source/src/cldr_serde/displaynames/mod.rs
sffc: This is a good suggestion. I wish I had thought of it

provider/source/src/cldr_serde/displaynames/mod.rs
```diff
@@ -51,13 +100,13 @@ where
                     Ok(WithAlt {
                         subtag,
                         alt: None,
-                        menu: Some(std::borrow::Cow::Owned(menu_str.to_string())),
+                        menu: Some(Menu::from_str(menu_str).unwrap_or(Menu::Unknown)),
```
sffc: Make Unknown the first variant of the enum, and if you encounter it, please `log::warn` the actual string before returning Unknown here

sffc:
Ping @Manishearth, @robertbastian approved this, would you like to stamp it?

Manishearth (approved):

provider/source/src/cldr_serde/displaynames/mod.rs
robertbastian: please resolve comments where you're basically talking to yourself, they're distracting during review

provider/source/src/displaynames/mod.rs
robertbastian: right, you usually call this `multi` though

PR #8091, opened 2026-06-16
Fix CI by running apt-get update before installing LLVM 20

sffc:
Fixes the CI failures on the `test-c` and `test-js` jobs, which were failing with 404 errors during the installation of `lld-20` and `clang-20`. 

By running `sudo apt-get update` before `apt-get install`, we ensure the package cache is refreshed and the runner can successfully locate and download the LLVM 20 packages.


robertbastian:
weird

PR #8095, opened 2026-06-17
Fix a safety issue in the Japanese calendar

robertbastian:
`PackedEra` construction unsafely assumed that `day > 0`, which could lead to undefined behavior when calling `NonZeroU8::new_unchecked` with 0.

This PR makes `PackedEra::pack` fallible, returning a `Result`. ~This change propagates to `JapaneseEras::last_after_reiwa` and the `Japanese` constructors.~

## Changelog

`icu_calendar`
* Fix safety issue in `Japanese::try_new_with_buffer_provider`


components/calendar/src/cal/japanese.rs
sffc, line 51: Nit: It's been a few months and I don't remember exactly how this `post_reiwa_era` is supposed to behave. Please write some internal docs here. (follow-up OK)

components/calendar/src/cal/japanese.rs
```diff
@@ -51,8 +51,14 @@ impl Japanese {
...
         const {
             Self {
-                post_reiwa_era: crate::provider::Baked::SINGLETON_CALENDAR_JAPANESE_MODERN_V1
-                    .last_after_reiwa(),
+                post_reiwa_era: const {
+                    match crate::provider::Baked::SINGLETON_CALENDAR_JAPANESE_MODERN_V1
+                        .last_after_reiwa()
+                    {
+                        Ok(era) => era,
+                        Err(_) => panic!("Invalid compiled data for Japanese calendar"),
```
sffc: Nit: since we have a reasonable fallback behavior (returning `None`), I prefer for this to be a debug_assert instead of a hard panic. (follow-up OK)

components/calendar/src/provider.rs
```diff
@@ -121,24 +121,28 @@ icu_provider::data_struct!(
 );
 
 impl JapaneseEras<'_> {
```
sffc, line 123: Nit: Update the PR description. You aren't changing the data struct here.

sffc (approved):

sffc (approved):

Manishearth (approved):
Hooray for one less use of unsafe

components/calendar/src/cal/japanese.rs
robertbastian: this might get you incorrect behaviour if you have corrupted data, so I don't think this is a reasonable fallback

Manishearth (approved):

components/calendar/src/cal/japanese.rs
sffc: It will just show the date in Reiwa or whatever the latest hard-coded era is. That's fine

PR #8102, opened 2026-06-18
Infer region in calendar resolution

### LINKED ISSUE #8096: Datetime formatting for "fa" outputs gregorian rather than persian calendar as default
I used the following setting `fieldsets::Y::short()`  with `locale!("fa")`. This was the output:

۲۰۲۰  2020 (Gregorian year)

The output for `locale!("fa-u-ca-persian")` was:

۱۳۹۸ 1398 (Persian / Solar Hijri year)

ICU4C, however, outputs the persian calendar for both locales ("fa" and "fa-u-ca-persian").

robertbastian:
Fixes #8096

## Changelog

`icu_locale_core`:
* `preferences` types now implement `databake` (feature-gated)

`icu_calendar`:
* Add `AnyCalendarKind::try_new` and deprecate  `AnyCalendarKind::new`. The new version uses locale data to infer calendars from locales.
* Deprecate `CalendarPreferences::resolve_calendar`. This method did not perform likely-subtags expansion.

`icu_datetime`:
* Use the correct calendar even if the region is only implied by the language (i.e. `fa`)


sffc:
Closely related discussion: https://github.com/unicode-org/icu4x/issues/6326

sffc:
The main question arising from that thread: should we put this new marker in `icu_locale` instead of `icu_calendar`? If the data key is purely a locale preference resolution, then maybe the locale crate is the right place for it.

components/calendar/src/any_calendar.rs
```diff
@@ -748,6 +748,7 @@ pub enum AnyCalendarKind {
 impl AnyCalendarKind {
     /// Selects the [`AnyCalendarKind`] appropriate for the given [`CalendarPreferences`].
     pub fn new(prefs: CalendarPreferences) -> Self {
+        #[allow(deprecated)] // TODO: what to do here?
```
sffc: probably mark this fn as deprecated, too, and optionally add dataful versions

components/calendar/src/preferences.rs
```diff
@@ -50,24 +52,73 @@ impl CalendarPreferences {
...
+    /// # assert_eq!(CalendarPreferences::from(&"und-US-u-ca-islamic-foo".parse::<icu::locale::Locale>().unwrap()).resolved_algorithm_2(),
     /// #     CalendarAlgorithm::Gregory
     /// # );
-    /// # assert_eq!(CalendarPreferences::from(&"und-US-u-ca-hebrew-foo".parse::<icu::locale::Locale>().unwrap()).resolved_algorithm(),
+    /// # assert_eq!(CalendarPreferences::from(&"und-US-u-ca-hebrew-foo".parse::<icu::locale::Locale>().unwrap()).resolved_algorithm_2(),
     /// #     CalendarAlgorithm::Gregory
     /// # );
     /// ```
+    #[cfg(feature = "compiled_data")]
+    pub fn resolved_algorithm_2(&self) -> CalendarAlgorithm {
```
sffc: Discussion: I might not be opposed to just making the existing fn use compiled data, and the power users that disabled the compiled data feature get a breakage. We can add `resolved_algorithm_no_data` for them to make it a quick and easy fix.

components/locale_core/Cargo.toml
```diff
@@ -27,7 +27,7 @@ tinystr = { workspace = true }
 writeable = { workspace = true }
 
 databake = { workspace = true, features = ["derive"], optional = true }
-serde = { workspace = true, optional = true }
+serde = { workspace = true, features = ["derive"], optional = true }
```
sffc, line 30: Observation: it would be nice to not link in `syn` from a low-level crate, but it's already there with `databake`

sffc (commented):
Thanks

robertbastian:
> The main question arising from that thread: should we put this new marker in `icu_locale` instead of `icu_calendar`? If the data key is purely a locale preference resolution, then maybe the locale crate is the right place for it.

Well, there are more architecture questions. We could also keep the existing code and just expand the locale before looking at the region. The data provider does that for us with fallback, but it's not inherently necessary (although maybe a bit nicer if CLDR changes preferred calendars for some region).

components/calendar/src/preferences.rs
Manishearth: > I might not be opposed to just making the existing fn use compiled data, and the power users that disabled the compiled data feature get a breakage

I'd rather not break a real API. 

I checked and it seems like Temporal doesn't actually use this, which surprises me a little: aren't you able to trigger this behavior somehow?

Another option on the table is silent degradation: have `resolved_algorithm` call `resolved_algorithm_2()` in data mode and `resolved_algorithm_no_data` in no-data mode. We document its behavior as such, this becomes intentional API. We don't have to deprecate then. Or we have that behavior on `resolved_algorithm_auto` and still deprecate this.

I think having the `no_data` version is useful.

components/calendar/src/preferences.rs
```diff
@@ -50,24 +52,73 @@ impl CalendarPreferences {
...
-    /// # assert_eq!(CalendarPreferences::from(&"und-US-u-ca-islamic-rgsa".parse::<icu::locale::Locale>().unwrap()).resolved_algorithm(),
+    /// # assert_eq!(CalendarPreferences::from(&"und-US-u-ca-islamic-rgsa".parse::<icu::locale::Locale>().unwrap()).resolved_algorithm_2(),
     /// #     CalendarAlgorithm::Hijri(Some(HijriCalendarAlgorithm::Civil))
     /// # );
-    /// # assert_eq!(CalendarPreferences::from(&"und-US-u-ca-islamic-foo".parse::<icu::locale::Locale>().unwrap()).resolved_algorithm(),
+    /// # assert_eq!(CalendarPreferences::from(&"und-US-u-ca-islamic-foo".parse::<icu::locale::Locale>().unwrap()).resolved_algorithm_2(),
     /// #     CalendarAlgorithm::Gregory
     /// # );
-    /// # assert_eq!(CalendarPreferences::from(&"und-US-u-ca-hebrew-foo".parse::<icu::locale::Locale>().unwrap()).resolved_algorithm(),
+    /// # assert_eq!(CalendarPreferences::from(&"und-US-u-ca-hebrew-foo".parse::<icu::locale::Locale>().unwrap()).resolved_algorithm_2(),
```
Manishearth: issue: if we're adding a `_2` we should somewhere explain the difference. I think the cleanest way to do this is adding a `_no_data` and sticking all the caveats in its documentation

Manishearth (approved):
don't like that we need to add new data for this. oh well

Manishearth (approved):

Manishearth (approved):

components/locale_core/src/preferences/extensions/unicode/macros/enum_keyword.rs
```diff
@@ -71,6 +71,9 @@ macro_rules! __enum_keyword {
         #[derive(Debug, Clone, Eq, PartialEq, Copy, Hash)]
         $(#[derive($derive_attrs)])?
         $(#[$doc])*
+        #[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
```
sffc, line 74: Nit/Issue: the serialized forms of these should probably be the BCP-47. It's hard to justify something other than that. Currently they are the EnumVariant names.

sffc (changes requested):
LGTM but a small issue involving the serde

components/locale_core/src/preferences/extensions/unicode/macros/enum_keyword.rs
robertbastian: I'm not sure I agree. Preferences are strongly typed, and they can also be strongly typed in serialization. `Value` is BCP-47.

sffc: ok, please file a follow-up issue and put it in the 2.3 milestone. Hopefully we can resolve on the thread without needing meeting time.

sffc (approved):

sffc:
This PR appears to have regressed `create_and_format` performance by about 20%

https://unicode-org.github.io/icu4x/benchmarks/perf/components/datetime/index.html

No impact on `format_only` or `fixed` calendar.

Manishearth:
Gemini investigation

It's the extra data load.

<details>

* **PR Mission**: *"Infer region in calendar resolution"* ([PR #8102](https://github.com/unicode-org/icu4x/pull/8102)).
* **Affected Benchmark**: `construct_and_format` on `AnyCalendar` fieldsets (e.g. [`ymd_short/any/construct_and_format/10_cases`](file:///usr/local/google/home/manishearth/dev/worktrees/icu4x-2/components/datetime/benches/datetime.rs#L175-L184)). No impact on `fixed` calendar benchmarks or `format_only`.

#### Root Cause
In [`components/datetime/src/scaffold/calendar.rs`](file:///usr/local/google/home/manishearth/dev/worktrees/icu4x-2/components/datetime/src/scaffold/calendar.rs#L470-L484):
* **Before PR #8102**: [`FormattableAnyCalendar`](file:///usr/local/google/home/manishearth/dev/worktrees/icu4x-2/components/datetime/src/scaffold/calendar.rs#L346) resolved preferences via in-memory matching ([`CalendarPreferences::resolved_algorithm`](file:///usr/local/google/home/manishearth/dev/worktrees/icu4x-2/components/calendar/src/preferences.rs#L71-L100)) with **0** data provider lookups.
* **In PR #8102**: To handle language-only locales like `fa` (Persian) and `th` (Buddhist) without explicit `-u-ca-` tags (fixing issue [#8096](https://github.com/unicode-org/icu4x/issues/8096)), PR #8102 added a new data key [`CalendarPreferredV1`](file:///usr/local/google/home/manishearth/dev/worktrees/icu4x-2/components/calendar/src/provider.rs#L101-L109).
* When `prefs.calendar_algorithm` is `None` (the default for any standard locale like `locale!("fr")`), [`FormattableAnyCalendar::try_new_unstable`](file:///usr/local/google/home/manishearth/dev/worktrees/icu4x-2/components/datetime/src/scaffold/calendar.rs#L450-L485) now performs a provider lookup:
  ```rust
  let algorithm = match prefs.calendar_algorithm {
      Some(algorithm) if !matches!(algorithm, Hijri(None)) => algorithm,
      unresolved => DataProvider::<CalendarPreferredV1>::load(
          &provider,
          DataRequest {
              id: DataIdentifierBorrowed::for_locale(&CalendarPreferredV1::make_locale(
                  prefs.locale_preferences,
              )),
              metadata: Default::default(),
          },
      )?
      .payload
      .get()
      .resolve(unresolved),
  };
  ```
* **Why it regressed by ~20%**:
  * Every `DateTimeFormatter::try_new` instantiation (which constructs an `AnyCalendar`) now runs an additional `CalendarPreferredV1` data load with `LocaleFallbackPriority::Region` and trie traversal.
  * [`FixedCalendarDateTimeFormatter`](file:///usr/local/google/home/manishearth/dev/worktrees/icu4x-2/components/datetime/src/neo.rs#L209) does not construct `FormattableAnyCalendar` and was unaffected, matching Shane's observation.

#### Recommended Fix
* In CLDR, the calendar preference list is sparse: only `TH` (Buddhist), `AF`/`IR` (Persian), and `AE`/`BH`/`KW`/`QA`/`SA` (Islamic-Umalqura) have non-Gregorian defaults; all other regions and the `und` root default to Gregorian.
* A fast-path can check the region (or the language subtag for `fa`, `th`, `ps`) directly in [`CalendarPreferences::resolved_algorithm`](file:///usr/local/google/home/manishearth/dev/worktrees/icu4x-2/components/calendar/src/preferences.rs#L71), or avoid provider fallback when `prefs.locale_preferences` matches standard Gregorian regions.
* 
</details>

PR #8104, opened 2026-06-18
Use `001` as default region for `und` units

robertbastian:
This was a side effect of `und` expanding to `en-Latn-US`. #8110 preserves behaviour, but makes it explicit that we choose `US` as the default region.

meter and kilogram become core units, foot and pound become extended units.

This significantly reduces data size, because the US's special-snowflaking was not great for fallback

```diff
-units/names/area/core/v1, <lookup>, 32936B, 2233 identifiers
+units/names/area/core/v1, <lookup>, 16794B, 1112 identifiers

-units/names/area/extended/v1, <lookup>, 25723B, 2286 identifiers
+units/names/area/extended/v1, <lookup>, 15796B, 1315 identifiers

-units/names/length/core/v1, <lookup>, 34216B, 2375 identifiers
+units/names/length/core/v1, <lookup>, 24255B, 1684 identifiers

-units/names/length/extended/v1, <lookup>, 26077B, 2403 identifiers
+units/names/length/extended/v1, <lookup>, 25257B, 2152 identifiers

-units/names/mass/core/v1, <lookup>, 33003B, 3101 identifiers
+units/names/mass/core/v1, <lookup>, 17354B, 1658 identifiers

-units/names/mass/extended/v1, <lookup>, 16789B, 1445 identifiers
+units/names/mass/extended/v1, <lookup>, 10676B, 935 identifiers

-units/names/volume/core/v1, <lookup>, 38249B, 2898 identifiers
+units/names/volume/core/v1, <lookup>, 21271B, 1546 identifiers

-units/names/volume/extended/v1, <lookup>, 62909B, 5710 identifiers
+units/names/volume/extended/v1, <lookup>, 40657B, 3546 identifiers
```

## Changelog

N/A


components/experimental/src/dimension/units/categorized_formatter.rs
robertbastian, line 1: @younies why are the english long names changing?

younies: This change is a side effect of changing `und`'s default region from `US` to `001`, which exposes a bug where `try_new_core` incorrectly succeeds for non-core units, subsequently shadowing `try_new_extended`.

### Why the test expectations changed (falling back to short names):
1. **Before the PR**: `square-meter` was not in `en-US` core. `try_new_core` for `en-US` and `square-meter` **correctly failed** (as asserted by the tests) because it fell back to `und` core, which also didn't have it (since `und` was US-centric).
2. **After the PR**: `und` now maps to `001` (metric). At the locale fallback level (`en-US` -> `en` -> `und`), `try_new_core` now **incorrectly succeeds** because it finds `square-meter` in `und` core. 
3. **The Shadowing Issue**: Because `und` is a fallback of last resort, it only contains short names/symbols (like `m²`) to save space. Since `try_new_core` now succeeds via fallback, `try_new_extended` (which checks `Core` first to avoid duplication) also succeeds at the `Core` check. It uses the fallback `"m²"` symbol and **never reaches the `en-US` Extended data** where the actual English long name (`square meters`) resides.

### How we should address this:
This behavior should be prevented. **`try_new_core` should fail** for units that are not in the target locale's core set, and users should be advised/directed to use the **`Extended Formatter`** for those units. We need to ensure that the fallback mechanism doesn't silently return "worse" data (short names) when better localized data is available in `Extended`.

I have filed an issue to track this fallback shadowing behavior: **#8125**. We should add a `TODO(#8125)` in the code to address this (e.g., by designing an interleaved fallback mechanism that checks both `Core` and `Extended` for a locale before falling back to the parent locale).

younies:
@robertbastian: thanks for the PR!

Can you change the title to: **Use 001 (World) as default region for und units**

or **fix: Use 001 (World) as default region for und units**

younies (approved):
look at my comments

younies (approved):

sffc (commented):
ok

PR #8105, opened 2026-06-18
Make cross-crate internal API style apply to component crates

sffc:
I was thinking about ICU4X component crates when I wrote this. It wasn't intended to apply to unstable crates (those versioned out of sync with ICU4X).

## Changelog

N/A


Manishearth (approved):
This matches my read of the policy when it was proposed. Me applying it to utils crates was me extending the motivation to utils crates, not attempting to apply the policy

PR #8109, opened 2026-06-19
writeable: impl TryWriteable on references, Either

sffc:
Writeable already had this.

## Changelog

writeable:
- impl TryWriteable on references
- impl TryWriteable on Either


robertbastian (approved):
we have `Writeable` on `&` and `Either`

robertbastian (approved):
we have `Writeable` on `&` and `Either`

PR #8110, opened 2026-06-19
Remove `und` -> `en-Latn-US` expansion

robertbastian:
As the data struct requires us to have a `und -> aa-Bbbb-CC` mapping, this is actually a change to the code. Instead of interpreting this entry as `und -> aa-Bbbb-CC`, we now interpret it as `und-Bbbb -> aa-Bbbb-CC`, `und-CC -> aa-Bbbb-CC`, and `und-Bbbb-CC -> aa-Bbbb-CC`, and we put the `Latn-US` mappings in it for data backward compatibility.

Datagen was updated to generate the same data from either `und -> aa-Bbbb-CC`, or `und-Bbbb -> aa-Bbbb-CC`, `und-CC -> aa-Bbbb-CC`, and `und-Bbbb-CC -> aa-Bbbb-CC`, as this will change in the next CLDR version (https://unicode-org.atlassian.net/browse/CLDR-14524)

## Changelog

`icu_locale`:
* `LocaleExpander::maximize` no longer maximizes `und` to `en-Latn-US`

### DIFF
--- a/components/locale/src/expander.rs
+++ b/components/locale/src/expander.rs
@@ -399,16 +399,23 @@ impl LocaleExpander {
         }
         if let Some(script) = langid.script {
             if let Some(region) = langid.region
-                && let Some(language) = data.get_sr(script, region)
+                && let Some(language) = data
+                    .get_sr(script, region)
+                    .or_else(|| ((script, region) == (und_s, und_r)).then_some(und_l))
             {
                 return update_langid(language, None, None, langid);
             }
-            if let Some((language, region)) = data.get_s(script) {
+            if let Some((language, region)) = data
+                .get_s(script)
+                .or_else(|| (script == und_s).then_some((und_l, und_r)))
+            {
                 return update_langid(language, None, Some(region), langid);
             }
         }
         if let Some(region) = langid.region
-            && let Some((language, script)) = data.get_r(region)
+            && let Some((language, script)) = data
+                .get_r(region)
+                .or_else(|| (region == und_r).then_some((und_l, und_s)))
         {
             return update_langid(language, Some(script), None, langid);
         }
@@ -417,9 +424,7 @@ impl LocaleExpander {
         // to fall back to bare "und"
         debug_assert!(langid.language.is_unknown());
 
-        // und expands to `und_l-und_s-und_r`
-        // TODO: remove this behavior: https://unicode-org.atlassian.net/browse/CLDR-14524
-        update_langid(und_l, Some(und_s), Some(und_r), langid)
+        TransformResult::Unmodified
     }
 
     /// This returns a new Locale that is the result of running the
@@ -543,7 +548,7 @@ impl LocaleExpander {
     fn infer_likely_script(&self, language: Language, region: Option<Region>) -> Option<Script> {
         // This is an inlined and simplified `maximize`.
         let data = self.as_borrowed();
-        let (_, und_s, _) = data.get_und();
+        let (und_l, und_s, und_r) = data.get_und();
 
         if !language.is_unknown() {
             if let Some(region) = region
@@ -558,7 +563,9 @@ impl LocaleExpander {
             return None;
         }
         if let Some(region) = region
-            && let Some((_, script)) = data.get_r(region)
+            && let Some((_, script)) = data
+                .get_r(region)
+                .or_else(|| (region == und_r).then_some((und_l, und_s)))
         {
             return Some(script);
         }
@@ -567,9 +574,7 @@ impl LocaleExpander {
         // to fall back to bare "und"
         debug_assert!(language.is_unknown());
 
-        // und expands to `und_l-und_s-und_r`
-        // TODO: remove this behavior: https://unicode-org.atlassian.net/browse/CLDR-14524
-        Some(und_s)
+        None
     }
 }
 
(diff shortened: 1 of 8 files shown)


Manishearth (commented):
Partial review: The components code looks good, and this change makes sense to me. Did not get around to reviewing provider code, it's a much bigger diff.

I am not as familiar with this code, Shane may be a better reviewer, but I'll see what I can do.

robertbastian:
the first three commits are cleanups, the actual behaviour changes are the last two and they shouldn't be too big

provider/source/src/locale/likely_subtags.rs
```diff
@@ -336,16 +337,12 @@ pub(crate) fn transform<'x>(
...
-                // Some of the target regions here are not equal to the source, such as und-Latn-001 -> en-Latn-US.
-                // However in the `maximize` method we do not replace tags, so we don't need to store the region.
-                with_diff!((l, None, _) => script_region.insert((s, r), l));
+                with_diff!((l, None, None) => script_region.insert((s, r), l));
             } else {
                 with_diff!((l, None, Some(r)) => script.insert(s, (l, r)));
             }
         } else if let Some(r) = entry.0.region {
-            // Some of the target regions here are not equal to the source, such as und-002 -> en-Latn-NG.
-            // However in the `maximize` method we do not replace tags, so we don't need to store the region.
```
sffc, line 347: deleted comment

components/locale/src/provider.rs
```diff
@@ -331,7 +331,9 @@ pub struct LikelySubtagsForLanguage<'data> {
     /// Undefined.
     ///
     /// This entry is treated as the following expansions:
-    /// * und->LSR
+    /// * und-S->LSR
+    /// * und-R->LSR
+    /// * und-S-R->LSR
```
sffc, line 336: I'm a little confused by this part. Shouldn't we just store und-Latn and und-US (and implied und-Latn-US) in the data if they aren't already there? We can either ignore the unused `und` field in the data struct, or it might work to set Script and Region to Zzzz-ZZ.

provider/source/src/locale/likely_subtags.rs
robertbastian: I changed the behaviour so this comment no longer applies

components/locale/src/provider.rs
robertbastian: that would require a bump to the data marker, as old data/new code and new code/old data won't work anymore

robertbastian: in particular new code old data would not have the und-US/und-Latn fallbacks

sffc: The only place that line of data is used in the current code is

```
        debug_assert!(langid.language.is_unknown());
        update_langid(
            data.get_und().0,
            Some(data.get_und().1),
            Some(data.get_und().2),
            langid,
        )
```

I don't understand the code and data shift risk.

If new code starts ignoring the field, then old code gets the old behavior, which is fine.

Or, we could change the data to be und-Zzzz-ZZ instead of en-Latn-US, which should make new-data-old-code also work fine.

robertbastian: * there is no way to get old code to not expand `und`
* old code should continue expanding `und` to `en-Latn-US`, not to `und-Zzzz-ZZ`
* if we already store the `en-Latn-US` expansion in the und field, we should not duplicate that data in other fields

sffc: > * if we already store the `en-Latn-US` expansion in the und field, we should not duplicate that data in other fields

ok, this is where you've been coming from then. My instinct is to just store new entries if necessary in the main list data, which duplicates a small number of bytes, but keeps the code working mostly the same way, which seems cleaner.

In other words, if we were making this from scratch, I don't think this behaving involving the `und` data field is behavior we would have invented as a way to save a few bytes. So, I think I prefer to keep things simpler at the cost of a few bytes of data until we can delete it in 3.0. Right?

robertbastian: > My instinct is to just store new entries if necessary in the main list data, which duplicates a small number of bytes, but keeps the code working mostly the same way, which seems cleaner.

this will break new code old data though. if the new code only looks at the main list, not at the und entry, then `und-US` won't maximise at all with the data that is being generated today

the way I've structured this is that we keep the data struct stable across CLDR 49/50 (https://github.com/unicode-org/icu4x/pull/8110/changes/606be4f5b504577270ec98a0c3cf2c68dd742d5e), and change the implementation to adopt the new behaviour given no data struct changes (https://github.com/unicode-org/icu4x/pull/8110/changes/44fc7c963e94c8906621a5f99d4e3444dfd4a364)

components/locale/src/expander.rs
```diff
@@ -417,9 +424,7 @@ impl LocaleExpander {
         // to fall back to bare "und"
         debug_assert!(langid.language.is_unknown());
 
-        // und expands to `und_l-und_s-und_r`
-        // TODO: remove this behavior: https://unicode-org.atlassian.net/browse/CLDR-14524
-        update_langid(und_l, Some(und_s), Some(und_r), langid)
```
sffc, line 422: Would another way to write this be to call `update_langid` only if langid != LanguageIdentifier::default() ?

components/locale/src/provider.rs
sffc: ok, I see

sffc (approved):

components/locale/src/expander.rs
robertbastian: I think so

PR #8112, opened 2026-06-19
Fix typo in complex script segmentation

Xuanwo:
`complex_language_segment_str` and `complex_language_segment_utf16` split complex-script spans by language, but then segmented the full input for each per-language slice. In mixed complex-script text, that can duplicate segmentation over the entire span and combine those boundaries with the slice offset, producing incorrect or risky offsets.

This changes the per-language dispatch to segment only the current slice while preserving the accumulated offset, and adds regression coverage for Thai+CJK and Japanese+Thai mixed spans across UTF-8 and UTF-16.

## Changelog:

N/A

### DIFF
--- a/components/segmenter/src/complex/mod.rs
+++ b/components/segmenter/src/complex/mod.rs
@@ -464,112 +464,73 @@ fn try_load_static<M: DataMarker, P: DataProvider<M> + ?Sized>(
 mod tests {
     use super::*;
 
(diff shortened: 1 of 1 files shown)


robertbastian (approved):
ah yes, this was a typo in https://github.com/unicode-org/icu4x/pull/8076. thanks for the test

PR #8113, opened 2026-06-19
Improve complex segmenter hot paths

Xuanwo:
This PR improves ICU4X segmenter hot paths observed in dictionary-backed complex-script segmentation. The profile that motivated this work showed a large share of CPU time under `RuleBreakIterator::next`, complex-language word segmentation, and `Char16TrieIterator::next16`.

What changed:

- Replaced repeated cached-break rebuilding with a `ResultCache` cursor. Previously each cached break consumption rebuilt a shortened `Vec` with adjusted offsets. The new cache keeps the original dictionary break offsets and tracks the consumed cursor/offset.
- Added zero-copy complex-run handling for `segment_str` in word and line segmenters. Since `segment_str` receives well-formed UTF-8, the complex-language run can be borrowed from the original `&str`. `segment_utf8` and `segment_utf16` keep their existing allocation paths to preserve ill-formed UTF-8 and UTF-16 behavior.
- Marked `Char16TrieIterator` hot traversal methods as inlineable to reduce call overhead in dictionary trie walking.
- Kept public APIs unchanged. The new state and helpers are internal to the iterator implementation.

Why:

Dictionary-backed segmentation can enter a complex-language run, compute all break offsets for that run, and then yield those offsets one at a time from the rule iterator. The previous implementation allocated and copied while consuming cached offsets, and `segment_str` also copied complex runs into temporary strings even though the input was already a valid `&str`. These costs show up strongly in mixed Thai/CJK/Japanese workloads such as full-text indexing tokenization.

Performance results:

Measured with a temporary release-mode probe on the same machine, comparing this branch against a clean `origin/main` worktree. Each case runs 10,000 iterations; timing values are the median of 3 runs. Allocation numbers are total allocated bytes during those 10,000 iterations, excluding result collection by the probe.

| Case | Time before | Time after | Speedup | Allocation before | Allocation after | Allocation reduction |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `word_english` | 8.95 ms | 8.81 ms | 1.02x | 0 B | 0 B | 0% |
| `word_thai` | 34.93 ms | 25.84 ms | 1.35x | 49.76 MB | 4.80 MB | 90.4% |
| `word_japanese` | 47.34 ms | 33.03 ms | 1.43x | 30.24 MB | 7.68 MB | 74.6% |
| `word_han` | 45.51 ms | 32.08 ms | 1.42x | 34.88 MB | 8.96 MB | 74.3% |
| `word_mixed` | 85.25 ms | 69.39 ms | 1.23x | 693.68 MB | 40.64 MB | 94.1% |
| `line_english` | 13.80 ms | 13.44 ms | 1.03x | 0 B | 0 B | 0% |
| `line_thai` | 36.25 ms | 25.84 ms | 1.40x | 49.76 MB | 4.80 MB | 90.4% |
| `line_mixed` | 82.12 ms | 60.51 ms | 1.36x | 99.52 MB | 9.60 MB | 90.4% |

Validation:

- `cargo fmt -p icu_segmenter -p icu_collections`
- `git diff --check`
- `cargo test -p icu_segmenter --features unstable`
- `cargo test -p icu_collections --features serde,databake`

## Changelog

icu_segmenter/icu_collections: Improve dictionary-backed word and line segmentation performance for complex scripts without public API changes.

### DIFF
--- a/components/segmenter/src/rule_segmenter.rs
+++ b/components/segmenter/src/rule_segmenter.rs
@@ -80,9 +80,11 @@ pub trait RuleBreakType: crate::private::Sealed + Sized {
 #[derive(Debug)]
 pub struct RuleBreakIterator<'data, 's, Y: RuleBreakType> {
     pub(crate) iter: Y::IterAttr<'s>,
+    // Original UTF-8 input for `segment_str`, used by complex handlers to borrow run slices.
+    pub(crate) str_input: Option<&'s str>,
     pub(crate) len: usize,
     pub(crate) current_pos_data: Option<(usize, Y::CharType)>,
-    pub(crate) result_cache: Vec<usize>,
+    pub(crate) result_cache: ResultCache,
     pub(crate) data: &'data RuleBreakData<'data>,
     pub(crate) complex: Option<ComplexPayloadsBorrowed<'data>>,
     // The property associated with the previous break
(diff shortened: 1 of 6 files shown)


components/segmenter/src/rule_segmenter.rs
```diff
@@ -80,9 +80,11 @@ pub trait RuleBreakType: crate::private::Sealed + Sized {
 #[derive(Debug)]
 pub struct RuleBreakIterator<'data, 's, Y: RuleBreakType> {
     pub(crate) iter: Y::IterAttr<'s>,
+    // Original UTF-8 input for `segment_str`, used by complex handlers to borrow run slices.
+    pub(crate) str_input: Option<&'s str>,
```
robertbastian, line 84: there's no need for this. the input can already be accessed as `iter.as_str()` for `str`, or `iter.as_slice()` for UTF-16 and UTF-8 (although you have to compensate the offset). if you rewrite your code in terms of operations on `Y::IterAttr`, it should be useable for UTF-16 and UTF-8 as well

components/segmenter/src/rule_segmenter.rs
```diff
@@ -93,6 +95,72 @@ pub struct RuleBreakIterator<'data, 's, Y: RuleBreakType> {
...
+/// Cached dictionary break offsets for the current complex-language run.
+///
+/// Dictionary segmentation returns offsets from the start of the complex run.
+/// The rule iterator yields one break at a time, so this cache keeps the
+/// original offsets plus a cursor instead of rebuilding a shortened `Vec` after
+/// every yielded break.
+#[derive(Debug)]
+pub(crate) struct ResultCache {
+    breaks: Vec<usize>,
+    index: usize,
```
robertbastian, line 107: a Vec with an index is pretty much the same as `vec::IntoIter`, but with the downside that the access is panicky. If you store `Peekable<vec::IntoIter<usize>>` instead, you'll be able to do a 0(1) `next_if`, which is what you need

components/segmenter/src/rule_segmenter.rs
```diff
@@ -93,6 +95,72 @@ pub struct RuleBreakIterator<'data, 's, Y: RuleBreakType> {
...
+    offset: usize,
```
robertbastian, line 108: I haven't looked at how this is used in detail, but can you not apply it to the break indices earlier or later to not have to compensate for it all the time?

robertbastian (commented):
Okay, so there are three independent optimisations here:
* Inlining CPT methods which improves performance of the dictionary itself
* Not allocating the substring in the complex `str` path
* A better `ResultCache` which avoids reallocations for some operations

So
* I would like to see benchmarks for these separately. In particular, the CPT improvement benefits both the published segmenter, and the in-progress `neo` rewrite
* Please publish your benchmark code. I would like to see how this all compares to the new segmenter
* I don't think the second optimisation makes much sense. `complex_script_segment_utf16` already accepts a slice of unvalidated UTF-16, I don't understand why we're allocating here in the first place. You've gotten rid of the allocation in the `str` path, but the same should be possible in the UTF-16 and UTF-8 paths (it's trivial to add a `complex_script_segment_utf8` method for that to use).

Xuanwo:
Thank you @robertbastian for the review, I will split the PR and address the comments.

sffc:
Closing in favor of the split-out PRs.

PR #8114, opened 2026-06-20
zerovec: Fix pointer provenance issues found by miri

Manishearth:
This is mostly "minor UB but still worth fixing" since eventually rustc will do more optimizations based on pointer provenance.

The general issue with pointer provenance is you have to be pretty careful of the pedigree of your pointers, making sure they were calculated through the right "path", even if the value is correct.

## Changelog: N/A

### DIFF
--- a/utils/zerovec/src/cow.rs
+++ b/utils/zerovec/src/cow.rs
@@ -193,7 +193,8 @@ impl<'a, V: VarULE + ?Sized> VarZeroCow<'a, V> {
     pub fn new_owned(val: Box<V>) -> Self {
         let raw_box: *mut V = Box::into_raw(val);
         // SAFETY: raw_box is a valid pointer to V as it comes from Box::into_raw.
-        let slice_ref: &[u8] = unsafe { (*raw_box).as_bytes() };
+        let raw_ref: &V = unsafe { &*raw_box };
+        let slice_ref: &[u8] = raw_ref.as_bytes();
         // SAFETY: We construct the NonNull pointer directly from raw_box (which has unique owning
         // provenance) cast to u8, using the length from slice_ref. This avoids losing unique provenance.
         let buf = unsafe {
--- a/utils/zerovec/src/cow.rs
+++ b/utils/zerovec/src/cow.rs
@@ -191,8 +191,17 @@ impl<'a, V: VarULE + ?Sized> VarZeroCow<'a, V> {
     /// ✨ *Enabled with the `alloc` Cargo feature.*
     #[cfg(feature = "alloc")]
     pub fn new_owned(val: Box<V>) -> Self {
-        let val = ManuallyDrop::new(val);
-        let buf: NonNull<[u8]> = val.as_bytes().into();
+        let raw_box: *mut V = Box::into_raw(val);
+        // SAFETY: raw_box is a valid pointer to V as it comes from Box::into_raw.
+        let slice_ref: &[u8] = unsafe { (*raw_box).as_bytes() };
+        // SAFETY: We construct the NonNull pointer directly from raw_box (which has unique owning
+        // provenance) cast to u8, using the length from slice_ref. This avoids losing unique provenance.
+        let buf = unsafe {
+            NonNull::new_unchecked(core::ptr::slice_from_raw_parts_mut(
+                raw_box.cast::<u8>(),
+                slice_ref.len(),
+            ))
+        };
         let raw = RawVarZeroCow {
             // Invariants upheld:
             // 1 & 3: The bytes came from `val` so they're a valid value and byte slice
--- a/utils/zerovec/src/ule/encode.rs
+++ b/utils/zerovec/src/ule/encode.rs
@@ -95,12 +95,10 @@ pub fn encode_varule_to_box<S: EncodeAsVarULE<T> + ?Sized, T: VarULE + ?Sized>(x
     // zero-fill the vector to avoid uninitialized data UB
     let mut vec: Vec<u8> = vec![0; x.encode_var_ule_len()];
     x.encode_var_ule_write(&mut vec);
-    let raw = Box::into_raw(vec.into_boxed_slice());
     // SAFETY:
-    // - raw is a valid pointer to [u8] from Box::into_raw of a valid Box.
     // - T::from_bytes_unchecked is safe because the bytes were written by x.encode_var_ule_write
     //   which guarantees a valid representation of T.
-    unsafe { box_from_raw_bytes(raw) }
+    unsafe { box_from_bytes(vec.into_boxed_slice()) }
 }
 
 unsafe impl<T: VarULE + ?Sized> EncodeAsVarULE<T> for T {
--- a/utils/zerovec/src/ule/encode.rs
+++ b/utils/zerovec/src/ule/encode.rs
@@ -95,15 +95,12 @@ pub fn encode_varule_to_box<S: EncodeAsVarULE<T> + ?Sized, T: VarULE + ?Sized>(x
     // zero-fill the vector to avoid uninitialized data UB
     let mut vec: Vec<u8> = vec![0; x.encode_var_ule_len()];
     x.encode_var_ule_write(&mut vec);
-    let boxed = core::mem::ManuallyDrop::new(vec.into_boxed_slice());
-    unsafe {
-        // Safety: `ptr` is a box, and `T` is a VarULE which guarantees it has the same memory layout as `[u8]`
-        // and can be recouped via from_bytes_unchecked()
-        let ptr: *mut T = T::from_bytes_unchecked(&boxed) as *const T as *mut T;
-
-        // Safety: we can construct an owned version since we have mem::forgotten the older owner
-        Box::from_raw(ptr)
-    }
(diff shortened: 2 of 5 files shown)


utils/zerovec/src/zerovec/mod.rs
```diff
@@ -175,7 +175,7 @@ impl<U> EyepatchHackVector<U> {
         //   smaller, from the same pointer, so it will be valid as well, and similarly non-null.
         self.buf = unsafe {
             NonNull::new_unchecked(core::ptr::slice_from_raw_parts_mut(
-                self.buf.as_mut().as_mut_ptr(),
+                self.buf.as_ptr().cast::<U>(),
```
robertbastian: why this? where did the mut go?

utils/zerovec/src/cow.rs
```diff
@@ -191,8 +191,17 @@ impl<'a, V: VarULE + ?Sized> VarZeroCow<'a, V> {
     /// ✨ *Enabled with the `alloc` Cargo feature.*
     #[cfg(feature = "alloc")]
     pub fn new_owned(val: Box<V>) -> Self {
-        let val = ManuallyDrop::new(val);
-        let buf: NonNull<[u8]> = val.as_bytes().into();
+        let raw_box: *mut V = Box::into_raw(val);
+        // SAFETY: raw_box is a valid pointer to V as it comes from Box::into_raw.
+        let slice_ref: &[u8] = unsafe { (*raw_box).as_bytes() };
```
robertbastian:
```suggestion
        let slice_ref: &[u8] = unsafe { *raw_box }.as_bytes();
```

utils/zerovec/src/ule/mod.rs
```diff
@@ -450,3 +446,51 @@ impl UleError {
...
+///    simply the byte length of the allocation. We must use `T::from_bytes_unchecked` to construct
+///    a reference with the correct metadata, and then combine that metadata with the unique thin
+///    pointer from `Box::into_raw` without clobbering the provenance.
+///
+/// ### Safety
+///
+/// - `raw` must be a valid pointer to `[u8]` obtained from `Box::into_raw` of a box that contains
+///   a valid representation of `T`.
+/// - `T::from_bytes_unchecked` must be safe to call on the slice representation of `raw`.
+pub(crate) unsafe fn box_from_raw_bytes<T: VarULE + ?Sized>(raw: *mut [u8]) -> Box<T> {
```
robertbastian: suggestion: make this take a `Box` so you have one fewer safety requirements

Manishearth: Yeah I considered that and then went to do other things. I'll try

utils/zerovec/src/cow.rs
Manishearth: I would prefer this to not get wrapped, which I think this does. Generally I don't scope unsafe beyond the line level.

utils/zerovec/src/zerovec/mod.rs
Manishearth: as_mut returns a reference, as_ptr returns a mutable raw pointer. Constructing a reference here is UB.

utils/zerovec/src/cow.rs
robertbastian: I think it's more important to keep unsafe blocks as local as possible. if you have formatting concerns you can create another variable for the result of `as_bytes()`

utils/zerovec/src/zerovec/mod.rs
robertbastian: ok, but please update the safety comment. there's a `cast<U>` here that wasn't there before; why, and why is this correct

Manishearth: Previously we were going `NonNull<[U]> -> &mut [U] -> *mut U`. Now we're going `NonNull<[U]> ->  *mut [U] -> *mut U`

utils/zerovec/src/cow.rs
Manishearth: I generally think that they should be as small as possible except not so much as to affect the readability of the code. I think this is more readable.

Manishearth: (Usually this means that I don't stick unsafe blocks within a line)

As for splitting the code, I don't think it helps readability here.

robertbastian: I had to think about whether the unsafe applies to the `*` or the `as_bytes` call. I don't want to think about that

utils/zerovec/src/ule/encode.rs
```diff
@@ -95,12 +95,10 @@ pub fn encode_varule_to_box<S: EncodeAsVarULE<T> + ?Sized, T: VarULE + ?Sized>(x
     // zero-fill the vector to avoid uninitialized data UB
     let mut vec: Vec<u8> = vec![0; x.encode_var_ule_len()];
     x.encode_var_ule_write(&mut vec);
-    let raw = Box::into_raw(vec.into_boxed_slice());
     // SAFETY:
-    // - raw is a valid pointer to [u8] from Box::into_raw of a valid Box.
     // - T::from_bytes_unchecked is safe because the bytes were written by x.encode_var_ule_write
     //   which guarantees a valid representation of T.
-    unsafe { box_from_raw_bytes(raw) }
+    unsafe { box_from_bytes(vec.into_boxed_slice()) }
```
robertbastian: now that the input is also a `Box`, `cast_box` might be a better name

utils/zerovec/src/ule/mod.rs
```diff
@@ -363,11 +363,9 @@ pub unsafe trait VarULE: 'static {
     fn to_boxed(&self) -> Box<Self> {
         use alloc::borrow::ToOwned;
         let bytesvec = self.as_bytes().to_owned().into_boxed_slice();
```
robertbastian: nit: `bytesvec` is not a `Vec`

utils/zerovec/src/ule/mod.rs
```diff
@@ -450,3 +444,50 @@ impl UleError {
...
+///    a reference with the correct metadata, and then combine that metadata with the unique thin
+///    pointer from `Box::into_raw` without clobbering the provenance.
+///
+/// ### Safety
+///
+/// - `T::from_bytes_unchecked` must be safe to call on the slice behind bytes
+pub(crate) unsafe fn box_from_bytes<T: VarULE + ?Sized>(bytes: Box<[u8]>) -> Box<T> {
+    use core::alloc::Layout;
+    let raw = Box::into_raw(bytes);
+    // SAFETY: caller guarantees `raw` is valid.
```
robertbastian: issue: not the caller anymore, but the previous line

utils/zerovec/src/ule/mod.rs
```diff
@@ -450,3 +444,50 @@ impl UleError {
 }
 
 impl core::error::Error for UleError {}
+
+#[cfg(feature = "alloc")]
+/// Reconstructs a `Box<T>` from an allocated byte slice while preserving unique pointer provenance
```
robertbastian: be exact with types

```suggestion
/// Reconstructs a `Box<T>` from a `Box<[u8]>` while preserving unique pointer provenance
```

utils/zerovec/src/ule/mod.rs
```diff
@@ -450,3 +444,50 @@ impl UleError {
...
+/// and correctly resolving DST metadata.
+///
+/// This is an internal helper used to safely convert a heap-allocated byte buffer (e.g., from a `Vec<u8>`)
```
robertbastian:
```suggestion
/// This is an internal helper used to safely convert a heap-allocated byte buffer
```

utils/zerovec/src/zerovec/mod.rs
```diff
@@ -175,7 +178,8 @@ impl<U> EyepatchHackVector<U> {
         //   smaller, from the same pointer, so it will be valid as well, and similarly non-null.
         self.buf = unsafe {
             NonNull::new_unchecked(core::ptr::slice_from_raw_parts_mut(
-                self.buf.as_mut().as_mut_ptr(),
+                // Remove the DST
+                buf_ptr.cast::<U>(),
```
robertbastian, line 182: we usually write pointer casts as `as *mut U`, this is the first time I'm seeing this syntax. I'd prefer to use `as` casts, as a `cast` function could do god-knows-what

utils/zerovec/src/cow.rs
Manishearth: Okay, fine.

utils/zerovec/src/zerovec/mod.rs
Manishearth: `as` casts are far more flexible and can do scarier things. `cast` is very single-purpose. If anything, we probably should be clippy linting `as` casts into `.cast()`.

robertbastian (approved):

utils/zerovec/src/zerovec/mod.rs
sffc: I didn't know about `.cast()`. I like fns with smaller contracts. We should use `.cast()` more.

sffc (commented):
If @robertbastian has reviewed this I'm happy with that.

PR #8115, opened 2026-06-22
Add iterator-based parse_ucd_lines helper

### LINKED ISSUE #7947: Collect UCD-format parsing code in one file

Manishearth:
Take 2 at https://github.com/unicode-org/icu4x/pull/8077

Fixes https://github.com/unicode-org/icu4x/issues/7947

My original attempt was predicated on "I want all of this parsery code in one file", and _then_ doing cleanups on top of that. But I think after going through that, it seems like we can actually design a pretty lightweight internal iteratory abstraction that handles all of the parsing complexities (including `@missing`), so that the caller can just have the processing step.

I've tried to make it so that it's abundantly clear when e.g. an `@missing` line is being explicitly ignored.

I think this code is much easier to read: the processing code doesn't need to muck with file syntax anymore. We still need to e.g. make it clear that we expect `@missing` rules before the data in the ranges they represent, but now it's much clearer that we make that assumption.

Also the main property parsing function becomes single-pass.

### DIFF
--- a/provider/source/src/properties/bidi.rs
+++ b/provider/source/src/properties/bidi.rs
@@ -46,8 +46,7 @@ impl DataProvider<PropertyEnumBidiMirroringGlyphV1> for SourceDataProvider {
         let bidi_mirroring = self
             .parse_ucd_lines("ucd/BidiMirroring.txt")?
             .filter_map(|line| {
-                let line = line.skip_missing_rule()?;
-                let mut fields = line.fields();
+                let mut fields = line.skip_missing_rule()?.fields();
                 let cp_range = fields.next().unwrap().trim();
                 let prop_value = fields.next().unwrap().trim();
                 let value = ucd_helpers::parse_cp(prop_value);
@@ -57,8 +56,7 @@ impl DataProvider<PropertyEnumBidiMirroringGlyphV1> for SourceDataProvider {
             .collect::<HashMap<_, _>>();
 
         let paired_brackets = self.parse_ucd_lines("ucd/BidiBrackets.txt")?.filter_map(|line| {
-                let line = line.skip_missing_rule()?;
-                let mut parts = line.fields();
+                let mut parts = line.skip_missing_rule()?.fields();
                 let cp = ucd_helpers::parse_cp(parts.next().unwrap().trim());
                 let mirror = ucd_helpers::parse_cp(parts.next().unwrap().trim());
 
--- a/provider/source/src/properties/bidi.rs
+++ b/provider/source/src/properties/bidi.rs
@@ -43,34 +43,21 @@ impl DataProvider<PropertyEnumBidiMirroringGlyphV1> for SourceDataProvider {
         let bidi_m_cpinvlist = self.get_binary_prop("Bidi_Mirrored", "Bidi_M")?;
 
         let bidi_mirroring = self
-            .unicode()?
-            .read_to_string("ucd/BidiMirroring.txt")?
-            .lines()
+            .parse_ucd_lines("ucd/BidiMirroring.txt")?
             .filter_map(|line| {
-                let line = line.split('#').next().unwrap().trim();
-                if line.is_empty() {
-                    return None;
-                }
-                let mut fields = line.split(';');
+                let line = line.skip_missing_rule()?;
+                let mut fields = line.fields();
                 let cp_range = fields.next().unwrap().trim();
                 let prop_value = fields.next().unwrap().trim();
                 let value = u32::from_str_radix(prop_value, 16).expect(prop_value);
-
                 let cp = u32::from_str_radix(cp_range, 16).unwrap();
                 Some((cp, char::from_u32(value).unwrap()))
             })
             .collect::<HashMap<_, _>>();
 
-        let paired_brackets = self
-           .unicode()?
-            .read_to_string("ucd/BidiBrackets.txt")?
-            .lines()
-            .filter_map(|line| {
-                let line = line.split('#').next().unwrap().trim();
-                if line.is_empty() {
-                    return None;
-                }
-                let mut parts = line.split(';');
+        let paired_brackets = self.parse_ucd_lines("ucd/BidiBrackets.txt")?.filter_map(|line| {
+                let line = line.skip_missing_rule()?;
+                let mut parts = line.fields();
                 let cp = u32::from_str_radix(parts.next().unwrap().trim(), 16).unwrap();
                 let mirror = u32::from_str_radix(parts.next().unwrap().trim(), 16).unwrap();
 
@@ -88,8 +75,8 @@ impl DataProvider<PropertyEnumBidiMirroringGlyphV1> for SourceDataProvider {
                     _ => unreachable!(),
                 };
                 Some((cp, typ))
-            })
-            .collect::<HashMap<_, _>>();
+
+        }).collect::<HashMap<_, _>>();
 
         let mut builder = CodePointTrieBuilder::new(
             BidiMirroringGlyph::default(),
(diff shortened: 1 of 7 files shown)


provider/source/src/properties/ucd_helpers.rs
```diff
@@ -0,0 +1,71 @@
...
+        if let Self::Fields(c) = *self {
+            Some(c)
+        } else {
+            None
+        }
+    }
+}
+
+impl SourceDataProvider {
+    /// Helper to parse UCD files line-by-line, providing an iterator over the fields of each line.
```
robertbastian, line 47: docs are all wrong

provider/source/src/properties/bidi.rs
```diff
@@ -43,34 +43,21 @@ impl DataProvider<PropertyEnumBidiMirroringGlyphV1> for SourceDataProvider {
...
-            .read_to_string("ucd/BidiMirroring.txt")?
-            .lines()
+            .parse_ucd_lines("ucd/BidiMirroring.txt")?
             .filter_map(|line| {
-                let line = line.split('#').next().unwrap().trim();
-                if line.is_empty() {
-                    return None;
-                }
-                let mut fields = line.split(';');
+                let line = line.skip_missing_rule()?;
```
robertbastian: nit: `line` should be inlined, also elsewhere

provider/source/src/properties/bidi.rs
```diff
@@ -43,34 +43,21 @@ impl DataProvider<PropertyEnumBidiMirroringGlyphV1> for SourceDataProvider {
...
+                let mut fields = line.fields();
                 let cp_range = fields.next().unwrap().trim();
                 let prop_value = fields.next().unwrap().trim();
                 let value = u32::from_str_radix(prop_value, 16).expect(prop_value);
-
                 let cp = u32::from_str_radix(cp_range, 16).unwrap();
```
robertbastian: parse_cp?

provider/source/src/properties/enum_codepointtrie.rs
```diff
@@ -57,82 +58,82 @@ impl SourceDataProvider {
...
+                        // Don't log an error for every code point, the name data marker code
+                        // will log an error that there's an unknown variant.
+                        continue;
+                    };
+
+                    if cps == "0000..10FFFF" {
+                        // This is a statement of default. just check that we're using the same one
+                        assert_eq!(value, T::default());
+                    } else {
+                        let range = ucd_helpers::parse_range(cps);
```
robertbastian: compare the parsed range to not have strings logic in here
```suggestion
                    let range = ucd_helpers::parse_range(cps);
                    if range == 0..=10FFFF {
                        // This is a statement of default. just check that we're using the same one
                        assert_eq!(value, T::default());
                    } else {
```

provider/source/src/properties/enum_codepointtrie.rs
```diff
@@ -57,82 +58,82 @@ impl SourceDataProvider {
...
-                if fields.next().unwrap().trim() != short_name {
-                    continue;
+        let mut last_seen_cp = -1i32;
+
+        for line in self.parse_ucd_lines(&file)? {
+            match line {
+                UcdLine::Missing(fields) => {
+                    let mut fields = fields.fields();
+                    let cps = fields.next().unwrap();
+                    if &file == "ucd/DerivedCoreProperties.txt" {
```
robertbastian, line 74: suggestion: move the file selection to `ucd_helpers` as well, and let it handle treating `DerivedCoreProperties.txt` differently.

provider/source/src/properties/enum_codepointtrie.rs
```diff
@@ -57,82 +58,82 @@ impl SourceDataProvider {
...
+                        .expect("file should only use names from PropertyValueAliases.txt")
+                        .0;
+                    let Some(&value) = short_name_to_t.get(value) else {
+                        // Don't log an error for every code point, the name data marker code
+                        // will log an error that there's an unknown variant.
+                        continue;
+                    };
+
+                    if cp_range.contains("..") {
+                        let range = ucd_helpers::parse_range(cp_range);
```
robertbastian: suggestion: let `parse_range` handle singleton ranges as well

provider/source/src/properties/ucd_helpers.rs
Manishearth: Yeah, I copied it from the old one and then didn't change it when I updated its architecture. Oops.

provider/source/src/properties/bidi.rs
Manishearth: It felt too oneliner for me but yeah I should add it.

provider/source/src/properties/enum_codepointtrie.rs
Manishearth: Actually, it already does. I just don't use it here :facepalm:

robertbastian (approved):

Manishearth:
Ah, the skip_missing_rule() on a separate line was loadbearing

robertbastian:
it really shouldn't be

provider/source/src/properties/ucd_helpers.rs
```diff
@@ -0,0 +1,73 @@
...
+pub(crate) struct UcdFields<'a>(&'a str);
+
+#[derive(Copy, Clone, Debug)]
+pub(crate) enum UcdLine<'a> {
+    Fields(UcdFields<'a>),
+    Missing(UcdFields<'a>),
+}
+
+impl<'a> UcdFields<'a> {
+    pub(crate) fn fields(&self) -> impl Iterator<Item = &'a str> {
```
robertbastian:
```suggestion
    pub(crate) fn fields(self) -> impl Iterator<Item = &'a str> {
```

provider/source/src/properties/ucd_helpers.rs
```diff
@@ -0,0 +1,73 @@
...
+        self.0.split(';').map(str::trim)
+    }
+}
+
+impl<'a> UcdLine<'a> {
+    pub(crate) fn skip_missing_rule(&self) -> Option<UcdFields<'a>> {
```
robertbastian:
```suggestion
    pub(crate) fn skip_missing_rule(self) -> Option<UcdFields<'a>> {
```

Manishearth:
Ah. Pesky `+ 'a`.

robertbastian (approved):

PR #8117, opened 2026-06-22
Pin cargo-rdme to v1.5.1 in tidy job

sffc:
Fixes tidy CI failures by pinning `cargo-rdme` to `v1.5.1` and splitting the installation into a separate task.

Recently, `cargo-rdme` released `v2.0.0` which introduced incompatibilities with the rustdoc JSON generated by our current toolchain, causing tidy CI to fail in all PRs.

To resolve this:
1. We pin `cargo-rdme` to `v1.5.1` (the last known working version).
2. We split the installation of `cargo-rdme` into a separate `install-cargo-rdme` task using the `@shell` runner. This ensures that the installation command runs in a proper shell environment, while the main `generate-readmes` task can continue to run in `@duckscript` without "command not found" errors.

## Changelog (N/A)

Internal CI configuration change.

### DIFF
--- a/tools/make/tidy.toml
+++ b/tools/make/tidy.toml
@@ -126,9 +126,17 @@ end
 echo "License header check complete"
 '''
 
+[tasks.install-cargo-rdme]
+description = "Install cargo-rdme v1.5.1"
+category = "ICU4X Development"
+script_runner = "@shell"
+script = '''
+cargo rdme --version | grep 1.5.1 || cargo install cargo-rdme --version 1.5.1
+'''
+
 [tasks.generate-readmes]
 description = "Automatically generate README.md for each component."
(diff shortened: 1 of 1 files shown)


sffc:
this is 100% AI-generated. I do not vouch for this being the right fix.

Manishearth:
Ah, I had a fix for this locally already and hadn't pushed it. My worry about this is that it forces a local version.

https://github.com/unicode-org/icu4x/pull/8118 is my approach

No strong opinion, a reason to prefer this approach over mine is that this will _never_ lead to a "use nightly" error locally. But I think that's fine, if I get that warning I'll just `cargo +nightly rdme`

sffc:
I prefer sticking with `install_crate`

PR #8118, opened 2026-06-22
Install cargo-rdme at 1.5.1

Manishearth:
Different from https://github.com/unicode-org/icu4x/pull/8117 in that it doesn't _force_ a particular version, but if it is installing fresh it installs the right one. Allows users to have flexibility in their local environments. But not as watertight.

## Changelog: N/A

### DIFF
--- a/tools/make/tidy.toml
+++ b/tools/make/tidy.toml
@@ -129,6 +129,8 @@ echo "License header check complete"
 [tasks.generate-readmes]
 description = "Automatically generate README.md for each component."
 install_crate = { crate_name = "cargo-rdme", test_arg = ["--help"] }
+# cargo-rdme 2.0 uses rustdoc-json and is tied to Rust versions
+install_crate_args = ["--version", "=1.5.1"]
 category = "ICU4X Development"
 script_runner = "@duckscript"
 script = '''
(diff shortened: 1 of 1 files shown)


sffc (approved):

tools/make/tidy.toml
```diff
@@ -129,6 +129,8 @@ echo "License header check complete"
 [tasks.generate-readmes]
 description = "Automatically generate README.md for each component."
 install_crate = { crate_name = "cargo-rdme", test_arg = ["--help"] }
+# cargo-rdme 2.0 uses rustdoc-json and is tied to Rust versions
+install_crate_args = ["--version", "=1.5.1"]
```
sffc: can you make it from the 1.x stream instead of =1.5.1?

sffc:
Is this temporary, or are we locked into an old version of this tool until further notice?

tools/make/tidy.toml
```diff
@@ -129,6 +129,8 @@ echo "License header check complete"
 [tasks.generate-readmes]
 description = "Automatically generate README.md for each component."
 install_crate = { crate_name = "cargo-rdme", test_arg = ["--help"] }
```
sffc, line 131: Can you put `version = "1.5.1"` on this line ?

Manishearth: No

Manishearth:
Going to proactively merge to fix CI.

tools/make/tidy.toml
Manishearth: (only a min version)

Manishearth: Proabably a good feature request, can't see why he wouldn't want to add that

robertbastian:
lgtm

PR #7243, opened 2025-11-13
Update to dart 3.10

robertbastian:
Native assets are now stable.

PR #8119, opened 2026-06-22
Upgrade dependencies for Dart ICU4X package

mosuem:
To not fall too far behind the current implementation.

## Changelog
FFI:
* Update supported Dart toolchain for `record_use` to `3.13.0-215.0.dev`

### DIFF
--- a/.github/workflows/artifacts-build.yml
+++ b/.github/workflows/artifacts-build.yml
@@ -238,9 +238,6 @@ jobs:
 
     - name: Install Dart
       uses: dart-lang/setup-dart@0a8a0fc875eb934c15d08629302413c671d3f672 # v1.6.5
-      with:
-        # Keep in sync with build-test.yml
-        sdk: 3.13.0-215.0.dev
         
     - name: Build docs
       run: |
--- a/.github/workflows/artifacts-build.yml
+++ b/.github/workflows/artifacts-build.yml
@@ -238,6 +238,9 @@ jobs:
 
     - name: Install Dart
       uses: dart-lang/setup-dart@0a8a0fc875eb934c15d08629302413c671d3f672 # v1.6.5
+      with:
+        # Keep in sync with build-test.yml
+        sdk: 3.13.0-215.0.dev
         
     - name: Build docs
       run: |
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -339,12 +339,11 @@ jobs:
     # Job-specific dependencies
     - uses: dart-lang/setup-dart@0a8a0fc875eb934c15d08629302413c671d3f672 # v1.6.5
       with:
-        # Keep in sync with artifacts-build.yml
         sdk: 3.13.0-215.0.dev
         # Currently disabled: There are a lot of incoming changes to Dart native-assets' JSON format,
         # so we won't do nightly builds for now. The first failing build is 3.12.0-62.0.dev.
         # Tracked upstream in https://github.com/orgs/dart-lang/projects/99
-        # Revisit in April 2026.
+        # Revisit in August 2026.
         # sdk: ${{ github.event_name == 'schedule' && github.event.schedule == '0 14 * * *' && 'dev' || '3.11.0-46.0.dev' }}
     - name: Install yq if Windows
       if: ${{ matrix.os == 'windows-latest' }}
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -340,7 +340,7 @@ jobs:
     - uses: dart-lang/setup-dart@0a8a0fc875eb934c15d08629302413c671d3f672 # v1.6.5
       with:
         # Keep in sync with artifacts-build.yml
-        sdk: 3.11.0-46.0.dev
+        sdk: 3.13.0-215.0.dev
         # Currently disabled: There are a lot of incoming changes to Dart native-assets' JSON format,
         # so we won't do nightly builds for now. The first failing build is 3.12.0-62.0.dev.
         # Tracked upstream in https://github.com/orgs/dart-lang/projects/99
--- a/ffi/dart/analysis_options.yaml
+++ b/ffi/dart/analysis_options.yaml
@@ -10,8 +10,6 @@ analyzer:
   language:
     strict-casts: true
     strict-inference: true
-  errors:
-    use_null_aware_elements: ignore
 
 linter:
   rules:
--- a/ffi/dart/analysis_options.yaml
+++ b/ffi/dart/analysis_options.yaml
@@ -11,9 +11,7 @@ analyzer:
     strict-casts: true
     strict-inference: true
   errors:
-    # this lint is overzealous (fixing would break semver)
-    # in 3.10, which breaks our perfect pub.dev score
-    deprecated_member_use_from_same_package: false
+    use_null_aware_elements: ignore
 
 linter:
   rules:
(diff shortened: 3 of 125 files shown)


Manishearth (approved):
Can you also update the versions used in the Diplomat repo?

mosuem:
> Can you also update the versions used in the Diplomat repo?

I don't see any Dart version specified there, would you mind pointing me to the version? Or do you mean the minimum version?

Also, I think the diplomat version must be updated to include https://github.com/rust-diplomat/diplomat/pull/1161 in the CI here.

robertbastian:
The generated code doesn't seem to compile, so we have to make changes in Diplomat first.

mosuem:
I think it's just that already landed PR which has to be pulled in here - could you point me to where the diplomat version is set? Or is that change not published yet?

robertbastian:
https://github.com/unicode-org/icu4x/blob/c6fd4da3460f6a99e8f11c913955ee902c451aa5/Cargo.toml#L221

robertbastian:
`cargo run diplomat-gen`, but we might want to bump the version in a separate PR

ffi/dart/analysis_options.yaml
```diff
@@ -30,7 +31,8 @@ linter:
     - prefer_final_locals
     - prefer_relative_imports
     - prefer_single_quotes
-    - public_member_api_docs
+    # Disable for now while not all members have docs
```
robertbastian: did this become breaking? if it just warns we should not disable it

ffi/dart/pubspec.yaml
```diff
@@ -5,7 +5,7 @@
 name: icu4x
 description: > 
   Dart bindings for the Rust `icu` internationalization crate.
-version: 2.2.0-dev.1
+version: 2.2.0-dev.2
```
robertbastian: I don't think there's any hurry to release this, can we wait for `2.3.0-dev.1`?

ffi/dart/analysis_options.yaml
```diff
@@ -13,7 +13,8 @@ analyzer:
   errors:
     # this lint is overzealous (fixing would break semver)
     # in 3.10, which breaks our perfect pub.dev score
-    deprecated_member_use_from_same_package: false
+    deprecated_member_use_from_same_package: ignore
+    use_null_aware_elements: ignore
```
robertbastian: where does this trigger?

.github/workflows/build-test.yml
```diff
@@ -340,7 +340,7 @@ jobs:
     - uses: dart-lang/setup-dart@0a8a0fc875eb934c15d08629302413c671d3f672 # v1.6.5
       with:
         # Keep in sync with artifacts-build.yml
-        sdk: 3.11.0-46.0.dev
+        sdk: 3.13.0-215.0.dev
```
robertbastian, line 342:
```yml
        # Keep in sync with artifacts-build.yml
```

ffi/dart/pubspec.yaml
mosuem: Sure, just wanted to rev to something. I will use the `-wip` tag for now

ffi/dart/analysis_options.yaml
mosuem: https://github.com/mosuem/icu4x/blob/3a2a802fa6f600773d2a390b34d0fb1f562600da/ffi/dart/lib/src/bindings/DateFields.g.dart#L91

tools/make/ffi.toml
```diff
@@ -86,7 +86,7 @@ cd examples/dart
 exec --fail-on-error dart pub get
 exec --fail-on-error dart format --set-exit-if-changed .
 exec --fail-on-error dart analyze
-exec --fail-on-error dart --enable-experiment=record-use build cli bin/example.dart
+exec --fail-on-error dart build cli --target bin/example.dart
```
robertbastian: if we don't have to enable an experiment anymore, can we use a stable toolchain?

.github/workflows/artifacts-build.yml
```diff
@@ -238,6 +238,9 @@ jobs:
 
     - name: Install Dart
       uses: dart-lang/setup-dart@0a8a0fc875eb934c15d08629302413c671d3f672 # v1.6.5
+      with:
```
robertbastian: ah ok, we seem to use the stable toolchain for docs since https://github.com/unicode-org/icu4x/pull/7243, let's keep doing that then and remove the comment in `build-test.yml`.

ffi/dart/analysis_options.yaml
robertbastian: can we fix this in diplomat (later)? what's the fix?

robertbastian: I think it's this
```dart
  core.List<Object> get _fieldsForLifetimeA => [?era, ?monthCode];
```

mosuem: Yes exactly.

tools/make/ffi.toml
mosuem: The stable record_use will be on Dart stable in August/end of July. Then intl4x can also go into `1.0` without an alpha.

robertbastian (approved):

robertbastian (commented):
this seems to break tree-shaking

ffi/dart/analysis_options.yaml
robertbastian: https://github.com/rust-diplomat/diplomat/pull/1178

tools/make/ffi.toml
robertbastian: if it's not stable we need to keep the experiment flag

```suggestion
exec --fail-on-error dart --enable-experiment=record-use build cli --target bin/example.dart
```

ffi/dart/pubspec.yaml
```diff
@@ -28,11 +28,11 @@ dependencies:
   crypto: ^3.0.7
   ffi: ^2.1.4
   # Pinned as we use unstable APIs
-  hooks: '>=1.0.0 <1.0.1'
+  hooks: ^2.0.2
```
robertbastian: issue: needs to be pinned

ffi/dart/analysis_options.yaml
robertbastian: let's remove the ignore and live with the warning until we get the diplomat change in

ffi/dart/pubspec.yaml
mosuem: package:hooks is stable by now, so this should be fine

mosuem:
> this seems to break tree-shaking

Apparently package:record_use works very differently now. I filed https://github.com/rust-diplomat/diplomat/pull/1179

ffi/dart/hook/link.dart
```diff
@@ -24,25 +25,28 @@ Future<void> main(List<String> args) async {
...
-        )]
-        ?.whereType<record_use.InstanceConstantReference>()
-        .map((instance) => instance.instanceConstant)
-        .whereType<record_use.InstanceConstant>()
-        .map(
-          (instanceConstant) =>
-              instanceConstant.fields['symbol'] as record_use.StringConstant,
+        .recordedUses;
+    if (recordedUses == null) {
+      throw ArgumentError(
```
robertbastian: issue: don't throw an error here. the library should be useable on stable without tree shaking

ffi/dart/pubspec.yaml
robertbastian: there's still an `// ignore: experimental_member_use` in our code

mosuem: This is only about `recordedUses`, which is going to stable in the next days/weeks. It's unlikely that will change. I can still pin the dependency.

robertbastian (approved):

Manishearth (approved):

PR #8121, opened 2026-06-22
Fix `CodePointInversionList::contains_set`

robertbastian:
A single range in the superset can contain multiple ranges of the subset.

## Changelog

`icu_collections`:
* Fix a bug in `CodePointInversionList::contains_set`


sffc (approved):
oof. Let's land this but let's also add more thorough tests on this fn. This is old code so this type of bug should not have sit around so long.

PR #8122, opened 2026-06-22
Remove unused PatternBackend::Error associated type

sffc:
There were two associated types, one on the pattern, one on the placeholder value provider.

Right now, we only have use cases for an error on the placeholder value provider:

1. A missing placeholder value
2. Bubbling up errors from inner writeables (new)

The associated error on the pattern _might_ come up in the future if a particular pattern type does some sort of lazy validation of the syntax. But, _it wasn't being used in that way_, I would rather add that back in when we actually need it. We don't need it now, so let's simplify the traits.

This is technically semver-breaking. Code such as the following will break:

```rust
let err: DoublePlaceholder::Error = // ...
```

However, I don't know why anyone would write that, at least for Single and Double, since ::Error was set to Infallible. This is most likely to break generic code, of which we have none in ICU4X (since this PR passes CI).

If you want to reduce the semver breakage, I can leave the associated type, mark it deprecated, and not use it anywhere. I'm not actually sure if this makes the PR 100% semver-safe, though, because the bounds are still changing.

## Changelog

icu_pattern: Remove public associated Error type from sealed trait PatternBackend

### DIFF
--- a/components/pattern/src/common.rs
+++ b/components/pattern/src/common.rs
@@ -78,9 +78,6 @@ pub trait PatternBackend: crate::private::Sealed + 'static + core::fmt::Debug {
     #[cfg(feature = "alloc")]
     type PlaceholderKeyCow<'a>;
 
-    /// The type of error that the [`TryWriteable`] for this backend can return.
-    type Error<'a>;
-
     /// The unsized type of the store required for this backend, usually `str` or `[u8]`.
     // Note: it is not good practice to feature-gate trait types, but this trait is sealed
     #[doc(hidden)] // TODO(#4467): Should be internal
--- a/components/pattern/src/double.rs
+++ b/components/pattern/src/double.rs
@@ -309,7 +309,6 @@ impl PatternBackend for DoublePlaceholder {
     type PlaceholderKey<'a> = DoublePlaceholderKey;
     #[cfg(feature = "alloc")]
     type PlaceholderKeyCow<'a> = DoublePlaceholderKey;
-    type Error<'a> = Infallible;
     type Store = str;
     type Iter<'a> = DoublePlaceholderPatternIterator<'a>;
 
--- a/components/pattern/src/frontend/mod.rs
+++ b/components/pattern/src/frontend/mod.rs
@@ -258,9 +258,9 @@ where
     pub fn try_interpolate<'a, P>(
         &'a self,
         value_provider: P,
-    ) -> impl TryWriteable<Error = B::Error<'a>> + fmt::Display + 'a
+    ) -> impl TryWriteable<Error = P::Error> + fmt::Display + 'a
     where
-        P: PlaceholderValueProvider<B::PlaceholderKey<'a>, Error = B::Error<'a>> + 'a,
+        P: PlaceholderValueProvider<B::PlaceholderKey<'a>> + 'a,
     {
         WriteablePattern::<B, P> {
             store: &self.store,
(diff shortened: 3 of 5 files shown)


Manishearth:
Not fully sure on whether we should do this without a corresponding semver breakage. I think we proooobably can given the way this crate is used and where this type is. Deprecation may be slightly safer and I think the bounds change is definitely too niche to matter.

Manishearth (approved):

robertbastian (commented):
can we not just bump the crate version? I don't think we have icu_pattern types on public APIs outside of provider types

Manishearth:
It's an annoying cost during google3 and chromium upgrades so I have a slight preference to not (I go into this in depth on the thread about the pattern extraction code)

but this is a decently motivated breakage imo. only reason I'm trying to avoid it is that there are decent routes like deprecation available to us.

but just removing it would be nice.

robertbastian:
I guess I should have left a blocking review. Can we not break semver please

Manishearth:
Right, sorry, if we're deleting the assoc type (not deprecating it) we should do a server bump. I don't like it but it's well motivated.

sffc:
> I guess I should have left a blocking review. Can we not break semver please

Thank you for not blocking this, because this blocks display names, and I shouldn't be blocked on display names due to a change in an unstable crate.

I think the smallest change that would unblock displaynames is changing all the bounds that this PR does, but leaving the associated type as deprecated. I offered to do that in the OP. It makes the PR slightly less likely to break, but there might still be code that does things like

```rust
pub fn custom_write_with_single_pattern<P>(values: P)
where
    P: PlaceholderValueProvider<SinglePlaceholder::PlaceholderKey<'a>, Error = SinglePlaceholder::Error<'a>>
{ ... }
```

In other words, code that copied our bound out into userland in order to make their code generic. However, that code will still compile, and I think it will still work, because the bound has gotten weaker (we no longer require `Error =`, which is the constraint I needed to relax for display names).

sffc:
I made https://github.com/unicode-org/icu4x/issues/8130 to track whether to consider this semver or not.

sffc:
> I think the smallest change that would unblock displaynames is changing all the bounds that this PR does, but leaving the associated type as deprecated.

I think the smallest _smallest_ change would be changing just this bound:

```rust
impl<'a, B, P> TryWriteable for WriteablePattern<'a, B, P>
```

and leaving the fn bounds intact, because I can add new fns with the correct bounds.

And the change here that completely avoids semver is to export a bunch of new types; we should do that for a stable crate but I don't want to do that for an unstable crate.

Manishearth:
> I think the smallest change that would unblock displaynames is changing all the bounds that this PR does, but leaving the associated type as deprecated. 

Yes, as I said earlier I think this is fine. The bounds changing is minor enough for an internal crate.

Manishearth:
> don't want to do that for an unstable crate.

Can we not use the term "unstable crate" here? It is clear the team is not aligned on what our policies are for crates like this so using the term is making things more murky imo.

sffc:
@Manishearth, I'm having trouble following what you are asking for.

Among these options for code:

1. Option 1: what landed in this PR (remove the associated type)
2. Option 2: deprecate the associated type; relax all of the bounds
3. Option 3: keep the associated type; relax the bounds only on WriteablePattern; add new fns with the correct bounds to be used by TryWriteable
4. Option 4: add sibling types and fns with the correct bounds, leaving all existing types and fns the same

And among these options for versioning:

1. Keep the crate at its current semver
2. Bump to the next 0.x semver

can you please clarify which set of options are _acceptable_ and _preferred_ to you?

And also @robertbastian.

Manishearth:
Option 1: Acceptable with a semver bump. Potentially acceptable without a semver bump, much dispreferred.

 
Option 2: Acceptable without a semver bump. Preferred. I do not consider the bounds change to be severe enough to need a bump. 


I generally want to avoid a semver bump, and if we are avoiding a semver bump, I want to avoid actual breakages.

I hadn't particularly considered Option 3 and 4, but:

Option 3: Acceptable, dispreferred since it sounds like a worse API (I could be wrong)

Option 4: Acceptable, dispreferred for the same reason as 3, dispreferred more than 3

sffc:
Discussion:

- @robertbastian This is an unstable crate, for this reason. So just bump semver. Unless bumping semver would cause other crates to break, if they export these types in their APIs?
- @sffc I don't know. I don't think so but I'm not sure. Will cargo-semver-checks fail?
- @robertbastian I'm pretty sure it will, because you change the return type of the method
- @sffc Hopefully it's smart enough to track workspace dependency version breakages
- @robertbastian It will complain about the provider types.
- @sffc I had some back-and-forth with @Manishearth about whether or not we should consider it a cost to bump semver on this crate
- @robertbastian I hope we've been careful enough with this being unstable that we can just bump the version

Conclusion: bump the version of icu_pattern, check the semver report, and consider another option if bumping is infeasible due to breaking other crates

PR #8123, opened 2026-06-22
Support TryWriteable inputs in Single and Double placeholder patterns

sffc:
Depends on #8122

There are multiple ways to do this. With this PR, the user needs to wrap the TryWriteables in a named tuple struct, since it needs a different implementation of PlaceholderValueProvider than the plain tuple. I tried another approach where `interpolate` and `try_interpolate` accept _different traits_, which means that you can pass a plain tuple into both. You can see that approach [here](https://github.com/sffc/omnicu/tree/pattern-try). I am not sure which approach I prefer. I put up this PR because it doesn't involve semver breakage, although the diffbase PR #8122 is technically a semver breakage.

## Changelog

icu_pattern: new input structs DoublePlaceholderValueProviderTry, SinglePlaceholderValueProviderTry

### DIFF
--- a/components/pattern/src/double.rs
+++ b/components/pattern/src/double.rs
@@ -7,8 +7,8 @@
 use core::convert::Infallible;
 use core::{cmp::Ordering, str::FromStr};
 use either::Either;
-use writeable::Writeable;
 use writeable::adapters::WriteableAsTryWriteableInfallible;
+use writeable::{TryWriteable, Writeable};
 
 use crate::Error;
 use crate::common::*;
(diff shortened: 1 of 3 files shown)


components/pattern/src/double.rs
```diff
@@ -131,6 +131,60 @@ where
...
+///
+/// ```
+/// use either::Either;
+/// use icu_pattern::DoublePlaceholderPattern;
+/// use icu_pattern::DoublePlaceholderValueProviderTry;
+/// use writeable::assert_try_writeable_eq;
+///
+/// let pattern = DoublePlaceholderPattern::try_from_str("{0}-{1}", Default::default()).unwrap();
+/// assert_try_writeable_eq!(
+///     pattern.try_interpolate(DoublePlaceholderValueProviderTry(Ok::<&str, &str>("xxx"), Err::<&str, &str>("yyy"))),
```
Manishearth, line 148: issue: it's hard for me to understand what this example is doing; can you use `let simple_writeable = Ok(..)` and `let simple_writeable_errory = Err(..)` or something to make it clear that these are placeholder TryWriteables (and include a comment)

components/pattern/src/double.rs
```diff
@@ -131,6 +131,60 @@ where
...
+///     "xxx-yyy",
+///     Err(Either::Right("yyy"))
+/// );
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[allow(clippy::exhaustive_structs)] // holds two placeholder values
+pub struct DoublePlaceholderValueProviderTry<W0, W1>(pub W0, pub W1);
```
Manishearth, line 155: thought: not sure I like the name. TryWrap maybe? It's already pretty long, though.

this crate likes long names so not a big deal

components/pattern/src/double.rs
```diff
@@ -131,6 +131,60 @@ where
     }
 }
 
+/// A pair of values for [`DoublePlaceholder`] that may bubble up errors.
```
Manishearth, line 134: nit: we should link to these two types in the docs for `try_interpolate`

Manishearth (approved):
This seems like a fine approach. The other one feels heavier, but it might be easier to "trace" for someone reading the docs. I'm not opposed to either, I think this is simpler.

components/pattern/src/double.rs
sffc: Were you thinking:

```rust
pub struct TryWrap<T>(pub T);

impl<W> PlaceholderValueProvider<SinglePlaceholder> for TryWrap<W> ...

impl<W0, W1> PlaceholderValueProvider<DoublePlaceholder> for TryWrap<(W0, W1)> ...
```

Manishearth: No I was thinking DoublePlaceholderProviderTryWrap, but ProviderTryWrap/TryWrapProvider/TryWrap implemented the way you put forth might actually be nicer.

sffc:
I'm going to land this and use #8130 to follow up on the types.

PR #8126, opened 2026-06-23
Validate japanese era data struct earlier

robertbastian:
Follow-up to #8095 

## Changelog

N/A

### DIFF
--- a/components/calendar/src/provider.rs
+++ b/components/calendar/src/provider.rs
@@ -15,6 +15,7 @@
 // Provider structs must be stable
 #![allow(clippy::exhaustive_structs, clippy::exhaustive_enums)]
 
+use core::fmt::Debug;
 use core::num::NonZeroU8;
 
 use crate::types::Weekday;
@@ -51,7 +52,7 @@ icu_provider::data_marker!(
     /// Modern Japanese era names
     CalendarJapaneseModernV1,
     "calendar/japanese/modern/v1",
-    JapaneseEras<'static>,
+    JapaneseEras,
     is_singleton = true
 );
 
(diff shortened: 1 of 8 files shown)


components/calendar/src/provider.rs
```diff
@@ -105,78 +105,235 @@ pub struct EraStartDate {
...
 
-impl JapaneseEras<'_> {
-    pub(crate) const fn last_after_reiwa(&self) -> Result<Option<PackedEra>, DataError> {
-        let Some(&zerovec::ule::tuple::Tuple2ULE(start, code)) =
-            self.dates_to_eras.as_slice().as_ule_slice().last()
-        else {
-            return Ok(None);
+impl JapaneseEras {
+    /// Creates a new [`JapaneseEras`] without a post-Reiwa era.
+    pub const fn up_to_reiwa() -> Self {
```
sffc, line 118: Thought/Issue: right now this is `reiwa` but really it is "the latest era known to this icu_calendar release"

sffc (approved):
Question, this serializes only the post-reiwa eras, but allows additional eras when deserializing?

Seems good, slight preference to not change the serialized form

robertbastian:
no, this serialises all eras, otherwise there'd be a diff in this PR (both in test data and the baked data hash)

it only deserialises the last era, as we only store that

components/calendar/src/provider.rs
robertbastian: yes and we can change that because it's an unstable provider struct

sffc:
Yes, because there is a data diff in this PR? (baked only, not json)

robertbastian:
when you say "serialize" and "deserialize" I interpret that as serde, for which my comment stands.

it only *bakes* the last era

Manishearth:
Looks like this means `unstable`'s signature will change if we add more eras (with a grace period of 1). I think that's fine.

robertbastian:
how so? if there's a new era, we change the `JapaneseEras` constructors and its `eras` method to take the new era into account, but the data marker stays stable. if there are two new eras before we make any code changes, we only respect the last one, but that's not new in this PR

Manishearth:
Oh, true, `JapaneseEras` has private fields.

PR #8127, opened 2026-06-23
Loose matching for property names in unicodeset

robertbastian:
Might not be the best way to do this, but I need this for segmenter and it's still unstable code anyway.

## Changelog

N/A


Manishearth (approved):
ick. but sure.

if you can think about a way to keep it cleaner that would be nice. Maybe a `set_field(&short, &long, &mut field)` helper function

PR #8129, opened 2026-06-24
Use pseudo symbols for complex language detection

robertbastian:
This continues #8076. Instead of having constants for complex symbols, we use the pseudo-symbol mechanism that was introduced for tailorings, and associate each pseudo symbol with a `Language`. This scales better than constants, because the state machines don't have a fixed number of complex classes, and it's also nice to not have a third type of state.

## Changelog

N/A


provider/source/src/segmenter/mod.rs
```diff
@@ -993,7 +993,7 @@ impl SourceDataProvider {
             .sentence
             .get_or_init(|| {
                 self.build_segmenter(&neo_sources(), "SentenceBreak", |s| {
-                    if s == "EOL" { 1 } else { 0 }
+                    if s == "Nonterminated" { 1 } else { 0 }
```
Manishearth, line 994: explain?

components/segmenter/src/neo/line.rs
```diff
@@ -568,55 +568,22 @@ impl<Y: RuleBreakType> ComplexHandler<Y> for ComplexLine<Y> {
...
     fn handle<'data, 's>(
-        symbol: Symbol,
-        dfa: &RuleBreakIterator<'_, '_, Y, Self>,
-        data: &Self::Data<'data>,
+        data: &Self::LanguageData<'data>,
         iter: Y::IterAttr<'s>,
-    ) -> Option<(ComplexIterator<'data, 's, Y>, Y::IterAttr<'s>)> {
-        use crate::provider::Language;
-
-        let data = Y::select_complex(
```
Manishearth, line 597: observation: this part isn't necessary anymore since the language is passed in

the next part is preserved in the `next()` impl

Manishearth (approved):
Couldn't review datagen diff in depth, but this seems fine.

provider/source/src/segmenter/mod.rs
robertbastian: https://github.com/unicode-org/icu4x/pull/8129/changes/7971fe0322c030f2688827a51be70da8d990d43b#diff-47727a3338fdfa3f71e170c3b7e7d213b1cbe08dcbb1d17866eb9f105ac06747

Manishearth: ah just a rename

Manishearth (approved):

robertbastian:
I think you might have had a race condition with the last two commits?

components/segmenter/src/complex/script.rs
```diff
@@ -2,14 +2,14 @@
 // called LICENSE at the top level of the ICU4X source tree
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
-use crate::provider::Language;
+use crate::provider::ComplexScript;
 
 // TODO: Use data provider
-pub(crate) fn get_language(codepoint: u32) -> Language {
-    // For Thai, Burmese, Lao and Khmer, these are the intersections
+pub(crate) fn get_script(codepoint: u32) -> ComplexScript {
```
Manishearth: I thought our LSTMs were language-specific, not script-specific

(though everyone probably uses them in a script-specific way)

robertbastian: actually let's make that a separate PR

PR #8006, opened 2026-05-29
Split singular display names into owned and borrowed types

sffc:
This PR splits `ScriptDisplayName` and `RegionDisplayName` in the `icu_experimental::displaynames` module into owned and borrowed variants. This follows a pattern where the owned type holds the `DataPayload` and the borrowed type holds a reference to the data string. This is useful for scenarios where only a borrowed reference is needed, avoiding cloning or owning the payload.

🤖 This pull request was created by an AI agent working with @sffc.

## Changelog
icu_experimental/displaynames: Split singular display names into owned and borrowed types
  * New types: `ScriptDisplayNameOwned`, `ScriptDisplayName<'a>`, `RegionDisplayNameOwned`, `RegionDisplayName<'a>`
  * New methods: `ScriptDisplayNameOwned::as_borrowed()`, `RegionDisplayNameOwned::as_borrowed()`

PR #8132, opened 2026-06-24
Implement BCP-47 fallback for LanguageIdentifierDisplayName

sffc:
Depends on #8135

This pull request is part of the implementation for the main locale display names tracking issue: #7825.

This pull request implements full BCP-47 subtag fallback hierarchy for the new `LanguageIdentifierDisplayName` (and its owned version), resolving the core fallback requirements for the experimental single displaynames module.

### Key Changes
1.  **BCP-47 Fallback Hierarchy**: Implements step-by-step subtag fallback (e.g., language+script+region -> language+region -> language -> code) during formatting.
2.  **LossyWrap Dual-Trait Implementation (API Design)**: 
    *   Following design feedback from the `writeable` crate, `LanguageIdentifierDisplayName` now implements both **`TryWriteable`** (for strict error detection and error part annotation) and **`Writeable`** / **`Display`** (for standard lenient formatting with raw code fallback) directly on the struct.
    *   Under the hood, `LanguageIdentifierDisplayName` is structured as a tuple newtype wrapping an inner struct in `LossyWrap`, leveraging `writeable::impl_try_writeable_delegate!` and `writeable::impl_writeable_delegate!`.
    *   Removed `Writeable` and `TryWriteable` from `LanguageIdentifierDisplayNameOwned` (callers use `.as_borrowed()` to format).
3.  **Renamed Fallback Error Type**:
    *   Renamed `DisplayNamesFallbackError` to **`LanguageIdentifierNameFallbackError`** (correcting the spelling of "Identifier" to match the formatter).
    *   Removed the `pub const fn new()` constructor and adopted struct syntax `LanguageIdentifierNameFallbackError` for construction, while suppressing the `clippy::exhaustive_structs` warning.
4.  **Idiomatic Result Formatting**:
    *   Leveraged the built-in `TryWriteable` implementation for `Result<T, E>` in the `writeable` crate to simplify the manual match-writing and `Part::ERROR` annotation blocks in both `QualifiersWriteable` and `LanguageIdentifierDisplayName` formatting code, making it extremely concise and clean.
5.  **Boilerplate Elimination**: Reuses the standalone constructors of `ScriptDisplayNameOwned`, `RegionDisplayNameOwned`, and `VariantDisplayNameOwned` in the main constructor to load sub-parts, eliminating hundreds of lines of duplicate data-loading code.
6.  **Error Annotation**: Annotates missing subtags in the output with `Part::ERROR` during strict formatting, enabling advanced styling in downstream applications.
7.  **Polished Tests**: Rewrote the integration test loop and doc tests to utilize direct formatting and `borrowed.try_write_to_string()` for clean, idiomatic fallback verification.

## Changelog

writeable: Add standard derives and `repr(transparent)` to `LossyWrap`
  * `LossyWrap<T>` now derives `Clone`, `Copy`, `PartialEq`, `Eq`, `PartialOrd`, `Ord`, and `Hash`, and is marked `#[repr(transparent)]`, enabling its use inside `Copy` structs and with delegate macros.
  * Added error mapping fn to `impl_try_writeable_delegate!`

icu_experimental/displaynames: Implement BCP-47 fallback for `LanguageIdentifierDisplayName`
  * New types: `LanguageIdentifierNameFallbackError`
  * Trait implementations:
    * `LanguageIdentifierDisplayName` implements `TryWriteable`, `Writeable`, and `Display` directly using `writeable::impl_delegate` macros over an internal `LossyWrap` field.
    * Removed `Writeable` and `TryWriteable` from `LanguageIdentifierDisplayNameOwned` (use `.as_borrowed()` to format).
  * Constructor behavior: The constructors `try_new()`, `try_new_with_buffer_provider()`, and `try_new_unstable()` no longer fail when the display name is not found.


components/experimental/src/displaynames/single/language.rs
```diff
@@ -87,20 +119,21 @@ impl LanguageIdentifierDisplayNameOwned {
...
                 (subject.language, Some(subject.script), Some(subject.region)),
                 (subject.language, Some(subject.script), None),
                 (subject.language, None, Some(subject.region)),
             ] {
                 // For Script and Region:
-                // - Some in the first position means "this should be present"
+                // - Some(Some(subtag)) in the first position means "this should be present"
+                // - Some(None) in the first position means "this must be absent"
                 // - None in the first position means "skip this field"
-                // We skip Some(None) because that case will be handled in a subsequent iteration
```
sffc, line 105: Restore this comment back to the way it was

components/experimental/src/displaynames/single/language.rs
```diff
@@ -308,11 +401,17 @@ impl<'a> writeable::Writeable for QualifiersWriteable<'a> {
...
                 sink.write_str(separator_str)?;
             }
-            sink.write_str(item)?;
+            match res {
+                Ok(s) => sink.write_str(s)?,
+                Err(code) => {
+                    fallback_occurred = true;
+                    sink.with_part(Part::ERROR, |sink| sink.write_str(code))?;
+                }
+            }
```
sffc: If possible, please try to use `res.try_write_str` here (using impl TryWriteable for Result)

components/experimental/src/displaynames/single/language.rs
```diff
@@ -325,38 +424,72 @@ impl<'a> writeable::Writeable for QualifiersWriteable<'a> {
...
 
+        let mut fallback_occurred = false;
+
         if !has_qualifiers {
-            sink.write_str(self.base_name)
+            match self.base_name {
+                Ok(s) => sink.write_str(s)?,
+                Err(code) => {
+                    fallback_occurred = true;
+                    sink.with_part(Part::ERROR, |sink| sink.write_str(code))?;
```
sffc: please leverage  impl TryWriteable for Result

components/experimental/src/displaynames/single/language.rs
```diff
@@ -325,38 +424,72 @@ impl<'a> writeable::Writeable for QualifiersWriteable<'a> {
...
-                .write_to(sink)
+
+            let interpolated =
+                self.locale_pattern
+                    .try_interpolate(DoublePlaceholderValueProviderTry(
+                        &self.base_name,
+                        &qualifiers,
+                    ));
+
+            fallback_occurred = interpolated.try_write_to_parts(sink)?.is_err();
```
sffc: Try chaining these methods instead of a local variable `interpolated`

components/experimental/src/displaynames/single/language.rs
sffc: I had to do this manually

sffc:
I've gone over this multiple times. I'm happy with this now and ready for a review.

Manishearth (approved):
Got a little confused around the error types for a bit but no, this is correct.

components/experimental/src/displaynames/single/language.rs
```diff
@@ -35,18 +54,20 @@ use tinystr::TinyAsciiStr;
...
 #[derive(Debug)]
 pub struct LanguageIdentifierDisplayNameOwned {
     formatting_locale: DataLocale,
     options: LanguageIdentifierDisplayNameOptions,
-    language_payload: DataPayload<LocaleNamesLanguageMediumV1>,
-    script_payload: DataPayloadOr<LocaleNamesScriptMediumV1, ()>,
-    region_payload: DataPayloadOr<LocaleNamesRegionMediumV1, ()>,
-    variant_payloads:
-        DataPayloadOr<LocaleNamesVariantMediumV1, Vec<DataPayload<LocaleNamesVariantMediumV1>>>,
+    language_payload: DataPayloadOr<LocaleNamesLanguageMediumV1, TinyAsciiStr<4>>,
```
robertbastian: strong types please

```suggestion
    language_payload: DataPayloadOr<LocaleNamesLanguageMediumV1, Language>,
```

components/experimental/src/displaynames/single/language.rs
```diff
@@ -154,57 +175,85 @@ impl LanguageIdentifierDisplayNameOwned {
...
-                    .payload
+                let result = provider.load(DataRequest {
+                    id: DataIdentifierBorrowed::for_marker_attributes_and_locale(
+                        attrs,
+                        &formatting_locale,
+                    ),
+                    ..Default::default()
+                });
+                match result {
+                    Ok(response) => DataPayloadOr::from_payload(response.payload),
```
robertbastian: `allow_identifier_not_found`, didn't we already go through 4 rounds of this in the last PR?

components/experimental/src/displaynames/single/language.rs
```diff
@@ -154,57 +175,85 @@ impl LanguageIdentifierDisplayNameOwned {
...
                     .collect::<Result<Vec<_>, _>>()?;
                 DataPayloadOr::from_other(payload_vec)
             } else {
                 // 1 variant
-                // TODO(#8100): Fall back to the code instead of dropping it if the variant name is not found
-                DataPayloadOr::from_payload(result?.payload)
+                match first_val.into_inner() {
+                    Ok(payload) => DataPayloadOr::from_payload(payload),
+                    Err(fallback_code) => {
+                        DataPayloadOr::from_other(vec![DataPayloadOr::from_other(fallback_code)])
```
robertbastian, line 282: where did the single-variant optimisation go? you're again allocating

components/experimental/src/displaynames/single/language.rs
```diff
@@ -232,44 +281,55 @@ impl LanguageIdentifierDisplayNameOwned {
...
-        self.as_borrowed().write_to(sink)
-    }
 
+    /// Returns a writeable that formats the display name.
+    ///
+    /// Missing display names will fall back to the raw BCP-47 code.
     #[inline]
-    fn writeable_length_hint(&self) -> writeable::LengthHint {
-        self.as_borrowed().writeable_length_hint()
+    pub fn as_borrowed_with_fallback(&self) -> LossyWrap<LanguageIdentifierDisplayName<'_>> {
```
robertbastian: I'd prefer if we returned `impl Writeable` here instead of exposing the `LossyWrap` implementation detail

components/experimental/src/displaynames/single/language.rs
```diff
@@ -9,11 +9,30 @@ use crate::displaynames::single::{
...
+#[non_exhaustive]
+enum AbsentOrFallback {
+    /// The subtag was not present in the subject.
+    Absent,
+    /// The subtag was present, but its display name was not found, so we fall back to the code.
+    Fallback(TinyAsciiStr<4>),
+}
+
+/// Represents a payload that was either successfully loaded or has fallen back to its code.
+type PayloadOrFallback<M> = DataPayloadOr<M, TinyAsciiStr<8>>;
```
robertbastian: this is only used for variants


```suggestion
type PayloadOrFallback<M> = DataPayloadOr<M, Variant>;
```

components/experimental/src/displaynames/single/language.rs
```diff
@@ -9,11 +9,30 @@ use crate::displaynames::single::{
...
+
+/// Represents a subtag that is either absent or has fallen back to its code.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+enum AbsentOrFallback {
+    /// The subtag was not present in the subject.
+    Absent,
+    /// The subtag was present, but its display name was not found, so we fall back to the code.
+    Fallback(TinyAsciiStr<4>),
+}
```
robertbastian: please use strong typing to not accidentally mix subtags

```suggestion
enum AbsentOrFallback<T> {
    /// The subtag was not present in the subject.
    Absent,
    /// The subtag was present, but its display name was not found, so we fall back to the code.
    Fallback(T),
}
```

components/experimental/tests/displaynames/tests.rs
```diff
@@ -177,19 +181,48 @@ fn test_concatenate() {
...
+        let borrowed = single_display_name.as_borrowed();
+        assert_writeable_eq!(borrowed.with_fallback(), cas.expected);
+        match borrowed.try_write_to_string() {
+            Ok(s) => {
                 assert_eq!(cas.single_should_err, false, "{cas:?}");
-                assert_writeable_eq!(single_display_name, cas.expected);
+                assert_eq!(s, cas.expected);
             }
             Err(_) => {
                 assert_eq!(cas.single_should_err, true, "{cas:?}");
```
robertbastian: this also returns a string, which you throw away instead of asserting. we have `assert_try_writeable_eq` for `TryWriteable`s, use it

components/experimental/tests/displaynames/tests.rs
```diff
@@ -177,19 +181,48 @@ fn test_concatenate() {
...
+#[test]
+fn test_fallback_parts() {
+    let locale = locale!("en-001");
+    let options = LanguageIdentifierDisplayNameOptions::default();
+
+    // xx-YY has both language and region missing in CLDR en data.
+    // It should fall back to "xx (YY)" and annotate "xx" and "YY" with Part::ERROR.
+    let display_name = LanguageIdentifierDisplayNameOwned::try_new(
+        locale.into(),
+        "xx-YY".parse().unwrap(),
```
robertbastian: can you add one non-error subtag to this test as well

components/experimental/src/displaynames/single/README.md
```diff
@@ -100,6 +100,47 @@ All owned constructors take their target subtag or `LanguageIdentifier` **by val
 
 ---
 
+## Fallback & `TryWriteable`
+
+The `LanguageIdentifierDisplayName` formatter supports robust BCP-47 subtag fallback hierarchy when localized display names are missing from the data provider (resolving [#8100](https://github.com/unicode-org/icu4x/issues/8100)).
```
robertbastian: what makes this "robust"? what would be non-robust? please no LLM fluff

robertbastian: documentation should be stateless and not explain which issues it closes

components/experimental/src/displaynames/single/README.md
```diff
@@ -100,6 +100,47 @@ All owned constructors take their target subtag or `LanguageIdentifier` **by val
...
+```rust
+use writeable::TryWriteable;
+use icu::experimental::displaynames::single::LanguageIdentifierNameFallbackError;
+
+let mut sink = String::new();
+let result = display_name.as_borrowed().try_write_to(&mut sink);
+
+if let Ok(Err(LanguageIdentifierNameFallbackError)) = result {
+    println!("Fallback occurred! Output: {}", sink);
+}
```
robertbastian: is this code executed? will it keep working?

components/experimental/src/displaynames/single/README.md
```diff
@@ -100,6 +100,47 @@ All owned constructors take their target subtag or `LanguageIdentifier` **by val
 
 ---
 
+## Fallback & `TryWriteable`
+
+The `LanguageIdentifierDisplayName` formatter supports robust BCP-47 subtag fallback hierarchy when localized display names are missing from the data provider (resolving [#8100](https://github.com/unicode-org/icu4x/issues/8100)).
+Unlike the old multi-formatter (`LocaleDisplayNamesFormatter`), the single formatter does not expose a `fallback` toggle in its options bag (`LanguageIdentifierDisplayNameOptions`).
+Instead, it always performs BCP-47 fallback (lenient mode) under the hood to ensure a display name can be formatted.
+
+### Detecting Fallback with `TryWriteable`
```
robertbastian, line 107: this is useful information, why is it not in rustdoc?

components/experimental/src/displaynames/single/README.md
```diff
@@ -227,8 +268,10 @@ The following features defined in UTS #35 are currently not supported and are pl
     In `LanguageIdentifierDisplayNameOwned::try_new(prefs, locale_id, options)`, we have placed `options` last.
     *   *Resolution*: This aligns with the standard ICU4X API style, as `options` behaves like a trailing optional bag.
 3.  **Fallback Behavior in Single Formatters**:
-    Currently, single formatters (`Script`, `Region`, `Variant`, `Language`) fail-fast in the constructor (`try_new` returns `Err(DataError)`) if the specific subtag data is missing from the provider (e.g., `xx` or an untranslated language).
-    *   *Resolution*: We will redesign the single formatters to support falling back to the code instead of failing with `DataError`. This has been deferred to a follow-up issue: [#8100](https://github.com/unicode-org/icu4x/issues/8100).
+    We have implemented full fallback support in `LanguageIdentifierDisplayName` (and `LanguageIdentifierDisplayNameOwned` via adapters) using `TryWriteable` (resolving [#8100](https://github.com/unicode-org/icu4x/issues/8100) for the main formatter).
```
robertbastian: funny that this PR has a bigger "Open questions" section than before. LLMs sure love to write

components/experimental/src/displaynames/single/language.rs
```diff
@@ -154,57 +175,85 @@ impl LanguageIdentifierDisplayNameOwned {
...
             }
         };
 
         // Step 2: Load script name (if present in subject)
-        // TODO(#8100): Fall back to the code instead of failing with DataError if the script name is not found
         let script_payload = if let Some(script) = subject.script {
-            DataPayloadOr::from_payload(
-                ScriptDisplayNameOwned::try_new_unstable(provider, prefs, script)?.payload,
-            )
+            let res = ScriptDisplayNameOwned::try_new_unstable(provider, prefs, script);
```
robertbastian: for language you call your intermediate result `result`, but here it's `res`. a minimum of consistency please

components/experimental/src/displaynames/single/language.rs
robertbastian: this API doesn't seem to add anything over `.as_borrowed().with_fallback()` (or `.with_fallback().as_borrowed()`?), why does it exist?

components/experimental/src/displaynames/single/language.rs
```diff
@@ -282,26 +342,42 @@ impl BorrowedVariants<'_> {
...
-    region_name: Option<&'a str>,
+    base_name: Result<&'a str, &'a str>,
+    script_name: Option<Result<&'a str, &'a str>>,
+    region_name: Option<Result<&'a str, &'a str>>,
     variants: BorrowedVariants<'a>,
     locale_pattern: &'a DoublePlaceholderPattern,
     locale_separator: &'a DoublePlaceholderPattern,
 }
 
+impl<'a> LanguageIdentifierDisplayName<'a> {
```
robertbastian: why is this in its own impl block? it's hard to follow which methods are on which type, this should be next to `as_borrowed_with_fallback`

robertbastian (changes requested):

components/experimental/src/displaynames/single/README.md
Manishearth: isn't the goal of this document to be an LLM designlog primarily for feeding back to LLMs?

thought: we should adopt some prefix, README.agents.md, DESIGN.agents.md, for stuff like this.

"Have the LLM write design docs primarily for LLM consumption" is a useful tactic. I've mostly been maintaining these docs out of tree when it comes to ICU4X but it makes sense to be able to check stuff in at least temporarily.

It also makes sense to have per-module LLM-generated design descriptions that reduce the amount of code-reading the LLM has to do so there's a use case for more permanent `README.agents.md` files.

robertbastian: > thought: we should adopt some prefix, README.agents.md, DESIGN.agents.md, for stuff like this.

I don't think we should have special information just for agents in the repo. There is useful stuff in here that should be in rustdoc, and available to human users and human contributors as well, and there's stuff in here that probably should not be checked in.

Manishearth: I think it makes sense to extract some of this out into docs. I think this is less "special info just for agents" and more "documents that are written in a verbose style that mention a lot of less-important nitpicks that a human would not need, making them only _useful_ to agents".

robertbastian: I don't want to review such a file though

Manishearth: It was marked as generated, which was why I skipped it entirely.

We would consider `.agents.md` files to be generated and not in need of review. Though the person submitting them should probably do a cursory scan.

robertbastian: `README.md` files are marked as generated because our crate-level ones are. We should probably improve the `.gitattributes`, or not call non-generated files `README.md`

Manishearth: Yes. 

Let me rephrase: it was marked as generated, AND I knew it was an llm devlog, so I skipped it. 

If we had README.agents.md, then that would be even clearer. Could also be AGENTS.md or AGENTS_README.md. Don't care too much about the specifics, I know "AGENTS.md" has preexisting meaning which may not be exactly what we want.

I agree that the gitattributes should be improved.

components/experimental/src/displaynames/single/language.rs
sffc: It only allocates in the error case. The happy path is the same.

I chose this design (the AI didn't suggest it) because I didn't feel it was necessary to create an optimization for the case of a "single variant subtag for which we didn't find data". We have only one special case, which is "single variant subtag for which we have data".

If we want to not allocate in the single-variant error case, I would need to change the field from

```rust
DataPayloadOr<
    LocaleNamesVariantMediumV1,
    Vec<PayloadOrFallback<LocaleNamesVariantMediumV1, Variant>>,
>
```

to

```rust
DataPayloadOr<
    LocaleNamesVariantMediumV1,
    // not actually Either; for illustration purposes only
    Either<Vec<PayloadOrFallback<LocaleNamesVariantMediumV1, Variant>>, Variant>,
>
```

and also add a variant to `BorrowedVariants`.

I can do this if you want me to, but I think I prefer the simpler one.

components/experimental/src/displaynames/single/language.rs
sffc: I told the agent to use a downgraded type here because we don't care what type the subtag is anymore. It's nice for `AbsentOrFallback` to be just one single type. Sure, I can technically upgrade Language and Variant, but then either I need to split `AbsentOrFallback` into two types (for Region and Script, or give it a generic), or I live with the inconsistency that some of the subtags have their upgraded type and some have their downgraded type.

But I'll go ahead and use the upgraded types.

components/experimental/tests/displaynames/tests.rs
sffc: ok, I was trying to reduce the diff in the test file. I agree `assert_try_writeable_eq` is better, it's just a larger diff

components/experimental/src/displaynames/single/README.md
sffc: 1. In this particular PR, I thought it was important for the component's design to explicitly talk about the TryWriteable stuff. I therefore included the design doc update in this PR. If I were _fully_ sequencing the PRs, I would have sent a standalone PR for just the design doc change.
2. I think this should be called DESIGN.md or AGENT.md or similar, not README.md. I plan to rename it later. I don't want to rename it while I have open PRs.
3. I am trying to strike a delicate balance between a doc that is useful for humans and a doc that is useful for LLMs (I've talked with @Manishearth about this before). I fully expect my reviewers to review changes to this doc, and I want it to have a high signal-to-noise ratio.
4. To go one step further, I want to get to a point where we can _primarily_ review a human-readable md file, and the code review is a "check check check" to make sure it implemented the design doc correctly.
5. I agree with @robertbastian that fluffy words like "robust" are not useful except to make an LLM take it more seriously. They come up so often that rather than whack-a-moling them I just kind-of ignore them.

In any event, I slimmed down the change to README.md

components/experimental/src/displaynames/single/README.md
sffc: ugh, I told the LLM to not put code in this file but it ignored me and I didn't catch it

components/experimental/src/displaynames/single/language.rs
sffc: Fixed, and I'm adding this to my code reviewer skill.

components/experimental/src/displaynames/single/language.rs
sffc: I think because `.as_borrowed().with_fallback()` has an intermediate type that can't be dropped due to lifetimes

components/experimental/src/displaynames/single/language.rs
sffc: sorry.

I added this to my reviewer skill.

components/experimental/src/displaynames/single/language.rs
sffc: I thought you might suggest changing how this works, and I'm glad that you did.

The advantage of returning `LossyWrap` is that we don't need to invent more new types, and we benefit from things like a concrete `to_string` impl (https://github.com/unicode-org/icu4x/pull/8140). But, I'll make it return `impl Writeable` for now.

---

Do you mind suggesting a specific type architecture that you will be happy with?

One design I was musing about that puts the Writeable impl first and delegates the TryWriteable to a power user thing:

- LanguageIdentifierDisplayName implements Writeable
- It has a fn `.strict(&self)` that returns something that implements TryWriteable (maybe even an `impl TryWriteable`)

Or, LanguageIdentifierDisplayName can just implement both Writeable and TryWriteable?

If this is going to be a long discussion, I prefer to merge the PR and leave this as an open question.

components/experimental/tests/displaynames/tests.rs
sffc: the lack of this test kept me up last night after I went home and realized it was missing from the PR. Fixed.

components/experimental/src/displaynames/single/language.rs
sffc: `with_fallback` is the only associated method on this type. The type also implements `TryWriteable`.

components/experimental/src/displaynames/single/language.rs
sffc: done

components/experimental/src/displaynames/single/language.rs
sffc: done

components/experimental/src/displaynames/single/language.rs
sffc: It seems we don't need the extra fn, because the intermediate type is `Copy` so we don't need to keep it around for anchoring anything.

I'm angry at my agent because I literally gave it this prompt yesterday:

> remove the Writeable impls, and instead, add a fn to LanguageIdentifierDisplayName (only the borrowed version) called .with_fallback that returns a writeable::adapters::LossyWrap. if you really really need to for ergonomics, you can add .as_borrowed_with_fallback to the Owned struct

When it added `.as_borrowed_with_fallback`, I wrongly took on faith that it needed it for ergonomics, i.e. needing to anchor the intermediate type, but I guess the LLM thought a single function call was nonetheless more ergonomic than `.as_borrowed().with_fallback()`. Or it just didn't listen or care. 🤷

components/experimental/src/displaynames/single/README.md
sffc: fixed

components/experimental/src/displaynames/single/README.md
sffc: Added to rustdoc

components/experimental/src/displaynames/single/README.md
sffc: fixed

sffc:
Of Robert's 16 comments:

- 6 were intentional choices by me
    - I fixed 4 of them the way Robert preferred
    - I left 2 of them open
- 5 were in the fluffy AI-generated README.md file, which I should have scrutinized more
- 2 were AI mistakes that I didn't catch
- 1 was a mistake by me
- 1 was an issue that I noticed but didn't fix in time
- 1 was a question

I think I've fixed them or commented on them.

components/experimental/src/displaynames/single/language.rs
```diff
@@ -24,18 +27,21 @@ pub struct LanguageIdentifierNameFallbackError;
...
     /// The subtag was not present in the subject.
     Absent,
     /// The subtag was present, but its display name was not found, so we fall back to the code.
-    Fallback(TinyAsciiStr<4>),
+    Fallback(S),
 }
 
 /// Represents a payload that was either successfully loaded or has fallen back to its code.
-type PayloadOrFallback<M> = DataPayloadOr<M, TinyAsciiStr<8>>;
+type PayloadOrFallback<M, S> = DataPayloadOr<M, S>;
```
robertbastian: at this point you're just renaming the type, I don't think this alias pulls its weight. you're using `DataPayloadOr` directly for the language, script, and region fields already

components/experimental/src/displaynames/single/language.rs
robertbastian: > It only allocates in the error case. The happy path is the same.

I wouldn't consider the code fallback an "error case", given that it's the default option to use code fallback in the old implementation.

> I can do this if you want me to, but I think I prefer the simpler one.

Yes the `DataPayloadOr` design makes this surprisingly hard, but using `DataPayloadOr` to implement what is basically a `SmallVec` was a bit of a hack in the first place, and just because that hack doesn't work anymore doesn't mean we should throw our hands up and allocate. We already have a `SmallVec` dependency in this crate, so just use that:

```rust
variant_payloads: SmallVec<1, DataPayloadOr<LocaleNamesRegionMediumV1, Variant>>
```
Alternatively, you can split it into two fields:

```rust
first_variant_payload: 
    DataPayloadOr<LocaleNamesRegionMediumV1, Option<Variant>>
additional_variant_payloads: 
    Vec<DataPayloadOr<LocaleNamesRegionMediumV1, Variant>>,
```

components/experimental/tests/displaynames/tests.rs
```diff
@@ -214,15 +214,37 @@ fn test_fallback_parts() {
...
+    options.language_display = Some(LanguageDisplay::Standard);
+
+    // This should format "zh-Hant-HK" to "Chinese (Traditional, Hong Kong SAR China)"
+    // in "en-001" using LanguageDisplay::Standard
+    let lang_id = langid!("zh-Hant-HK");
+    let lang_name = LanguageIdentifierDisplayNameOwned::try_new(locale.into(), lang_id, options)
+        .expect("Data should load successfully");
+
+    assert_writeable_eq!(
+        lang_name.as_borrowed().with_fallback(),
```
robertbastian: you're not testing fallback here, so use `assert_try_writeable_eq`

components/experimental/src/displaynames/single/language.rs
robertbastian: ah, what confused me here is that usually our types are called `Foo` and `FooBorrowed`. Here you have `FooOwned` and `Foo` – why is that?

components/experimental/tests/displaynames/tests.rs
```diff
@@ -174,22 +179,72 @@ fn test_concatenate() {
...
+        assert_writeable_eq!(borrowed.with_fallback(), cas.expected);
+        if cas.single_should_err {
+            assert_try_writeable_eq!(
+                borrowed,
+                cas.expected,
+                Err(LanguageIdentifierNameFallbackError)
+            );
+        } else {
+            assert_try_writeable_eq!(borrowed, cas.expected, Ok(()));
         }
```
robertbastian:
```suggestion
        assert_try_writeable_eq!(borrowed, cas.expected, cas.single_result);
```

and update `single_should_err: false` -> `single_result: Ok(())` etc.

components/experimental/tests/displaynames/tests.rs
robertbastian, line 173: issue (follow-up because I don't think this is implemented): this should also be tested for the new implementation

components/experimental/src/displaynames/single/language.rs
```diff
@@ -9,14 +9,39 @@ use crate::displaynames::single::{
...
+
+/// Display name fallback occurred
+#[derive(displaydoc::Display, Debug, Copy, Clone, PartialEq, Eq, Default)]
+#[allow(clippy::exhaustive_structs)]
+pub struct LanguageIdentifierNameFallbackError;
+
+/// Represents a subtag that is either absent or has fallen back to its code.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+enum AbsentOrFallback<S> {
```
robertbastian: I mean... this is just `Option` now. seeing `Option<Variant>` instead of `AbsentOrFallback<Variant>` in the or place of `DataPayloadOr` communicates pretty much the same thing to the reader, and reduces complexity

components/experimental/src/displaynames/single/language.rs
```diff
@@ -9,14 +9,39 @@ use crate::displaynames::single::{
...
+use icu_locale_core::LanguageIdentifier;
+use icu_locale_core::subtags::{Language, Region, Script, Variant};
+use icu_pattern::{DoublePlaceholderPattern, DoublePlaceholderValueProviderTry, PatternItem};
 use icu_provider::DataPayloadOr;
 use icu_provider::prelude::*;
 use tinystr::TinyAsciiStr;
+use writeable::Writeable;
+use writeable::{PartsWrite, TryWriteable, adapters::LossyWrap};
+
+/// Display name fallback occurred
```
robertbastian: nit: better docs. what did it fall back to?

components/experimental/src/displaynames/single/language.rs
```diff
@@ -280,44 +375,65 @@ impl BorrowedVariants<'_> {
...
     locale_pattern: &'a DoublePlaceholderPattern,
     locale_separator: &'a DoublePlaceholderPattern,
 }
 
+impl<'a> LanguageIdentifierDisplayName<'a> {
+    /// Returns a [`Writeable`](writeable::Writeable) that formats the display name.
+    ///
+    /// Missing display names will fall back to the raw BCP-47 code.
+    #[inline]
+    pub fn with_fallback(&self) -> impl Writeable + core::fmt::Display + 'a {
```
robertbastian: unfortunately the `#[deny(missing_debug_implementations)]` doesn't catch this

```suggestion
    pub fn with_fallback(&self) -> impl Writeable + core::fmt::Display + core::fmt::Debug + 'a {
```

components/experimental/src/displaynames/single/language.rs
```diff
@@ -327,38 +443,61 @@ impl<'a> writeable::Writeable for QualifiersWriteable<'a> {
...
+        &self,
+        sink: &mut S,
+    ) -> Result<Result<(), Self::Error>, core::fmt::Error> {
         let has_variants = !self.variants.is_empty();
         let has_qualifiers =
             self.script_name.is_some() || self.region_name.is_some() || has_variants;
 
-        if !has_qualifiers {
-            sink.write_str(self.base_name)
+        let fallback_occurred = if !has_qualifiers {
```
robertbastian: why do you convert the results of the `try_write_to_parts` calls to a bool, to later convert them back to a result?

components/experimental/src/displaynames/single/language.rs
```diff
@@ -9,14 +9,39 @@ use crate::displaynames::single::{
...
+    Fallback(S),
+}
+
+/// Represents a payload that was either successfully loaded or has fallen back to its code.
+type PayloadOrFallback<M, S> = DataPayloadOr<M, S>;
 
 /// A localized display name for a language identifier, owned version.
 ///
+/// The formatter falls back to the BCP-47 subtag when localized display names are missing
+/// from the data provider. Fallback can be detected using [`TryWriteable`].
```
robertbastian, line 29: issue: fallback *must* be detected using `TryWriteable`, because the thing here doesn't implement `Writeable`. it's more important to say that `.with_fallback()` gives you a `Writeable` and a `to_string`.

components/experimental/src/displaynames/single/language.rs
```diff
@@ -35,24 +60,66 @@ use tinystr::TinyAsciiStr;
...
+/// )
+/// .expect("Data should load successfully");
+///
+/// let borrowed = display_name.as_borrowed();
+///
+/// // TryWriteable returns an error:
+/// let mut sink = String::new();
+/// let result = borrowed.try_write_to(&mut sink).expect("core::fmt::Write succeeded");
+/// assert_eq!(sink, "Italian (Qabc, Europe)");
+/// assert_eq!(result, Err(LanguageIdentifierNameFallbackError));
```
robertbastian: I would say use `assert_try_writeable_eq`, but you already have `assert_try_writeable_parts_eq`, so just remove this

components/experimental/src/displaynames/single/language.rs
```diff
@@ -35,24 +60,66 @@ use tinystr::TinyAsciiStr;
 /// )
 /// .expect("Data should load successfully");
 ///
-/// assert_writeable_eq!(display_name, "Canadian French");
+/// assert_writeable_eq!(display_name.as_borrowed().with_fallback(), "Canadian French");
```
robertbastian: issue: you're using `with_fallback` before talking about it. I'd use `assert_try_writeable_eq` here, and then in the next section show that calling `with_fallback` lets you use `Writeable`.

that is, of course, for the current architecture.

components/experimental/src/displaynames/single/language.rs
robertbastian: this is probably the only interesting discussion on this PR so far.

I think err-on-unknown-code is very much a power user feature (I can't really think of a use case for it). Also, in the multi code path, the default is falling back to the code instead of erroring, I would assume that's well-motivated.

So I'd prefer `.as_borrowed()` to return a `Writeable`. I don't have a strong preference between that thing also implementing `TryWriteable`, or having to call `.strict()` to get a `TryWriteable`. The latter requires an additional type, with the former we don't even need the borrowed type (format can return `impl Writeable + TryWriteable`), although it might still make sense to name it.

components/experimental/src/displaynames/single/README.md
robertbastian: I think the existence of this doc goes in the wrong direction. We should write self-documenting code, and you can explain architecture choices in code.

> To go one step further, I want to get to a point where we can primarily review a human-readable md file, and the code review is a "check check check" to make sure it implemented the design doc correctly.

Seeing the code that your LLM currently produces, we're years away from reviewing just the instructions. Until then, I will review the code, and only the code.

components/experimental/src/displaynames/single/language.rs
robertbastian: I don't really need the backstory of how you ended up with this code, here and elsewhere. You're sending me code to review, if you're using some LLMs to write that code, I'm still holding you responsible for it.

And just to be clear, theses LLM PRs are of significantly worse quality than what I'm used to getting from you, and a lot more work for me. And judging by the amount of changes I need to request, they _must_ be more work for you as well.

components/experimental/src/displaynames/single/language.rs
```diff
@@ -35,24 +60,66 @@ use tinystr::TinyAsciiStr;
...
 pub struct LanguageIdentifierDisplayNameOwned {
     formatting_locale: DataLocale,
     options: LanguageIdentifierDisplayNameOptions,
-    language_payload: DataPayload<LocaleNamesLanguageMediumV1>,
-    script_payload: DataPayloadOr<LocaleNamesScriptMediumV1, ()>,
-    region_payload: DataPayloadOr<LocaleNamesRegionMediumV1, ()>,
-    variant_payloads:
-        DataPayloadOr<LocaleNamesVariantMediumV1, Vec<DataPayload<LocaleNamesVariantMediumV1>>>,
+    language_payload: DataPayloadOr<LocaleNamesLanguageMediumV1, Language>,
+    script_payload: DataPayloadOr<LocaleNamesScriptMediumV1, AbsentOrFallback<Script>>,
```
robertbastian: is there a technical reason why you nested it like this? I would find this more logical, because you have an optional script, which is either a name or the code:

```suggestion
    script_payload: Option<DataPayloadOr<LocaleNamesScriptMediumV1, Script>>,
```

components/experimental/src/displaynames/single/language.rs
sffc: > I wouldn't consider the code fallback an "error case", given that it's the default option to use code fallback in the old implementation.

Maybe a "warning case". Still, not something I necessarily want to optimize.

> using DataPayloadOr to implement what is basically a SmallVec was a bit of a hack in the first place

I don't agree. DataPayloadOr is designed to reduce the stack size of formatters by leveraging more niches. SmallVec adds at least 1 and probably 2 or 3 usizes to the stack (haven't measured).

I'll edit this to do the DataPayloadOr structure I suggested above.

components/experimental/src/displaynames/single/language.rs
sffc: `DataPayloadOr<LocaleNamesScriptMediumV1, AbsentOrFallback<Script>>` is smaller on the stack than `Option<DataPayloadOr<LocaleNamesScriptMediumV1, Script>>`

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=fdb73e9e224e48162c2e6897266fa8a8

components/experimental/src/displaynames/single/language.rs
sffc: I proposed it in https://github.com/unicode-org/icu4x/pull/8006 and no-one pushed back (but I didn't request a broad review). We should look at this again before graduating.

Previous discussion: https://github.com/unicode-org/icu4x/issues/6468

robertbastian: I didn't push back on #8006 because it was opened and merged outside my working hours, and my review wasn't requested. Please stop presenting this as universal consensus.

The conclusion in #6468 is pretty clear: we use `Foo` and `FooBorrowed`, because even if `FooOwned` and `Foo` would be better, it's not possible to migrate all usages of this pattern, and we want to be consistent. You agreed with this conclusion.

components/experimental/src/displaynames/single/language.rs
sffc: because the Error types of the two inner TryWriteables are not compatible.

I could have written it by having two `.map_err`s but 🤷 

Note: one side's error is `Either<&str, LanguageIdentifierNameFallbackError>` and the other side's error is `&str`

components/experimental/src/displaynames/single/language.rs
sffc: (I wrote this code to show how to use TryWriteable because it is less obvious and less well-documented than Writeable. The only other place where we expose TryWriteable is in power-user APIs like DateTimePattern.)

sffc:
I changed it to implement both Writeable and TryWriteable on the same type, instead of making a function that returns `impl Writeable + '_`.

components/experimental/src/displaynames/single/language.rs
sffc: This PR does not change the status quo here. Follow-up: https://github.com/unicode-org/icu4x/issues/8180

components/experimental/src/displaynames/single/language.rs
sffc: If you're okay with one type implementing both traits, I prefer that approach.

components/experimental/src/displaynames/single/language.rs
sffc: I ran out of time to do this today. I don't think it's worth blocking the PR on this since it only impacts a warning case.

components/experimental/tests/displaynames/tests.rs
sffc: https://github.com/unicode-org/icu4x/issues/8181

components/experimental/src/displaynames/single/language.rs
sffc: (and now there is just one type to implement both traits)

components/experimental/tests/displaynames/tests.rs
sffc: 429d795d6fa6c5ab0faa14f8f6f31d788c903652

components/experimental/tests/displaynames/tests.rs
sffc: sffc/displaynames-fallback

components/experimental/src/displaynames/single/language.rs
sffc: be2f6bab1680e1a5f62cf126b6f61c8ef5ab4a4b

components/experimental/src/displaynames/single/language.rs
sffc: A good follow-up issue for this thread would be to actually measure the stack size of the different options, weigh their tradeoffs, and implement a consensus solution. I think that is a very good scope of work for a follow-up, and it doesn't impact API.

components/experimental/src/displaynames/single/language.rs
```diff
@@ -88,31 +74,35 @@ type PayloadOrFallback<M, S> = DataPayloadOr<M, S>;
...
-///
 /// // The fallback string is identified with a [`Part::ERROR`](writeable::Part::Error):
 /// assert_try_writeable_parts_eq!(
 ///     borrowed,
 ///     "Italian (Qabc, Europe)",
 ///     Err(LanguageIdentifierNameFallbackError),
 ///     [(9, 13, Part::ERROR)] // the span of Qabc
 /// );
+///
+/// // To format in lossy mode (ignoring fallback errors), use standard `Writeable` or `Display`:
```
robertbastian:
```suggestion
/// // To format in lossy mode (ignoring fallback errors), use [`Writeable`] or [`Display`]:
```

components/experimental/src/displaynames/single/language.rs
```diff
@@ -88,31 +74,35 @@ type PayloadOrFallback<M, S> = DataPayloadOr<M, S>;
...
 #[derive(Debug)]
 pub struct LanguageIdentifierDisplayNameOwned {
     formatting_locale: DataLocale,
     options: LanguageIdentifierDisplayNameOptions,
     language_payload: DataPayloadOr<LocaleNamesLanguageMediumV1, Language>,
-    script_payload: DataPayloadOr<LocaleNamesScriptMediumV1, AbsentOrFallback<Script>>,
-    region_payload: DataPayloadOr<LocaleNamesRegionMediumV1, AbsentOrFallback<Region>>,
+    /// In the fallback (`Option<Script>`) state:
+    /// - `None` indicates the script subtag was absent from the subject.
+    /// - `Some(script)` indicates the script subtag was present, but display name data was not found, so it falls back to the raw code.
```
robertbastian: this is pointless. both `None` and `Some(script)` indicate that "display name data was not found", as it's the or branch of `DataPayloadOr`. and then `None` vs `Some` is the same as `LanguageIdentifier.script`, so this is not worth documenting

components/experimental/src/displaynames/single/language.rs
```diff
@@ -280,44 +365,70 @@ impl BorrowedVariants<'_> {
...
 
+writeable::impl_try_writeable_delegate!(
+    LanguageIdentifierDisplayName<'_>,
+    |&self| &self.0.0,
+    Error = LanguageIdentifierNameFallbackError
+);
+
+writeable::impl_writeable_delegate!(LanguageIdentifierDisplayName<'_>, |&self| &self.0);
+
+writeable::impl_display_with_writeable!(LanguageIdentifierDisplayName<'_>);
```
robertbastian, line 401: I don't think you need a separate inner type. just implement `TryWriteable` directly on `LanguageIdentifierDisplayName`, and then

```suggestion
writeable::impl_writeable_delegate!(LanguageIdentifierDisplayName<'_>, |&self| LossyWrap(&self.0));

writeable::impl_display_with_writeable!(LanguageIdentifierDisplayName<'_>);
```

components/experimental/src/displaynames/single/language.rs
robertbastian: where does `Either` come from here?

this would be more readable with `map_err`s, because you can match on the type and I could see what's going on. as-is you're discarding errors and reconstructing them and a reader has no idea why

components/experimental/src/displaynames/single/language.rs
```diff
@@ -89,8 +146,9 @@ impl LanguageIdentifierDisplayNameOwned {
         // and we "consume" the corresponding subtags so they are not repeated in the qualifiers.
         // If none are found, we fall back to the base language name (e.g., "zh") and all
         // present subtags (script, region, variants) will be formatted as qualifiers.
-
-        let mut language_payload = None;
+        //
+        // Prefer dialect handling if requested and available.
```
robertbastian: duplicate comment, should only be on the line 154

components/experimental/src/displaynames/single/language.rs
```diff
@@ -143,8 +201,7 @@ impl LanguageIdentifierDisplayNameOwned {
         }
 
         // If the language name is not loaded yet, try loading it from the language subtag alone.
-        // TODO(#8100): Fall back to the code instead of failing with DataError if the language name is not found
-        let language_payload = match language_payload {
+        let language_payload = match language_payload_or {
```
robertbastian: you renamed the first `language_payload` to `language_payload_or`, but not the second one. iiuc both of these change type to `DataPayloadOr` in this PR. I don't think renaming is necessary, but if you do one you should do both

you also don't use `script_payload_or` et al

components/experimental/src/displaynames/single/language.rs
```diff
@@ -154,57 +211,81 @@ impl LanguageIdentifierDisplayNameOwned {
...
+        let load_variant = |variant: Variant| -> Result<
+            DataPayloadOr<LocaleNamesVariantMediumV1, Variant>,
+            DataError,
+        > {
+            match VariantDisplayNameOwned::try_new_unstable(provider, prefs, variant)
+                .allow_identifier_not_found()?
+            {
+                Some(obj) => Ok(DataPayloadOr::from_payload(obj.payload)),
+                None => Ok(DataPayloadOr::from_other(variant)),
+            }
```
robertbastian, line 261:
```suggestion
            Ok(VariantDisplayNameOwned::try_new_unstable(provider, prefs, variant)
                .allow_identifier_not_found()?
                .map_or_default(
                    || DataPayloadOr::from_other(variant),
                    |r| DataPayloadOr::from_payload(r.payload)))
```

also elsewhere

components/experimental/src/displaynames/single/language.rs
```diff
@@ -154,57 +211,81 @@ impl LanguageIdentifierDisplayNameOwned {
...
+                let payload_vec = core::iter::once(Ok(first_val))
                     .chain(variant_results)
-                    .map(|result| result.map(|obj| obj.payload))
                     .collect::<Result<Vec<_>, _>>()?;
                 DataPayloadOr::from_other(payload_vec)
             } else {
                 // 1 variant
-                // TODO(#8100): Fall back to the code instead of dropping it if the variant name is not found
-                DataPayloadOr::from_payload(result?.payload)
+                match first_val.into_inner() {
```
robertbastian: this does not need `peek`
```suggestion
            if let Some(second) = variant_results.next() {
                // 2 or more variants
                let payload_vec = [result, second].into_iter()
                    .chain(variant_results)
                    .collect::<Result<Vec<_>, _>>()?;
                DataPayloadOr::from_other(payload_vec)
            } else {
                // 1 variant
                match result?.into_inner() {
```

components/experimental/src/displaynames/single/language.rs
sffc: Unfortunately this pattern doesn't work because `write_to_string` borrows from the temporary `LossyWrap`. The Writeable trait doesn't have a way to track the inner lifetime.

components/experimental/src/displaynames/single/language.rs
sffc: Also, this is parallel with the previous TryWriteable impl, which also has

```rust
        if fallback_occurred {
            Ok(Err(LanguageIdentifierNameFallbackError))
        } else {
            Ok(Ok(()))
        }
```

sffc: I rewrote the error handling by adding a little helper type `NameOrFallback` which implements TryWriteable with the correct error type, so now I can just chain the errors and don't need any funny mapping.

components/experimental/src/displaynames/single/language.rs
sffc: f0e79e5504

components/experimental/src/displaynames/single/language.rs
sffc: changed back in c9345cb667

components/experimental/src/displaynames/single/language.rs
sffc: `map_or_default` is too new

https://github.com/rust-lang/rust/issues/138099

components/experimental/src/displaynames/single/language.rs
sffc: ok, `[result, second].into_iter()` as opposed to `core::iter::once(Ok(first_val))` 🤷

sffc:
There are a lot of comments so I'm going to click resolve on the ones that are either fixed or have follow-up issues filed.

components/experimental/src/displaynames/single/language.rs
sffc: 1ac53e1

components/experimental/src/displaynames/single/language.rs
sffc: > where does `Either` come from here?

https://unicode-org.github.io/icu4x/rustdoc/icu_pattern/struct.DoublePlaceholderValueProviderTry.html#associatedtype.Error

sffc:
I think the last round of comments was more about subjective style and commenting choices than substance, with the possible exception of the error handling, because I agree it's better style to track errors.

Variable names and things like that being self-inconsistent in the same file is, well, regrettable but 🤷 not really worth the ping-pong to fix IMO.

This PR has been in a mergeable state 3 times. Let's please favor follow-up issues.

robertbastian (approved):
> Variable names and things like that being self-inconsistent in the same file is, well, regrettable but 🤷 not really worth the ping-pong to fix IMO.

You could just not send code like that for review. It might involve hand-writing it.

PR #8135, opened 2026-06-25
Introduce dedicated LanguageIdentifierDisplayNameOptions struct

sffc:
This pull request is part of the implementation for the main locale display names tracking issue: #7825.

This pull request introduces a dedicated options bag `LanguageIdentifierDisplayNameOptions` specifically for the single formatter `LanguageIdentifierDisplayNameOwned` (and its borrowed version).

Previously, the single formatter reused the general `DisplayNamesOptions` bag, which contains options like `style` that are only supported by the multi-formatters (since single formatters are currently hardcoded to the `Medium` width data markers).

By introducing a dedicated options bag, we:
1. Cleanly separate the options supported by the single formatter (currently only `language_display` / dialect resolution).
2. Prevent confusion regarding unsupported options like `style`.
3. Lay the foundation for adding fallback options specifically to the single formatter in a subsequent step.

All docstring examples, constructors, and integration tests have been updated to use the new options bag.

## Changelog

icu_experimental/displaynames: Introduce dedicated `LanguageIdentifierDisplayNameOptions` struct
  * New types: `LanguageIdentifierDisplayNameOptions`

### DIFF
--- a/components/experimental/src/displaynames/mod.rs
+++ b/components/experimental/src/displaynames/mod.rs
@@ -66,4 +66,5 @@ pub use displaynames::DisplayNamesPreferences;
 pub use options::DisplayNamesOptions;
 pub use options::Fallback;
 pub use options::LanguageDisplay;
+pub use options::LanguageIdentifierDisplayNameOptions;
 pub use options::Style;
--- a/components/experimental/src/displaynames/options.rs
+++ b/components/experimental/src/displaynames/options.rs
@@ -35,6 +35,14 @@ pub struct DisplayNamesOptions {
     pub language_display: LanguageDisplay,
 }
 
+/// A bag of options defining how a language identifier display name will be formatted.
+#[derive(Copy, Debug, Eq, PartialEq, Clone, Default)]
+#[non_exhaustive]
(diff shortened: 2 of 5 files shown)


Manishearth (approved):

robertbastian:
I pity the person who will have to consolidate all those AI-generated changelog entries. Is it necessary to list the *fields* of new types? And the methods it listed as new are not new.

Manishearth:
When we discussed the changelog rules we explicitly agreed to not include methods, fields, etc on new types, and [that exception is in our policy](https://github.com/unicode-org/icu4x/blob/main/documents/process/changelog.md#what-to-include).

I edited the changelog entry.


I did find that the PR body in https://github.com/unicode-org/icu4x/pull/8132 had a much less "density of usefulness" than I typically get from sffc PRs. It still had everything I wanted so I didn't complain.

sffc:
The agent wrote both the PR description and the changelog entry. I could have slimmed down the PR description but figured it wasn't a good use of my time (perhaps for a similar reason as @robertbastian sometimes doesn't include enough detail in the PR description). I probably should have slimmed down the changelog entry.

PR #8136, opened 2026-06-25
Publish icu4x

mosuem:
To bring the changes into `package:intl4x`.

## Changelog

none

### DIFF
--- a/ffi/dart/pubspec.yaml
+++ b/ffi/dart/pubspec.yaml
@@ -5,7 +5,7 @@
 name: icu4x
 description: > 
   Dart bindings for the Rust `icu` internationalization crate.
-version: 2.2.0-dev.1-wip
+version: 2.2.0-dev.2
 repository: https://github.com/unicode-org/icu4x/tree/main/ffi/dart
 issue_tracker: https://github.com/unicode-org/icu4x/issues?q=is%3Aissue%20is%3Aopen%20label%3AU-flutter
 


robertbastian (changes requested):
you shouldn't be publishing from main. this will pull in unreleased changes, and we don't have published artifacts at this commit.

do we urgently need to publish?

mosuem:
> do we urgently need to publish?

Define urgently - but we do want to try to get the newest version ready for testing as soon as possible.

Ideally, we should publish after every commit to have the published versions following the main branch as close as possible. Could we not set up CI to publish artifacts with the ICU4X package publish, i.e. on tag?

Manishearth:
I think we already do that, no? Publishing a package is a part of the release process, and CI publishes artefacts on tags.

robertbastian:
Yes we have that. The process for this is
* generate a tag on the branch that will be released
* run `regenerate_hashes.dart` with that tag
* create a PR with the updated hashes, the Dart package version bump, and reenabling the fetch CI (https://github.com/unicode-org/icu4x/blame/1569b93140c14c1261dd47f3e0e7bdc889280ddf/.github/workflows/build-test.yml#L371)

We don't generally publish patch releases from main, as that will pull in unpublished changes as well. A quick check showed that the Julian calendar is currently unpublished, we made a breaking change to the ScriptExtension Dart bindings, behaviour changes to fallback, and probably more.

Now, Dart being an unstable package, this probably doesn't matter much, but it conflicts with documentation such as the changelog: Dart `2.2.0-dev.2` will have API and behaviour differences from Rust `2.2.0`, and that would all be undocumented.

So I think we should either
* cherry-pick #8119 to the 2.2 release branch and publish from there
* wait for 2.3 to be published from main, which is targeted for July

robertbastian:
(if you cherry-pick this to the release branch, you won't need to publish artifacts, as this change itself doesn't need artifact changes)

robertbastian:
Published `2.2.0-dev.2` in #8138

PR #8137, opened 2026-06-25
Rename complex languages to complex scripts

robertbastian:
In the context of segmentation, there are no complex languages, only complex scripts. The segmenter looks at the Unicode `Script` property to determine which dictionary model to load.

Languages *also* appear in the context of segmentation, for tailorings like Greek or Japanese/Chinese, so we should use accurate names.

## Changelog

N/A


Manishearth (approved):
I know some of our *models* are language-specific, so I'd like @sffc to weigh in as well.

Fine to land if you're depending on it somewhere.

sffc (approved):
I'm assuming this is a clean rename.

"complex scripts" is verbiage we agreed on in the constructor, https://unicode-org.github.io/icu4x/rustdoc/icu/segmenter/struct.LineSegmenter.html#method.new_for_non_complex_scripts

It's a bit more _complex_ than that (😆), since most of our models are language-based, not script-based, but the thing that defines when to use a model is the script, not the language, and we should eventually strive to make the models script-based.

Manishearth:
Oh, yes, we did use the naming in the constructor. Then yeah this is straightforwardly better.

robertbastian:
yes this is a clean rename. the reason why there's a baked diff is that the Burmese -> Myanmar renaming upsets some BTreeMap during pseudo-symbol generation.

PR #8139, opened 2026-06-26
Add writeable delegate macros and support where clauses

sffc:
This PR is 80% human, 20% Gemini.

I added `writeable::impl_writeable_delegate!` and `writeable::impl_try_writeable_delegate!`, using them throughout the codebase.

They support generics via a `where` clause at the end of the macro invocation. I experimented with several ways to get the generics into `macro_rules!`, and I settled on this approach as the cleanest. If you have another suggestion that works, feel free to offer it.

I also went ahead and modified the existing `writeable::impl_display_with_writeable` to support the same `where` clause syntax.

This PR fixes several call sites, both inside the writeable crate and outside the writeable crate, by delegating all of the trait functions, not just one or two. There is never a case with a pure delegation Writeable impl where all the functions should not be delegated in their entirety.

Although displaynames is what gave me the motivation to work on this PR, the PR is not motivated solely because of displaynames. I did not touch any of the code in displaynames because I didn't want to introduce merge conflicts, but I will migrate them soon.

## Changelog

writeable:
    - New `impl_writeable_delegate!` macro to delegate `Writeable` implementations
    - New `impl_try_writeable_delegate!` macro to delegate `TryWriteable` implementations
    - Support for `where` clause in `impl_display_with_writeable!`

### DIFF
--- a/components/locale_core/src/helpers.rs
+++ b/components/locale_core/src/helpers.rs
@@ -146,20 +146,7 @@ macro_rules! impl_tinystr_subtag {
             }
         }
 
-        impl writeable::Writeable for $name {
-            #[inline]
-            fn write_to<W: core::fmt::Write + ?Sized>(&self, sink: &mut W) -> core::fmt::Result {
-                sink.write_str(self.as_str())
-            }
-            #[inline]
-            fn writeable_length_hint(&self) -> writeable::LengthHint {
-                writeable::LengthHint::exact(self.0.len())
-            }
-            fn writeable_borrow(&self) -> Option<&str> {
-                Some(self.0.as_str())
-            }
-        }
-
+        writeable::impl_writeable_delegate!($name, |&self| &self.0, #[cfg(feature = "alloc")]);
         writeable::impl_display_with_writeable!($name, #[cfg(feature = "alloc")]);
 
         #[doc = concat!("A macro allowing for compile-time construction of valid [`", stringify!($name), "`] subtags.")]
--- a/utils/writeable/src/try_writeable.rs
+++ b/utils/writeable/src/try_writeable.rs
@@ -370,7 +370,7 @@ where
 /// );
 /// ```
 ///
-/// With a cfg on fn write_to_string:
+/// With a cfg on fn `write_to_string`:
 ///
 /// ```
 /// struct MyStruct(Result<String, String>);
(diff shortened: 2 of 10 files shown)


Manishearth (commented):
Generally like this idea. Will review in depth later.

utils/writeable/src/try_writeable.rs
```diff
@@ -354,37 +354,114 @@ where
...
+/// struct MyStruct(Result<String, String>);
+/// writeable::impl_try_writeable_delegate!(MyStruct, |&self| &self.0, Error = String);
+///
+/// writeable::assert_try_writeable_eq!(
+///     MyStruct(Ok("hello".to_string())),
+///     "hello"
+/// );
+/// ```
+///
+/// With a cfg on fn `write_to_string`:
```
Manishearth, line 373: thought: I found it hard to figure out what the cfg was doing just by looking at the callsites. this is _fine_, but we might be able to do better, maybe something like

`#[cfg(....)] on write_to_string`, or just `#[cfg(..)] fn write_to_string`?

That makes callsites self-documenting.

Manishearth (approved):
Looks cleaner!

components/locale_core/src/helpers.rs
```diff
@@ -146,20 +146,7 @@ macro_rules! impl_tinystr_subtag {
...
-            #[inline]
-            fn writeable_length_hint(&self) -> writeable::LengthHint {
-                writeable::LengthHint::exact(self.0.len())
-            }
-            fn writeable_borrow(&self) -> Option<&str> {
-                Some(self.0.as_str())
-            }
-        }
-
+        writeable::impl_writeable_delegate!($name, |&self| &self.0, #[cfg(feature = "alloc")]);
```
robertbastian, line 149: thought: this delegates to `TinyAsciiStr`, the old code delegated to `str`. This has more room for error, because we maintain `str` in the `writeable` crate, but `TinyAsciiStr` in its crate. I'm not even sure if `TinyAsciiStr`'s impl has all concrete implementations



```suggestion
        writeable::impl_writeable_delegate!($name, |&self| self.as_str(), #[cfg(feature = "alloc")]);
```

robertbastian (approved):

PR #8140, opened 2026-06-26
Add LossyWrap::to_string impl

sffc:
Adds a fn that should have been there but was missing.

No AI was used in this pull request.

## Changelog

writeable:

- New concrete fn LossyWrap::to_string

### DIFF
--- a/utils/writeable/src/lib.rs
+++ b/utils/writeable/src/lib.rs
@@ -135,6 +135,19 @@ pub mod adapters {
             Ok(())
         }
     }
+
+    impl<T: TryWriteable> LossyWrap<T> {
+        /// Converts the given value to a `String`.
+        ///
+        /// Under the hood, this uses an efficient [`Writeable`] implementation.
+        /// However, in order to avoid allocating a string, it is more efficient
+        /// to use [`Writeable`] directly.
+        #[cfg(feature = "alloc")]
+        #[inline]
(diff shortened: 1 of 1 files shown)


Manishearth (approved):

utils/writeable/src/lib.rs
```diff
@@ -135,6 +135,20 @@ pub mod adapters {
             Ok(())
         }
     }
+
+    impl<T: TryWriteable> LossyWrap<T> {
+        /// Converts the given value to a `String`.
+        ///
+        /// Under the hood, this uses an efficient [`Writeable`] implementation.
+        /// However, in order to avoid allocating a string, it is more efficient
+        /// to use [`Writeable`] directly.
```
robertbastian, line 144: yeah but if you need a `String` you can't avoid allocating one

sffc: These docs are copied verbatim from `impl_display_with_writeable`. I find them clear enough.

robertbastian: well then they should also be updated.

```rust
        /// Under the hood, this uses an efficient [`Writeable`] implementation.
        /// 
        /// If you don't need an allocated [`String`], but e.g. need to write this
        /// to some sink, it is more efficient to use [`Writeable`] directly.
```

PR #8143, opened 2026-06-26
Optimize segmenter result cache consumption

Xuanwo:
This PR removes the repeated allocation performed when the v1 complex-break cache consumes a boundary.

## Design

Complex-run-relative offsets are converted once into distances between adjacent breaks. The cache keeps natural order in `vec::IntoIter<usize>`: `as_slice().first()` peeks at the next distance and `next()` consumes it in O(1).

This follows the v2/neo cache representation and addresses the review feedback directly: there is no reversed vector, `Vec + index` state, or per-boundary offset compensation. Public API and segmentation behavior are unchanged.

## Performance

Measured on an Apple M4 Max, comparing `main@3f2215305` with `PR@7a5fdbc4c`.

| 1M-character full traversal | main | PR |
| --- | ---: | ---: |
| Word dictionary, continuous Han | 19.063 s | 31.859 ms (598x) |
| Line dictionary, continuous Thai | 6.576 s | 22.834 ms (288x) |

At 1M characters, allocation calls and cumulative requested bytes drop from `524,325 / 1.10 TB` to `38 / 16.8 MB` for word/Han, and from `299,630 / 359 GB` to `38 / 16.8 MB` for line/Thai. Stopping after 10 boundaries remains neutral.

The merged PR #8141 workloads show the same direction on cache-heavy paths:

- Thai line: LSTM `-5.4%`, dictionary `-7.1%`
- Japanese/Han word: auto neutral, dictionary `-15.0%`
- Mixed: line LSTM `-8.5%`, line dictionary `-2.4%`, word dictionary `-10.8%`

Mixed word auto is about `+4%` in the latest paired rerun; it is a small non-dictionary workload and remains the observed residual regression.

```sh
ICU4X_SEGMENTER_BENCH_LARGE=1 cargo bench -p icu_segmenter --bench bench -- "Complex Break Cache" --noplot
```

Validation: `cargo test -p icu_segmenter --features unstable` and Clippy with `-D warnings` pass.


components/segmenter/src/rule_segmenter.rs
```diff
@@ -9,10 +9,36 @@ use crate::complex::ComplexPayloadsBorrowed;
 use crate::complex::{ComplexIterator, ComplexPayloadBorrowed};
 use crate::indices::{Latin1Indices, Utf16Indices};
 use crate::provider::*;
-use alloc::vec::Vec;
+use alloc::vec::{IntoIter, Vec};
+use core::iter::Peekable;
 use core::str::CharIndices;
 use utf8_iter::Utf8CharIndices;
 
+pub(crate) type ResultCache = Peekable<IntoIter<usize>>;
```
robertbastian: nit: don't import a common name like `IntoIter`, use it qualified

```suggestion
pub(crate) type ResultCache = Peekable<vec::IntoIter<usize>>;
```

components/segmenter/src/rule_segmenter_v1.rs
```diff
@@ -9,10 +9,36 @@ use crate::complex::ComplexPayloadsBorrowed;
...
+    // Complex segmenters return offsets relative to the start of the complex run.
+    // The cache stores distances from the iterator's current position to the next
+    // break, then from each cached break to the following cached break.
+    for break_offset in &mut break_offsets {
+        let current_offset = *break_offset;
+        assert!(
+            current_offset >= previous_offset,
+            "complex break offsets must be monotonically increasing"
+        );
+        *break_offset = current_offset - previous_offset;
```
robertbastian, line 28: I think precomputing the offsets might actually hurt performance:
* you need to load the whole cache from memory
* you need to write the whole cache to memory
* if the cache is large, you *actually* have to write to memory (instead of the CPU cache)
* this cannot use SIMD, because there's a data dependency on the previous element

if you instead do this in the `next()` method, you never need to write back to memory, and the data you need is already in registers

components/segmenter/src/line.rs
```diff
@@ -8,6 +8,7 @@ use crate::provider::*;
 use crate::rule_segmenter::*;
 use alloc::string::String;
 use alloc::vec;
+#[cfg(test)]
 use alloc::vec::Vec;
```
robertbastian: move into `mod test`

robertbastian (commented):
I'm not sure about this. We don't have benchmarks for large texts, where I expect this to become slow.

components/segmenter/src/rule_segmenter.rs
Xuanwo: removed.

components/segmenter/src/rule_segmenter_v1.rs
Xuanwo: You're right that normalization adds a scalar read/write pass over the full cache, so I benchmarked this concern directly.

I added opt-in continuous-script benchmarks up to 1M characters, covering both full traversal and stopping after 10 boundaries. Against main, the 1M full traversal improves from 18.688 s to 32.230 ms for word/Han and from 6.187 s to 21.861 ms for line/Thai. The prefix-10 cases remain neutral.

Separately, I implemented the suggested lazy `Peekable<vec::IntoIter<usize>>` approach and ran the same 1M workloads. It was 5-6% slower than the current in-place normalization for full traversal and 4-5% slower for prefix-10, with identical allocation counts.

The complex segmenter already materializes the full `Vec` of offsets, so the current approach adds one linear in-place pass and then consumes every cached boundary with `last()`/`pop()`. Based on these results, I kept the precomputed distances. The large-input benchmark and reproduction command are now included in this PR.

---

Do you think this addressed your concern? I'm open to start another PR with the lazy version along with the benchmark result.

components/segmenter/src/rule_segmenter_v1.rs
```diff
@@ -237,3 +258,34 @@ impl<Y: RuleBreakType> RuleBreakIterator<'_, '_, Y> {
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn result_cache_stores_distances_between_breaks() {
+        let mut cache = result_cache_from_offsets(vec![12, 21, 33], 3);
+
+        assert_eq!(
+            core::iter::from_fn(|| cache.pop()).collect::<Vec<_>>(),
+            [9, 9, 12]
```
robertbastian: this is a really convoluted way of reversing a vector

components/segmenter/src/rule_segmenter_v1.rs
```diff
@@ -237,3 +258,34 @@ impl<Y: RuleBreakType> RuleBreakIterator<'_, '_, Y> {
...
+        let mut cache = result_cache_from_offsets(vec![3, 12], 3);
+
+        assert_eq!(
+            core::iter::from_fn(|| cache.pop()).collect::<Vec<_>>(),
+            [0, 9]
+        );
+    }
+
+    #[test]
+    #[should_panic(expected = "complex break offsets must be monotonically increasing")]
```
robertbastian: document this invariant on `result_cache_from_offsets`

components/segmenter/src/rule_segmenter_v1.rs
```diff
@@ -7,6 +7,27 @@ use crate::provider::*;
...
+pub(crate) fn result_cache_from_offsets(
+    mut break_offsets: Vec<usize>,
+    mut previous_offset: usize,
+) -> Vec<usize> {
+    // Complex segmenters return offsets relative to the start of the complex run.
+    // Store distances from the iterator's current position and between later breaks,
+    // reversed so that each next distance can be consumed in O(1) with Vec::pop().
+    for break_offset in &mut break_offsets {
+        let current_offset = *break_offset;
+        assert!(
```
robertbastian:
```suggestion
        debug_assert!(
```

components/segmenter/src/rule_segmenter_v1.rs
```diff
@@ -7,6 +7,27 @@ use crate::provider::*;
 use crate::scaffold::RuleBreakType;
 use alloc::vec::Vec;
 
+#[inline]
+pub(crate) fn result_cache_from_offsets(
```
robertbastian, line 18: document this function better. what does it return? why does it reverse?

components/segmenter/src/word.rs
```diff
@@ -952,13 +953,12 @@ where
     iter.current_pos_data = start_point;
     #[expect(clippy::unwrap_used)] // iter.complex present for word segmenter
     let breaks = iter.complex.unwrap().segment_str(&s);
-    iter.result_cache = breaks;
-    let first_pos = *iter.result_cache.first()?;
-    let mut i = left_codepoint.len_utf8();
+    iter.result_cache = result_cache_from_offsets(breaks, left_codepoint.len_utf8());
+    let first_pos = *iter.result_cache.last()?;
```
robertbastian: working with a reversed vector makes this hard to understand, and you pay the cost of reversing the vector. these data types would be better suited:
* `alloc::vec::IntoIter`, constant-time front-popping with `.next()`
* `alloc::collectsion::VecDeque`, constant-time front-popping with `.pop_front()`

the neo implementation uses `IntoIter`, so try using that as well

components/segmenter/benches/bench.rs
```diff
@@ -18,6 +19,11 @@ const TEST_STR_TH: &str =
 const TEST_STR_JA: &str =
     "こんにちは世界こんにちは世界こんにちは世界こんにちは世界こんにちは世界こんにちは世界";
 const TEST_STR_LONG_MIXED: &str = include_str!("../tests/testdata/SegmenterBenchMixed.txt");
+const TEST_STR_CONTINUOUS_TH: &str = "ภาษาไทย";
```
robertbastian, line 23: why not use `TEST_STR_TH`?

components/segmenter/src/rule_segmenter_v1.rs
Xuanwo: agreed, I have changed it!

components/segmenter/src/rule_segmenter_v1.rs
Xuanwo: changed.

components/segmenter/src/rule_segmenter_v1.rs
Xuanwo: added

components/segmenter/src/rule_segmenter_v1.rs
Xuanwo: added

components/segmenter/src/word.rs
Xuanwo: Ah, this idea is great. changed my impl.

components/segmenter/benches/bench.rs
Xuanwo: `TEST_STR_TH` has an unexpected space 😆

robertbastian (approved):

PR #8145, opened 2026-06-26
feat(currency): Migrate CurrencyFormatter to adopt the new design with generic ValueRepresentation (Decimal) and add short/narrow constructors

younies:
Migrate the currency formatter based on the design doc:
https://hackmd.io/@younies/number_formatter_4x

### Overview
This PR refactors the `CurrencyFormatter` in `icu_experimental` to align with the proposed design for modular, zero-copy currency formatting (Option 1 in the design doc).

Specifically, it introduces the `ValueRepresentation` typestate pattern to partition constructors by capability and representation, starting with the standard `Decimal` representation, while ensuring consistent locale and numbering system resolution across both short and narrow widths.

TAG=agy
CONV=0f8e0e91-077a-4704-b63f-0c02f37cbc41

## Changelog

icu_experimental/currency:

- Implement `ValueRepresentation` trait and `Decimal` marker struct to support typestate-based currency formatting.
- Refactor `CurrencyFormatter` to be generic over `ValueRepresentation`: `pub struct CurrencyFormatter<V: ValueRepresentation>`.
- Migrate the old short and narrow currency formatter to `CurrencyFormatter<Decimal>` with new `try_new_short` and `try_new_narrow` constructors (and their unstable/buffer variants), removing the old non-generic `try_new` constructors.
- **Fix Narrow Fallback**: Update both compiled and unstable versions of `try_new_narrow` to use robust fallback loading, ensuring they correctly support numbering system overrides consistently with `try_new_short`.
- Update experimental currency tests and doctests to use the new generic API, including comprehensive test cases for narrow formatting and numbering system overrides.

### DIFF
--- a/components/experimental/src/dimension/currency/format.rs
+++ b/components/experimental/src/dimension/currency/format.rs
@@ -9,61 +9,144 @@ mod tests {
     use tinystr::*;
     use writeable::assert_writeable_eq;
 
-    use crate::dimension::currency::{CurrencyCode, formatter::CurrencyFormatter};
+    use crate::dimension::currency::{
+        CurrencyCode,
+        formatter::{CurrencyFormatter, CurrencyFormatterPreferences, Decimal},
+    };
 
     #[test]
     pub fn test_en_us() {
-        let locale = locale!("en-US").into();
+        let prefs: CurrencyFormatterPreferences = locale!("en-US").into();
         let currency_code = CurrencyCode(tinystr!(3, "USD"));
-        let fmt = CurrencyFormatter::try_new(locale, Default::default()).unwrap();
 
-        // Positive case
+        // Short
+        let fmt_short = CurrencyFormatter::<Decimal>::try_new_short(prefs, &currency_code).unwrap();
         let positive_value = "12345.67".parse().unwrap();
-        let formatted_currency = fmt.format_fixed_decimal(&positive_value, &currency_code);
-        assert_writeable_eq!(formatted_currency, "$12,345.67");
-
-        // Negative case
+        assert_writeable_eq!(
+            fmt_short.format_fixed_decimal(&positive_value),
+            "$12,345.67"
+        );
         let negative_value = "-12345.67".parse().unwrap();
-        let formatted_currency = fmt.format_fixed_decimal(&negative_value, &currency_code);
-        assert_writeable_eq!(formatted_currency, "-$12,345.67");
+        assert_writeable_eq!(
+            fmt_short.format_fixed_decimal(&negative_value),
+            "-$12,345.67"
+        );
+
+        // TODO(#8151): This should format to 2 decimal places ($123.46 or $123.00) once we use currency patterns.
+        // Currently it uses decimal patterns which do not pad '123' to 2 decimal places, and do not round '123.4567'.
+        let value_no_decimals = "123".parse().unwrap();
+        assert_writeable_eq!(fmt_short.format_fixed_decimal(&value_no_decimals), "$123");
+        let value_4_decimals = "123.4567".parse().unwrap();
+        assert_writeable_eq!(
+            fmt_short.format_fixed_decimal(&value_4_decimals),
+            "$123.4567"
+        );
+
+        // Narrow
+        let fmt_narrow =
+            CurrencyFormatter::<Decimal>::try_new_narrow(prefs, &currency_code).unwrap();
+        assert_writeable_eq!(
+            fmt_narrow.format_fixed_decimal(&positive_value),
+            "$12,345.67"
+        );
+        assert_writeable_eq!(
+            fmt_narrow.format_fixed_decimal(&negative_value),
+            "-$12,345.67"
+        );
+
+        // TODO(#8151): This should format to 2 decimal places ($123.46 or $123.00) once we use currency patterns.
+        assert_writeable_eq!(fmt_narrow.format_fixed_decimal(&value_no_decimals), "$123");
+        assert_writeable_eq!(
+            fmt_narrow.format_fixed_decimal(&value_4_decimals),
+            "$123.4567"
+        );
     }
 
     #[test]
     pub fn test_fr_fr() {
-        let locale = locale!("fr-FR").into();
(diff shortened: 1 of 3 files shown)


components/experimental/src/dimension/currency/format.rs
```diff
@@ -9,62 +9,124 @@ mod tests {
...
-    use crate::dimension::currency::{CurrencyCode, formatter::CurrencyFormatter};
+    use crate::dimension::currency::{
+        CurrencyCode,
+        formatter::{CurrencyFormatter, CurrencyFormatterPreferences, Decimal},
+    };
 
     #[test]
     pub fn test_en_us() {
-        let locale = locale!("en-US").into();
+        let locale: CurrencyFormatterPreferences = locale!("en-US").into();
```
robertbastian: nit: 
```suggestion
        let prefs: CurrencyFormatterPreferences = locale!("en-US").into();
```

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -164,9 +252,10 @@ impl CurrencyFormatter {
     /// ```
     pub fn format_fixed_decimal<'l>(
         &'l self,
-        value: &'l Decimal,
+        value: &'l FixedDecimal,
```
robertbastian, line 275: issue, pre-existing: this should probably take some integer numeric type (which is capable of representing exact decimal fractions) (like https://github.com/googleapis/googleapis/blob/fdd6a6e4ef12853400539139d014d96e34253166/google/type/money.proto#L30-L40), and use some logic on the `ValueRepresentation` to work with that

it does not make sense to accept a `FixedDecimal`, which is basically an int with a user-selected magnitude, if we use currency-specific magnitude and rounding anyway

components/experimental/src/dimension/currency/format.rs
```diff
@@ -74,18 +136,60 @@ mod tests {
...
+        // Narrow
+        let fmt_narrow = CurrencyFormatter::<Decimal>::try_new_narrow(locale).unwrap();
+        assert_writeable_eq!(
+            fmt_narrow.format_fixed_decimal(&value, &currency_code),
+            "$12,345.67"
+        );
+    }
+
+    #[test]
+    pub fn test_en_us_aud() {
```
robertbastian: this doesn't test anything that the CAD test doesn't test

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -45,39 +56,50 @@ prefs_convert!(CurrencyFormatterPreferences, PluralRulesPreferences);
...
+    );
+
     icu_provider::gen_buffer_data_constructors!(
-        (prefs: CurrencyFormatterPreferences, options: CurrencyFormatterOptions) -> error: DataError,
+        (prefs: CurrencyFormatterPreferences) -> error: DataError,
         functions: [
-            try_new: skip,
-            try_new_with_buffer_provider,
-            try_new_unstable,
+            try_new_narrow: skip,
```
robertbastian, line 92: why not use the macro-generated version?

younies: Thanks for the suggestion! We have refactored the constructors to use the macro-generated versions as much as possible (for the buffer and unstable constructors).

However, we still have to skip the compiled constructors (`try_new_short`/`try_new_narrow`) in the macro and implement them manually. This is because `CurrencyFormatter` is a composite formatter that internally instantiates `DecimalFormatter` (from `icu_decimal`). Since `icu_experimental`'s local compiled data provider (`Baked`) only contains experimental keys and does not implement the stable decimal keys, the macro-generated compiled constructors fail to compile due to unsatisfied trait bounds on `Baked`.

Manually implementing them allows us to initialize `DecimalFormatter` using its own compiled data, while loading the currency-specific data from `icu_experimental`'s `Baked` provider. This is the same established pattern used in `RelativeTimeFormatter` (see `relativetime.rs:178`).

We have added a `TODO` comment in the code to re-evaluate and try to use the macro-generated compiled versions once `CurrencyFormatter` is stabilized and migrated out of experimental.

Additionally, while refactoring, we noticed that `try_new_narrow` was missing the numbering system fallback loading logic that `try_new_short` had. We have updated both the compiled and unstable versions of `narrow` to use the fallback loader, so they are now fully consistent and correct.

components/experimental/src/dimension/currency/formatter.rs
younies: This is a very good point. The question of what the optimal input representation for currency formatting should be—whether `FixedDecimal` is the right fit, or if we need an exact decimal/money representation (like `units`/`nanos`)—is a crucial architectural question.

However, since this is a pre-existing design aspect (as you noted), changing it would require a much broader discussion that is out of scope for this migration PR.

To ensure we don't lose track of this, I have created a dedicated issue to host this discussion: **#8146**. I have also added a `TODO(#8146)` in `format_fixed_decimal` to track this and ensure we address it before stabilizing the formatter.

components/experimental/src/dimension/currency/format.rs
younies: Good point! It is indeed redundant with the CAD test. I have removed the AUD test case to keep the test suite clean and DRY. Thanks!

components/experimental/src/dimension/currency/format.rs
younies: thanks, done

younies:
We have updated this PR to refactor the `CurrencyFormatter` design:
- The constructors (`try_new_short`, `try_new_narrow`, and their unstable/buffer variants) now take `currency_code: &CurrencyCode` as input.
- The `CurrencyFormatter` struct now stores this as `bound_currency`.
- The `format_fixed_decimal` method signature has been simplified to remove the `currency_code` parameter, as it now uses the `bound_currency` stored in the struct.

This aligns the `Short`/`Narrow` formatter API with the `Long` formatter design (in the stacked PR #8150), ensuring a consistent API across all widths.

younies:
@robertbastian : Ping

robertbastian (approved):

Issue #7693, opened 2026-02-25
Bikeshed Month::as_input()


Manishearth:
@sffc:

> Call it just `.month()` since that is the type it returns?

@robertbastian:

> that'd give use `date.month().month()`

_Originally posted by @sffc in https://github.com/unicode-org/icu4x/pull/7667#discussion_r2824522842_

Manishearth:
This should be a quick bikeshed.

@sffc and @robertbastian can ideally come to a conclusion without meeting time


I think both options here are fine. `.month().month()` is indeed ugly and confusing.

sffc:
`.month().standard_month()` ?

`.month().identity()` ?

Manishearth:
standard = leap vs standard to me

identity is confusing to me. I know what it means because we have been discussing "month identity" but it's a pretty technical topic and not one most will use.

sffc:
`.month.id()`

Manishearth:
I don't hate that.

robertbastian:
it's not an identifier. `MonthInfo` is the output month type, `Month` is the input month type. `MonthInfo::as_input` goes from `MonthInfo` to `Month`.

sffc:
Context: we have a type [`MonthInfo`](https://unicode-org.github.io/icu4x/rustdoc/icu/calendar/types/struct.MonthInfo.html) that has a getter that returns [`Month`](https://unicode-org.github.io/icu4x/rustdoc/icu/calendar/types/struct.MonthInfo.html#method.as_input). This will sometimes be used in a chain with [`Date::month`](https://unicode-org.github.io/icu4x/rustdoc/icu/calendar/struct.Date.html#method.month), which returns MonthInfo.

WG minutes not recorded.

Options:

1. `to_input()` / `as_input()` (according to rust style)
2. `id()`
3. `AsRef<Month>` impl without concrete function
4. `to_month()` / `as_month()` (according to rust style)
5. `month()`

Vote on the thread.

sffc:
My ballot: 2 ~> 4 ~> 5 > 1 > 3

robertbastian:
1 > 4 >> 5 > 3 > 2

Manishearth:
1 ~> 4 ~> 5 > 2 >> 3

Veto reason: I think it's not discoverable and somewhat confusing.

[...]

PR #8147, opened 2026-06-27
Implement difference resolution on DateTimeInput

Manishearth:
Part of #5448

This implements the routine that can compare two dates and figure out what the greatest different field is, including handling flexible day period differences.

I figured this is a smaller piece that we can land early reducing the size of the final set of changes. I don't _love_ landing unused code, but "unused and tested" is fine and a good way to build things up piece by piece.

This also creates the `range` module where all the rest of this is going to live.

## Changelog: N/A

### DIFF
--- a/components/datetime/src/format/input.rs
+++ b/components/datetime/src/format/input.rs
@@ -154,4 +154,10 @@ impl DateTimeInputUnchecked {
                 .into_option(),
         }
     }
+
+    /// Checks if all the time zone fields (currently: offset and ID) are the same.
+    #[allow(dead_code)] // TODO(#5448): range formatting is WIP
+    pub(crate) fn has_same_zone(&self, other: &Self) -> bool {
+        self.zone_offset == other.zone_offset && self.zone_id == other.zone_id
+    }
 }
--- a/components/datetime/src/format/input.rs
+++ b/components/datetime/src/format/input.rs
@@ -155,9 +155,11 @@ impl DateTimeInputUnchecked {
         }
     }
 
-    /// Checks if all the time zone fields (currently: offset and ID) are the same.
+    /// Checks if all the time zone fields (currently: offset, ID, and name timestamp) are the same.
     #[allow(dead_code)] // TODO(#5448): range formatting is WIP
     pub(crate) fn has_same_zone(&self, other: &Self) -> bool {
-        self.zone_offset == other.zone_offset && self.zone_id == other.zone_id
+        self.zone_offset == other.zone_offset
+            && self.zone_id == other.zone_id
+            && self.zone_name_timestamp == other.zone_name_timestamp
     }
 }
--- a/components/datetime/src/range/difference.rs
+++ b/components/datetime/src/range/difference.rs
@@ -112,7 +112,8 @@ pub(crate) fn resolve_difference(
                 }
 
                 // Check flexible day period (DayPeriodB)
-                if dayperiod_names.is_some_and(|dp| has_flexible_day_period_difference(dp, h1, h2)) {
+                if dayperiod_names.is_some_and(|dp| has_flexible_day_period_difference(dp, h1, h2))
+                {
                     return Difference::DayPeriodB;
                 }
                 return Difference::Hour;
--- a/components/datetime/src/range/difference.rs
+++ b/components/datetime/src/range/difference.rs
@@ -89,6 +89,10 @@ pub(crate) fn resolve_difference(
     let month1 = input1.month.map(|m| m.to_input().code());
     let month2 = input2.month.map(|m| m.to_input().code());
     if month1 != month2 {
+        // This also catches the Some != None case in case
+        // one input chooses to not use month codes. This is
+        // expected: date time formatting typically needs month codes
+        // to work and "unspecified" should count as a difference.
         return Difference::Month;
     }
 
--- a/components/datetime/src/range/difference.rs
+++ b/components/datetime/src/range/difference.rs
@@ -11,7 +11,6 @@
 )]
 
 use crate::format::DateTimeInputUnchecked;
-use crate::provider::day_periods::DayPeriodRules;
 use crate::provider::names::DayPeriodNames;
 use icu_calendar::types::YearInfo;
 use icu_time::Hour;
@@ -129,12 +128,7 @@ fn has_flexible_day_period_difference(
     hour1: Hour,
     hour2: Hour,
 ) -> bool {
-    if let Some(rules) = dayperiod_names
-        .names
-        .get(4)
-        .and_then(|s| s.split_at_checked(4))
-        .and_then(|(b, _)| DayPeriodRules::decode_from_str(b))
-    {
+    if let Some(rules) = dayperiod_names.day_period_rules() {
         return rules.name_offset(hour1) != rules.name_offset(hour2);
     }
     false
--- a/components/datetime/src/range/difference.rs
+++ b/components/datetime/src/range/difference.rs
@@ -55,22 +55,22 @@ pub(crate) fn resolve_difference(
     input2: &DateTimeInputUnchecked,
     dayperiod_names: Option<&DayPeriodNames<'_>>,
 ) -> Difference {
-    if input1.zone_offset != input2.zone_offset || input1.zone_id != input2.zone_id {
+    if !input1.has_same_zone(input2) {
         return Difference::Mixed;
     }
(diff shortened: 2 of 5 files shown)


components/datetime/src/range/difference.rs
```diff
@@ -0,0 +1,237 @@
...
+        // One input has a year and the other does not. This is a major structural
+        // difference, so we fall back to Era (the largest date-specific difference).
+        return Difference::Era;
+    }
+
+    // Since we have already established that the years are equal,
+    // comparing month ordinals is sufficient to detect a month difference.
+    let month1 = input1.month.map(|m| m.ordinal);
+    let month2 = input2.month.map(|m| m.ordinal);
+    if month1 != month2 {
```
sffc, line 89: Suggestion: this also catches Some != None. Good noting that in a comment.

components/datetime/src/range/difference.rs
```diff
@@ -0,0 +1,237 @@
...
+    } else if input1.year.is_some() != input2.year.is_some() {
+        // One input has a year and the other does not. This is a major structural
+        // difference, so we fall back to Era (the largest date-specific difference).
+        return Difference::Era;
+    }
+
+    // Since we have already established that the years are equal,
+    // comparing month ordinals is sufficient to detect a month difference.
+    let month1 = input1.month.map(|m| m.ordinal);
+    let month2 = input2.month.map(|m| m.ordinal);
```
sffc: Issue for investigation: I'm not sure that `.ordinal` will consistently be populated. One input could have a MonthCode and the other could have an ordinal month, and I'm not sure whether they are resolved by this point. I think not, because MonthCode is the only thing we should be formatting with.

components/datetime/src/range/difference.rs
```diff
@@ -0,0 +1,237 @@
...
+                }
+            }
+            (YearInfo::Cyclic(_), YearInfo::Cyclic(_)) => {}
+            _ => {
+                return Difference::Era;
+            }
+        }
+        if y1.extended_year() != y2.extended_year() {
+            return Difference::Year;
+        }
```
sffc: Suggestion: write this as something more like,

```rust
match (input1.year, input2.year) {
    (Some(YearInfo::Era(y1)), Some(YearInfo::Era(y2))) => // compare as era years
    (Some(YearInfo::Cyclic(y1)), Some(YearInfo::Cyclic(y2))) => // compare as cycle years
    (None, None) => // no year in input; continue to checking months, I guess?
    _ => // different year types, or Some/None; bail with Difference::Era
}
```

components/datetime/src/range/difference.rs
```diff
@@ -0,0 +1,237 @@
...
+
+    if input1.day_of_month != input2.day_of_month {
+        return Difference::Day;
+    }
+
+    // Compare Time fields
+    if input1.hour != input2.hour {
+        // Hours are different. Check day periods.
+        if let (Some(h1), Some(h2)) = (input1.hour, input2.hour) {
+            // Check standard AM/PM (DayPeriodA)
```
sffc: Nit: This style is inconsistent with other field comparisons. Also, it seems redundant to check `input1.hour != input2.hour` and _only then_ pull them out of the Option. Please implement a match more like the one I suggested above for Year comparison.

components/datetime/src/range/difference.rs
```diff
@@ -0,0 +1,237 @@
...
+
+fn has_flexible_day_period_difference(
+    dayperiod_names: &DayPeriodNames<'_>,
+    hour1: Hour,
+    hour2: Hour,
+) -> bool {
+    if let Some(rules) = dayperiod_names
+        .names
+        .get(4)
+        .and_then(|s| s.split_at_checked(4))
```
sffc: Nit: Don't hardcode magic numbers

Is this the best place for this fn? Seems like it could be an instance method on DayPeriod something?

sffc (changes requested):

components/datetime/src/range/difference.rs
Manishearth: Yeah I split this out with intent to put it somewhere but didn't end up deciding where it should go.

Manishearth: I think DayPeriodNames is right., that's where the rest of this logic lives.

Manishearth: Argh, I had fixed this, I hadn't pushed. My bad, this will be a force push.

Manishearth: Used `jj` to avoid a force push

components/datetime/src/range/difference.rs
```diff
@@ -0,0 +1,245 @@
...
+        return Difference::Mixed;
+    }
+
+    // Compare Date fields
+    match (input1.year, input2.year) {
+        (Some(y1 @ YearInfo::Era(e1)), Some(y2 @ YearInfo::Era(e2))) => {
+            if e1.era != e2.era {
+                return Difference::Era;
+            }
+            if y1.extended_year() != y2.extended_year() {
```
sffc: Nit: read the era year directly, not the extended year. The eras are the same

components/datetime/src/range/difference.rs
```diff
@@ -0,0 +1,245 @@
...
+                return Difference::Year;
+            }
+        }
+        (Some(y1 @ YearInfo::Cyclic(_)), Some(y2 @ YearInfo::Cyclic(_))) => {
+            if y1.extended_year() != y2.extended_year() {
```
sffc: Nit: read related_iso (or whatever we ended up naming the field), not extended_year

components/datetime/src/range/difference.rs
```diff
@@ -0,0 +1,245 @@
...
+            // This is a major structural difference, so we fall back to Era
+            // (the largest date-specific difference).
+            return Difference::Era;
+        }
+    }
+
+    // Since we have already established that the years are equal,
+    // comparing month codes is sufficient to detect a month difference.
+    let month1 = input1.month.map(|m| m.to_input().code());
+    let month2 = input2.month.map(|m| m.to_input().code());
```
sffc: Vent: `to_input()` is such a confusing name in this context. `.to_identity()` would have been so much more clear.

https://github.com/unicode-org/icu4x/issues/7693

sffc: Issue: use `impl PartialEq for Month`

components/datetime/src/range/difference.rs
```diff
@@ -0,0 +1,245 @@
...
+/// If it is `None`, flexible day periods will be assumed to be the same, and only
+/// standard AM/PM (`a`) will be compared.
+pub(crate) fn resolve_difference(
+    input1: &DateTimeInputUnchecked,
+    input2: &DateTimeInputUnchecked,
+    dayperiod_names: Option<&DayPeriodNames<'_>>,
+) -> Difference {
+    if input1.zone_offset != input2.zone_offset || input1.zone_id != input2.zone_id {
+        return Difference::Mixed;
+    }
```
sffc: Suggestion: this looks fragile, because if there are more zone fields added to DateTimeInputUnchecked, this will become stale. Maybe move this to be an instance method on DateTimeInputUnchecked?

I think it should be basically "true or false: everything other than year/month/day/hour/minute/second is equal"

sffc (changes requested):

components/datetime/src/range/difference.rs
Manishearth: I think this is internal code, not user code, and it's okay for choices made with user code in mind to make internal code a little bit more confusing

components/datetime/src/format/input.rs
```diff
@@ -154,4 +154,10 @@ impl DateTimeInputUnchecked {
                 .into_option(),
         }
     }
+
+    /// Checks if all the time zone fields (currently: offset and ID) are the same.
+    #[allow(dead_code)] // TODO(#5448): range formatting is WIP
+    pub(crate) fn has_same_zone(&self, other: &Self) -> bool {
+        self.zone_offset == other.zone_offset && self.zone_id == other.zone_id
```
sffc: Nit: Now that this is in this file you can see that you are missing `zone_name_timestamp`, which you should probably include in this, since it can make the display names be different.

sffc (approved):

sffc:
Praise: Thanks for the small PRs

sffc (approved):

PR #8148, opened 2026-06-27
Move PatternsByGreatestDifference::get_*_pattern resolvers to PatternsByGreatestDifferenceULE

Manishearth:
This is annoying.

Normally, our VarULE types are the "internal" types and we implement complex logic on our regular types, because it's easy, and you can always go from a VarULE type to a regular one but the other way around isn't always possible.

So having methods on `PatternsByGreatestDifference` is great: they can refer to the "nice" decoded fields, and if we have a `PatternsByGreatestDifferenceULE` (which we will) we can always `ZeroFrom` it to the other type.

Unfortunately, there's one downside: because `PatternsByGreatestDifference` _could_ be allocated, we cannot return long lived `'data` references to its patterns. This is why [`PatternsByGreatestDifference::get_*_pattern`](https://unicode-org.github.io/icu4x/rustdoc/icu/datetime/provider/range_patterns/struct.PatternsByGreatestDifference.html#method.get_date_pattern) returns things with lifetime `'a` not `'data`.

So we can't actually write code that e.g. returns an appropriately resolved `FormattedDateTimeRange` that contains `'data` patterns.

`PatternsByGreatestDifference*` lives inside a VarZeroVec in [`GenericPackedPatterns`](https://unicode-org.github.io/icu4x/rustdoc/icu/datetime/provider/packed_pattern/struct.GenericPackedPatterns.html) anyway, which means ICU4X code will _always_ have this type as `PatternsByGreatestDifferenceULE` (even a runtime mutated VarZeroVec contains encoded data). So there's no API downside of moving these methods over to `PatternsByGreatestDifferenceULE` where they will be actually callable.

The main downsides are:

 - The code is slightly worse since it needs to decode some AsULEs. Not a big deal
 - The tests are slightly worse because they need to convert to `PatternsByGreatestDifferenceULE`. _shrug_.

### DIFF
--- a/components/datetime/src/provider/pattern/runtime/pattern.rs
+++ b/components/datetime/src/provider/pattern/runtime/pattern.rs
@@ -40,8 +40,8 @@ pub struct Pattern<'data> {
 }
 
 /// Fully borrowed version of [`Pattern`].
-#[derive(Debug, Copy, Clone)]
-pub(crate) struct PatternBorrowed<'data> {
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct PatternBorrowed<'data> {
     pub(crate) items: &'data ZeroSlice<PatternItem>,
     pub(crate) metadata: PatternMetadata,
 }
@@ -146,6 +146,17 @@ impl<'data> PatternBorrowed<'data> {
         Pattern {
             items: self.items.as_zerovec(),
             metadata: self.metadata,
+        }
+    }
+}
+
+impl<'data> zerofrom::ZeroFrom<'data, PatternULE> for PatternBorrowed<'data> {
+    #[inline]
+    fn zero_from(ule: &'data PatternULE) -> Self {
+        use zerovec::ule::AsULE;
+        Self {
+            items: &ule.items,
+            metadata: <PatternMetadata as AsULE>::from_unaligned(ule.metadata),
         }
     }
 }
--- a/components/datetime/src/provider/range_patterns.rs
+++ b/components/datetime/src/provider/range_patterns.rs
@@ -7,6 +7,7 @@
 #[cfg(feature = "datagen")]
 use crate::provider::packed_pattern::GenericUnpackedPatterns;
 use crate::provider::packed_pattern::{GenericPackedPatterns, PackedPatternsBuilderHelper};
+use crate::provider::pattern::runtime::PatternBorrowed;
 use crate::provider::pattern::runtime::{Pattern, PatternULE};
 use icu_provider::prelude::*;
 use zerovec::VarZeroVec;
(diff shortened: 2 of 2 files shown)


components/datetime/src/provider/pattern/runtime/pattern.rs
```diff
@@ -40,7 +40,7 @@ pub struct Pattern<'data> {
 }
 
 /// Fully borrowed version of [`Pattern`].
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
```
Manishearth, line 43: I need this later. Can remove from this PR if desired.

components/datetime/src/provider/range_patterns.rs
```diff
@@ -235,55 +258,51 @@ fn resolve_fallback(header: GreatestDifferenceHeader, requested: u8, max_value:
...
-                .unwrap_or_default()
+                .map(<PatternBorrowed as zerofrom::ZeroFrom<PatternULE>>::zero_from)
+                .unwrap_or(PatternBorrowed::DEFAULT)
         };
 
         match state {
-            RangeStructure::Absent => RangePatternInfo::Symmetric(Pattern::default()),
-            RangeStructure::Symmetric => RangePatternInfo::Symmetric(get_pat(0)),
-            RangeStructure::FullRange => RangePatternInfo::FullRange(get_pat(0)),
+            RangeStructure::Absent => RangePatternInfoBorrowed::Symmetric(PatternBorrowed::DEFAULT),
```
Manishearth, line 288: observation: PatternBorrowed::DEFAULT was useful here

components/datetime/src/provider/range_patterns.rs
```diff
@@ -540,23 +563,35 @@ mod tests {
                 .is_date_field_present(DateGreatestDifferenceField::Year)
         );
 
+        let pgd_ule_box = zerovec::ule::encode_varule_to_box(&pgd);
```
Manishearth, line 576: this boxing is unfortunately inevitable, see PR body

sffc:
I have not yet looked at the code, but based on the PR description:

The way I solved this problem for non-range formatting was to introduce a runtime::PatternBorrowed type which differs from runtime::Pattern in that it doesn't have a Cow and therefore can return the longer lifetime. DateTimeFormatter::format borrows from the receiver and stores a PatternBorrowed (wrapped in some enums) in the return value, a FormattedDateTime.

PatternBorrowed is nice because it can contain all the methods, and it can be created from a bunch of different sources and forms of data (VarULE or Cow).

In non-range formatting, I pull the pattern out of the PluralElements in the .format function. I think you can probably do this in range formatting, too, so PatternBorrowed is all you need. But, if you wanted to instead retain the PatternsByGreatestDifference into the impl Writeable, then instead of PatternBorrowed, you would use PatternsByGreatestDifferenceBorrowed.

Manishearth:
Hmm, the tricky thing here is that PGD is an array.

However we know it is an array of length 4, so in theory we could return `[PatternBorrowed; 4]`. That might bloat some stack sizes.

I'm not really sure if a Borrowed type is better than doing this on the ULE type.

sffc:
We don't need to return an array of all 4 patterns. We can pick the pattern in `.format_range` and then only store the pattern we actually need

Manishearth:
@sffc this _is_ the "pick a pattern" code.

I believe my WIP changes on top of this are using PatternBorrowed. For them to be able to use PatternBorrowed, the "pick a pattern" code needs to live on PatternsByGreatestDifferenceULE.

sffc:
So are you basically getting ready to implement the function for range patterns?

https://github.com/unicode-org/icu4x/blob/f4b2756bb5ec5c1a918863247fa3550716de610a/components/datetime/src/provider/packed_pattern.rs#L543

That fn calls `PluralElementsPackedULE::get_default()` (should in the future pass in the plural form), so adding a fn to get the pattern for the given GD onto the VarULE type seems consistent with the non-range formatting.

sffc:
That fn uses an internal type `PluralElementsUnpacked` which unpacks the VarULE with the correct lifetime; the `Unpacked` element is borrowed-only, so that could be an abstraction layer to consider

Manishearth:
Yes. See https://github.com/unicode-org/icu4x/pull/8149, specifically the `fn select` impl: https://github.com/unicode-org/icu4x/pull/8149/changes/5fef8c8d31871264ac0d6ef96b8c6a96bec61622#diff-a7ba40b6e383f5e3f2781d50fa6c8d0a036aaab7f6ec0fc0211069004fa24332R273

Manishearth:
I basically don't think we need another abstraction layer here: this is doable within VarULE and we should probably just do that. We have other abstraction layers like RangePatternBorrowed that pull their weight here.

sffc (approved):
Thanks for the small PRs

PR #8149, opened 2026-06-29
Add unstable date range (interval) formatter

Manishearth:
Part of https://github.com/unicode-org/icu4x/issues/5448

Depends on <s>https://github.com/unicode-org/icu4x/pull/8147, https://github.com/unicode-org/icu4x/pull/8148, #8172,  https://github.com/unicode-org/icu4x/pull/8173</s>

Can be reviewed commit by commit

This adds FixedCalendarDateRangeFormatter and DateRangeFormatter, matching the existing date formatter types. There is no subslicing into date/time/zoned types, I don't think we need it here.

`format_impl_shared` is the main juicy bit: it handles the dispatch to the different types of formatted date ranges. The rest is mostly just gluing everything together, though there is some nontrivial skeleton match logic in `DateRangePatternSelectionData::select`. It might be possible to copy that code out.

This does not add FFI. 

Hopefully this can make the 2.3 release. Unfortunately I will have limited capacity to handle feedback for the next 3 weeks, so it might not, it might be useful to try and land this and file many small followups if the approach looks good.

## Changelog

icu_datetime: Add unstable range formatter
  - New types: `DateRangeFormatter`, `FixedCalendarDateRangeFormatter`, `FormattedDateRange`


Manishearth:
Pre rebase state

<details>

```
├─╯  (empty) agent-driven: Address review comments on PR #8172
│ ×  uvvqsvty me 2 days ago range-formatter* ca09c21aba7c35 (conflict)
│ │  agent-driven: feat(datetime): Add range scaffolding traits and implement on fieldsets
│ ×  lnstpsqz me 2 days ago d74f0e858f9b9e (conflict)
│ │  agent-driven: feat(datetime): Implement DateRangeFormatter with dynamic calendar conversion
│ ×  prozyzsx me 2 days ago 5cdc4359ed1b35 (conflict)
│ │  agent-driven: feat(datetime): Implement FixedCalendarDateRangeFormatter
│ ×  tvpuomoq me 2 days ago 5f423471b81040 (conflict)
╭─┤  feat(datetime): Introduce shared range formatting infrastructure
○ │  stslkwnm me 3 days ago 05006045341806
│ │  agent-driven: feat(datetime): Implement range formatting result writeables
│ │ ◆  nnturkqp me 1 hour ago main* 3452708d1d4719
│ │ │  Writeables for range formatting (#8172)
│ │ ~  (elided revisions)
├───╯
│ │ ○  lztuvvwr me 3 days ago range-scaffold b22cdf2faf4000
│ ├─╯  agent-driven: fix(ci): fix FormattableHijriRules doctest for RangeSkeletaV1
│ ○  mmkoomtq me 3 days ago 4eafffef9ca673
│ │  agent-driven: feat(datetime): Implement range pattern loading
│ ○  zryoputo me 3 days ago 855c1505cba7c0
├─╯  agent-driven: feat(datetime): Add range scaffolding traits and implement on fieldsets
│ ○  xxzokkvy me 3 days ago dbb0b17fbe8c14
├─╯  agent-driven: feat(datetime): Implement range pattern loading
◆  xnnvmnmv me 5 days ago aef08cb11c4695
```

</details>

components/datetime/src/raw/neo.rs
```diff
@@ -272,6 +274,64 @@ impl DateRangePatternSelectionData {
...
+            }
+            (YearStyle::Auto | YearStyle::NoEra, _) => {
+                debug_assert!(false, "unreachable");
+                PackedSkeletonVariant::Standard
+            }
+        };
+
+        let ule = payload.get_element(options.length(), variant)?;
+
+        let field = match diff {
```
sffc: The function above this line seems like it is duplicated

components/datetime/src/range/formatter.rs
```diff
@@ -0,0 +1,723 @@
...
+};
+use icu_provider::prelude::*;
+
+// ================= Private Shared Infrastructure =================
+
+/// An internal trait that abstracts the core formatting capabilities needed for range formatting.
+///
+/// This allows shared range formatting infrastructure to work over both
+/// `FixedCalendarDateTimeFormatter` and `DateTimeFormatter`.
+trait RangeFormatterCore {
```
sffc: Question: why did you use a trait here; the non-range formatter don't have a trait at this spot. They call down to the raw neo formatter very quickly. I would like for the range formatting code to be basically the same as the non-range formatting code (with common parts refactored out to shared code).

components/datetime/src/raw/neo.rs
Manishearth: Oh, yeah, this was one of the refactors on my list of prerequisites that I guess I never got around to. Fixed

components/datetime/src/range/formatter.rs
Manishearth: This is mostly because of `format_impl_shared`: the non-range formatter basically very quickly constructs a formattable, the range formatter has complicated specced logic before it can do that. I'd rather use a trait than duplicate `format_impl_shared`.

There may be other ways to do this, I reached for the easy option, and it's an internal trait so we can experiment with cleanups later.

Manishearth: (The initial "get something working end-to-end" pass didn't have traits, I added traits on purpose because this complicated logic was duplicated)

Cargo.toml
```diff
@@ -417,6 +417,7 @@ copy_impl_added = "allow"
 enum_no_repr_variant_discriminant_changed = "allow" # we don't consider this breaking https://github.com/obi1kenobi/cargo-semver-checks/issues/1376
 enum_variant_marked_deprecated = "allow"
 struct_field_marked_deprecated = "allow" # allow deprecation in patch releases
+trait_associated_type_added = "allow"
```
sffc: Is this safe?

components/datetime/src/lib.rs
```diff
@@ -152,6 +152,10 @@ pub use neo::DateTimeFormatterPreferences;
 pub use neo::FixedCalendarDateTimeFormatter;
 pub use neo::FormattedDateTime;
 pub use neo::NoCalendarFormatter;
+#[cfg(feature = "unstable")]
+pub use range::DateRangeFormatter;
```
sffc: Discussion: I think I might prefer to keep this in the module

components/datetime/src/provider/pattern/runtime/pattern.rs
```diff
@@ -149,9 +149,6 @@ impl<'data> PatternBorrowed<'data> {
         }
     }
 
-    /// Returns the index of the first repeated field in the pattern, if any.
-    ///
-    /// This is used to find the split point for range patterns.
```
sffc, line 154: Deleted comments

components/datetime/src/range/formatter.rs
```diff
@@ -0,0 +1,723 @@
...
+    let end_formatted = format_datetime(core, end, names);
+    let glue = range_selection.range_glue.get();
+    FormattedDateRangeInner::Fallback(FormattedRangeFallback {
+        start: start_formatted,
+        end: end_formatted,
+        glue,
+    })
+}
+
+// ================= Public Formatters =================
```
sffc: Nit: this file should be split

components/datetime/src/range/formatter.rs
```diff
@@ -0,0 +1,723 @@
...
+
+/// The shared implementation of range formatting, used by both `DateTimeFormatter`
+/// and `FixedCalendarDateTimeFormatter`.
+///
+/// This function orchestrates the formatting by:
+/// 1. Resolving the greatest difference between the start and end dates.
+/// 2. Selecting the appropriate pattern (date range, time range, or fallback).
+/// 3. Formatting the sides and wrapping them in the appropriate result type.
+fn format_impl_shared<'a>(
+    core: &'a impl RangeFormatterCore,
```
sffc: Issue: `impl` causes this code to be duplicated. We want this code to be a single "raw"

components/datetime/src/range/formatter.rs
```diff
@@ -0,0 +1,723 @@
...
+    let is_mixed = range_selection.date_range.payload.is_payload()
+        && range_selection.time_range.payload.is_payload();
+
+    // Early fallback for mixed date-time formatter with date difference.
+    // UTS 35: If date differs in a mixed skeleton, fall back to range fallback (Case 4).
+    if is_mixed
+        && !diff.is_time_diff()
+        && !matches!(
+            diff,
+            Difference::None | Difference::Mixed | Difference::Second
```
sffc: I'm confused: this is checking for date difference of second or less, but the difference can still be in hours or minutes or day period. This seems to contradict the comment

components/datetime/src/range/formatter.rs
```diff
@@ -0,0 +1,723 @@
...
+    start: &DateTimeInputUnchecked,
+    end: &DateTimeInputUnchecked,
+) -> FormattedDateRange<'a> {
+    let names = core.names();
+    let dayperiods = names.dayperiod_names();
+
+    // 1. Resolve difference
+    let diff = resolve_difference(start, end, dayperiods);
+
+    let is_mixed = range_selection.date_range.payload.is_payload()
```
sffc: Issue: I think the word "mixed" is being used to mean two different things in the same fn: a time zone difference, and a formatter with both date and time.

components/datetime/src/range/formatter.rs
```diff
@@ -0,0 +1,723 @@
...
+    fn select_time_only(
+        &self,
+        input: &DateTimeInputUnchecked,
+    ) -> DateTimeZonePatternDataBorrowed<'_> {
+        DateTimeZonePatternDataBorrowed {
+            date: None,
+            time: self
+                .selection
+                .time
+                .select(input, self.selection.options, self.selection.prefs),
```
sffc: Nit/Issue: I still think the trait adds unnecessary complexity. It is like multiple layers of abstraction over basically just this one line, which is not that bad of a line to inline to the call site.

components/datetime/src/range/formatter.rs
```diff
@@ -0,0 +1,723 @@
...
+    core: &'a impl RangeFormatterCore,
+    range_selection: &'a DateTimeZoneRangePatternSelectionData,
+    start: &DateTimeInputUnchecked,
+    end: &DateTimeInputUnchecked,
+    names: RawDateTimeNamesBorrowed<'a>,
+) -> FormattedDateRangeInner<'a> {
+    let start_formatted = format_datetime(core, start, names);
+    let end_formatted = format_datetime(core, end, names);
+    let glue = range_selection.range_glue.get();
+    FormattedDateRangeInner::Fallback(FormattedRangeFallback {
```
sffc: Suggestion: this and the other fns could be constructors

sffc (changes requested):

Cargo.toml
Manishearth: No, mistake.

components/datetime/src/provider/pattern/runtime/pattern.rs
Manishearth: Intentional: Duplicate comment cleaned up

components/datetime/src/lib.rs
Manishearth: I'm fine with that.

components/datetime/src/range/formatter.rs
Manishearth: Renamed Difference::Mixed to Difference::Incomparable

components/datetime/src/range/formatter.rs
Manishearth: The goal is to have `format_impl_shared` (complicated specced logic) be one routine. I attempted some ways of avoiding this, but they were less reviewable and I didn't want to further complicate this PR. I have some ideas, I might still try them but would prefer they do not block this as release-blockers.

components/datetime/src/range/formatter.rs
Manishearth: See comment below about having some ideas here but not wanting to complicate this PR. This should be a non-release-blocking followup given the timeline here.

components/datetime/src/range/formatter.rs
Manishearth: It's checking "not a time diff and not things that are neither time diffs nor date diffs". That's indeed confusing, I just created an `is_date_diff` to clean it up.

Manishearth:
I had to update this PR to the MinuteOptional coarse hour cycle stuff from https://github.com/unicode-org/icu4x/pull/8237, and this uncovered a bug with pattern selection. I fixed it, however I discovered some interesting behavior where if you ask for a range difference between "7am" and "7:15am", in the case where minutes are supposed to be omitted, you'll still get "7-7:15am" because there is no "omit minutes for first one" range pattern.

In theory we could treat "MinuteOptional behavior differs" as a special type of `Difference` and fall back to glue in that case.

This is probably a nitpicky CLDR discussion. What we have now works.

components/datetime/src/range/formatter.rs
Manishearth: I think this may also work better as a followup since I'm hoping to attempt various cleanups here.

components/datetime/src/provider/range_patterns.rs
```diff
@@ -451,6 +452,29 @@ impl<'a>
...
+    #[cfg(feature = "unstable")]
+    pub(crate) fn get_coarse_hour_cycle(
+        &self,
+        length: crate::options::Length,
+        diff: TimeGreatestDifferenceField,
+    ) -> Option<crate::provider::pattern::CoarseHourCycle> {
+        use crate::provider::packed_pattern::PackedSkeletonVariant;
+        // To get the hour cycle without reading the whole pattern, we can check the time
+        // granularity of the hour pattern (packed variant `Standard`).
+        let element = self.get_element(length, PackedSkeletonVariant::Standard)?;
```
sffc, line 466: Question: In the packed range patterns, does `Standard` still mean "skeleton H or ha"?

components/datetime/src/provider/range_patterns.rs
```diff
@@ -451,6 +452,29 @@ impl<'a>
...
+        // Query the pattern with maximal difference
+        let pattern = element.get_time_pattern(diff)?;
```
sffc, line 468: Observation: with skeleton H or ha, will there ever be more than 1 pattern inside this store? It might be more clear to load the first pattern instead of using `diff` which does do anything here.

components/datetime/src/range/formatter.rs
```diff
@@ -0,0 +1,398 @@
...
+
+        Ok(Self {
+            datetime_formatter,
+            range_selection,
+        })
+    }
+
+    /// Constructor taking a data provider.
+    ///
+    /// [📚 Help choosing a constructor](icu_provider::constructors)
```
sffc: Nit: if you can't generate these fns with the macro, please at a minimum use the docs macro

```
#[doc = icu_provider::gen_buffer_unstable_docs!(UNSTABLE, Self::try_new)]
```

components/datetime/src/range/formatter.rs
```diff
@@ -0,0 +1,398 @@
...
+        )
+    }
+
+    #[cfg(feature = "serde")]
+    /// Constructor taking a buffer provider.
+    ///
+    /// ✨ *Enabled with the `serde` Cargo feature.*
+    ///
+    /// [📚 Help choosing a constructor](icu_provider::constructors)
+    pub fn try_new_with_buffer_provider<P>(
```
sffc, line 157: Thought: you could use `gen_buffer_constructors_with_external_loader!` like in the other file. But, that macro might not carry its weight, especially without `any` constructors.

components/datetime/src/range/formatter.rs
```diff
@@ -0,0 +1,398 @@
...
+            field_set_with_options,
+        )
+    }
+
+    #[cfg(feature = "serde")]
+    /// Constructor taking a buffer provider.
+    ///
+    /// ✨ *Enabled with the `serde` Cargo feature.*
+    ///
+    /// [📚 Help choosing a constructor](icu_provider::constructors)
```
sffc: docs macro

components/datetime/src/range/formatter.rs
```diff
@@ -0,0 +1,398 @@
...
+        external_loader: &(impl FormattableAnyCalendarLoader + DecimalFormatterLoader),
+        prefs: DateTimeFormatterPreferences,
+        field_set_with_options: FSet,
+    ) -> Result<Self, DateTimeFormatterLoadError>
+    where
+        P: ?Sized
+            + AllAnyCalendarFormattingDataMarkers<FSet>
+            + AllAnyCalendarRangePatternDataMarkers<FSet>,
+    {
+        let field_set_with_options_clone = field_set_with_options.clone();
```
sffc: Issue: this clone is used only for calling `.get_field()`, which takes `&self`, so you shouldn't need to clone. This is probably a cheap clone, but we shouldn't use `.clone()` in ICU4X library code except in very deliberate situations.

components/datetime/src/range/formatter.rs
```diff
@@ -0,0 +1,398 @@
...
+
+        let range_selection = DateTimeZoneRangePatternSelectionData::try_new_with_skeleton(
+            &FormattableAnyCalendarNamesLoader::<<FSet::D as DateDataMarkers>::RangeSkel, _>::new(
+                provider,
+                &datetime_formatter.calendar,
+            ),
+            &<FSet::T as TimeMarkers>::TimeRangeSkeletonPatternsV1::bind(provider),
+            &DatetimePatternsRangeGlueV1::bind(provider),
+            prefs,
+            field_set_with_options.get_field(),
```
sffc: Suggestion: I think the return value of `get_field` is the same both here and above, so you should avoid calling the fn twice. I believe this is the fn that converts the FSet to a CompositeFieldSet, which is not free.

components/datetime/src/range/formatter.rs
```diff
@@ -0,0 +1,398 @@
...
+        external_loader: &impl DecimalFormatterLoader,
+        prefs: DateTimeFormatterPreferences,
+        field_set_with_options: FSet,
+    ) -> Result<Self, DateTimeFormatterLoadError>
+    where
+        P: ?Sized
+            + AllFixedCalendarFormattingDataMarkers<C, FSet>
+            + AllFixedCalendarRangePatternDataMarkers<C, FSet>,
+    {
+        let field_set_with_options_clone = field_set_with_options.clone();
```
sffc: ditto

components/datetime/src/range/formatter.rs
```diff
@@ -0,0 +1,398 @@
...
+            prefs,
+            field_set_with_options_clone.get_field(),
+        )?;
+
+        let range_selection = DateTimeZoneRangePatternSelectionData::try_new_with_skeleton(
+            &<FSet::D as TypedDateDataMarkers<C>>::DateRangeSkeletonPatternsV1::bind(provider),
+            &<FSet::T as TimeMarkers>::TimeRangeSkeletonPatternsV1::bind(provider),
+            &DatetimePatternsRangeGlueV1::bind(provider),
+            prefs,
+            field_set_with_options.get_field(),
```
sffc: ditto

components/datetime/src/range/formatter_impl.rs
```diff
@@ -0,0 +1,327 @@
...
+    for FixedCalendarDateTimeFormatter<C, FSet>
+{
+    fn names(&self) -> RawDateTimeNamesBorrowed<'_> {
+        self.names.as_borrowed()
+    }
+    fn select_datetime(
+        &self,
+        input: &DateTimeInputUnchecked,
+    ) -> DateTimeZonePatternDataBorrowed<'_> {
+        self.selection.select(input)
```
sffc: Issue: I know there are already two other threads about this, but the _only_ thing this trait does is to abstract over `self.selection` and `self.names` between FixedCalendarDateTimeFormatter and DateTimeFormatter, but those fields _are already the same types_. On both of the types, they are `DateTimeZonePatternSelectionData` and `RawDateTimeNames<FSet>`, respectively. I feel like Robert when nitpicking about this, but I don't think we should willingly add a new trait to the codebase that literally does nothing.

components/datetime/src/range/formatter_impl.rs
```diff
@@ -0,0 +1,327 @@
...
+    };
+    FormattedDateRange(inner)
+}
+
+/// Formats the range using the fallback range pattern (gluing the fully formatted
+/// start and end datetimes together).
+///
+/// This is used as a final fallback when no specific greatest difference pattern
+/// is available, or when the difference requires a full fallback (Case 4).
+fn fallback_format_shared<'a>(
```
sffc: Nit: The `_shared` in the function name is distracting. Reading this, what comes to mind is that an LLM made this fn to be "shared" between all types of fallback. We don't use a `_shared` suffix on other fns that have more than one call site, though.

sffc (commented):
I'm trying to focus on the big things, but it's hard to not notice the small things.

components/datetime/src/raw/neo.rs
```diff
@@ -288,7 +320,10 @@ impl DateTimeInputUnchecked {
             TimePrecision::Subsecond(f) => (PackedSkeletonVariant::Variant1, Some(f)),
             TimePrecision::MinuteOptional => {
                 let minute = self.minute.unwrap_or_default();
-                if minute.is_zero() && matches!(coarse_hour_cycle, Some(CoarseHourCycle::H11H12)) {
+                if minute.is_zero()
+                    && second_range_minute_is_zero.unwrap_or(true)
```
sffc, line 324: Issue: If in doubt on a particular CLDR behavior, I prefer to do the thing that causes less code complexity first, and change later. From reviewing the test cases, it seems like you are outputting either "7-8 AM" or "7:00-8:15 AM". I think it is fine to output "7-8:15 AM" if it makes the code simpler.

components/datetime/src/provider/range_patterns.rs
Manishearth: Yes the selection code is exactly the same.

components/datetime/src/raw/neo.rs
Manishearth: No, that is the case that makes the code less simple. There is no pattern that produces "7-8:15 AM", the only way to produce that is to have a glue pattern, and this means you need to hoist the MinuteOptional checks out into `format_impl` code.

Without this, the code outputs "7-7AM" when given 7 - 7:15 as a time. That's just incorrect. The current output is correct (including according to the spec) but _potentially_ not ideal.

The open CLDR question is basically which is the "worse" of the two options: having to fall back to glue pattern, or not handling MinuteOptional for half of a pattern. Our current code hits the latter case, which AIUI matches the spec (which does not mention MinuteOptional during range pattern resolution). An argument could be made for the former case, but that's more complicated to implement.

components/datetime/src/range/formatter_impl.rs
Manishearth: Okay, then can you please say why my proposed plan in the other two threads is not acceptable? This is a new _private_ trait. I am _already_ proposing to clean this up in a separate PR. It does not seem to be the case that this design is unreviewable, just something you don't want in the code. We have been building up a norm of moving things into followups to smoothen review, why are you not letting me invoke that here?

They are not "already the same types", DateTimeFormatter types are different, and they no longer wrap "raw" types since your neo changes. I could "splat" them but that starts pulling in DateTimeFormatter complexity.

I want to do this in a way that gets rid of any monomorphization here as well, and that's not 100% straightforward.

In earlier variations of this PR I _tried this_ and the code got messy. It's possible that other changes to this code have since simplified things: I know that having this trait made it easy for me to tweak things. Given the experience with the range datagen PR I've been opting for reviewability over all else here: this is a large PR, I'm making choices that are easy to follow.

I may still work on this if I have the time, but I'd really, really prefer to land this large change _that we want to get into a release_ and then perform internal refactoring followups as needed.

Manishearth: I ended up doing this, my partial branch experimenting this was pretty close to done anyway. Split it into two commits to make it readable, one is a pure code move.

components/datetime/src/provider/range_patterns.rs
Manishearth: I think this is more clear: I would either have to justify that getting the first element is "fine", or I can perform a call I _know_ will return something sensible in context and just use that. The higher-level `get_time_pattern` API seems harder to make mistakes with.

Manishearth: Basically, clean API call > lower level `.get(0)` with big justification comment

components/datetime/src/range/formatter_impl.rs
Manishearth: I would still prefer that when I ask for something to be a followup that reasons be given why that is not acceptable before continuing to ask for it.

components/datetime/src/provider/range_patterns.rs
sffc: I think `.get(0)` with a 2-3-line comment is clearer than "passing down a variable as a function argument that literally does nothing", but ok

sffc: What I mean is: this fn takes `diff` as an argument

```
    pub(crate) fn get_coarse_hour_cycle(
        &self,
        length: crate::options::Length,
        diff: TimeGreatestDifferenceField,
    ) 
```

But, there can be only 1 CourseHourCycle to describe the entire `GenericPackedPatterns`. It is considered an invariant of the data. By accepting that argument to this fn, it makes it look like the CourseHourCycle depends on the Difference, which it doesn't.

components/datetime/src/raw/neo.rs
```diff
@@ -273,13 +239,79 @@ impl DateRangePatternSelectionData {
...
+
+    /// Resolve the variant to select based on time precision
+    ///
+    /// In case of range patterns, `second_range_minute_is_zero` should be set to whether or not
+    /// the second input has a nonzero `minute` field. Set to None for non range patterns.
     fn resolve_time_precision(
         &self,
         time_precision: TimePrecision,
         coarse_hour_cycle: Option<CoarseHourCycle>,
+        second_range_minute_is_zero: Option<bool>,
```
sffc, line 314: Suggestion (optional): take a `range_second_input: Option<&Self>` instead of `Option<bool>`

components/datetime/src/range/formatter_impl.rs
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        // 2. Select pattern and format
+        let inner = match diff {
+            Difference::None => FormattedDateRangeInner::Single(self.format_datetime(start)),
+            Difference::Incomparable | Difference::Second => self.format_fallback(start, end),
+            diff => {
+                if is_mixed {
+                    // Case 3: Mixed range, only time differs (date diff was handled by early fallback).
+                    let date_formatted = self.format_date_only(start);
```
sffc, line 151: Please test what happens with `fieldsets::E::with_time(...)`, which loads the `ej` pattern. If it is broken, you can fix it in a follow-up.

sffc (approved):

components/datetime/src/range/formatter_impl.rs
sffc: Thanks for fixing it. The new code is what I had in mind. It is easier to reason about and less likely to cause binary code duplication.

I wouldn't have pressed on this if it seemed like the refactor was going to take a lot of time. But, it always looked to me like a 5-minute change that makes the state of the repository substantially better.

Issue #8146, opened 2026-06-26
currency: Design of input numeric representation for CurrencyFormatter


younies:
During the migration of `CurrencyFormatter` to the typestate `ValueRepresentation` design (see [PR #8145](https://github.com/unicode-org/icu4x/pull/8145)), a design discussion arose regarding the optimal input numeric representation for currency formatting.

### The Issue
Currently, `CurrencyFormatter::format_fixed_decimal` takes `FixedDecimal` as input:
```rust
pub fn format_fixed_decimal<'l>(&'l self, value: &'l FixedDecimal) -> ...
```
However, `FixedDecimal` represents a number with a *user-selected* magnitude and fractional length. Currency formatting, on the other hand, is governed by strict currency-specific rules regarding:
1. **Magnitude/Decimal Places**: e.g., USD typically has 2 decimal places, JPY has 0, IQD has 3.
2. **Rounding**: Currencies have specific rounding increments (e.g., cash rounding to the nearest 0.05 in some locales).

Because the formatter must enforce these currency-specific rules anyway, it may not make sense to accept a `FixedDecimal` where the user has already determined the magnitude and rounding. 

### Proposed Alternatives
It might be more robust to accept an **exact decimal/money representation** that preserves the exact value without pre-determined formatting constraints, and let the formatter handle the magnitude and rounding internally.

Possible models to consider:
*   An integer-based `units` + `nanos` representation (similar to Google's [`google.type.Money`](https://github.com/googleapis/googleapis/blob/fdd6a6e4ef12853400539139d014d96e34253166/google/type/money.proto#L30-L40)).
*   A new trait or struct representing exact decimals.
*   Integrating with a broader ICU4X exact-decimal proposal if one exists.

### Key Questions for Discussion
*   Is `FixedDecimal` the right fit for `CurrencyFormatter`, or does it introduce confusing semantics (e.g., what happens if a user passes `1.2345 USD` when USD only supports 2 decimal places)?
*   How would a different input type integrate with the rest of the ICU4X number formatting pipeline, which is heavily built around `FixedDecimal`?
*   What are the API ergonomics and performance implications of introducing a new representation specifically for currency?

This issue is created to host the discussion for the long-term API design of `CurrencyFormatter` before it is stabilized.

*Context: This was originally raised by @robertbastian in [this PR review comment](https://github.com/unicode-org/icu4x/pull/8145#discussion_r3482839372).*

sffc:
[...]

PR #8150, opened 2026-06-29
feat/currency: Migrate LongCurrencyFormatter and add try_new_long

younies:
This PR migrates the `LongCurrencyFormatter` functionality into the main `CurrencyFormatter`, building on top of the refactoring in #8145.

- Added tests for long currency formatting, including numbering system overrides.
- Removed redundant `long_formatter.rs` and `long_format.rs` files.

Diff Base: #8145

## Changelog

icu_experimental/currency:

- Migrated `LongCurrencyFormatter` to `CurrencyFormatter`.
- Added `try_new_long` and `try_new_long_unstable` constructors.


younies:
@robertbastian : ping

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -274,24 +425,45 @@ impl CurrencyFormatter<Decimal> {
...
+                let pattern = pattern.unwrap_or_else(|| <&DoublePlaceholderPattern>::default());
+
+                self.decimal_formatter.format_sign(
+                    value.sign,
+                    pattern.interpolate((
+                        self.decimal_formatter
+                            .format_unsigned(icu_decimal::Cow::Borrowed(&value.absolute)),
+                        currency_str,
+                    )),
+                )
```
robertbastian: this is the same across both branches, just make the `match` return `(pattern, display_name)`

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -244,8 +264,139 @@ impl CurrencyFormatter<Decimal> {
...
+            .payload;
+
+        let patterns = provider.load(Default::default())?.payload;
+
+        let plural_rules = PluralRules::try_new_cardinal_unstable(provider, (&prefs).into())?;
+
+        Ok(Self {
+            // Width option is not used for Long formatting (which always uses long names),
+            // so we initialize with default options.
+            options: Default::default(),
```
robertbastian: move the width into the data enum then, to the `Essential` branch

robertbastian (commented):
looks good, just two nits

components/experimental/src/dimension/currency/formatter.rs
younies: Done. I've updated the PR to extract the common formatting logic, so the `match` now returns `(pattern, display_name)`.

components/experimental/src/dimension/currency/formatter.rs
younies: Done. I've moved the `width` into the `Essential` variant of `CurrencyFormatterData` and updated the constructors and formatting logic accordingly. 

Since `CompactCurrencyFormatter` is not yet migrated to specific constructors and still relies on `options.width`, I've kept `width` in `CurrencyFormatterOptions` for now, but `CurrencyFormatter` no longer uses it (it now has no `options` field in this PR, though it will be re-introduced in a subsequent PR to support the `usage` option).

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -48,6 +51,19 @@ pub struct Decimal;
 
 impl ValueRepresentation for Decimal {}
 
+#[derive(Debug)]
+enum CurrencyFormatterData {
+    Essential {
+        essential: DataPayload<CurrencyEssentialsV1>,
+        width: Width,
```
robertbastian, line 58: `bound_currency` should also be in here, as it's implicit in the payloads for `Long` (and unused in that branch).

I'd just call it `currency`

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -274,22 +416,36 @@ impl CurrencyFormatter<Decimal> {
...
-        let (currency_str, pattern, _pattern_selection) = self
-            .essential
-            .get()
-            .name_and_pattern(self.options.width, &self.bound_currency);
-
-        let pattern = pattern.unwrap_or_else(|| <&DoublePlaceholderPattern>::default());
+        let (pattern, display_name) = match &self.data {
+            CurrencyFormatterData::Essential { essential, width } => {
+                // TODO(#6064): Support plural-specific patterns and full currency formatting spec.
+                // TODO(#8146): Evaluate if FixedDecimal is the correct input type or if we should use
```
robertbastian: nit: this todo should not be in this branch, it applies to the whole function

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -274,22 +416,36 @@ impl CurrencyFormatter<Decimal> {
...
+            }
+        };
 
         self.decimal_formatter.format_sign(
             value.sign,
             pattern.interpolate((
                 self.decimal_formatter
                     .format_unsigned(icu_decimal::Cow::Borrowed(&value.absolute)),
-                currency_str,
+                display_name,
```
robertbastian: nit: needless diff, and it might not be a "display name", could be an ISO code

components/experimental/src/dimension/currency/formatter.rs
younies: Done! Moved currency into Essential and removed bound_currency from Long and struct CurrencyFormatter.

components/experimental/src/dimension/currency/formatter.rs
younies: Done! Moved TODO(#8146) out of the match branch to the top of format_fixed_decimal.

components/experimental/src/dimension/currency/formatter.rs
younies: Done! Renamed display_name to currency_str across both branches in format_fixed_decimal.

robertbastian (approved):

PR #8152, opened 2026-06-29
Use neo grapheme segmenter for neo complex segmentation

robertbastian:
So far, the neo segmenters have used the existing complex segmentation code, which uses a (non-neo) `GraphemeClusterBreakSegmenter` internally. This PR makes the whole complex code path generic in a GCBS, so that neo can use neo.

#7962 

## Changelog

N/A

### DIFF
--- a/components/segmenter/src/complex/dictionary.rs
+++ b/components/segmenter/src/complex/dictionary.rs
@@ -2,7 +2,7 @@
 // called LICENSE at the top level of the ICU4X source tree
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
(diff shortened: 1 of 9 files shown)


Manishearth (approved):
Why is ComplexPayloads (and thus AbstractGCSB) hidden-public?

robertbastian:
Because of these: https://github.com/unicode-org/icu4x/blob/20105543d08bcd024df1ba86688e9ee933fb1767/components/segmenter/src/rule_segmenter.rs#L36-L56

Manishearth:
Ahh. I wonder if we can make the concrete types fully private while the assoc types stay hidden.

PR #8154, opened 2026-06-29
Make ComplexPayloads private, not hidden

Manishearth:
Noticed in https://github.com/unicode-org/icu4x/pull/8152

ComplexPayloads is public-hidden because it's used in a public-hidden associated type. But we don't need it to be: the associated type can be public-hidden and this type can be private and it should work just fine.


Manishearth:
I think the way to do this is to use the public-type-in-private-module trick

Issue #8035, opened 2026-06-04
About the annotation in library crates


xfq:
https://github.com/unicode-org/icu4x/blob/main/documents/process/boilerplate.md#library-annotations

> If the crate has an std feature, specify this in the first line:

appears identical to the code snippet shown afterward, failing to clearly illustrate the condition. Is this an oversight or a deliberate design choice?

sffc:
It appears that it changed in https://github.com/unicode-org/icu4x/commit/263fdc2e572b8e0e107507d190d5290d6a2c6b0a#diff-608d8e7dbd13ab7d89b8525eadee573ada2f652cd62b4bc3c99265b1c1937666

CC @robertbastian

robertbastian:
Yeah that should be 

```
    #![cfg_attr(not(any(test, doc, feature = "std")), no_std)]
```

polarathene:
@robertbastian Could you provide context for the addition of the feature constraint?

https://github.com/unicode-org/icu4x/blob/51de0b8144429e2a56d5a21ff3567d5d14e98a81/ffi/capi/src/lib.rs#L5-L6

https://github.com/unicode-org/icu4x/blob/51de0b8144429e2a56d5a21ff3567d5d14e98a81/ffi/capi/src/lib.rs#L53-L55

Wouldn't it be better to take this approach?:

```rust
#![cfg_attr(not(any(test, doc)), no_std)]

#[cfg(feature = "std")]
extern crate std;
```

If you want to ensure that `core`/`alloc` is used instead when viable, wouldn't these lints for `clippy` be a more meaningful way to communicate the concern to anyone when they're encountered?:

```rust
// Warn if alloc/core types are imported through std.
#![warn(
    clippy::std_instead_of_core,
    clippy::std_instead_of_alloc,
    // Only relevant if you don't (always) need alloc.
    clippy::alloc_instead_of_core,
)]
```

---

That advice to use clippy is referenced from [this comment](https://www.reddit.com/r/rust/comments/1ncdev1/comment/nd8kqo5/).

The other advice to shift the feature gate is from [this discussion](https://www.reddit.com/r/rust/comments/1hs6spy/psa_for_std_feature_in_no_std_libraries/) (_notably for the consistency in prelude behaviour_):

> **Tl;dr:** don't use `#![cfg_attr(not(feature = "std"), no_std)]` to have an `std` feature, always use:
> 
> ```
> #![no_std]
> 
> #[cfg(feature = "std")]
[...]

PR #4123, opened 2023-10-06
Rename `ffi/diplomat` to `ffi/capi`

robertbastian:
The crate is called `icu_capi`, and it's confusing to have two crates in `diplomat` directories in my workspace.

PR #3802, opened 2023-08-07
Deprecating `cpp_default` and `wasm_default` cargo features

robertbastian:
With the disappearance of testdata, these become the same, `["logging"]`. Instead I made `logging` a default feature.

I also moved the WASM logging setup from `wasm_glue` into `ICU4XLogger`, to reduce wasm glue.

PR #8155, opened 2026-06-30
refactor: Adapt most crates to unconditionally use `#![no_std]`

polarathene:
This minor refactor across crates is [at the request](https://github.com/unicode-org/icu4x/issues/8035#issuecomment-4799490007) of @robertbastian to PR the relevant changes.

I've tried to make it a bit easier to review by scoping related changes across commits.

I've additionally revised the boilerplate docs, and shuffled some extern imports around. There were some minor changes I made that I'm assuming are valid but as I'm not an actual user of any of these crates I'm deferring that to test coverage 😅 (_I'll highlight these via review comments_)

Any crate with boilerplate for `no_std` commented out has been left that way as I assume there is implicit reliance upon the `std` prelude and I lack time to pursue updating those crates. I have documented that discrepancy at the end of the boilerplate docs, should anyone wish to tackle that going forward 💪

I did all these changes manually (_well where possible with the editor search find/replace functionality_), but did consult Gemini to assist as peer review when revising my wording in the boilerplate docs.

**NOTE:** I've not been able to follow the [contributor checklist](https://github.com/unicode-org/icu4x/blob/main/CONTRIBUTING.md#checklist). As I'm pressed for time and my current system is at 77/88GB of committed memory, I've pieced this PR together via the `github.dev` web editor (_now rebranded as `vscode.dev`_). It may not be perfect but at least outlines the scope of change.

### DIFF
--- a/documents/process/boilerplate.md
+++ b/documents/process/boilerplate.md
@@ -4,40 +4,54 @@ This file contains explanations of ICU4X-specific conventions on boilerplate (ex
 
 ## Library annotations
 
-In order to assert that library crates conform to the ICU4X style guide, the following annotation should be present in every *lib.rs* file. The annotations do the following:
+To ensure the ICU4X library crates conform to the project's style guide, every `lib.rs` file must include the standard boilerplate annotations.
 
-1. Set the crate to be `no_std`
-2. Enforce no unannotated panicky behavior, except in test mode
-3. Require every exported method to be documented, implement `Debug`, and be `non_exhaustive` or annotated as exhaustive
+These annotations enforce the following rules:
 
-If the crate has no `std` feature:
+1. Configure the crate as `no_std`.
+2. Enforce annotating any panicking behavior, except during tests.
+3. Require every exported item to be documented, implement `Debug`, and where appropriate marked as `non_exhaustive` (_unless annotated to permit as exhaustive_).
 
-    // https://github.com/unicode-org/icu4x/blob/main/documents/process/boilerplate.md#library-annotations
-    #![cfg_attr(not(any(test, doc)), no_std)]
-    #![cfg_attr(
-        not(test),
-        deny(
-            clippy::indexing_slicing,
-            clippy::unwrap_used,
-            clippy::expect_used,
-            clippy::panic,
-        )
-    )]
-    #![warn(missing_docs)]
+> [!NOTE]
+> While the majority of lints are configured globally via the workspace `Cargo.toml`, some [boilerplate-level overrides remain necessary][gh-issues::boilerplate-lints] due to upstream tooling limitations.
 
-Not all crates are yet able to be annotated in this way. Annotations may be omitted and added when able.
+Include the following at the top of each `lib.rs` file:
 
-If the crate has an `std` feature, specify this in the first line:
+```rust
+// https://github.com/unicode-org/icu4x/blob/main/documents/process/boilerplate.md#library-annotations
+#![no_std]
+#![cfg_attr(
+    not(test),
+    deny(
+        clippy::indexing_slicing,
+        clippy::unwrap_used,
+        clippy::expect_used,
+        clippy::panic,
+    )
+)]
+#![warn(missing_docs)]
 
-    // https://github.com/unicode-org/icu4x/blob/main/documents/process/boilerplate.md#library-annotations
-    #![cfg_attr(not(any(test, doc)), no_std)]
-    #![cfg_attr(
-        not(test),
-        deny(
-            clippy::indexing_slicing,
-            clippy::unwrap_used,
-            clippy::expect_used,
-            clippy::panic,
-        )
-    )]
-    #![warn(missing_docs)]
+#[cfg(test)]
+extern crate std;
+```
+
+By convention, `extern crate std;` should be placed immediately following the crate-level attributes. If the file also requires `extern crate alloc;`, place `extern crate std;` after the `alloc` declaration.
+
+Not all crates are compatible with the full boilerplate. Incompatible annotations should be included but commented out (_if that is `#![no_std]`, the associated `extern crate std` lines should be excluded from the file_).
+
+Mandating `#![no_std]` ensures that the [`core` prelude is always used instead][prelude-nostd] of the [`std` prelude][prelude-std]. This provides a consistent prelude across environments, ensuring that `std` is only explicitly referenced where required.
+
+> [!NOTE]
+> When a crate optionally supports `std`, it should be gated via a `std` feature and additionally enabled when generating documentation:
+> 
+> ```rust
+> #[cfg(any(test, doc, feature = "std"))]
+> extern crate std;
+> ```
+> 
+> - Some crates may **only support `std`**, in which case `extern crate std;` is enforced without any conditional constraints.
+> - Any crate missing this extern presently relies upon an implicit `std` prelude, but it's boilerplate should have `#![no_std]` commented out, until that crate has been migrated to explicitly import from `std`.
+
+[gh-issues::boilerplate-lints]: https://github.com/unicode-org/icu4x/issues/5974#issuecomment-3775926163
+[prelude-nostd]: https://doc.rust-lang.org/reference/names/preludes.html#the-no_std-attribute
+[prelude-std]: https://doc.rust-lang.org/std/prelude/index.html
--- a/ffi/capi/src/lib.rs
+++ b/ffi/capi/src/lib.rs
@@ -3,7 +3,7 @@
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
 // https://github.com/unicode-org/icu4x/blob/main/documents/process/boilerplate.md#library-annotations
-#![cfg_attr(not(any(test, feature = "std")), no_std)]
+#![no_std]
 #![cfg_attr(
     not(test),
     deny(
@@ -50,9 +50,8 @@
 //! semver version, as are the bindings in the <code>bindings</code> folder.
 //! </div>
 
-// Renamed so you can't accidentally use it
-#[cfg(target_arch = "wasm32")]
-extern crate std as rust_std;
+#[cfg(any(test, doc, feature = "std"))]
+extern crate std;
 
 #[cfg(all(not(feature = "std"), feature = "looping_panic_handler"))]
 #[panic_handler]
--- a/ffi/capi/src/lib.rs
+++ b/ffi/capi/src/lib.rs
@@ -50,18 +50,19 @@
 //! semver version, as are the bindings in the <code>bindings</code> folder.
 //! </div>
 
+extern crate alloc;
+
 #[cfg(any(test, doc, feature = "std"))]
 extern crate std;
+
+#[cfg(all(not(feature = "std"), feature = "libc_alloc"))]
+extern crate libc_alloc;
 
 #[cfg(all(not(feature = "std"), feature = "looping_panic_handler"))]
 #[panic_handler]
 fn panic(_info: &core::panic::PanicInfo) -> ! {
     loop {}
 }
-
-extern crate alloc;
-#[cfg(all(not(feature = "std"), feature = "libc_alloc"))]
-extern crate libc_alloc;
 
 #[cfg(feature = "datetime")]
 pub(crate) mod datetime_helpers;
--- a/ffi/freertos/src/lib.rs
+++ b/ffi/freertos/src/lib.rs
@@ -3,7 +3,7 @@
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
 // https://github.com/unicode-org/icu4x/blob/main/documents/process/boilerplate.md#library-annotations
-#![cfg_attr(not(any(test, doc)), no_std)]
+#![no_std]
 #![cfg_attr(
     not(test),
     deny(
@@ -32,6 +32,9 @@
     feature(alloc_error_handler)
 )]
 
+#[cfg(test)]
+extern crate std;
+
 // Necessary to for symbols to be linked in
 extern crate icu_capi;
 
@@ -59,9 +62,3 @@ mod stuff {
         loop {}
     }
 }
-
-// Needed for rust runtime stuff
-//
-// renamed so you can't accidentally use it
-#[cfg(not(target_os = "none"))]
-extern crate std as rust_std;
--- a/provider/core/src/lib.rs
+++ b/provider/core/src/lib.rs
@@ -3,7 +3,7 @@
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
 // https://github.com/unicode-org/icu4x/blob/main/documents/process/boilerplate.md#library-annotations
-#![cfg_attr(not(any(test, doc, feature = "std")), no_std)]
+#![no_std]
 #![cfg_attr(
     not(test),
     deny(
@@ -92,6 +92,9 @@
 
 #[cfg(feature = "alloc")]
 extern crate alloc;
+
+#[cfg(any(test, doc, feature = "std"))]
+extern crate std;
 
 #[cfg(feature = "baked")]
 pub mod baked;
@@ -192,10 +195,10 @@ pub use log;
 #[doc(hidden)] // internal
 #[cfg(all(
     not(feature = "logging"),
-    all(debug_assertions, feature = "alloc", not(target_os = "none"))
+    debug_assertions,
+    feature = "std",
 ))]
 pub mod log {
-    extern crate std;
     pub use std::eprintln as error;
     pub use std::eprintln as warn;
     pub use std::eprintln as info;
@@ -203,11 +206,12 @@ pub mod log {
     pub use std::eprintln as trace;
 }
 
+#[doc(hidden)] // internal
 #[cfg(all(
     not(feature = "logging"),
-    not(all(debug_assertions, feature = "alloc", not(target_os = "none"),))
+    not(debug_assertions),
+    not(feature = "std"),
 ))]
-#[doc(hidden)] // internal
 pub mod log {
     #[macro_export]
     macro_rules! _internal_noop_log {
--- a/components/calendar/src/lib.rs
+++ b/components/calendar/src/lib.rs
@@ -3,7 +3,7 @@
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
 // https://github.com/unicode-org/icu4x/blob/main/documents/process/boilerplate.md#library-annotations
-#![cfg_attr(not(any(test, doc)), no_std)]
+#![no_std]
 #![cfg_attr(
     not(test),
     deny(
@@ -91,6 +91,9 @@
 
 #[cfg(feature = "alloc")]
 extern crate alloc;
+
+#[cfg(test)]
+extern crate std;
 
 // Make sure inherent docs go first
 mod date;
--- a/components/casemap/src/lib.rs
+++ b/components/casemap/src/lib.rs
@@ -28,7 +28,7 @@
 //! [`ICU4X`]: ../icu/index.html
 
 // https://github.com/unicode-org/icu4x/blob/main/documents/process/boilerplate.md#library-annotations
-#![cfg_attr(not(any(test, doc)), no_std)]
+#![no_std]
 #![cfg_attr(
     not(test),
     deny(
@@ -43,6 +43,9 @@
 #![allow(confusable_idents, uncommon_codepoints)]
 
 extern crate alloc;
+
+#[cfg(test)]
+extern crate std;
 
 mod casemapper;
 mod closer;
--- a/components/collator/src/lib.rs
+++ b/components/collator/src/lib.rs
@@ -7,7 +7,7 @@
 // described in LICENSE.
 
 // https://github.com/unicode-org/icu4x/blob/main/documents/process/boilerplate.md#library-annotations
-#![cfg_attr(not(any(test, doc)), no_std)]
+#![no_std]
 #![cfg_attr(
     not(test),
     deny(
@@ -342,6 +342,9 @@
 //! [`CollatorOptions`]: options::CollatorOptions
 
 extern crate alloc;
+
+#[cfg(test)]
+extern crate std;
 
 mod comparison;
 #[cfg(doc)]
--- a/components/collections/codepointtrie_builder/src/lib.rs
+++ b/components/collections/codepointtrie_builder/src/lib.rs
@@ -3,7 +3,7 @@
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
 // https://github.com/unicode-org/icu4x/blob/main/documents/process/boilerplate.md#library-annotations
-// #![cfg_attr(not(any(test, doc)), no_std)]
+// #![no_std]
 // #![cfg_attr(
 //     not(test),
 //     deny(
--- a/components/collections/src/lib.rs
+++ b/components/collections/src/lib.rs
@@ -3,7 +3,7 @@
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
 // https://github.com/unicode-org/icu4x/blob/main/documents/process/boilerplate.md#library-annotations
-#![cfg_attr(not(any(test, doc)), no_std)]
+#![no_std]
 #![cfg_attr(
     not(test),
     deny(
@@ -35,6 +35,9 @@
 
 #[cfg(feature = "alloc")]
 extern crate alloc;
+
+#[cfg(test)]
+extern crate std;
 
 pub mod char16trie;
 pub mod codepointinvlist;
--- a/components/datetime/src/lib.rs
+++ b/components/datetime/src/lib.rs
@@ -3,7 +3,7 @@
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
 // https://github.com/unicode-org/icu4x/blob/main/documents/process/boilerplate.md#library-annotations
-#![cfg_attr(not(any(test, doc)), no_std)]
+#![no_std]
 #![cfg_attr(
     not(test),
     deny(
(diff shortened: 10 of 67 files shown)


ffi/capi/src/lib.rs
```diff
@@ -50,20 +50,20 @@
 //! semver version, as are the bindings in the <code>bindings</code> folder.
 //! </div>
 
-// Renamed so you can't accidentally use it
-#[cfg(target_arch = "wasm32")]
-extern crate std as rust_std;
+extern crate alloc;
+
+#[cfg(any(test, doc, feature = "std"))]
+extern crate std;
```
polarathene, line 56: I don't see a need for the `target_arch = "wasm32"` constraint for `std`, or the need to rename it.

With `#![no_std]` always being present now, there will be no `std` prelude implicitly available, which I assume this workaround was for to ensure the use of the `core` types perhaps? Whereas now an explicit `std` would be required otherwise. I didn't see any references for `rust_std` so I assume it is no longer used.

This `rust_std` rename was [introduced in Aug 2021](https://github.com/unicode-org/icu4x/pull/941) (_when the `capi` crate declared `#![no_std]` unconditonally_). Later in [Sep 2022 with Rust 1.64](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0/#c-compatible-ffi-types-in-core-and-alloc), `core` gained the FFI type aliases (satisfying `c_char`) and `alloc` gained `CString`. However by this time it seems the `wasm_glue.rs` file had already dropped these lines.

**Reference:** File history is a bit messy to track for `wasm_glue.rs`, but here's the relevant outline proceeding that Aug 2021 PR:
- Sep 2021: [`ffi/capi/src/wasm_glue.rs` renamed to `ffi/diplomat/src/wasm_glue.rs`](https://github.com/unicode-org/icu4x/commit/007b532720771ef334a02d8c4704db6f689b8493#diff-44aba2ef2d53a5bc99d24e93cf106ef59d9edcbc5c72e9dd815ef8fabfe82cc6)
- July 2022: [All `rust_std` lines dropped](https://github.com/unicode-org/icu4x/pull/2248) from `ffi/diplomat/src/wasm_glue.rs`.
- Aug 2023: [Further slimming as the target becomes feature gated](https://github.com/unicode-org/icu4x/commit/5cc4ad9780f120cff275febe55bd3c308fecd094#diff-44aba2ef2d53a5bc99d24e93cf106ef59d9edcbc5c72e9dd815ef8fabfe82cc6) and code is shifted ([PR](https://github.com/unicode-org/icu4x/pull/3802)).
- Oct 2023: [Renamed again in Oct 2023](https://github.com/unicode-org/icu4x/pull/4123/changes#diff-9ce580a5e1080ba6c0ffb9d639179dd1af042cf8c4df0a938f59c3ede3622d3c) (_`ffi/diplomat/src/wasm_glue.rs` => `ffi/capi/src/wasm_glue.rs`_). The file was then shortly after [removed in a PR bumping the `diplomat` dep](https://github.com/unicode-org/icu4x/pull/4196/changes#diff-9ce580a5e1080ba6c0ffb9d639179dd1af042cf8c4df0a938f59c3ede3622d3c). Using `diplomat-runtime` for WASM logging.

ffi/freertos/src/lib.rs
```diff
@@ -59,9 +62,3 @@ mod stuff {
         loop {}
     }
 }
-
-// Needed for rust runtime stuff
-//
-// renamed so you can't accidentally use it
-#[cfg(not(target_os = "none"))]
-extern crate std as rust_std;
```
polarathene, line 67: This is similar to the WASM workaround with `rust_std` and other crates that had adopted that convention previously.

This file was introduced via [this March 2022 PR](https://github.com/unicode-org/icu4x/pull/1747) where it [extracted out `diplomat/src/freertos_glue.rs` to `ffi/freetos/src/lib.rs`](https://github.com/unicode-org/icu4x/pull/1747/changes#diff-c3483707e0d09b1bfe9bc8bd799f5dfb211e38b440011766cd2a25b0a9883553) and the `extern crate std as rust_std` line with constraint was added in a [follow-up April 2022 PR](https://github.com/unicode-org/icu4x/pull/1818). It's ambiguous what was considered "rust runtime stuff" and no context was given or discussed in the PR.

Like with the other review comment regarding WASM, this workaround should not be necessary, as `std` won't implicitly be available with explicit unconditional `#![no_std]`. It is now only in the scope of of `cfg(test)` where it's presumably needed. Should there ever be a need for `std` outside of that it can be feature gated as expected 👍

provider/core/src/lib.rs
```diff
@@ -192,22 +195,23 @@ pub use log;
 #[doc(hidden)] // internal
 #[cfg(all(
     not(feature = "logging"),
-    all(debug_assertions, feature = "alloc", not(target_os = "none"))
+    debug_assertions,
+    feature = "std",
```
polarathene, line 199: I might be making a mistake here, presumably with the `--all-features` flag for CI testing? (_the constraint to prevent `target_os = "none"` even with the `std` feature gate is presumably important for testing only?_)

Given the use of `std`, the reliance upon `alloc` didn't seem correct here, so I swapped that for the `std` feature that is already in use. I've not looked through `git blame` history to understand the motive of why the `alloc` feature was used instead but I assume the `std` feature arrived later?

Since the conditions were already wrapped in an `all()` predicate, I split them onto separate lines, there wasn't any meaningful grouping I could infer to keep the nested `all()`.

provider/core/src/lib.rs
```diff
@@ -192,22 +195,23 @@ pub use log;
...
     pub use std::eprintln as debug;
     pub use std::eprintln as trace;
 }
 
+#[doc(hidden)] // internal
 #[cfg(all(
     not(feature = "logging"),
-    not(all(debug_assertions, feature = "alloc", not(target_os = "none"),))
+    not(debug_assertions),
+    not(feature = "std"),
```
polarathene, line 213: Likewise, the nested `all()` didn't seem to be of value and added friction to groking the difference, instead I just mirrored the `std` variant conditionals to be wrapped in `not()`.

That prior `all(not(all(not(target_os = "none"))))` may still be relevant, but could be better expressed on a separate line as just `target_os = "none",` if it is needed for tests?

documents/process/boilerplate.md
```diff
@@ -4,40 +4,54 @@ This file contains explanations of ICU4X-specific conventions on boilerplate (ex
...
-        )
-    )]
-    #![warn(missing_docs)]
+To ensure the ICU4X library crates conform to the project's style guide, every `lib.rs` file must include the standard boilerplate annotations.
+
+These annotations enforce the following rules:
+
+1. Configure the crate as `no_std`.
+2. Enforce annotating any panicking behavior, except during tests.
+3. Require every exported item to be documented, implement `Debug`, and where appropriate marked as `non_exhaustive` (_unless annotated to permit as exhaustive_).
```
polarathene, line 13: This point seems a bit outdated. Perhaps the boilerplate had relevant lints in the past, but now there is a mismatch as some of these are delegated to the workspace `Cargo.toml`.

I didn't drop these descriptions to defer to review feedback. In the meantime I added a note below to add context to the reader that other lints are configured at the workspace and these are purely workaround lints, with an issue link to associate them to.

documents/process/boilerplate.md
```diff
@@ -4,40 +4,54 @@ This file contains explanations of ICU4X-specific conventions on boilerplate (ex
...
+        clippy::panic,
+    )
+)]
+#![warn(missing_docs)]
+
+#[cfg(test)]
+extern crate std;
+```
+
+By convention, `extern crate std;` should be placed immediately following the crate-level attributes. If the file also requires `extern crate alloc;`, place `extern crate std;` after the `alloc` declaration.
```
polarathene, line 38: Since the change proposed by this PR does introduce some distance from the `extern` line in some files, I figured this was worth documenting.

I've adjusted source in crates to elevate any `extern crate alloc` to the top, and where `extern crate std` is present it made sense to have it occur after. If this extra guidance seems redundant feel free to drop it 👍

documents/process/boilerplate.md
```diff
@@ -4,40 +4,54 @@ This file contains explanations of ICU4X-specific conventions on boilerplate (ex
...
+
+Not all crates are compatible with the full boilerplate. Incompatible annotations should be included but commented out (_if that is `#![no_std]`, the associated `extern crate std` lines should be excluded from the file_).
```
polarathene, line 40: This is revised from the original line on the same topic. I'm not sure if you'd want a `extern crate std` added when the crate isn't adjusted to use or require it for anything else.

Keeping a `// #![no_std]` line at least makes it easy to filter a search query for until such crates are adapted to not rely on an implicit `std` prelude.

documents/process/boilerplate.md
```diff
@@ -4,40 +4,54 @@ This file contains explanations of ICU4X-specific conventions on boilerplate (ex
...
+
+Mandating `#![no_std]` ensures that the [`core` prelude is always used instead][prelude-nostd] of the [`std` prelude][prelude-std]. This provides a consistent prelude across environments, ensuring that `std` is only explicitly referenced where required.
```
polarathene, line 42: This is just documentation for better awareness of why it's done this way, as it seems in the past there's been some switching between the usage and workarounds without the understanding at the time about issues experienced.

I provided more detail about why the change proposed by this PR is the better approach over in [this issue comment](https://github.com/unicode-org/icu4x/issues/8035#issuecomment-4776767152) of mine.

documents/process/boilerplate.md
```diff
@@ -4,40 +4,54 @@ This file contains explanations of ICU4X-specific conventions on boilerplate (ex
...
+> [!NOTE]
+> When a crate optionally supports `std`, it should be gated via a `std` feature and additionally enabled when generating documentation:
+> 
+> ```rust
+> #[cfg(any(test, doc, feature = "std"))]
+> extern crate std;
+> ```
+> 
+> - Some crates may **only support `std`**, in which case `extern crate std;` is enforced without any conditional constraints.
+> - Any crate missing this extern presently relies upon an implicit `std` prelude, but it's boilerplate should have `#![no_std]` commented out, until that crate has been migrated to explicitly import from `std`.
```
polarathene, line 53: I'm not entirely confident about this advice, but my understanding is that for tests not in a `tests/` directory, you may need to ensure `std` is present, so that is always toggling it. While `doc` may only be relevant if the crate actually uses `std` itself?

**NOTE:** You have a crate (`provider/blob/src/lib.rs`) that uses `export` as the feature name instead of `std`. I've not mentioned that here, and I'm not sure if there's any desire/intent to adjust that feature gate for the crate 🤷‍♂️ 

The later bullet points in this admonition is for the current crates incompatible with running without `std`, while none are presently configured to use `#![no_std]` they could presumably add `extern crate std` without any annotation and be fixed up to explicitly reference `std` where necessary. Presently all those crates have `// #![no_std]` but once they've been adapted, there might be some misunderstanding at a glance about their ability to support `no_std`? 🤷‍♂️ (_or maybe it's not a concern to worry about_)

Since the line may be further apart from the boilerplate snippet and have no annotation to communicate the `std`-only support, someone may likewise think it's redundant without a contextual comment, but reviewers and CI would likely address such misunderstandings. Since the `std` prelude won't be available, I don't think they'll be any real issues that trip up any contributor besides mandatory imports from `std`.

polarathene (commented):
The bulk of the PR is effectively:

```diff
-#![cfg_attr(not(any(test, doc)), no_std)]
+#![no_std]

+#[cfg(test)]
+extern crate std;
```

or:

```diff
-// #![cfg_attr(not(any(test, doc)), no_std)]
+// #![no_std]
```

With the removal of `extern crate std` from locations where it was added at the crate root instead. And additionally moving `extern crate alloc` to be declared earlier in a file as a convention.

Any other difference has an attached comment for context 👍

ffi/capi/src/lib.rs
Manishearth: Yeah, I think this makes sense

provider/core/src/lib.rs
Manishearth: No, the `alloc` feature arrived later. Not sure what happened here.

Manishearth: and the `target_os = none` is loadbearing I think. We ship to freertos.

I don't recall the details here though. It's certainly strange that we're exposing items from `std` like this

documents/process/boilerplate.md
Manishearth: > I'm not sure if you'd want a extern crate std added when the crate isn't adjusted to use or require it for anything else

Yes, please don't add `extern crate std` where not needed.

polarathene:
Oh awesome... I'll go over the Clippy lint failures tomorrow I guess. I'm not sure if I'll be able to do a clone locally to run and resolve lints any time soon, I really don't want to risk exceeding the memory commit ceiling on my system and triggering an OOM crash elsewhere right now 😓 

If anyone would like to take over this PR they're welcome to.

documents/process/boilerplate.md
Manishearth: > While doc may only be relevant if the crate actually uses std itself?

`doc` is needed for intra-doc links. But also I'm not entirely sure what you're asking in this paragraph, can you be a bit more explicit about your antecedents?

> NOTE: You have a crate (provider/blob/src/lib.rs) that uses export as the feature name instead of std. I've not mentioned that here, and I'm not sure if there's any desire/intent to adjust that feature gate for the crate 🤷‍♂️


Please do not change public feature sets. This is intentional; `blob`'s export functionality is not needed by normal ICU4X users.

In general, we have ICU4X runtime code which is expected to be no_std, and "datagen", which has no such expectation. Datagen crates don't have `no_std`.

Manishearth:
@polarathene I would recommend splitting this into smaller sets of changes: one with the noncontroversial cleanups (where crates that do follow the boilerplate are updated to the better boilerplate), and one with the changes that are likely to actually affect CI.

Manishearth:
Thanks for this PR, though! Working towards more consistency here is good.

provider/core/src/lib.rs
Manishearth: FWIW I would recommend not changing cfg gates like this in this PR, or at least not in the same commit. It is much harder to review if I have to go looking at each line to ensure I have noticed these cases.

This is currently a behavior change since it removes the `target_os` thing. This type of PR should not change public API behavior, even if it is unintentional.

documents/process/boilerplate.md
polarathene: > `doc` is needed for intra-doc links. But also I'm not entirely sure what you're asking in this paragraph, can you be a bit more explicit about your antecedents?

Presently you have crates configured with:

```rust
#![cfg_attr(not(any(test, doc)), no_std)]
```

So if building docs, `no_std` is not used and `std` as an implicit prelude is used. My understanding is this was only meaningful if the crate had something related to `std` to document? (_as in the crate actually uses `std` beyond tests_)

As such any crate that had no `std` or equivalent feature to opt-in to `std` presumably generates docs fine with `#![no_std]`, without needing annotation to enable `std` during docs generation?:

```rust
#[cfg(doc)]
extern crate std;
```

I could add that constraint for all uses regardless, like I had done with `cfg(test)` which from what I understand is necessary for `cargo test` to work (_not verified but I've seen various resources online have a problem with it for unit tests but not integration tests_), it just was not clear for `cfg(doc)`.

---

> Please do not change public feature sets. This is intentional; `blob`'s export functionality is not needed by normal ICU4X users.

I have not done that AFAIK.

> In general, we have ICU4X runtime code which is expected to be no_std, and "datagen", which has no such expectation. Datagen crates don't have `no_std`.

I specifically only targeted crates with the boilerplate present.

documents/process/boilerplate.md
polarathene: > Yes, please don't add `extern crate std` where not needed.

Just to clarify, the intention was on files that already have `// #![no_std]` (_or the equivalent prior boilerplate definition_), if you uncomment that line you would need `extern crate std` to be present and likely make adjustments.

The only motivation for such would be for consistency across crates, which conventionally for someone working across them might be saner. Some types that exist in both `core`/`alloc` and `std` however do have additional methods/support in their `std` variant, being aware of those differences can be useful in `std` only crates where those features could be leveraged.

Either way, I've not introduced changes like that to those crates functionality.

documents/process/boilerplate.md
Manishearth: > So if building docs, no_std is not used and std as an implicit prelude is used. My understanding is this was only meaningful if the crate had something related to std to document? (as in the crate actually uses std beyond tests)


Yes, this means that an intra doc link, like `[std::rc::Rc]`, will work in the markdown docs.

I think it's fine to unconditionally enable std in doc mode.

I'd be careful here, because broken intra-doc links do not always cause CI errors. They certainly won't cause CI errors in anythng but a `rustdoc` build, too.

> I have not done that AFAIK.



I know, I wanted to ward you off considering that as an option :smile: 

> I specifically only targeted crates with the boilerplate present.


Good to know, thanks!

documents/process/boilerplate.md
Manishearth: This is a good idea

documents/process/boilerplate.md
Manishearth: Yeah we should clean this up. Probably worth doing separately.

documents/process/boilerplate.md
```diff
@@ -4,40 +4,54 @@ This file contains explanations of ICU4X-specific conventions on boilerplate (ex
...
+    deny(
+        clippy::indexing_slicing,
+        clippy::unwrap_used,
+        clippy::expect_used,
+        clippy::panic,
+    )
+)]
+#![warn(missing_docs)]
+
+#[cfg(test)]
```
robertbastian, line 34: this used to be `any(test, doc)` because we did need this somewhere

ffi/capi/src/lib.rs
robertbastian: I believe what we had previously was equivalent to `not(target_arch = "wasm32")` – we _didn't_ want to use std on wasm32. IIRC it's a weird target as it has a partially implemented std, so we renamed it to not accidentally use it, even if the std feature is enabled.

sffc:
The main thing to _avoid_ is: a no_std crate has `extern crate std;` and then someone checks in code using an std-specific thing that looks benign, something like `io::Write`. To avoid this, we only enable `std` in test mode and when the feature is enabled. Relying on human or AI review is not sufficient.

Manishearth:
> To avoid this, we only enable std in test mode and when the feature is enabled.

We have a nostd CI job that is expected to fail if you accidentally enable std anywhere in the "core" ICU4X tree.

provider/core/src/lib.rs
polarathene: > It is much harder to review if I have to go looking at each line to ensure I have noticed these cases.

You do not need to go looking? I explicitly pointed all such changes out via these earlier review comments for you to help reduce such a review burden?...

> or at least not in the same commit.

They're all staged in the [same commit](https://github.com/unicode-org/icu4x/pull/8155/changes/6109f1275575f44d403d873954eeb0e050690d34) for these changes. The other two commits are focused on one overall changeset as well.

I suppose I could split that linked commit up as the notable changes like this one are only 4 of files in the commit:

- **`ffi/capi/src/lib.rs‎`:** (_[associated review comment](https://github.com/unicode-org/icu4x/pull/8155#discussion_r3496174837) provides change history to establish confidence_)

  ```diff
  -#![cfg_attr(not(any(test, feature = "std")), no_std)]
  +#![no_std]

  -// Renamed so you can't accidentally use it
  -#[cfg(target_arch = "wasm32")]
  -extern crate std as rust_std;
  +#[cfg(any(test, feature = "std"))]
  +extern crate std;
  ```
  
- **`ffi/freertos/src/lib.rs‎`:** (_[associated review comment](https://github.com/unicode-org/icu4x/pull/8155#discussion_r3496227455) provides context to establish confidence_)

  ```diff
  -#![cfg_attr(not(any(test, doc)), no_std)]
  +#![no_std]

  -// Needed for rust runtime stuff
  -//
  -// renamed so you can't accidentally use it
  -#[cfg(not(target_os = "none"))]
  -extern crate std as rust_std;

  +#[cfg(test)]
  +extern crate std;
  ```

- **`‎provider/blob/src/lib.rs`:** (_and a few others with a similar feature gated `no_std` toggle_)

  ```diff
  -#![cfg_attr(not(any(test, doc, feature = "export")), no_std)]
  +#![no_std]

  +#[cfg(any(test, feature = "export"))]
  +extern crate std;
  ```

Other than these, the other files in that linked commit dropped no longer relevant `extern crate std` declarations, which seemed relevant to the commit scope as that was associated to the highlighted changes above.

I do understand the issue with noisy diffs, if it wasn't as much of a hassle to manage with the web UI I'd have followed up with an interactive rebase.

**EDIT:** You were right however to raise a concern with this specific cfg change to `provider/core/src/lib.rs`, where it turned out avoiding a `std` feature gate was intentional (_unclear if still relevant today, presumably it is_). For context another [associated review comment I left](https://github.com/unicode-org/icu4x/pull/8155#discussion_r3496247351) did raise a concern about being cautious with the change.

---

> This is currently a behavior change since it removes the `target_os` thing. This type of PR should not change public API behavior, even if it is unintentional.

There was related changes similar to the one discussed here (as noted above), although I had provided detailed history (_`std` rename for [`wasm32`](https://github.com/unicode-org/icu4x/pull/8155#discussion_r3496174837) + [`freertos`](https://github.com/unicode-org/icu4x/pull/8155#discussion_r3496227455)_) to instill confidence. ~~I have not taken the time to dig through the history of this particular decision with `target_os`~~ (**EDIT:** [I have now](https://github.com/unicode-org/icu4x/issues/8178#issuecomment-4888404864)), so I understand your concern.

From what I can infer this change won't break any real users, but may have been a workaround for CI tests? (_**EDIT:** As per update this seems to be accurate_)

These two specific variations for when the `icu_provider/logging` feature is **not enabled** are both marked with hidden docs (_with an "internal" comment implying they're **not for public API use** outside of this project workspace crates?_).
- **EDIT:** [Confirmed via style guide document](https://github.com/unicode-org/icu4x/blob/main/documents/process/style_guide.md#implement-copy-on-error-types-and-log-details--suggested)?

---

**UPDATE:** I've [looked into this further](https://github.com/unicode-org/icu4x/issues/8178#issuecomment-4888404864):

> The use of a `not(target_os = "none")` cfg condition here was a workaround so that component crates do not need to have a forwarding `std` feature only to support toggling this `eprintln!` output (_rather than the stubbed fallback of no-op logging_).
> - Constraints for `log` module variant `eprintln!` appear to have been chosen for these reasons:
>   - **`debug_assertions`:** Debug only builds (_relevant for CI `cargo` subcommands like `check`/`test`/etc_)
>   - **`not(target_os = "none")`:** No OS equates to no `std` support.
>     - `target_os = "unknown"` targets like `wasm32-unknown-unknown` have a partial `std` implementation, even if by default it's `eprintln!` isn't wired to anything and thus a no-op.
>     - The only other target with an `unknown` OS appears to be `x86_64-fortanix-unknown-sgx`.
>   - **`alloc = "feature"`:** Prevents no-alloc crates (_`alloc` compatible will bring in `std` unintentionally but permitted as only for debug builds_)
> - Incompatible build targets (_`std` crate is unavailable but using the `alloc` feature in debug builds_) **will fail** builds.
> 
>   The selective gates avoid this in CI (_unclear if intentionally covered_), for example the crate [`tools/noalloctest`](https://github.com/unicode-org/icu4x/tree/main/tools/noalloctest) (_used in CI with `--target x86_64-unknown-linux-gnu`_) would fail if it ever disabled the [`icu_provider/logging` feature](https://github.com/unicode-org/icu4x/blob/d0e721bf145e38953df94d2d22098d6e9f30c423/tools/noalloctest/Cargo.toml#L27) and (in)directly enabled the `icu_provider/alloc` feature (_**NOTE:** The `icu_freertos` package does so implicitly via transitive deps, but avoids the `eprintln!` variant due to `none` OS_).

So `feature = "std"` would appear to go against the intent of subsequent changes in the past. Some inline commentary for context would better ensure that intention to not feature gate on `std` is more easily visible to the reader 😓

provider/core/src/lib.rs
polarathene: > Not sure what happened here.

[I looked into it](https://github.com/unicode-org/icu4x/issues/8178#issuecomment-4888404864)
- There previously was a `std` feature gate but it was dropped in favor `not(target_os = "none")` of not requiring component crates to not require a `std` feature from what I could gather.
- Later to better support `no-alloc` builds, the `feature = "alloc"` gate was also added, but this is a bit misleading to read for the `cfg` as it's not intended to be meaningful for `eprintln!` log variant having `extern crate std`, but rather the absence of the feature for ensuring the no-op stub fallback is matched instead as when this condition is `false`, the whole `all(...)` predicate for `eprintln!` fails (_which in turn ensures it succeeds for the no-op variant_).

> the `target_os = none` is loadbearing I think. We ship to freertos.

That remains relevant alongside `feature = "alloc"`.

For the same intent of ensuring any `false` condition resolves to the no-op variant instead of `eprintln!`, since a `no_std` + `alloc` build target lacks `std` with `target_os = "none"` 👍 

---

> I don't recall the details here though. It's certainly strange that we're exposing items from `std` like this

Hope the above info provides clarity. It'd be good to add a contextual comment for maintainers to not lose track of the decisions behind this, since it's not immediately apparent and [digging through history for answers](https://github.com/unicode-org/icu4x/issues/8178#issuecomment-4888404864) each time is not ideal.

provider/core/src/lib.rs
Manishearth: > You do not need to go looking? I explicitly pointed all such changes out via these earlier review comments for you to help reduce such a review burden?...


Right, but it's not clear to me if you got them all (even if you think you did), and it's a lot of work to verify. Big PRs are an easy place to lose yourself when reviewing.

PLUS every time the PR is changed I have to page all of this context back.

This PR still doesn't pass CI, and some of it is due to there being actual behavior changes in it. It will be easier for both you as a PR author trying to get CI to pass and us as reviewers if you first land the mechanical refactor, and then try to identify places that could be improved and make those as separate PRs. A lot of these are subtly deliberate, some of them aren't, and most of that history isn't documented. The "big refactor" should not need to be bogged down with things that need discussion.

PR #8163, opened 2026-07-01
Graduate DataPayloadOr

sffc:
We already use it in icu_datetime, and I want to use it in icu_locale_names. The design is good enough; I would like to graduate this. It's hard for us to change it anyway because it isn't behind an unstable feature.

The only real downside is that it's not obvious when to use this type instead of `Option<DataPayload>`. It exists to reduce stack sizes, full stop. It deviates from Rust style, but has convenient constructors and methods that make it feel like an `Option<DataPayload>`.

## Changelog

icu_provider: Stabilize type DataPayloadOr, added as draft in 1.5

### DIFF
--- a/provider/core/src/lib.rs
+++ b/provider/core/src/lib.rs
@@ -123,7 +123,6 @@ mod request;
 pub use request::{DataLocale, DataMarkerAttributes, DataRequest, DataRequestMetadata, *};
 
 mod response;
-#[doc(hidden)] // TODO(#4467): establish this as an internal API
 pub use response::DataPayloadOr;
 pub use response::{Cart, DataPayload, DataResponse, DataResponseMetadata};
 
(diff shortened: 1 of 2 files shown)


Manishearth (approved):
I think this is fine, the API is pretty straightforward.

sffc:
https://icu4x-pr-artifacts.storage.googleapis.com/gha/275bee8edbf527f7b8e7e1080da47c62edc1fa9e/rustdoc/icu_provider/struct.DataPayloadOr.html

@robertbastian Docs explaining why this type exists and what it does

robertbastian (approved):
A comment on the docs:

Please don't use the term "word". Rust has a concrete type for what you mean: `usize`. Word is a bad term because it sometimes *doesn't* mean the architecture-specific bit-width, but a fixed size data type (and it's also a linguistic term).

Manishearth (approved):

PR #8168, opened 2026-07-02
litemap: return concrete iterator types from iter/iter_mut/keys/values

### LINKED ISSUE #8054: litemap: Consider returning concrete types on `LiteMap::{iter,iter_mut,keys,values}`

Chaganti-Reddy:
Makes LiteMap::iter/iter_mut/keys/values return concrete types (matching the associated IntoIterator::IntoIter) instead of impl Trait, so e.g. iter() is now Clone.

  Closes #8054

### DIFF
--- a/utils/litemap/src/map.rs
+++ b/utils/litemap/src/map.rs
@@ -896,35 +896,51 @@ where
     }
 }
 
+fn key_of_pair<'a, K, V>(pair: (&'a K, &'a V)) -> &'a K {
+    pair.0
+}
+
+fn value_of_pair<'a, K, V>(pair: (&'a K, &'a V)) -> &'a V {
+    pair.1
+}
+
+/// The iterator type returned by [`LiteMap::keys`] and [`LiteMap::iter_keys`].
+pub type LiteMapKeysIter<'a, K, V, S> =
+    core::iter::Map<<S as StoreIterable<'a, K, V>>::KeyValueIter, fn((&'a K, &'a V)) -> &'a K>;
+
+/// The iterator type returned by [`LiteMap::values`] and [`LiteMap::iter_values`].
+pub type LiteMapValuesIter<'a, K, V, S> =
+    core::iter::Map<<S as StoreIterable<'a, K, V>>::KeyValueIter, fn((&'a K, &'a V)) -> &'a V>;
+
 impl<'a, K: 'a, V: 'a, S> LiteMap<K, V, S>
 where
     S: StoreIterable<'a, K, V>,
 {
     /// Produce an ordered iterator over key-value pairs
-    pub fn iter(&'a self) -> impl DoubleEndedIterator<Item = (&'a K, &'a V)> {
+    pub fn iter(&'a self) -> S::KeyValueIter {
         self.values.lm_iter()
     }
 
     /// Produce an ordered iterator over keys
     #[deprecated = "use keys() instead"]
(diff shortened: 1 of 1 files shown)


Manishearth:
Note: there's already an open PR in the process of review that is implementing this: https://github.com/unicode-org/icu4x/pull/8072

Chaganti-Reddy:
Ah, missed that #8072 already covers this more thoroughly (custom ExactSizeIterator/FusedIterator impls) — closing in favor of that one. Thanks for the heads up

PR #8169, opened 2026-07-02
feat(currency): Resolve and apply locale-specific currency precision

### LINKED ISSUE #8151: [currency] CurrencyFormatter should use currency patterns instead of decimal patterns
### Description
Currently, `CurrencyFormatter` (in `icu_experimental::dimension::currency`) relies on `DecimalFormatter` and its associated decimal patterns (e.g., `#,##0.###` for `en`) to format the numeric part of the currency.

This leads to incorrect formatting for currencies:
- It does not apply currency-specific decimal padding (e.g., formatting `123` as `$123` instead of `$123.00` for USD).
- It does not apply currency-specific rounding/precision (e.g., formatting `123.4567` as `$123.4567` instead of `$123.46`).

### Proposed Solutions
**Option 1** `CurrencyFormatter` can depend on currency format patterns directly (e.g., `¤#,##0.00`) instead of reusing `DecimalFormatter` with decimal patterns. This will allow using the currency's default fraction digits and applying the appropriate padding and rounding.

**Option 2** If the `DecimalFormatter` and currency formatter number patterns are slightly different (for example, if only the rounding is different), should we just use the `DecimalFormatter` and adapt it to currency?

### Context
Temporary test cases asserting the current (incorrect) behavior have been added to `components/experimental/src/dimension/currency/format.rs` with `TODO`s.

> [!IMPORTANT]
> **Communication with CLDR**: We need to communicate with the CLDR team to ensure that the number patterns are the same for each locale across specific formatting modes (e.g., the currency representation part for `standard` should be the same as `standard-next-to-number`, `accounting`, and `accounting-negative`). If there are any differences, we must support them in both the formatting code and the data generator.

younies:
Implement locale- and currency-sensitive fraction precision resolution and rounding uniformly across standard and compact currency formatters ✨

Fixes #8151 🎯

TAG=agy
CONV=f93d26e4-1f87-41f6-88d1-9069dadd73ee

## Key Highlights & Architectural Improvements 🚀

- **Uniform Precision Handling** 🪙: Currency fraction precision and rounding are applied uniformly across all decimal formatters in `CurrencyFormatter::format_fixed_decimal`.
- **Constructor-Time Fraction Resolution** ⚡: Resolved review feedback by moving fraction lookup (`resolve_fraction_info`) directly into `try_new_*` constructors. The resulting immutable `FractionInfo` is cached on `CurrencyFormatter`, eliminating runtime data provider table lookups and enabling precision adjustment completely outside formatting match arms.
- **Test Suite Support** 🧪: Added comprehensive unit test coverage verifying TR35 Section 3.8 Rule 3 uncompacted transitions (positive and negative values below 1,000) and fraction resolution behaviors. All testing adheres strictly to `--all-features`.

## Changelog

icu_experimental/currency:

- Apply fraction precision and rounding uniformly across all currency formatters.
- Optimize runtime performance by resolving and caching `FractionInfo` during formatter construction.


components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -438,21 +470,57 @@ impl CurrencyFormatter<Decimal> {
...
+                    .map(|ext| ext.get().display_names.get(operands, plural_rules))
+                    .unwrap_or(self.bound_currency.0.as_str());
                 let pattern = patterns.get().patterns.get(operands, plural_rules);
 
                 self.decimal_formatter.format_sign(
                     value.sign,
                     pattern.interpolate((
                         self.decimal_formatter
-                            .format_unsigned(icu_decimal::Cow::Borrowed(&value.absolute)),
+                            .format_unsigned(icu_decimal::Cow::Owned(value.absolute)),
```
younies: why `Owned` ?

younies: We need `Cow::Owned` here because `value` is a local variable. 

In `format_fixed_decimal`, we apply precision which clones and modifies the input, creating a new local `FixedDecimal` (`value`). If we used `Cow::Borrowed(&value.absolute)`, the returned `FormattedDecimal` would borrow from this local variable, which would result in a lifetime error since the local variable is dropped when the function returns. 

By using `Cow::Owned(value.absolute)`, we move the owned data into the `Cow` so it can be safely returned. Since `UnsignedDecimal` uses inline storage for typical digit counts, this `Cow::Owned` usually does not trigger heap allocations.

An alternative to avoid `Cow::Owned` would be to define a custom `FormattedCurrency` struct that owns the modified `FixedDecimal` and implements `Writeable`, but that would add more complexity for now.

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -370,13 +394,16 @@ impl CurrencyFormatter<Decimal> {
                     &locale,
                 ),
                 ..Default::default()
-            })?
-            .payload;
+            })
+            .allow_identifier_not_found()?
+            .map(|res| res.payload);
```
younies, line 326: why do we allow this ?

younies: We allow `IdentifierNotFound` here to support fallback to the ISO code when the localized display name for a currency is missing.

According to [UTS #35 (LDML Part 3: Numbers) Section 3.11.1](https://www.unicode.org/reports/tr35/tr35-numbers.html#Plural_Rules_in_Currency_Formatting):
> *If no displayName element is found at any level in the fallback chain, the system uses the currency code itself (e.g., "ZWD") as the display name.*

If we didn't use `allow_identifier_not_found()?`, the formatter construction would fail completely for unsupported currencies. By allowing it, we can still format the number and gracefully fall back to the ISO code (e.g., `12,345.67 XYZ`).

components/experimental/src/dimension/currency/compact_formatter.rs
```diff
@@ -252,4 +272,33 @@ impl CompactCurrencyFormatter {
             )),
         )
     }
+
+    fn resolve_fraction_info(
```
younies: add a comment to this function.

younies: Done. Added doc comments to both \`compact_formatter.rs\` and \`formatter.rs\`.

provider/source/data/debug/currency/essentials/v1/und.json
```diff
@@ -988,5 +988,15 @@
     "narrow_pattern_selection": "StandardAlphaNextToNumber",
     "short_placeholder_value": null,
     "narrow_placeholder_value": null
-  }
+  },
+  "fractions_vec": [
```
robertbastian: this seems to be the same for all locales, and matches `CurrencyFractions`' default value.

if you just use `u8::MAX` for `standard_index` and `standard_next_to_alpha`, this it will fall back to this value from `CurrencyFractions`, without you having to store it here

components/experimental/src/dimension/provider/currency/essentials.rs
```diff
@@ -80,6 +81,16 @@ pub struct CurrencyEssentials<'data> {
...
+
+    /// Locale-specific pattern precision.
+    #[cfg_attr(feature = "serde", serde(borrow))]
+    pub fractions_vec: ZeroVec<'data, FractionInfo>,
+
+    /// Index into `fractions_vec` for the standard pattern.
+    pub standard_index: u8,
+
+    /// Index into `fractions_vec` for the standard alpha next to number pattern.
+    pub standard_next_to_alpha: u8,
```
robertbastian:
```suggestion
    pub standard_next_to_alpha_fractions_index: u8,
```

components/experimental/src/dimension/provider/currency/essentials.rs
```diff
@@ -80,6 +81,16 @@ pub struct CurrencyEssentials<'data> {
     /// The fallback currency pattern configuration used
     /// when a specific currency's pattern is not found in the currency patterns map.
     pub default_pattern_config: CurrencyPatternConfig,
+
+    /// Locale-specific pattern precision.
+    #[cfg_attr(feature = "serde", serde(borrow))]
+    pub fractions_vec: ZeroVec<'data, FractionInfo>,
+
+    /// Index into `fractions_vec` for the standard pattern.
+    pub standard_index: u8,
```
robertbastian:
```suggestion
    pub standard_fractions_index: u8,
```

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -500,3 +576,37 @@ fn load_with_fallback<'a, M: DataMarker>(
...
+///   e.g., 2 decimals: round to `50 * 10^-2 = 0.50` (magnitude -1, increment 5).
+pub(crate) fn apply_precision(value: FixedDecimal, fraction_info: FractionInfo) -> FixedDecimal {
+    let precision = fraction_info.digits as i16;
+    let rounding = fraction_info.rounding;
+
+    let (magnitude, increment) = match rounding {
+        Rounding::R50 => (-precision + 1, RoundingIncrement::MultiplesOf5),
+        Rounding::R20 => (-precision + 1, RoundingIncrement::MultiplesOf2),
+        Rounding::R5 => (-precision, RoundingIncrement::MultiplesOf5),
+        _ => (-precision, RoundingIncrement::MultiplesOf1),
```
robertbastian: be exhaustive here
```suggestion
        Rounding::R1 => (-precision, RoundingIncrement::MultiplesOf1),
```

provider/source/src/currency/essentials.rs
```diff
@@ -393,15 +394,57 @@ fn extract_currency_essentials<'data>(
...
+            has_decimal = true;
+        }
+    }
+    FractionInfo {
+        digits,
+        rounding: Rounding::R1,
+        cash_digits: None,
+        cash_rounding: None,
+    }
+}
```
robertbastian: why do you use this data type if all you store is `digits`?

younies: I was assuming I needed the first two fields and used `FractionInfo` just for consistency, but you are right. Indeed, we only extract the number of digits from the pattern. The other fields in `FractionInfo` (rounding, cash digits) always use default values (`Rounding::R1` and `None`) when extracted from the pattern.

I have updated `CurrencyEssentials` to store `u8` (digits) instead of `FractionInfo` in `fractions_vec`. We now construct `FractionInfo` on the fly in the formatter when falling back to the pattern precision. This saves space in the data provider.

provider/source/data/debug/currency/essentials/v1/und.json
younies: Done. I've updated the datagen to use `u8::MAX` as a sentinel value when the precision matches the default (2), leaving `fractions_vec` empty for most locales. The formatter now falls back to `CurrencyFractions`' default when it encounters `u8::MAX`.

components/experimental/src/dimension/currency/formatter.rs
younies: done

components/experimental/src/dimension/provider/currency/essentials.rs

components/experimental/src/dimension/provider/currency/essentials.rs

components/experimental/src/dimension/currency/compact_formatter.rs
```diff
@@ -252,4 +272,48 @@ impl CompactCurrencyFormatter {
...
         )
     }
+
+    /// Resolves the fraction/precision information for a given currency.
+    ///
+    /// The resolution follows a 3-step fallback:
+    /// 1. Try currency-specific override in the global map (e.g., JPY = 0 decimals).
+    /// 2. Try locale-specific pattern precision from `CurrencyEssentials` (if available).
+    /// 3. Fallback to the global default.
+    fn resolve_fraction_info(
```
robertbastian: issue: move this method to be on `CurrencyFractions` or `CurrencyEssentials`. it is not `CompactCurrencyFormatter` specific

components/experimental/src/dimension/currency/compact_formatter.rs
```diff
@@ -252,4 +272,48 @@ impl CompactCurrencyFormatter {
...
+        &self,
+        essentials: Option<&CurrencyEssentials>,
+        pattern_selection: Option<PatternSelection>,
+        currency_code: CurrencyCode,
+    ) -> FractionInfo {
+        let iso_code = currency_code.0.to_unvalidated();
+
+        // 1. Try currency-specific override in global map
+        if let Some(ule) = self.fractions.get().fractions.get(&iso_code) {
```
robertbastian: use `get_copied` to get the aligned type

components/experimental/src/dimension/provider/currency/essentials.rs
```diff
@@ -80,6 +80,16 @@ pub struct CurrencyEssentials<'data> {
...
+
+    /// Locale-specific pattern precision (number of decimal digits).
+    #[cfg_attr(feature = "serde", serde(borrow))]
+    pub fractions_vec: ZeroVec<'data, u8>,
+
+    /// Index into `fractions_vec` for the standard pattern.
+    pub standard_fractions_index: u8,
+
+    /// Index into `fractions_vec` for the standard alpha next to number pattern.
+    pub standard_next_to_alpha_fractions_index: u8,
```
robertbastian: issue: now that this is just a `u8`, you're not gaining anything by storing a list + 2 indices, just store two values.

in fact, you can probably pack both values into a single u8, as 4 bits should be more than enough fraction digits for any locale

provider/source/src/currency/essentials.rs
```diff
@@ -393,15 +393,59 @@ fn extract_currency_essentials<'data>(
         accounting_alpha_next_to_number_negative: accounting_alpha_neg_idx,
     };
 
+    const DEFAULT_PRECISION: u8 = 2;
```
robertbastian: issue: this needs to use the default value from the `CurrencyFractionsV1` data, as that's what this will fall back to

components/decimal/src/abstract_formatter.rs
```diff
@@ -28,7 +28,10 @@ pub trait AbstractFormatter: core::fmt::Debug + Sealed {
         Self: 'a;
 
     #[doc(hidden)]
-    fn format_unsigned<'a>(&'a self, value: &'a UnsignedDecimal) -> Self::FormattedUnsigned<'a>;
+    fn format_unsigned<'a>(
+        &'a self,
+        value: crate::Cow<'a, UnsignedDecimal>,
```
robertbastian: currency is the only caller of these, just take by value

components/experimental/src/dimension/provider/currency/essentials.rs
robertbastian: I also think it should be a hard error for alpha-next-to-number to use different fractional grouping than standard. For which locales do these differ?

In fact, looking at the test data, at least none of the test locales set this to a non-default value. Do we actually need locale-specific rounding? I think the rounding info should *always* come from the currency, and if CLDR doesn't have rounging information for a currency, that just means to use the default.

robertbastian (commented):
please let's get https://github.com/unicode-org/icu4x/pull/8235 in first, it's waiting for your review

components/decimal/src/abstract_formatter.rs
younies: Adopted! `AbstractFormatter::format_unsigned` now takes `value: UnsignedDecimal` directly by value.

provider/source/src/currency/essentials.rs
```diff
@@ -171,12 +171,47 @@ fn extract_currency_essentials<'data>(
         accounting_alpha_next_to_number_negative: accounting_alpha_neg_idx,
     };
 
+    let standard_fractions = extract_precision(standard);
+    for pattern in [
+        standard_alpha_next_to_number,
+        accounting,
+        accounting_alpha_next_to_number,
```
younies: where are the negatives, if they are exist too

younies: Updated `extract_currency_essentials` to check both `.positive` and optional `.negative` subpatterns across all currency pattern variants during datagen verification.

younies: done

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -374,9 +400,9 @@ impl CurrencyFormatter<DecimalFormatter> {
     /// use tinystr::*;
     /// use writeable::assert_writeable_eq;
     ///
-    /// let currency_preferences = locale!("en-US").into();
+    /// let locale = locale!("en-US").into();
     /// let currency_code = CurrencyCode(tinystr!(3, "USD"));
-    /// let fmt = CurrencyFormatter::try_new_long(currency_preferences, &currency_code).unwrap();
+    /// let fmt = CurrencyFormatter::try_new_long(locale, &currency_code).unwrap();
```
younies: here the correct is currency_preferences, right ?

younies: Yes, exactly right! Updated all doc test examples in `formatter.rs` to use `let currency_preferences = locale!(\"en-US\").into();` for accurate and consistent naming.

younies: done

components/decimal/src/abstract_formatter.rs
```diff
@@ -22,13 +22,16 @@ pub trait Sealed {}
 /// trait, please consider using a type from the implementors listed below.
 /// </div>
 pub trait AbstractFormatter: core::fmt::Debug + Sealed {
+    #[doc(hidden)]
+    const REQUIRES_CURRENCY_PRECISION: bool = true;
```
younies: @robertbastian : this constant is for having the same handling path for the currency and compact currency formatting, it works as a gate to allow applying precision or not. I am not sure if this is the best way for that.

components/decimal/src/provider.rs
```diff
@@ -306,7 +306,7 @@ impl<'a> core::ops::Deref for DecimalSymbols<'a> {
 
 impl DecimalSymbols<'static> {
     /// Create a new en-US format for use in testing
-    #[cfg(feature = "datagen")]
+    #[cfg(any(test, feature = "datagen"))]
```
younies: why this ?

younies: We need `any(test, feature = "datagen")` because `new_en_for_testing()` is invoked directly during standard unit test runs in `test_grouper()` (`components/decimal/src/grouper.rs:162`).

If gated exclusively on `#[cfg(feature = "datagen")]`, running tests without `--features datagen` enabled fails to compile with `E0599` (`no function or associated item named new_en_for_testing found for struct DecimalSymbols`). This mirror existing conventions elsewhere in the codebase (such as in `datetime/src/provider/range_patterns.rs`).

components/decimal/src/abstract_formatter.rs
robertbastian: * this type should not mention "currency" at all. if you're doing this, it should be modeled as some kind of minimum displayable magnitude, which would be 0 for compact
* for numbers below 1000, compact formatting and decimal formatting are the same, it's not correct to just say compact does not need to apply currency precision

components/decimal/src/provider.rs
robertbastian: revert this and tell your AI to always run tests with `--all-features`

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -52,11 +56,13 @@ pub(crate) enum CurrencyFormatterData {
     Essential {
         essential: DataPayload<CurrencyEssentialsV1>,
         symbol: DataPayload<CurrencySymbolsV1>,
+        currency: CurrencyCode,
     },
     Long {
-        extended: DataPayload<CurrencyExtendedDataV1>,
+        extended: Option<DataPayload<CurrencyExtendedDataV1>>,
```
robertbastian: issue: create a `::IsoExtended` variant instead of having optional `DataPayload`s

we should also figure out if we need to distinguish between `::Iso` and `::IsoExtended`

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -457,44 +484,82 @@ impl<V: AbstractFormatter> CurrencyFormatter<V> {
...
             CurrencyFormatterData::Iso {
                 essential,
                 currency,
             } => {
                 let pattern = essential.get().get_positive(true, true);
-                (pattern, currency.0.as_str())
+                let (value_to_format, sign) = if V::REQUIRES_CURRENCY_PRECISION {
+                    // Note: we assume that all currency formatting variants within a locale
+                    // share the same fraction digits as the standard pattern (verified during datagen).
+                    let fraction_info =
```
robertbastian: issue: this is independent of the value (and `Copy`), so look it up in the constructor. you can then pull the rounding code outside of the match

components/experimental/src/dimension/provider/currency/essentials.rs
```diff
@@ -52,6 +52,8 @@ pub struct CurrencyEssentials<'data> {
 
     /// Indices into `patterns` for each formatting variant.
     pub indices: PatternIndices,
+    /// Locale-specific standard fraction digits.
+    pub standard_fractions: u8,
```
robertbastian: what's "standard" about this? it applies to all patterns, so surely 
```suggestion
    pub fraction_digits: u8,
```
would be a better name?

provider/source/src/currency/essentials.rs
```diff
@@ -47,31 +47,55 @@ impl DataProvider<CurrencyEssentialsV1> for SourceDataProvider {
...
+            let default_numsys = &numbers.default_numbering_system;
+
+            for (nsname, patterns) in &numbers.numsys_data.currency_patterns {
+                if patterns.standard.positive.is_empty() {
+                    log::warn!("{locale}/{nsname}: empty standard currency pattern, skipping");
+                    continue;
+                }
+                if nsname == default_numsys {
+                    ids.insert(DataIdentifierCow::from_locale(locale));
+                } else if let Ok(attr) = DataMarkerAttributes::try_from_str(nsname) {
```
robertbastian: issue: silently discarding invalid numbering systems, this should be a loud error

components/experimental/src/dimension/currency/formatter.rs
robertbastian: please clean up conversations between you and your AI before requesting review, I don't think I need to see this?

provider/source/src/currency/essentials.rs
robertbastian: it might make sense to change this type to be
```rust
struct RoundingInfo { digits: u8, rounding: Rounding }
struct FractionInfo { normal: RoundingInfo, cash: RoundingInfo }
```

`RoundingInfo` can probably be packed into a `u8`, using 4 bits for `Rounding` and 4 bits for digits.

you'd store a `RoundingInfo` in the `CurrencyFormatter`

components/experimental/src/dimension/currency/compact_formatter.rs
robertbastian: please address

components/experimental/src/dimension/currency/compact_formatter.rs
robertbastian: please address

provider/source/src/currency/essentials.rs
```diff
@@ -47,31 +47,55 @@ impl DataProvider<CurrencyEssentialsV1> for SourceDataProvider {
...
+            let numbers_resource: &cldr_serde::numbers::Resource = self
+                .cldr()?
+                .numbers()
+                .read_and_parse(&locale, "numbers.json")?;
+            let numbers = &numbers_resource.main.value.numbers;
+            let default_numsys = &numbers.default_numbering_system;
+
+            for (nsname, patterns) in &numbers.numsys_data.currency_patterns {
+                if patterns.standard.positive.is_empty() {
+                    log::warn!("{locale}/{nsname}: empty standard currency pattern, skipping");
```
robertbastian: is this warn-worthy? plenty of locales don't have all data in CLDR, that's why there's a fall back strategy

components/decimal/src/abstract_formatter.rs
younies: Thanks @robertbastian, good points.
I removed the constant from `AbstractFormatter` entirely so the trait remains agnostic to currencies and type-level switches.
Currency precision (`apply_precision`) is now applied uniformly across all formatters in `CurrencyFormatter::format_fixed_decimal`. When `CompactDecimalFormatter` renders the rounded number, its internal scaling and `.trimmed_end()` logic automatically strips trailing fractional zeros. This directly implements **UTS #35 (LDML Part 3, §3.8 Rule 3)**:
* **$< 1,000$:** Falls back to standard patterns with 0 fractional digits by default (e.g., `12` USD $\rightarrow$ `$12`, not `$12.00`, and `-990` $\rightarrow$ `-$990`).
* **$\ge 1,000$:** Transitions to abbreviations and rounds to significant digits (`1200` USD $\rightarrow$ `$1.2K` and `1299` $\rightarrow$ `$1.3K`).
I've updated the documentation and added tests (`test_tr35_rule_3_below_1000` and `test_en_us`) verifying these fallback behaviors across positive and negative values.

components/decimal/src/provider.rs
younies: done

components/experimental/src/dimension/currency/formatter.rs
younies: done

components/experimental/src/dimension/currency/formatter.rs
younies: done :)

components/experimental/src/dimension/provider/currency/essentials.rs
younies: Done

provider/source/src/currency/essentials.rs
younies: done

provider/source/src/currency/essentials.rs
younies: done

components/experimental/src/dimension/currency/compact_format.rs
```diff
@@ -172,4 +175,59 @@ mod tests {
...
+
+    #[test]
+    pub fn test_tr35_rule_3_below_1000() {
+        // Per UTS #35 (LDML Part 3: Numbers, Section 3.8, Rule 3 in Compact Number Formatting):
+        // "If the element value of P is '0', then use the corresponding non-compact number formatting instead,"
+        //
+        // Specifically, for numbers below 1,000 (which map to the uncompacted pattern "0"):
+        // * Suffix division and compact abbreviation steps (like "K" or "M") are skipped.
+        // * The standard non-compact currency pattern is used instead.
+        // * By default, maximum fractional digits are set to 0, stripping trailing zeros (so 12 USD formats
```
robertbastian: This is the current behaviour of the `CompactDecimalFormatter`, but I've looked for this in UTS-35 and couldn't find anything suggesting that this is correct.

components/experimental/src/dimension/currency/compact_formatter.rs
younies: done

components/experimental/src/dimension/currency/compact_formatter.rs
younies: done

provider/source/src/currency/essentials.rs
younies: Added a TODO and will address this in the next PR.

components/experimental/src/dimension/provider/currency/fractions.rs
```diff
@@ -40,7 +40,44 @@ pub struct CurrencyFractions<'data> {
...
+
+        // 2. Try locale-specific pattern fraction digits
+        if let Some(essentials) = essentials {
+            return FractionInfo {
+                digits: essentials.fraction_digits,
+                rounding: Rounding::R1,
+                cash_digits: None,
+                cash_rounding: None,
+            };
+        }
```
robertbastian: I think it's weird to derive fractional digits from the patterns for unknown currencies instead of using the default:

```xml
            <info iso4217="DEFAULT" digits="2" rounding="0"/>
```

In fact, for currency names, UTS-35 states

> 5. As when formatting symbol currency values, reset the number of decimals according to the supplemental <currencyData> and use the currencyDecimal symbol if different from the decimal symbol.

which suggests this should be done for both symbols and names and that the two fractional digits that are in every currency pattern in CLDR are pointless.

It goes on to say 

> 1. The number of decimals should be overridable in an API, so that clients can choose between “2 US dollars” and “2.00 US dollars”.

so this should be done as a follow-up.

provider/source/src/currency/essentials.rs
robertbastian: don't log this at any level

components/experimental/src/dimension/currency/compact_format.rs
younies: Updated the comments to clearly distinguish between what is specified in UTS #35 and ICU4X's current zero-trimming behavior.

younies: Per **UTS #35 / TR35 (LDML Part 3, Section 3.8, Rule 3)**:
> "If the element value of P is *0*, then use the corresponding non-compact number formatting instead,"

* **What TR35 actually states:** For numbers below 1,000 (which map to the compact pattern `0`), compact abbreviation and suffix division steps (such as `K` or `M`) are skipped in favor of non-compact formatting. TR35 does not mandate stripping standard currency fraction digits.
* **How it is implemented here:** We apply currency precision uniformly across all formatters without type-level switches. For numbers under 1,000, `CompactDecimalFormatter` cleanly falls back without suffix division per TR35 Rule 3, while trimming trailing fractional zeros by default as part of its current implementation (rendering e.g. `$12` instead of `$12.00`). I've updated the comments to clearly separate the TR35 rule from this ICU4X zero-trimming behavior.

robertbastian: please link to this rule 3 that you keep mentioning

younies: Here is the direct link to the algorithm in **[UTS #35 / TR35, Part 3: Numbers, Section 3.8 (Compact Number Formatting)](https://www.unicode.org/reports/tr35/tr35-numbers.html#Compact_Number_Formatting)**.

Under the 9-step formatting algorithm in Section 3.8, **Step 3** literally states:
> 3. If the element value of P is "0", then use the corresponding non-compact number formatting instead, and skip the rest of these steps — but adjust the precision as described below.
>   * For example, instead of `currencyFormat` `<pattern type="10000" count="one">¤00K</pattern>`, use `<pattern>¤#,##0.00</pattern>`.

And two paragraphs below the algorithm, TR35 explicitly defines the **precision adjustment for the "0" case**, explicitly specifying the zero-trimming behavior for currencies:
> *"If the value is precisely '0', either explicit or defaulted, then the normal number format pattern for that sort of object is supplied — either `<decimalFormat>` or `<currencyFormat type="standard">` — with the normal formatting for the locale (such as the grouping separators). **However, for the '0' case by default the significant digits are adjusted for consistency, typically to 2 or 3 digits, and the maximum fractional digits are set to 0 (for both currencies and plain decimal). Thus the output would be $12, not $12.01.** APIs may, however, allow these default behaviors to be overridden."*

robertbastian: that link does not work. is it AI generated?

components/experimental/src/dimension/provider/currency/fractions.rs
younies: Done. Removed pattern-derived fraction fallback to rely directly on supplemental `<currencyData>` and its `DEFAULT` value. We will explore an API for overriding decimal digit count as a follow-up.

components/experimental/src/dimension/provider/currency/essentials.rs
```diff
@@ -52,6 +52,8 @@ pub struct CurrencyEssentials<'data> {
 
     /// Indices into `patterns` for each formatting variant.
     pub indices: PatternIndices,
+    /// Locale-specific pattern fraction digits.
+    pub fraction_digits: u8,
```
robertbastian: also remove the data

provider/source/src/currency/essentials.rs
younies: Removed logging completely.

components/experimental/src/dimension/provider/currency/essentials.rs

components/experimental/src/dimension/currency/compact_format.rs
younies: Screenshot 2026-07-22 at 18 27 22

I use AI, but the problem is that TR35 doesn't have enough good sections. That's something I want to work on, too.

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -494,10 +529,6 @@ impl CurrencyFormatter<DecimalFormatter> {
 
     /// Creates a new [`CurrencyFormatter`] for formatting with full currency display names from compiled locale data.
     ///
-    /// ✨ *Enabled with the `compiled_data` Cargo feature.*
```
robertbastian, line 497: issue: comment disappeared

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -335,11 +372,6 @@ impl CurrencyFormatter<DecimalFormatter> {
         ]
     );
 
-    // We manually implement the compiled constructors because of the cross-crate dependency
```
robertbastian, line 338: I believe this still applies?

robertbastian (commented):
overall lgtm, just two nits

components/experimental/src/dimension/currency/formatter.rs
younies: good catch, done

components/experimental/src/dimension/currency/formatter.rs
younies: Done, restored the comment.

younies: I removed the TODO because I don't think we'll implement it in the future, as CurrencyFormatter will depend on DecimalFormatter.

robertbastian (approved):

PR #8172, opened 2026-07-03
Writeables for range formatting

Manishearth:
Part of #5448

Split out of #8149

This PR just has the writeables: without figuring out data loading, _once_ you have the data loaded you can produce these types for formatting.

I could add tests for these, but I think they'd be more fiddly than it's worth. #8149 has tests showing that this works end-to-end.

You can see #8149 to understand how this is used: I'm trying to keep this work in smaller PRs as much as possible and at this point more of it is "land APIs that will be useful later" which is harder to holistically review but overall the code here is still pretty straightforward so it should be fine.

### DIFF
--- a/components/datetime/src/provider/day_periods.rs
+++ b/components/datetime/src/provider/day_periods.rs
@@ -63,7 +63,7 @@ impl DayPeriodRules {
     }
 
     /// Encodes rules into a string
-    #[cfg(feature = "datagen")]
+    #[cfg(any(feature = "datagen", test))]
     pub fn encode_to_string(&self) -> String {
         let [a, b, c] = self.transitions;
         let raw = u32::from_le_bytes([a, b, c, 0]);
@@ -80,7 +80,7 @@ impl DayPeriodRules {
     }
 }
 
-#[cfg(feature = "datagen")]
+#[cfg(any(feature = "datagen", test))]
 impl DayPeriodRules {
     /// Computes `DayPeriodRules` from a set of periods and names.
     ///
--- a/components/datetime/src/provider/day_periods.rs
+++ b/components/datetime/src/provider/day_periods.rs
@@ -63,7 +63,7 @@ impl DayPeriodRules {
     }
 
     /// Encodes rules into a string
-    #[cfg(any(feature = "datagen", test))]
+    #[cfg(feature = "datagen")]
     pub fn encode_to_string(&self) -> String {
         let [a, b, c] = self.transitions;
         let raw = u32::from_le_bytes([a, b, c, 0]);
@@ -80,7 +80,7 @@ impl DayPeriodRules {
     }
 }
 
-#[cfg(any(feature = "datagen", test))]
+#[cfg(feature = "datagen")]
 impl DayPeriodRules {
     /// Computes `DayPeriodRules` from a set of periods and names.
     ///
--- a/components/datetime/src/provider/fields/mod.rs
+++ b/components/datetime/src/provider/fields/mod.rs
@@ -14,7 +14,6 @@ pub use length::{FieldLength, FieldNumericOverrides, LengthError};
 pub use symbols::*;
 use writeable::Writeable;
 
-#[cfg(any(feature = "unstable", feature = "datagen"))]
 pub mod components;
 
 use core::{
--- a/components/datetime/src/provider/fields/mod.rs
+++ b/components/datetime/src/provider/fields/mod.rs
@@ -14,6 +14,7 @@ pub use length::{FieldLength, FieldNumericOverrides, LengthError};
 pub use symbols::*;
 use writeable::Writeable;
 
+#[cfg(any(feature = "unstable", feature = "datagen"))]
 pub mod components;
 
 use core::{
--- a/components/datetime/src/provider/pattern/runtime/pattern.rs
+++ b/components/datetime/src/provider/pattern/runtime/pattern.rs
@@ -156,7 +156,7 @@ impl<'data> PatternBorrowed<'data> {
     ///
     /// This is used to find the split point for range patterns.
     #[allow(dead_code, reason = "#5448")]
-    fn first_repeated_field_index(&self) -> Option<usize> {
+    fn first_repeated_field_index(&self) -> usize {
         struct Seen {
             // A bitset where the i-th bit is set if we have seen a field of type index i.
             // Since there are at most 16 field types, a u16 is sufficient.
@@ -167,8 +167,8 @@ impl<'data> PatternBorrowed<'data> {
                 Self { mask: 0 }
             }
             fn insert_or_contains(&mut self, symbol: crate::provider::fields::FieldSymbol) -> bool {
-                let val = symbol.type_idx();
-                let bit = 1 << val;
+                let type_idx = symbol.type_idx();
+                let bit = 1 << type_idx;
                 if (self.mask & bit) != 0 {
                     true
                 } else {
@@ -182,12 +182,12 @@ impl<'data> PatternBorrowed<'data> {
         for (idx, item) in self.items.iter().enumerate() {
             match item {
                 PatternItem::Field(field) if seen.insert_or_contains(field.symbol) => {
-                    return Some(idx);
+                    return idx;
                 }
                 _ => {}
             }
         }
-        None
+        self.items.len()
     }
 
     /// Splits the pattern into two halves at the first repeated field.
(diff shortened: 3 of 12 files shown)


components/datetime/src/provider/day_periods.rs
```diff
@@ -63,7 +63,7 @@ impl DayPeriodRules {
     }
 
     /// Encodes rules into a string
-    #[cfg(feature = "datagen")]
+    #[cfg(any(feature = "datagen", test))]
```
robertbastian: is this diff really needed? we run tests with all features

components/datetime/src/provider/fields/mod.rs
```diff
@@ -14,7 +14,6 @@ pub use length::{FieldLength, FieldNumericOverrides, LengthError};
 pub use symbols::*;
 use writeable::Writeable;
 
-#[cfg(any(feature = "unstable", feature = "datagen"))]
```
robertbastian, line 17: why? I guess the unstable feature is fairly pointless in provider code, but if this is only used by unstable code we should keep it

components/datetime/src/provider/pattern/runtime/pattern.rs
```diff
@@ -148,6 +148,75 @@ impl<'data> PatternBorrowed<'data> {
...
+            mask: u16,
+        }
+        impl Seen {
+            fn new() -> Self {
+                Self { mask: 0 }
+            }
+            fn insert_or_contains(&mut self, symbol: crate::provider::fields::FieldSymbol) -> bool {
+                // The top 4 bits of the serialized FieldSymbol index represent the field type
+                // (e.g., Year, Month, Day), as defined in FieldSymbol::idx.
+                let val = symbol.idx() >> 4;
```
robertbastian: can you move this to the symbol type?

components/datetime/src/provider/pattern/runtime/pattern.rs
```diff
@@ -148,6 +148,75 @@ impl<'data> PatternBorrowed<'data> {
...
+    /// Splits the pattern into two halves at the first repeated field.
+    ///
+    /// If there are no repeated fields, the second half will be empty.
+    #[allow(dead_code, reason = "#5448")]
+    pub(crate) fn split_on_repeated_field(&self) -> (Self, Self) {
+        let idx = self
+            .first_repeated_field_index()
+            .unwrap_or(self.items.len());
+        let ule_slice = self.items.as_ule_slice();
+        let (start_ule, end_ule) = ule_slice.split_at(idx);
```
robertbastian: issue: split_at_checked

components/datetime/src/range/write.rs
```diff
@@ -0,0 +1,233 @@
...
+use crate::provider::range_patterns::RangePatternInfoBorrowed;
+use crate::provider::semantic_skeletons::GluePattern;
+use core::fmt;
+use writeable::{PartsWrite, Writeable};
+use zerovec::ule::AsULE;
+
+/// The formatting result of a date/time range.
+#[derive(Debug)]
+pub struct FormattedDateRange<'l> {
+    pub(crate) inner: FormattedDateRangeInner<'l>,
```
robertbastian: nit: I'd prefer to use tuple structs for the private-enum pattern. the name "inner" doesn't add any more information than "0", and a tuple struct makes matching more concise and readable

components/datetime/src/range/write.rs
```diff
@@ -0,0 +1,233 @@
...
+
+writeable::impl_display_with_writeable!(FormattedDateRange<'_>);
+
+/// The formatting result of a date/time range where the start and end dates
+/// differ by a specific field (e.g., day, month, year) and are formatted
+/// using a greatest-difference pattern.
+#[derive(Debug)]
+pub(crate) struct FormattedGreatestDifference<'l> {
+    pub(crate) start: FormattedDateTime<'l>,
+    #[allow(dead_code, reason = "https://github.com/unicode-org/icu4x/issues/5448")]
```
robertbastian: nit: you use a URL here but a raw issue number elsewhere.

that's a good use of reason though!

components/datetime/src/range/write.rs
```diff
@@ -0,0 +1,233 @@
...
+    pub(crate) end: FormattedDateTime<'l>,
```
robertbastian, line 63: observations: this contains two decimal formatters, which adds size (do we care?) but also potential for mismatch

components/datetime/src/range/write.rs
```diff
@@ -0,0 +1,233 @@
...
+                    alignment: self.alignment,
+                };
+                write_glue_pattern(sink, self.glue, &start_side, &end_side)
+            }
+        }
+    }
+}
+
+impl<'l> FormattedGreatestDifference<'l> {
+    fn write_full_range<S: PartsWrite + ?Sized>(
```
robertbastian: what is this for? this just concats two dates

robertbastian (commented):
tests?

Manishearth:
@robertbastian 

> tests?

See PR body

> I could add tests for these, but I think they'd be more fiddly than it's worth. https://github.com/unicode-org/icu4x/pull/8149 has tests showing that this works end-to-end.

components/datetime/src/provider/day_periods.rs
Manishearth: I was getting a test failure without it, not sure why. Didn't look too deeply.

components/datetime/src/provider/fields/mod.rs
Manishearth: Yeah I think this happened in one of the "fix CI" cleanup passes. Worth trying to not have this diff.

components/datetime/src/range/write.rs
Manishearth: IMO good thing to try and address in a followup which may involve more architectural shifts.

Manishearth:
I addressed most of the comments. I can try to add tests later if you really want, but I'd prefer to not do a lot of  data mocking for testing. I'll think about it more.

components/datetime/src/provider/pattern/runtime/pattern.rs
```diff
@@ -167,9 +167,7 @@ impl<'data> PatternBorrowed<'data> {
                 Self { mask: 0 }
             }
             fn insert_or_contains(&mut self, symbol: crate::provider::fields::FieldSymbol) -> bool {
-                // The top 4 bits of the serialized FieldSymbol index represent the field type
-                // (e.g., Year, Month, Day), as defined in FieldSymbol::idx.
-                let val = symbol.idx() >> 4;
+                let val = symbol.type_idx();
```
robertbastian: nit: `val` is a bad name for an index

components/datetime/src/provider/pattern/runtime/pattern.rs
```diff
@@ -148,6 +148,73 @@ impl<'data> PatternBorrowed<'data> {
...
+    /// Splits the pattern into two halves at the first repeated field.
+    ///
+    /// If there are no repeated fields, the second half will be empty.
+    #[allow(dead_code, reason = "#5448")]
+    pub(crate) fn split_on_repeated_field(&self) -> (Self, Self) {
+        let idx = self
+            .first_repeated_field_index()
+            .unwrap_or(self.items.len());
+        let ule_slice = self.items.as_ule_slice();
+        let (start_ule, end_ule) = ule_slice.split_at_checked(idx).unwrap_or((ule_slice, &[]));
```
robertbastian: thought: can `first_repeated_field_index` return the split without having the error branch? _it_ knows that the index if valid, but that information is lost when it returns

components/datetime/src/provider/day_periods.rs
robertbastian: would still like to not have these diffs

robertbastian (approved):

components/datetime/src/provider/pattern/runtime/pattern.rs
Manishearth: heh I also wanted to fix this but had to head out

components/datetime/src/provider/day_periods.rs
Manishearth: yeah I'll probably manually clean this up when I'm back.

There's always a bit of work fixing up our all-features tests.

components/datetime/src/provider/pattern/runtime/pattern.rs
Manishearth: I had it return a usize. I thought about merging the two, but either way you need to take the index and call `split_at` since there's not really any all-encapsulated "iterate to a split point and then split" Rust helper.

robertbastian: yeah not yet https://doc.rust-lang.org/std/primitive.slice.html#method.split_once

robertbastian (approved):

PR #8173, opened 2026-07-03
Scaffolding traits for range formatting

Manishearth:
Part of #5448

Split out of #8149

This hooks up the scaffolding and pattern selection data loading framework for range formatting.

You can see #8149 to understand how this is used: I'm trying to keep this work in smaller PRs as much as possible and at this point more of it is "land APIs that will be useful later" which is harder to holistically review but overall the code here is still pretty straightforward so it should be fine.

## Changelog

icu_datetime: Add scaffolding for range formatting
  - New associated type: `TypedDateDataMarkers::RangeSkel`

### DIFF
--- a/components/datetime/src/fieldsets.rs
+++ b/components/datetime/src/fieldsets.rs
@@ -59,7 +59,10 @@ pub use crate::combo::Combo;
 use crate::{
     options::*,
     provider::semantic_skeletons::{DatetimePatternsGlueV1, GluePattern},
-    provider::{fields, names::*, semantic_skeletons::*, time_zones::tz},
+    provider::{
+        fields, names::*, range_patterns::DatetimePatternsRangeTimeV1, semantic_skeletons::*,
+        time_zones::tz,
+    },
     raw::neo::RawOptions,
     scaffold::*,
 };
@@ -537,14 +540,16 @@ macro_rules! impl_date_or_calendar_period_marker {
         }
         impl<C: CldrCalendar> TypedDateDataMarkers<C> for $type {
             type DateSkeletonPatternsV1 = datetime_marker_helper!(@dates/typed, yes);
-            type YearNamesV1 = datetime_marker_helper!(@years/typed, $($years_yes)?);
-            type MonthNamesV1 = datetime_marker_helper!(@months/typed, $($months_yes)?);
+            type DateRangeSkeletonPatternsV1 = datetime_marker_helper!(@dates/range/typed, yes);
+            type YearNamesV1 = datetime_marker_helper!(@years/typed, $($year_yes)?);
+            type MonthNamesV1 = datetime_marker_helper!(@months/typed, $($month_yes)?);
             type WeekdayNamesV1 = datetime_marker_helper!(@weekdays, $($weekdays_yes)?);
         }
         impl DateDataMarkers for $type {
             type Skel = datetime_marker_helper!(@calmarkers, yes);
-            type Year = datetime_marker_helper!(@calmarkers, $($years_yes)?);
-            type Month = datetime_marker_helper!(@calmarkers, $($months_yes)?);
+            type RangeSkel = datetime_marker_helper!(@calmarkers, yes);
+            type Year = datetime_marker_helper!(@calmarkers, $($year_yes)?);
+            type Month = datetime_marker_helper!(@calmarkers, $($month_yes)?);
             type WeekdayNamesV1 = datetime_marker_helper!(@weekdays, $($weekdays_yes)?);
         }
         impl DateTimeMarkers for $type {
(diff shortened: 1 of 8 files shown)


components/datetime/src/scaffold/fieldset_traits.rs
```diff
@@ -358,28 +393,89 @@ where
 {
 }
 
-#[rustfmt::skip]
```
sffc, line 361: Nit: Change back this formatting please

components/datetime/src/fieldsets.rs
```diff
@@ -537,14 +540,16 @@ macro_rules! impl_date_or_calendar_period_marker {
         }
         impl<C: CldrCalendar> TypedDateDataMarkers<C> for $type {
             type DateSkeletonPatternsV1 = datetime_marker_helper!(@dates/typed, yes);
-            type YearNamesV1 = datetime_marker_helper!(@years/typed, $($years_yes)?);
-            type MonthNamesV1 = datetime_marker_helper!(@months/typed, $($months_yes)?);
+            type DateRangeSkeletonPatternsV1 = datetime_marker_helper!(@dates/range/typed, yes);
+            type YearNamesV1 = datetime_marker_helper!(@years/typed, $($year_yes)?);
+            type MonthNamesV1 = datetime_marker_helper!(@months/typed, $($month_yes)?);
```
sffc: Issue: don't change "years_yes" to "year_yes", here and below. They are two different macro variables with subtly different meanings. They should have better names.

components/datetime/src/raw/neo.rs
```diff
@@ -384,6 +430,70 @@ impl<'a> TimePatternDataBorrowed<'a> {
...
+        provider: &(impl BoundDataProvider<ErasedPackedRangePatterns> + ?Sized),
+        prefs: DateTimeFormatterPreferences,
+        components: TimeFieldSet,
+    ) -> Result<Self, DataError> {
+        let locale = provider
+            .bound_marker()
+            .make_locale(prefs.locale_preferences);
+        let prefs = RawPreferences::from_prefs(prefs);
+        // First try to load with the explicit hour cycle. If there is no explicit hour cycle,
+        // or if loading the explicit hour cycle fails, then load with the default hour cycle.
```
sffc, line 463: Thought: This is more duplicated than I'd like but it's a small enough bit of code that I guess it's fine?

components/datetime/src/raw/neo.rs
```diff
@@ -687,6 +797,120 @@ impl DateTimeZonePatternSelectionData {
...
+#[cfg(feature = "unstable")]
+#[allow(dead_code)]
+impl DateTimeZoneRangePatternSelectionData {
+    pub(crate) fn try_new_with_skeleton(
+        date_range_provider: &(impl BoundDataProvider<ErasedPackedRangePatterns> + ?Sized),
+        time_range_provider: &(impl BoundDataProvider<ErasedPackedRangePatterns> + ?Sized),
+        range_glue_provider: &(impl BoundDataProvider<DatetimePatternsRangeGlueV1> + ?Sized),
+        prefs: DateTimeFormatterPreferences,
+        skeleton: CompositeFieldSet,
+    ) -> Result<Self, DataError> {
```
sffc, line 769: Suggestion (optional): this is a lot of duplicated code. It might be worth putting this into a helper fn that takes some closures for the context-specific things you need, like how to create the specific PatternSelectionData objects, if such a refactor can be done cleanly.

Manishearth: I threw this at the agent and it managed to do something. It looks clean at the range pattern end but the DateTimeZone code got changed a lot. I am still looking closely at the change.

Since this is an optional suggestion, if you don't like this refactor I think the best route is to undo it and file a followup. Happy to do that.

(In general for the range formatter I think a good route is to do as much as possible via followups because that biases towards having the core feature get landed and it's easier for me to work on the little followups later)

components/datetime/src/raw/neo.rs
```diff
@@ -537,182 +582,71 @@ impl DateTimeZonePatternSelectionData {
...
-                    date,
-                    time,
-                    zone: Some(zone),
-                    glue: Some(glue),
-                })
+                });
             }
         }
+
+        let options = skeleton.to_raw_options();
```
Manishearth, line 613: I think this resulting code is much cleaner and easier to read, but I'm not 100% sure if this has the same semantics as the older stuff

components/datetime/src/raw/neo.rs
Manishearth: Yeah. Worth looking at as a followup.

components/datetime/src/scaffold/fieldset_traits.rs
```diff
@@ -567,6 +656,7 @@ impl ZoneMarkers for () {
     type MetazonePeriodV1 = NeverMarker<tz::MzPeriod<'static>>;
 }
 
+#[allow(unused_macro_rules)]
```
sffc: ?

components/datetime/src/raw/neo.rs
```diff
@@ -427,182 +582,77 @@ impl DateTimeZonePatternSelectionData {
...
-                    glue: None,
-                })
-            }
-            CompositeFieldSet::DateTime(field_set) => {
-                let options = field_set.to_raw_options();
-                // TODO(#5387): load the patterns for custom hour cycles here
-                if let (Some(attributes), None) = (field_set.id_str(), prefs.hour_cycle) {
-                    // Try loading an overlap pattern.
-                    // Note: Overlap patterns are loaded from the date skeleton pattern provider
-                    // and then stored as a TimePatternSelectionData.
```
sffc, line 503: Nit: There are some helpful comments here that got deleted. Please restore them.

sffc (commented):
The refactor looks good

sffc (approved):

PR #8174, opened 2026-07-03
Make neo an alternative constructor on existing segmenter types

robertbastian:
This will be needed for releasing neo. We need to be able to keep handling the `V1` data structs, which means we'll have to keep carrying the code for them. I'm handling this by making each `FooSegmenter`, `FooSegmenterBorrowed`, `FooBreakIterator` have an internal enum over the data struct. Currently the neo variants are `#[cfg(feature = "unstable")]`, when we stabilise neo, we will make the legacy variant be `#[cfg(feature = "serde")]`, as it will only be reachable in the serde codepath (we'll have to change the buffer provider constructors to check both data structs).

## Changelog

N/A


components/segmenter/src/complex/lstm/mod.rs
```diff
@@ -52,12 +52,15 @@ pub(super) struct LstmSegmenter<'data, G: AbstractGraphemeClusterSegmenterBorrow
     timew_fw: MatrixZero<'data, 2>,
     timew_bw: MatrixZero<'data, 2>,
     time_b: MatrixZero<'data, 1>,
-    grapheme: Option<G>,
+    grapheme: Option<GraphemeClusterSegmenterBorrowed<'data>>,
```
Manishearth, line 55: thought: hooray for removing generics from types with complex logic and potentially getting codesize wins

components/segmenter/src/line.rs
```diff
@@ -323,20 +324,42 @@ impl LineBreakOptions<'_> {
...
 #[derive(Debug)]
-pub struct LineSegmenter {
-    options: ResolvedLineBreakOptions,
-    payload: DataPayload<SegmenterBreakLineV1>,
-    complex: ComplexPayloads,
+pub struct LineSegmenter(LineSegmenterInner);
+
+#[derive(Debug)]
+enum LineSegmenterInner {
+    Legacy {
```
Manishearth, line 331: thought: and i guess eventually this can be `cfg(serde)`'d

Manishearth (approved):

components/segmenter/src/line.rs
robertbastian: yes, as explained in the PR description

PR #8175, opened 2026-07-03
feat(currency): Add cash formatting option to CurrencyFormatter

younies:
This PR implements the cash formatting option for `CurrencyFormatter` and `CompactCurrencyFormatter`.

Diff Base: #8169.

## Changelog

1.  **Add `CurrencyUsage` Enum**:
    *   Defined `CurrencyUsage` in `options.rs` with `Standard` and `Cash` variants.
    *   Replaced the temporary `cash: bool` option in `CurrencyFormatterOptions` with `usage: CurrencyUsage`.

2.  **Implement Cash Rounding Fallback**:
    *   Updated `resolve_fraction_info` in both `CurrencyFormatter` and `CompactCurrencyFormatter` to check if `options.usage` is `CurrencyUsage::Cash`.
    *   If cash usage is specified, it overrides the standard precision (`digits` and `rounding`) with `cash_digits` and `cash_rounding` from the resolved `FractionInfo` (if they are present in the data).

3.  **Update Constructors**:
    *   Modified `CurrencyFormatter` constructors (`try_new_short`, `try_new_narrow`, `try_new_long` and their unstable variants) to accept `CurrencyFormatterOptions` so that callers can configure the usage.

4.  **Tests**:
    *   Updated existing tests and doc-tests to pass `Default::default()` (which uses `Standard` usage).
    *   Added a new test case `test_cash_rounding` in `format.rs` using `CHF` (Swiss Franc) to verify that `CurrencyUsage::Cash` correctly rounds values to the nearest 0.05 (e.g., `1.23` -> `1.25`, `1.22` -> `1.20`) while `Standard` usage keeps 2 decimal places (`1.23` and `1.22`).

TAG=agy


components/experimental/src/dimension/currency/compact_formatter.rs
```diff
@@ -287,24 +290,31 @@ impl CompactCurrencyFormatter {
...
+        };
+
+        if self.options.usage == CurrencyUsage::Cash {
+            if let Some(cash_digits) = fraction_info.cash_digits {
+                fraction_info.digits = cash_digits;
+            }
+            if let Some(cash_rounding) = fraction_info.cash_rounding {
+                fraction_info.rounding = cash_rounding;
             }
         }
```
robertbastian: instead of returning a mutated `FractionInfo` that depends on options, return just the `digits` and `rounding` values

components/experimental/src/dimension/currency/formatter.rs
robertbastian, line 606: wait why is this function duplicated?

components/experimental/src/dimension/currency/compact_formatter.rs
younies: Done. I've refactored `resolve_fraction_info` to return `(u8, Rounding)` instead of mutating and returning `FractionInfo`. `apply_precision` has also been updated to accept these resolved values directly.

components/experimental/src/dimension/currency/formatter.rs
younies: Initially, the two implementations had slightly different signatures because `CurrencyFormatter` is bound to a single currency (accessing it via `self.bound_currency`), whereas `CompactCurrencyFormatter` is unbound and requires the `currency_code` to be passed as an argument. This difference in context led to separate helper functions.

However, the underlying logic is indeed identical. I've de-duplicated this by extracting the common logic into a shared helper function `resolve_fraction_info` in `formatter.rs` (which now accepts the currency code as a parameter) and updated both formatters to use it.

components/experimental/src/dimension/currency/compact_formatter.rs
robertbastian: issue: inline into its one caller

components/experimental/src/dimension/currency/options.rs
```diff
@@ -17,14 +17,31 @@ pub struct CurrencyFormatterOptions {
     // to use specific constructors, as the width is determined at construction time.
     /// The width of the currency format.
     pub width: Width,
+    /// The usage of the currency format (standard or cash).
+    pub usage: CurrencyUsage,
```
robertbastian, line 17: issue: I think this should be `Option<CurrencyUsage>` (same for width)

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -78,6 +78,8 @@ enum CurrencyFormatterData {
 /// Read more about the options in the [`super::options`] module.
 #[derive(Debug)]
 pub struct CurrencyFormatter<V: ValueRepresentation> {
+    options: CurrencyFormatterOptions,
```
robertbastian: issue: only store relevant options, and resolve them early. in this case, this either means:
* store `CurrencyUsage`
* store `(u8, Rounding)`, as this is independent of the formatted value

components/experimental/src/dimension/currency/formatter.rs
robertbastian: remove this trivial redirecting function

also, this change needs to be in the parent PR, as that introduces the duplicated code

younies:

PR #8176, opened 2026-07-03
feat(currency): Support negative patterns in CurrencyFormatter

younies:
This PR implements support for negative patterns in `CurrencyFormatter` and `CompactCurrencyFormatter`, addressing part of #6064.

Diff Base: https://github.com/unicode-org/icu4x/pull/8175

## Changelog

1.  **Update `name_and_pattern`**:
    *   Modified `CurrencyEssentials::name_and_pattern` to return both the positive and (optional) negative patterns: `(currency_str, pos_pattern, neg_pattern, pattern_selection)`.

2.  **Implement Negative Pattern Logic**:
    *   In both `CurrencyFormatter` and `CompactCurrencyFormatter`, we now check if the value is negative.
    *   If a negative pattern exists (`neg_pattern` is `Some`), we use it and format the absolute value (passing `Sign::None` to `format_sign` to avoid double-signing).
    *   If no negative pattern exists, we fall back to the positive pattern and let `format_sign` handle the negative sign (passing `Sign::Negative`).

3.  **Tests**:
    *   Added `test_negative_values` and `test_compact_negative_values` in `format.rs` to verify that negative values are formatted correctly (e.g., `-12345.67` -> `-$12,345.67` and `-$12K` in `en` locale).

TAG=agy


younies:

PR #8177, opened 2026-07-03
feat(currency): Add accounting formatting and relax constructor validation

younies:
This PR addresses what was left in #8066 and adds support for accounting currency formatting.

## Changelog
1.  **Relaxed Formatter Validation**: Removed the strict validation in `CurrencyFormatter` and `CompactCurrencyFormatter` constructors that required the standard pattern to be present. This allows the formatters to initialize and rely on runtime fallback.
2.  **Accounting Formatting**:
    *   Added `Accounting` variant to `CurrencyUsage` enum.
    *   Updated `CurrencyEssentials::name_and_pattern` to retrieve accounting patterns (`accounting_positive`, `accounting_negative`, etc.) when `CurrencyUsage::Accounting` is selected.
    *   Updated formatters to pass `self.options.usage` to `name_and_pattern`.
3.  **GIGO Fallback**:
    *   Implemented GIGO fallback in `standard_pattern` to return `FALLBACK_PATTERN` (`"{0} {1}"`) if the pattern is missing (due to corrupt data), with a `debug_assert!` in debug builds.
    *   Updated positive pattern getters to return non-optional `&DoublePlaceholderPattern`.
4.  **Datagen**:
    *   Updated datagen fallback for `accounting_alpha_next_to_number_positive` to `standard_alpha_next_to_number`.
    *   Regenerated the baked data.
5.  **Tests**:
    *   Added tests for accounting formatting in `format.rs`.
    *   Added a test for `FALLBACK_PATTERN` in `essentials.rs`.

Diff Base: `currency-plural-patterns` (which is stacked on `currency-cash-formatting` #8175)

TAG=agy


robertbastian (commented):
needs rebase

robertbastian (changes requested):
I still think that the accounting style should be handled by the `DecimalFormatter`, as that code handles signs, and `(` `)` is just a sign. `DecimalFormatter` was designed from the beginning to support signs on both sides, basically for this.

Currency should not need to store separate non-accounting and accounting patterns. It should store the _sign part_ of the accounting pattern and initialise the `DecimalFormatter` with that.

In fact, it should also store the other parts of [DecimalSymbolsStrs](https://docs.rs/icu/latest/icu/decimal/provider/struct.DecimalSymbolsStrs.html), and use that data struct to initialise the `DecimalFormatter`.

younies:

PR #8186, opened 2026-07-07
Replace last vestiges of ICU4C properties

robertbastian:
Generates the deprecated normalization-inert properties from the normalizer implementation instead of ICU4C toml files.

⚠️ vibe-assisted coding

## Changelog

N/A


provider/source/src/normalizer/mod.rs
```diff
@@ -171,3 +171,179 @@ normalization_non_recursive_decomposition_supplement_provider!(
...
+                let decomp = decomp.as_borrowed();
+                let ccc = CodePointMapData::<CanonicalCombiningClass>::try_new_unstable(self)?;
+                let ccc = ccc.as_borrowed();
+
+                let mut builder = CodePointInversionListBuilder::new();
+                for cp in 0..=(char::MAX as u32) {
+                    if let Some(ch) = char::from_u32(cp) {
+                        if ccc.get(ch) == CanonicalCombiningClass::NotReordered {
+                            let mut iter = decomp.normalize_iter([ch].into_iter());
+                            if iter.next() == Some(ch) && iter.next() == None {
```
eggrobin: This looks like a rather convoluted way to say that the string consisting of ch is in NFD (decomp.is_normalized(ch.to_string().as_str()), or however that goes).

provider/source/src/normalizer/mod.rs
```diff
@@ -171,3 +171,179 @@ normalization_non_recursive_decomposition_supplement_provider!(
...
+                    provider::PropertyCodePointSet,
+                };
+
+                let decomp = icu::normalizer::DecomposingNormalizer::$try_new_decomp(self)?;
+                let decomp = decomp.as_borrowed();
+                let ccc = CodePointMapData::<CanonicalCombiningClass>::try_new_unstable(self)?;
+                let ccc = ccc.as_borrowed();
+
+                let mut builder = CodePointInversionListBuilder::new();
+                for cp in 0..=(char::MAX as u32) {
```
eggrobin, line 201: This could deserve a comment along the lines of

```
                // Add all code points that are starters and are not canonical
                // decomposable characters, including surrogates.
                for cp in 0..=(char::MAX as u32) {
```
(where [starter](https://unicode.org/versions/Unicode17.0.0/core-spec/chapter-3/#G49580) and [canonical decomposable character](https://unicode.org/versions/Unicode17.0.0/core-spec/chapter-3/#G24016) are defined terms).

EDIT: Nevermind, this can be compatibility decomposables depending on the normalization form we are dealing with; then something like

```suggestion
                // Add all code points that are starters and are not decomposable,
                // including surrogates.
                for cp in 0..=(char::MAX as u32) {
```

eggrobin (commented):
Not done yet, the NFC_inert gloubi-boulga will require a more thorough audit.

provider/source/src/normalizer/mod.rs
eggrobin: By the way, if you have Decomposition_Type handy, that is Decomposition_Type≠Canonical.

provider/source/src/normalizer/mod.rs
```diff
@@ -171,3 +171,178 @@ normalization_non_recursive_decomposition_supplement_provider!(
     NormalizerNfdSupplementV1,
     "decompositionex"
 );
+
+// These macros implement ICU4C-internal properties that we have accidentally exposed (#7892).
+// They are slated for removal, but the code might be useful for a future ICU4C-independent
+// normalization implementation, which is why they live in this file.
+
+macro_rules! impl_nfd_inert_property {
```
eggrobin: This is a misnomer, since it is used with either NFD or NFKD—and it is confusing enough that I did not realize that on first reading and assumed this was all canonical. Call it impl_decomposition_inert_property or something along those lines.

provider/source/src/normalizer/mod.rs
```diff
@@ -171,3 +171,178 @@ normalization_non_recursive_decomposition_supplement_provider!(
...
+                use icu::collections::codepointinvlist::CodePointInversionListBuilder;
+                use icu::properties::{
+                    CodePointMapData, props::CanonicalCombiningClass,
+                    provider::PropertyCodePointSet,
+                };
+
+                let nfc = icu::normalizer::ComposingNormalizer::$try_new_comp(self)?;
+                let nfc = nfc.as_borrowed();
+                let nfd = icu::normalizer::DecomposingNormalizer::$try_new_decomp(self)?;
+                let nfd = nfd.as_borrowed();
```
eggrobin: Misnomers as above, and particularly bad since the logic here actually involves real NFC (`comp` below is canonical composition), and _should_ involve real NFD.

Rename to composition and decomposition (or abbreviations thereof if `osition` is too many letters for you), and rename comp below to either say canonical, or just to nfc (well, comp might not survive to the other changes anyway; but call nfc things that are really nfc).

provider/source/src/normalizer/mod.rs
```diff
@@ -171,3 +171,178 @@ normalization_non_recursive_decomposition_supplement_provider!(
...
+                let comp =
+                    icu::normalizer::properties::CanonicalComposition::try_new_unstable(self)?;
+                let comp = comp.as_borrowed();
+                let ccc = CodePointMapData::<CanonicalCombiningClass>::try_new_unstable(self)?;
+                let ccc = ccc.as_borrowed();
+
+                let mut potential_seconds = HashSet::new();
+                for ch in (0..=char::MAX as u32).filter_map(char::from_u32) {
+                    let mut iter = nfd.normalize_iter([ch].into_iter());
```
eggrobin: This should use actual NFD, not NF(possibly K)D, since the purpose of this set is to check below whether they are trailing in a _canonical_ decomposition (composition is always based on _canonical_ decompositions, NFKC does not compose صلى الله عليه وسلم into ﷺ nor キロメートル into ㌖).

But taking a step back, this business of nested iteration over potential_seconds is a very convoluted way of constructing the sets of characters that combine forward or backward.

For every [primary composite](https://unicode.org/versions/Unicode17.0.0/core-spec/chapter-3/#G49608)[^primary-composite], if you take the first code point of the (non-recursive) canonical Decomposition_Mapping (which seems to be available here as CanonicalDecomposition), that code point is one that combines forward; you can obtain the set of code points that combine forward like this in a single pass.

For the set of code points that combine backward, those are the ones whose NFD (here you want the actual (recursive) [canonical decomposition](https://unicode.org/versions/Unicode17.0.0/core-spec/chapter-3/#G7425)) starts with a character that appears as the second character of the canonical Decomposition_Mapping of a primary composite; you can first construct the set of characters that appear at the end of the Decomposition_Mapping of a primary composite (you are looking at Decomposition_Mappings of primary composites for the forward-combining characters anyway), and then do a pass over the code space to construct the backward-combinings.

[^primary-composite]: If you have the right properties you can look over `[[:Decomposition_Type=Canonical:]-[:Full_Composition_Exclusion:]]`. If you do not, you could characterize those as code points that are in NFC and decompose under NFD, or using CanonicalComposition and CanonicalDecomposition.

provider/source/src/normalizer/mod.rs
```diff
@@ -171,3 +171,178 @@ normalization_non_recursive_decomposition_supplement_provider!(
...
+
+                let mut combines_forwards = HashSet::new();
+                let mut combines_backwards = HashSet::new();
+                let mut composes_with_ccc = HashSet::new();
+                for ch in (0..=char::MAX as u32).filter_map(char::from_u32) {
+                    for &second in &potential_seconds {
+                        if comp.compose(ch, second).is_some() {
+                            combines_forwards.insert(ch);
+                            combines_backwards.insert(second);
+                            composes_with_ccc.insert((ch, ccc.get(second)));
```
eggrobin: You only need to keep track of the lowest CCC encountered there, and you can then put that into an associative container (starter to lowest CCC of the trailing character of a canonical decomposition of a primary composite), and avoid the weird loop over ranges of CCC values below.

eggrobin (changes requested):

sffc (commented):
Seems good; I'll defer to @eggrobin on the detailed algorithm review

provider/source/src/normalizer/mod.rs
```diff
@@ -171,3 +171,194 @@ normalization_non_recursive_decomposition_supplement_provider!(
...
+                            if prev_ccc.0 < lowest_ccc && lowest_ccc < ccc.0 {
+                                continue 'cp;
+                            }
+                        }
+
+                        if let Some(composite) = canonical_comp.compose(starter, follow) {
+                            starter = composite;
+                            prev_ccc = CanonicalCombiningClass::NotReordered;
+                        } else {
+                            prev_ccc = ccc;
```
eggrobin: This branch, and the prev_ccc thing in general, does not make sense to me, but I think it is conveniently dead code: if the whole thing did not canonically compose back to one character, we would have exited at line 304 above.

Can you unwrap composite instead? And then you can replace the one usage of prev_ccc with `0`.

provider/source/src/normalizer/mod.rs
```diff
@@ -171,3 +171,194 @@ normalization_non_recursive_decomposition_supplement_provider!(
...
+
+                    let mut starter = decomposed.next().unwrap();
+                    let mut prev_ccc = CanonicalCombiningClass::NotReordered;
+
+                    if combines_backwards.contains(&starter) {
+                        continue;
+                    }
+
+                    for follow in decomposed {
+                        let ccc = ccc.get(follow);
```
eggrobin: Let’s handle CCC numerically as soon as possible, the .0 in the comparisons below don’t help with readability.
```suggestion
                        let ccc = ccc.get(follow).0;
```

robertbastian: or we just stay in the strongly typed realm

eggrobin: I tend to prefer that, but then we would really need comparison operators on CCC (that is a natural and well-defined operation for CCC).

eggrobin: Ah, apparently we have those? Nice.

robertbastian: yes, the only thing we don't have is the `Step` trait, needed for `a..=b`, because that's currently unstable, but we don't do the loop anymore

provider/source/src/normalizer/mod.rs
```diff
@@ -171,3 +171,188 @@ normalization_non_recursive_decomposition_supplement_provider!(
...
+
+                let canonical_comp = CanonicalComposition::try_new_unstable(self)?;
+                let canonical_comp = canonical_comp.as_borrowed();
+                let canonical_decomp = CanonicalDecomposition::try_new_unstable(self)?;
+                let canonical_decomp = canonical_decomp.as_borrowed();
+                let ccc = CodePointMapData::<CanonicalCombiningClass>::try_new_unstable(self)?;
+                let ccc = ccc.as_borrowed();
+
+                let mut combines_forwards = HashSet::new();
+                let mut potential_seconds = HashSet::new();
```
eggrobin: The adjective `potential` doesn’t make much sense to me here; these are actually second characters of the canonical decomposition of a primary composite. How about `canonical_comp_seconds`? (`canonical_decomp_seconds` sounds like it would include the seconds of the decompositions of composition exclusions, which is not the case.)

provider/source/src/normalizer/mod.rs
```diff
@@ -171,3 +171,188 @@ normalization_non_recursive_decomposition_supplement_provider!(
...
+                let mut composes_with_lowest_reordered_ccc = HashMap::new();
+
+                for ch in (0..=char::MAX as u32).filter_map(char::from_u32) {
+                    if let Decomposed::Expansion(starter, second) = canonical_decomp.decompose(ch)
+                        && canonical_comp.compose(starter, second) == Some(ch)
```
eggrobin, line 271:
```suggestion
                // Compute `combines_forwards`, `canonical_comp_seconds`, and
                // `composes_with_lowest_reordered_ccc` from the canonical
                // decompositions of all primary composites.
                for ch in (0..=char::MAX as u32).filter_map(char::from_u32) {
                    if let Decomposed::Expansion(starter, second) = canonical_decomp.decompose(ch)
                        && canonical_comp.compose(starter, second) == Some(ch)
```
Or something to that effect, so it is clear what the condition in the if statement actually means (maybe link to https://unicode.org/versions/Unicode17.0.0/core-spec/chapter-3/#G49608 if you feel like it).

provider/source/src/normalizer/mod.rs
```diff
@@ -171,3 +171,188 @@ normalization_non_recursive_decomposition_supplement_provider!(
...
+                                .entry(starter)
+                                .and_modify(|c| *c = std::cmp::min(*c, ccc))
+                                .or_insert(ccc);
+                        }
+                    }
+                }
+
+                let mut combines_backwards = HashSet::new();
+                for ch in (0..=char::MAX as u32).filter_map(char::from_u32) {
+                    let starter = nfd.normalize_iter([ch].into_iter()).next().unwrap();
```
eggrobin: Misnomer: this need not be a [starter](https://unicode.org/versions/Unicode17.0.0/core-spec/chapter-3/#G49580) (and many of the characters that combine backward are nonstarters).
```suggestion
                    let nfd_first = nfd.normalize_iter([ch].into_iter()).next().unwrap();
```
(or [`full_canonical_decomposition_first`](https://unicode.org/versions/Unicode17.0.0/core-spec/chapter-3/#G59401), that is probably more appropriate but equivalent on single characters and more verbose.)

This is in contrast to the loop over primary composites above, since characters with [non-starter decompositions](https://unicode.org/versions/Unicode17.0.0/core-spec/chapter-3/#G49603) are not primary composites.

provider/source/src/normalizer/mod.rs
```diff
@@ -171,3 +171,188 @@ normalization_non_recursive_decomposition_supplement_provider!(
...
+
+                    if combines_backwards.contains(&starter) {
+                        continue;
+                    }
+
+                    for follow in decomposed {
+                        if let Some(&lowest_ccc) = composes_with_lowest_reordered_ccc.get(&starter)
+                            && lowest_ccc < ccc.get(follow)
+                        {
+                            continue 'cp;
```
eggrobin, line 331:
```suggestion
                            // There exists a character C (with
                            // ccc=`lowest_ccc`) which composes with `starter`
                            // and which is not blocked from `starter` in the
                            // sequence <`starter`, `follow`, C>, nor therefore in
                            // <`starter`, (rest of `decomposed`), C>,
                            // thus the decomposition of <`ch`, C> would compose
                            // to something different and `ch` is not inert.
                            continue 'cp;
```
(Link to [the definition of blocked](https://unicode.org/versions/Unicode17.0.0/core-spec/chapter-3/#G49610) to taste.)

eggrobin (changes requested):
Looks good now. Some nitpicks about identifiers, and suggestions for comments to leave breadcrumbs for whoever needs to understand this code a few decades from now…

eggrobin (approved):

PR #8187, opened 2026-07-07
feat(currency): Add accounting support for symbol and code formatting

### LINKED ISSUE #8185: currency: Add accounting option support for compact currency formatting (ECMA-402 compliance)
### Summary
Currently, `CurrencyFormatter<Compact>` in ICU4X supports standard short and narrow compact currency formatting (e.g., `-$3.5M` in `en-US`). However, financial and bookkeeping applications require **accounting compact currency formatting**, where negative amounts are formatted according to locale-specific accounting conventions (typically using parentheses instead of minus signs, e.g., `($3.5M)`).

### Prior Art & Standards (ECMAScript 402)
The ECMAScript Internationalization API Specification (ECMA-402) explicitly supports combining `currencySign: 'accounting'` with `notation: 'compact'`.

From ECMA-402 `Intl.NumberFormat`:
- `currencySign`: May be set to `'standard'` (default) or `'accounting'` (enabling accounting format for negative values).
- `notation`: May be set to `'compact'` (with `compactDisplay: 'short'` or `'long'`).

#### JavaScript / Node.js Example
```javascript
const formatter = new Intl.NumberFormat('en-US', {
  style: 'currency',
  currency: 'USD',
  currencySign: 'accounting',
  notation: 'compact',
  compactDisplay: 'short'
});

console.log(formatter.format(-3500000)); 
// Output: "($3.5M)"

### PR DESCRIPTION
**Pull request by younies**
## Summary 📋

Add `CurrencyUsage::{Standard, Accounting}` via `CurrencyFormatterOptions` so negative values can use accounting patterns (e.g. `($12,345.67)` / `($12K)` in `en-US`) 💵.

Applies to symbol and code constructors, including compact and compact-long variants 🔣. Does not apply to name or compact-name constructors.

Fixes #8185 🎯

- **Usage Options** ⚙️: Extend `CurrencyFormatterOptions` with `usage: CurrencyUsage` (`Standard` default, `Accounting`).
- **Constructor Integration** 🔗: Thread options through symbol/code constructors and their compact / compact-long counterparts (plus unstable/buffer variants).
- **Pattern Resolution** 📊: Resolve accounting patterns from `CurrencyEssentials`, falling back to standard sign handling when needed.
- **Sign Handling** ➖: Suppress the decimal minus sign when an accounting negative pattern already encodes the sign.
- **Test Coverage** 🧪: Cover accounting beside existing standard cases in currency tests.

## Test plan ⚙️

- [x] `cargo fmt --all -- --check` 
- [x] `cargo clippy -p icu_experimental --all-features --all-targets` 
- [x] `cargo test -p icu_experimental --lib dimension::currency` 

## Changelog

icu_experimental/currency:

- Extend `CurrencyFormatterOptions` with `usage: CurrencyUsage` (`Standard` default, `Accounting`).

### CODE REVIEW

**Review by robertbastian — COMMENTED** [review]
needs rebase

File: components/experimental/src/dimension/currency/compact_formatter.rs
Comment (robertbastian) [inline components/experimental/src/dimension/currency/compact_formatter.rs:1]: at some point we should move the constructors into the `formatter` module, given that it's the same type now

File: components/experimental/src/dimension/currency/formatter.rs [hunk 1]
```diff
@@ -75,6 +77,7 @@ pub(crate) enum CurrencyFormatterData {
 pub struct CurrencyFormatter<V: AbstractFormatter> {
     value_formatter: V,
     currency_data: CurrencyFormatterData,
+    options: CurrencyFormatterOptions,
```
robertbastian: issue: don't store the full unresolved options bag. resolution should happen in the constructor, and you should only store the things you need

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -640,14 +693,42 @@ impl<V: AbstractFormatter> CurrencyFormatter<V> {
...
+    essentials: &'a super::super::provider::currency::essentials::CurrencyEssentials<'_>,
+    accounting: bool,
+    sign: Sign,
+    symbol_starts_with_letter: bool,
+    symbol_ends_with_letter: bool,
+) -> (&'a icu_pattern::DoublePlaceholderPattern, Sign) {
+    if accounting {
+        if sign == Sign::Negative
+            && let Some(pattern) = essentials
+                .get_negative_accounting(symbol_starts_with_letter, symbol_ends_with_letter)
```
robertbastian, line 1281: issue: you should also return `essentials.get_negative` at some point in this function

components/experimental/src/dimension/currency/compact_formatter.rs
younies: Done in 1ccb35615a: moved all the compact constructors into the `formatter` module and removed `compact_formatter.rs`.

components/experimental/src/dimension/currency/formatter.rs
younies: Done in 1ccb35615a: the constructors now resolve the options bag and the formatter only stores the resolved `usage`. We could resolve even further (pre-select the positive/negative pattern indices in the constructor and drop `usage` entirely, since only the value's sign is unknown until format time) — happy to do that here or as a follow-up if you prefer.

robertbastian (approved):

components/experimental/src/dimension/currency/formatter.rs
younies: Moved this to a separate PR: #8264 implements `get_negative` support on top of main (explicit negative patterns are used without applying the minus sign; sign-bearing negative subpatterns are not baked and fall back to the value formatter's localized minus sign — #8263 tracks rendering pattern sign characters with the localized signs). This PR keeps only the accounting pattern handling.

components/experimental/src/dimension/currency/formatter.rs
robertbastian: currently this would require a `map_project` of the data payload, which is too complicated.

components/experimental/src/dimension/currency/compact_formatter.rs
robertbastian: when I said "at some point" I meant it should be an independent change later. but don't revert it now

younies: okay

robertbastian (approved):

PR #8192, opened 2026-07-07
Add try_writeable_borrow and some impls for it

sffc:
Fixes #7004

I have a call site for this now. This is the correct signature.

## Changelog

writeable: add `TryWriteable::try_writeable_borrow`

PR #8193, opened 2026-07-07
Add borrowing and length hints for LanguageIdentifierDisplayName

### LINKED ISSUE #8181: Implement and test fully-borrowed language identifier display name
https://github.com/unicode-org/icu4x/pull/8132#discussion_r3481143874

If a language identifier resolves to a since data string, its dname Writeable should return a Cow::Borrowed.

sffc:
Fixes #8181

Depends on #8192

## Changelog

icu_pattern: manually implement `writeable_length_hint`


robertbastian (approved):

components/experimental/src/displaynames/single/language.rs
```diff
@@ -464,6 +472,42 @@ impl<'a> TryWriteable for QualifiersWriteable<'a> {
...
+        match self.variants {
+            BorrowedVariants::One(variant) => {
+                length_hint += variant.writeable_length_hint();
+                num_items += 1;
+            }
+            BorrowedVariants::Slice(slice) => {
+                for item in slice.iter() {
+                    length_hint += match item.get() {
+                        Ok(p) => writeable::Writeable::writeable_length_hint(&**p),
+                        Err(var) => writeable::Writeable::writeable_length_hint(var.as_str()),
```
robertbastian, line 496: if you're changing it in the other branch
```suggestion
                        Ok(p) => p.writeable_length_hint(),
                        Err(var) => var.as_str().writeable_length_hint(),
```

robertbastian (approved):

components/experimental/src/displaynames/single/language.rs
sffc: These two are `Writeable`, so they still need the fully qualified call. The change across the merge is that `variant` changes from `Writeable` to `TryWriteable`, which is imported.

Issue #7888, opened 2026-04-20
Potential quadratic behavior in dictionary segmenter


Manishearth:
This is from https://github.com/unicode-org/icu4x/security/advisories/GHSA-c8m9-943c-5xmw, which I mostly addressed but didn't look at this last option.

The bug report contained this:

components/segmenter/src/complex/dictionary.rs:57-73

Nested loop over dictionary matches and grapheme boundaries could theoretically exhibit O(n*m) behavior with adversarial text
combining complex scripts. In practice, grapheme boundaries are well-bounded, but pathological input with many combining characters
could amplify costs.

There are ways to do pathologically large graphemes in Unicode, but most of them will cause this algorithm to consume the grapheme and move on. I don't _think_ there are problems here, but I want to check this.

@sffc @aethanyc @makotokato @hsivonen

Manishearth:
@robertbastian can you confirm if the neo code has this problem?

robertbastian:
Neo uses the same complex segmentation code

Manishearth:
Sounds good. We should at some point investigate if this is real, and fix if so.

PR #8195, opened 2026-07-08
Fix grapheme iterator desync when dictionary segmenter rewinds (#7888)

Manishearth:
Sometimes `DictionaryBreakIterator::next()` does a speculative iterator advance, and then "rewinds" when it fails. The current code rewinds the contained iterator but not the contained grapheme iterator. Seems incorrect to me.

This was incidentally discovered by an LLM while I was trying to analyze this code for #7888.

<details><summary>LLM analysis</summary>
When exploring dictionary matches in `DictionaryBreakIterator::next()`, hitting an Intermediate match advances the `grapheme_iter` to check if the match aligns with a grapheme boundary. If subsequent trie traversal fails with NoMatch, self.iter is rewound to the previous match point, but previously `self.grapheme_iter` and `last_grapheme_offset` were not rewound.

This caused grapheme_iter to be out of sync with iter on subsequent next() calls, potentially rejecting valid grapheme boundaries after a backtrack. We implement Clone manually for RuleBreakIterator and GraphemeClusterBreakIterator (avoiding unnecessary Y: Clone bounds) and save/restore the grapheme iterator state alongside iter on backtrack.

Investigation of #7888 confirms there is no `O(n*m)` quadratic behavior between dictionary trie traversal and grapheme boundary checks: grapheme_iter moves monotonically across iterations and trie backtrack distance is bounded by maximum dictionary word length.

</details>

## Changelog

icu_segmenter:
 - Fix rewinding behavior in dictionary segmenter

### DIFF
--- a/components/segmenter/src/complex/dictionary.rs
+++ b/components/segmenter/src/complex/dictionary.rs
@@ -60,13 +60,14 @@ impl<Y: RuleBreakType> Iterator for DictionaryBreakIterator<'_, '_, Y> {
                     }
 
                     intermediate_length = next.0 + Y::char_len(next.1);
-                    previous_match = Some(self.iter.clone());
+                    previous_match = Some((self.iter.clone(), self.grapheme_iter.clone_internal()));
                 }
                 TrieResult::NoMatch => {
                     if intermediate_length > 0 {
-                        if let Some(previous_match) = previous_match {
+                        if let Some((prev_iter, prev_grapheme_iter)) = previous_match {
                             // Rewind previous match point
-                            self.iter = previous_match;
+                            self.iter = prev_iter;
+                            self.grapheme_iter = prev_grapheme_iter;
                         }
                         return Some(intermediate_length);
                     }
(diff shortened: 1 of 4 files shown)


components/segmenter/src/complex/dictionary.rs
```diff
@@ -60,13 +60,14 @@ impl<Y: RuleBreakType> Iterator for DictionaryBreakIterator<'_, '_, Y> {
                     }
 
                     intermediate_length = next.0 + Y::char_len(next.1);
-                    previous_match = Some(self.iter.clone());
+                    previous_match = Some((self.iter.clone(), self.grapheme_iter.clone_internal()));
```
Manishearth, line 63: I'm not sure if this is a cheap clone. There may be better ways to "rewind" here. Cloning iterators is the common way to do it.

robertbastian: given that you also have `iter` representing where in the input you are, you could have a function that sets the segmenter to that. this way you can do something more clever than cloning the whole thing

Manishearth: I'm kind of worried about the more clever thing being buggier since the grapheme segmenter has a bunch of state

robertbastian: it's the same as creating a new iterator:

https://github.com/unicode-org/icu4x/blob/7f5f30f975f89d7d0bc482b6c3ef5314476634f3/components/segmenter/src/grapheme.rs#L234-L259

Manishearth: A new iterator will be starting at the beginning, though. There's contextual state there that gets set once you start iterating.

robertbastian: at least with neo, you're in the same state after each break point, except for `iter`, which, being `char_indices`, encapsulates the offset. I think this is the case for legacy as well, conceptually you should be in the same state if you remove a prefix grapheme cluster or not

components/segmenter/src/neo/mod.rs
```diff
@@ -134,6 +134,22 @@ pub(crate) struct RuleBreakIterator<'data, 's, Y: RuleBreakType, C: ComplexHandl
     complex: Option<C::ComplexPayloads<'data>>,
 }
 
+impl<'data, 's, Y: RuleBreakType, C: ComplexHandler<Y>> Clone
```
robertbastian: issue: don't make this a general implementation on `RuleBreakIterator` (either of them), but instead inline this into `Clone for GraphemeClusterBreakIteratorInner`. it's easier to reason about this only in the context of grapheme, because we can ignore the complex stuff

components/segmenter/src/grapheme.rs
```diff
@@ -35,6 +35,23 @@ pub(crate) enum GraphemeClusterBreakIteratorInner<'data, 's, Y: RuleBreakType> {
     Neo(crate::neo::RuleBreakIterator<'data, 's, Y, crate::neo::NoComplexHandler>),
 }
 
+impl<'data, 's, Y: RuleBreakType> Clone for GraphemeClusterBreakIteratorInner<'data, 's, Y> {
```
robertbastian: just inline this into `clone_internal`, I don't want a `Clone` implementation floating around if we're only writing it for one specific use case.

components/segmenter/src/rule_segmenter.rs
```diff
@@ -35,7 +35,7 @@ pub trait RuleBreakType: crate::private::Sealed + Sized {
 
     #[doc(hidden)]
     #[cfg(feature = "unstable")]
-    type ComplexPayloads<'data>: core::fmt::Debug;
+    type ComplexPayloads<'data>: core::fmt::Debug + Clone;
```
robertbastian: this is a borrowed type so `Clone` should come with `Copy`
```suggestion
    type ComplexPayloads<'data>: core::fmt::Debug + Clone + Copy;
```

robertbastian (approved):

PR #8197, opened 2026-07-08
Add RuleBreakIterator::replace_input

robertbastian:
#8195 

## Changelog

N/A

### DIFF
--- a/components/segmenter/src/neo/mod.rs
+++ b/components/segmenter/src/neo/mod.rs
@@ -172,6 +172,21 @@ impl<'data, 's, Y: RuleBreakType, C: ComplexHandler<Y>> RuleBreakIterator<'data,
         }
     }
 
+    pub(crate) fn replace_input<'s2>(
+        self,
+        input: Y::IterAttr<'s2>,
+    ) -> RuleBreakIterator<'data, 's2, Y, C> {
+        RuleBreakIterator {
+            data: self.data,
+            pseudo_symbol_map: self.pseudo_symbol_map,
+            complex: self.complex,
+            cache: SmallVec::new().into_iter(),
+            lookahead_positions: SmallVec::from_elem(None, self.data.num_lookaheads),
+            last_accepting_status: 0,
+            remaining_input: input,
+        }
+    }
+
     pub(crate) fn last_accepting_status(&self) -> u8 {
         self.last_accepting_status
     }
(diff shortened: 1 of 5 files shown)


components/segmenter/src/neo/mod.rs
```diff
@@ -172,6 +172,21 @@ impl<'data, 's, Y: RuleBreakType, C: ComplexHandler<Y>> RuleBreakIterator<'data,
...
+    pub(crate) fn replace_input<'s2>(
+        self,
+        input: Y::IterAttr<'s2>,
+    ) -> RuleBreakIterator<'data, 's2, Y, C> {
+        RuleBreakIterator {
+            data: self.data,
+            pseudo_symbol_map: self.pseudo_symbol_map,
+            complex: self.complex,
+            cache: SmallVec::new().into_iter(),
+            lookahead_positions: SmallVec::from_elem(None, self.data.num_lookaheads),
```
Manishearth, line 184: I don't trust this, `lookahead_positions` and `last_accepting_status` appear to be stateful.

robertbastian: `lookahead_positions` gets cleared at the beginning of `next`, it's just not local to that function to avoid potentially reallocating it for every `next` call (in case the number of lookaheads exceeds the SmallVec).

`last_accepting_status` is stateful, yes, but all this means is that you get the wrong status if you call the status API (i.e. `LineBreak`'s `is_mandatory` and `WordBreaks` `word_type`). I guess this is worth documenting, but doesn't affect your use case.

Manishearth: Okay, please document  this behavior on this function, and for all stateful fields document why it's okay to reset them.

Manishearth: I still generally think this is more brittle than just cloning, though. Is cloning expensive?

robertbastian: there are two SmallVecs, but for grapheme they are always empty. for line/word, cloning might allocate

Manishearth (changes requested):
.

components/segmenter/src/neo/mod.rs
Manishearth: https://github.com/unicode-org/icu4x/pull/8195 doesn't need line/word, just grapheme. I think it's a good idea to document there that the clone is cheap (if it indeed is) though.

Issue #8184, opened 2026-07-07
currency: Revisit compact currency formatting logic after CLDR TR35 investigation regarding compact patterns vs gluing


younies:
### Background
In our initial migration of compact currency formatting to `CurrencyFormatter<Compact>` (see follow-up to #8145), we implemented a simplified "gluing" approach in `format_fixed_decimal`:
1. We resolve the standard currency pattern (or `alpha_next_to_number` pattern) from `CurrencyEssentialsV1` via `name_and_pattern`.
2. We format the numeric magnitude using the compact decimal formatter (`DecimalCompactShortV1`), producing e.g. `"12K"` or `"12万"`.
3. We interpolate the compact number and the currency symbol into that standard currency pattern (gluing them together).

Currently, CLDR also defines separate double-placeholder compact currency data (`ShortCurrencyCompactV1` / `currency_patterns`, such as `"{1}{0}K"` or `"{0} k {1}"`). 

### Ongoing CLDR Investigation
There is an active investigation and issue opened against CLDR by @younies evaluating whether the simpler "gluing" approach (combining standard/alpha-next-to-number currency patterns with decimal compact output) should be formally endorsed or standardized by TR35, rather than requiring separate double-placeholder compact currency data tables.

### Action Items / Next Steps
Once the CLDR investigation concludes and a consensus is reached (whether approved, rejected, or a hybrid approach):
- [ ] Revisit `CurrencyFormatter<Compact>::format_fixed_decimal` in `compact_formatter.rs` and adjust the formatting logic accordingly.
- [ ] **If CLDR approves the gluing approach**: Formally remove or deprecate `ShortCurrencyCompactV1` data requirements from `CompactData` to reduce data bundle size and simplify provider loading.
- [ ] **If CLDR requires explicit double-placeholder compact patterns**: Implement pattern resolution from `ShortCurrencyCompactV1` with proper fallback to decimal compact formatting (addressing the TODO in `compact_formatter.rs`).

PR #8198, opened 2026-07-08
fix(currency): evaluate plural rules against formatted amount

robertbastian:
Currently the currency formatter uses the input `Decimal` to determine the plural category. This is wrong, it needs to use the formatted `FormattedUnsignedDecimal`/`FormattedUnsignedCompactDecimal`. This is because these types might have a different plural category than the `Decimal` they're representing. For example in `fr`, `999_999` has category `other` ("dollars US"), but "1.0 million" has category `many` ("de dollars US").

## Changelog

`icu_plurals`:
* add `PluralElements::get`

### DIFF
--- a/components/plurals/src/lib.rs
+++ b/components/plurals/src/lib.rs
@@ -1080,6 +1080,20 @@ impl<T> PluralElements<T> {
     }
 
     /// Returns the value for the given [`PluralOperands`] and [`PluralRules`].
+    ///
+    /// # Example
+    /// ```
+    /// use icu::locale::locale;
+    /// use icu::plurals::{PluralCategory, PluralRules, PluralElements};
+    ///
+    /// let rules = PluralRules::try_new_cardinal(locale!("fr").into()).unwrap();
+    ///
+    /// let elements = PluralElements::new("chats").with_one_value(Some("chat"));
+    ///
+    /// assert_eq!(*elements.get(0_usize.into(), &rules), "chat");
+    /// assert_eq!(*elements.get(1_usize.into(), &rules), "chat");
+    /// assert_eq!(*elements.get(12_usize.into(), &rules), "chats");
+    /// ```
     pub fn get<'a>(&'a self, op: PluralOperands, rules: &PluralRules) -> &'a T {
         let category = rules.category_for(op);
 
--- a/components/plurals/src/lib.rs
+++ b/components/plurals/src/lib.rs
@@ -1078,6 +1078,32 @@ impl<T> PluralElements<T> {
     pub fn as_ref(&self) -> PluralElements<&T> {
         PluralElements(self.0.as_ref())
     }
+
+    /// Returns the value for the given [`PluralOperands`] and [`PluralRules`].
+    pub fn get<'a>(&'a self, op: PluralOperands, rules: &PluralRules) -> &'a T {
+        let category = rules.category_for(op);
+
+        if op.is_exactly_zero()
+            && let Some(value) = self.0.explicit_zero.as_ref()
+        {
+            return value;
+        }
+
+        if op.is_exactly_one()
+            && let Some(value) = self.0.explicit_one.as_ref()
+        {
+            return value;
+        }
+
+        match category {
+            PluralCategory::Zero => self.zero(),
+            PluralCategory::One => self.one(),
+            PluralCategory::Two => self.two(),
+            PluralCategory::Few => self.few(),
+            PluralCategory::Many => self.many(),
+            PluralCategory::Other => self.other(),
+        }
+    }
 }
 
 impl<T: PartialEq> PluralElements<T> {
--- a/components/plurals/src/lib.rs
+++ b/components/plurals/src/lib.rs
@@ -1110,13 +1110,14 @@ impl<T> PluralElements<T> {
         }
 
         match category {
-            PluralCategory::Zero => self.zero(),
-            PluralCategory::One => self.one(),
-            PluralCategory::Two => self.two(),
-            PluralCategory::Few => self.few(),
-            PluralCategory::Many => self.many(),
-            PluralCategory::Other => self.other(),
+            PluralCategory::Zero => self.0.zero.as_ref(),
+            PluralCategory::One => self.0.one.as_ref(),
+            PluralCategory::Two => self.0.two.as_ref(),
+            PluralCategory::Few => self.0.few.as_ref(),
+            PluralCategory::Many => self.0.many.as_ref(),
+            PluralCategory::Other => return &self.0.other,
         }
+        .unwrap_or(&self.0.other)
     }
 }
 
--- a/components/plurals/src/operands.rs
+++ b/components/plurals/src/operands.rs
@@ -258,6 +258,7 @@ impl From<u128> for PluralOperands {
 }
 
 impl PluralOperands {
+    #[doc(hidden)] // todo: figure out whether to expose this
     /// Creates a [`PluralOperands`] from a significand and power of ten.
     pub fn from_significand_and_exponent(dec: &UnsignedDecimal, exp: u8) -> PluralOperands {
         let exp_i16 = i16::from(exp);
--- a/components/plurals/src/operands.rs
+++ b/components/plurals/src/operands.rs
@@ -258,7 +258,8 @@ impl From<u128> for PluralOperands {
 }
 
 impl PluralOperands {
-    fn from_significand_and_exponent(dec: &UnsignedDecimal, exp: u8) -> PluralOperands {
+    /// Creates a [`PluralOperands`] from a significand and power of ten.
+    pub fn from_significand_and_exponent(dec: &UnsignedDecimal, exp: u8) -> PluralOperands {
         let exp_i16 = i16::from(exp);
 
         let mag_range = dec.magnitude_range();
--- a/components/decimal/src/abstract_formatter.rs
+++ b/components/decimal/src/abstract_formatter.rs
@@ -3,6 +3,7 @@
 // (online at: https://github.com/unicode-org/icu4x/blob/main/LICENSE ).
 
 use fixed_decimal::UnsignedDecimal;
+use icu_plurals::PluralOperands;
 use writeable::Writeable;
 
 use crate::{
@@ -15,6 +16,11 @@ pub trait Sealed {}
 /// A trait representing an abstract number formatter.
 ///
 /// This is a building block for more complicated formatters, like currency or units.
+///
+/// <div class="stab unstable">
+/// 🚫 This trait is sealed; it cannot be implemented by user code. If an API requests an item that implements this
+/// trait, please consider using a type from the implementors listed below.
+/// </div>
 pub trait AbstractFormatter: core::fmt::Debug + Sealed {
     #[doc(hidden)]
     type FormattedUnsigned<'a>: Writeable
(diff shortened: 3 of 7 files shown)


younies (approved):
Thanks for the PR, @robertbastian! This looks great and solves a subtle grammatical issue with compact currencies. LGTM! 🚀

I pushed a follow-up commit (397bf928d3) to your branch to update the Diplomat FFI coverage allowlist (`missing_apis.txt` for `PluralOperands::from_significand_and_exponent`), which resolves the failing CI check. I also tweaked the PR title and description to add changelog details and full context—feel free to adjust them if needed!

robertbastian:
The changelog you generated is way too verbose, and is not the right format either. Given this is all very much in flux, we should write changelog entries for this at the end.

components/plurals/src/operands.rs
```diff
@@ -258,7 +258,8 @@ impl From<u128> for PluralOperands {
 }
 
 impl PluralOperands {
-    fn from_significand_and_exponent(dec: &UnsignedDecimal, exp: u8) -> PluralOperands {
+    /// Creates a [`PluralOperands`] from a significand and power of ten.
+    pub fn from_significand_and_exponent(dec: &UnsignedDecimal, exp: u8) -> PluralOperands {
```
sffc, line 263: Nit:

```suggestion
    pub fn from_significand_and_compact_exponent(dec: &UnsignedDecimal, exp: u8) -> PluralOperands {
```

Please also put this in the changelog

sffc (changes requested):

sffc:
> Given this is all very much in flux, we should write changelog entries for this at the end.

_Especially_ for a stable function in a stable crate, _please_ fill in the changelog entries in the description of the PR that exports the function. I am policing the release this time and I am going to be relying on these entries.

components/plurals/src/lib.rs
```diff
@@ -1078,6 +1078,32 @@ impl<T> PluralElements<T> {
...
+            return value;
+        }
+
+        match category {
+            PluralCategory::Zero => self.zero(),
+            PluralCategory::One => self.one(),
+            PluralCategory::Two => self.two(),
+            PluralCategory::Few => self.few(),
+            PluralCategory::Many => self.many(),
+            PluralCategory::Other => self.other(),
```
sffc: Suggestion: the match should return an `Option<&T>` by directly accessing the fields, and then you can have a single `unwrap_or(&self.0.other)` at the end. It is _probably_ better than having 6 function calls, which even when inlined call `unwrap_or` each time.

sffc (changes requested):

components/plurals/src/lib.rs
```diff
@@ -1078,6 +1078,32 @@ impl<T> PluralElements<T> {
     pub fn as_ref(&self) -> PluralElements<&T> {
         PluralElements(self.0.as_ref())
     }
+
+    /// Returns the value for the given [`PluralOperands`] and [`PluralRules`].
+    pub fn get<'a>(&'a self, op: PluralOperands, rules: &PluralRules) -> &'a T {
```
sffc, line 1097: Huh, did we not actually have a public getter for this before?

We have lots of data provider impls, and a getter on PluralElementsPackedULE.

Please add this to the changelog and please add a docstest. I suggest adding the docstest to the PluralElements struct.

younies (dismissed):
I noticed an important architectural issue with our current constructor naming and behavior for compact currency formatting. 

Right now, we are conflating two orthogonal formatting dimensions into a single name:
1. **The Number Style:** Compact / compact-short (`1.2K`) vs. Verbose / compact-long (`1.2 thousand`)
2. **The Currency Style:** Short symbol (`$`), Narrow symbol (`$`), vs. Long display name (`US dollars`)

### Why this is currently so difficult and confusing
Because we use the terms "short" and "long" for *both* dimensions (i.e., short/long compact numbers and short/long currency names), a constructor named `try_new_compact_long` is ambiguous—does "long" refer to the number or the currency? Currently, `try_new_compact_long` formats *both* as long, producing `"1.2 thousand US dollars"`, leaving users with no way to get `"1.2K US dollars"`.

In fact, in my original [design doc](https://hackmd.io/@younies/number_formatter_4x), I specifically planned to introduce a `struct Verbose` alongside `struct Decimal` and `struct Compact` as a distinct `ValueRepresentation` marker to prevent this exact ambiguity:

```rust
/// Representation for compact long formatting (verbose notation)
/// (planned for when CLDR adds verbose compact currency support).
pub struct Verbose;
impl ValueRepresentation for Verbose {}
```

To eliminate this confusion, I propose using the wording **`verbose`** for compact-long numbers (`1.2 thousand`) and reserving **`compact`** strictly for compact-short numbers (`1.2K`). This allows us to cleanly distinguish between the two number styles without clashing with currency width names.

### Proposed 6-Function Architecture (2x3 Matrix)
In this architecture, the prefix (`compact` vs `verbose`) controls the **number**, and the suffix (`_short`, `_narrow`, `_long`) controls the **currency width**:

| Constructor | Number Style | Currency Style | Example Output (`en-US`, `USD`) | Status |
| :--- | :--- | :--- | :--- | :--- |
| **`try_new_compact_short`** | Compact (`1.2K`) | Short (`$`) | `"$1.2K"` | Already correct |
| **`try_new_compact_narrow`** | Compact (`1.2K`) | Narrow (`$`) | `"$1.2K"` | Already correct |
| **`try_new_compact_long`** | Compact (`1.2K`) | Long (`US dollars`) | `"1.2K US dollars"` | **Needs fix:** change inner formatter to `try_new_short` |
| **`try_new_verbose_short`** | Verbose (`1.2 thousand`) | Short (`$`) | `"$1.2 thousand"` | Add / track under #8184 |
| **`try_new_verbose_narrow`** | Verbose (`1.2 thousand`) | Narrow (`$`) | `"$1.2 thousand"` | Add / track under #8184 |
| **`try_new_verbose_long`** | Verbose (`1.2 thousand`) | Long (`US dollars`) | `"1.2 thousand US dollars"` | This is what `compact_long` currently does $\rightarrow$ rename to `verbose_long` |

### Action Items & Requested TODOs
1. **Fix `try_new_compact_long`:** Update line 147 to pass `CompactDecimalFormatter::try_new_short` for the number formatting so that it outputs `"1.2K US dollars"`.
2. **Add TODO(#8184) to `try_new_compact_*`:** Mark `try_new_compact_short`, `try_new_compact_narrow`, and `try_new_compact_long` with a TODO noting that the gluing architecture is under evaluation in [CLDR-19617](https://unicode-org.atlassian.net/browse/CLDR-19617).
3. **Add TODO(#8184) to `try_new_verbose_*`:** Rename the current written-number behavior to `try_new_verbose_long` (and add/track `verbose_short` and `verbose_narrow`), marking them with a TODO noting that verbose/written compact currency formatting is not currently defined in CLDR TR35.

---

### Suggested Code Snippets for `try_new_compact_*`

```rust
    // TODO(#8184): The gluing approach used here is currently under evaluation.
    // Revisit this implementation after an approved resolution in CLDR-19617 or when ICU4X
    // formally decides to standardize on the gluing architecture.
    pub fn try_new_compact_short(
```
```rust
    // TODO(#8184): The gluing approach used here is currently under evaluation.
    // Revisit this implementation after an approved resolution in CLDR-19617 or when ICU4X
    // formally decides to standardize on the gluing architecture.
    pub fn try_new_compact_narrow(
```
```rust
    // TODO(#8184): The gluing approach used here is currently under evaluation.
    // Revisit this implementation after an approved resolution in CLDR-19617 or when ICU4X
    // formally decides to standardize on the gluing architecture.
    pub fn try_new_compact_long(
```

### Suggested Code Snippets for `try_new_verbose_*`

```rust
    // TODO(#8184): Verbose (compact long) currency formatting is not currently defined in CLDR TR35.
    // Revisit this implementation after an approved resolution in CLDR-19617 or when ICU4X
    // formally decides to standardize on the gluing architecture.
    pub fn try_new_verbose_short(
```
```rust
    // TODO(#8184): Verbose (compact long) currency formatting is not currently defined in CLDR TR35.
    // Revisit this implementation after an approved resolution in CLDR-19617 or when ICU4X
    // formally decides to standardize on the gluing architecture.
    pub fn try_new_verbose_narrow(
```
```rust
    // TODO(#8184): Verbose (compact long) currency formatting is not currently defined in CLDR TR35.
    // Revisit this implementation after an approved resolution in CLDR-19617 or when ICU4X
    // formally decides to standardize on the gluing architecture.
    pub fn try_new_verbose_long(
```


Issue to track this: https://github.com/unicode-org/icu4x/issues/8206

components/plurals/src/operands.rs
```diff
@@ -258,7 +258,9 @@ impl From<u128> for PluralOperands {
 }
 
 impl PluralOperands {
-    fn from_significand_and_exponent(dec: &UnsignedDecimal, exp: u8) -> PluralOperands {
+    #[doc(hidden)] // todo: figure out whether to expose this
```
sffc, line 261: Suggestion (optional): add a non-hidden public version of this function behind an unstable feature

sffc (approved):

sffc:
@younies Your comment is not related to this current PR. But, to respond to it, I think "short" and "long" should refer to the compact notation length, since that is the terminology in both CLDR and ECMA. To refer to a currency long name, consider using something like "name", which is what ECMA uses.

younies:
@sffc Thanks for the feedback! While I agree with the ECMA/CLDR terminology preference (`short`/`long` for number notation vs. `symbol`/`name` for currency width) and have documented that in **[Issue #8206](https://github.com/unicode-org/icu4x/issues/8206)** to discuss in our next ICU4X meeting, I respectfully disagree that this is unrelated to the current PR.

Here is why it is directly relevant to PR #8198:
1. **The problem being solved here stems from a previous architectural conflation:** The entire motivating example of this PR (French `"1.0 million de dollars US"` requiring the `many` plural category) only exists because a previous PR implemented `try_new_compact_long` by calling `CompactDecimalFormatter::try_new_long(...)`. This caused `try_new_compact_long` to apply "long" to **both the number and the currency simultaneously** (`"12 thousand US dollars"`).
2. **Missing Specification:** Combining compact-long numbers with long currency display names is currently not defined anywhere in the CLDR TR35 specification. If `try_new_compact_long` had originally been implemented as compact-short number + long currency name (`"1.2K US dollars"`), this specific plural category divergence would not be occurring in the same way.

### Conclusion & Request for this PR
Robert's fix for evaluating plural operands against the formatted output (`plural_operands`) is fantastic and absolutely necessary for grammatical correctness whenever compact-long/verbose notation is used. I am 100% in favor of approving and merging this PR!

However, before we merge, we should **add a clear doc comment and a BIG TODO(#8206) to `try_new_compact_long` in this PR**, explicitly noting that:
- Currently, `try_new_compact_long` applies "long" to *both* the compact number and the currency display name at the same time.
- This dual-long behavior is not in CLDR TR35 and will be revisited/decoupled as part of #8206 following our discussions in the next ICU4X meeting.

---

### Suggested Code Snippet for `try_new_compact_long`

```rust
    // TODO(#8206, #8184): Currently, `try_new_compact_long` applies "long" to BOTH the number
    // notation ("12 thousand") and the currency display name ("US dollars") simultaneously.
    // Combining compact-long numbers with currency names is not currently defined in CLDR TR35.
    // Revisit and decouple this constructor into orthogonal number/currency dimensions as part of
    // issue #8206 following ICU4X architectural discussions.
    pub fn try_new_compact_long(
```

robertbastian:
> 1. **The problem being solved here stems from a previous architectural conflation:** The entire motivating example of this PR (French `"1.0 million de dollars US"` requiring the `many` plural category) only exists because a previous PR implemented `try_new_compact_long` by calling `CompactDecimalFormatter::try_new_long(...)`. This caused `try_new_compact_long` to apply "long" to **both the number and the currency simultaneously** (`"12 thousand US dollars"`).

This is not true. This solves a fundamental problem of what is being pluralized, which for all we know is _not_ unique to long compact formats. So while the long compact format might have been incorrectly added in the past, this PR applies to all formats.

PR #8205, opened 2026-07-08
Implement menu names for language identifiers

sffc:
#7825 

I attempted this with an LLM, #8204, but was unhappy with the result so I redid it without the LLM.

This is reviewable commit-by-commit.

## Changelog

icu_experimental::displaynames:

- New scaffolding trait `single::models::LanguageIdentifierDisplayNameModel`
- New marker type `single::models::Standard`
- New marker type `single::models::Menu`
- New constructors `LanguageIdentifierDisplayNameOwned::try_new_menu[_unstable|_with_buffer_provider]`

### DIFF
--- a/components/experimental/src/displaynames/provider.rs
+++ b/components/experimental/src/displaynames/provider.rs
@@ -160,6 +160,7 @@ pub struct VariantDisplayNames<'data> {
 #[cfg_attr(feature = "datagen", derive(serde::Serialize, databake::Bake))]
 #[cfg_attr(feature = "datagen", databake(path = icu_experimental::displaynames::provider))]
 #[zerovec::make_varule(MenuNamePartsULE)]
+#[zerovec::derive(Debug)]
 #[zerovec::skip_derive(Ord)]
 #[cfg_attr(feature = "serde", zerovec::derive(Deserialize))]
 #[cfg_attr(feature = "datagen", zerovec::derive(Serialize))]
@@ -281,6 +282,25 @@ icu_provider::data_marker!(
     LocaleNamesEssentials<'static>
 );
 
+/// A data struct that is either [`MenuNameParts`] or a string
+#[derive(Debug, PartialEq, Clone, yoke::Yokeable, zerofrom::ZeroFrom)]
+#[allow(clippy::exhaustive_enums)] // provider-unstable
+pub enum MenuNamePartsOrString<'a> {
+    /// A data struct that is [`MenuNameParts`]
+    MenuNameParts(VarZeroCow<'a, MenuNamePartsULE>),
+    /// A data struct that is a string
+    String(VarZeroCow<'a, str>),
+}
+
+/// A marker for [`MenuNamePartsOrString`]
+#[derive(Debug)]
+#[allow(clippy::exhaustive_structs)] // marker
+pub struct MenuNamePartsOrStringErased;
+
+impl DynamicDataMarker for MenuNamePartsOrStringErased {
+    type DataStruct = MenuNamePartsOrString<'static>;
+}
+
 impl LocaleNamesLanguageMediumV1 {
     /// Helper to construct infallible attributes from subtags.
     #[inline]
@@ -318,8 +338,21 @@ impl LocaleNamesLanguageMediumV1 {
     }
 }
 
+impl LocaleNamesLanguageMenuMediumV1 {
+    /// Helper to create data marker attributes from subtags.
+    #[inline]
+    pub(crate) fn make_attributes(
+        language: Language,
+        script: Option<Script>,
+        region: Option<Region>,
+        buffer: &mut tinystr::TinyAsciiStr<16>,
+    ) -> &DataMarkerAttributes {
+        LocaleNamesLanguageMediumV1::make_attributes(language, script, region, buffer)
+    }
+}
+
 impl LocaleNamesRegionMediumV1 {
-    /// Helper to create data marker attributes from a region.
+    /// Helper to construct infallible attributes from subtags.
     #[inline]
     pub(crate) fn make_attributes(region: &Region) -> &DataMarkerAttributes {
         // This is infallible (will not panic) because a validated `Region` is guaranteed to
--- a/components/experimental/src/displaynames/single/language.rs
+++ b/components/experimental/src/displaynames/single/language.rs
@@ -40,7 +40,9 @@ pub mod models {
     /// </div>
     pub trait LanguageIdentifierDisplayNameModel: private::Sealed {
         /// The payload storage type for this model.
-        type LanguagePayloadUnstable: 'static + core::fmt::Debug;
+        type LanguageMarkerUnstable: DynamicDataMarker<
+            DataStruct: for<'a> yoke::Yokeable<'a, Output: core::fmt::Debug>,
+        >;
     }
 
     /// Model for standard and dialect language display names.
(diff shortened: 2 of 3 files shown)


sffc:
I pulled out the first refactoring commit into #8207 if you prefer to review that first, but this PR is reviewable commit by commit.

components/experimental/src/displaynames/single/language.rs
```diff
@@ -25,8 +25,49 @@ use writeable::{PartsWrite, TryWriteable, adapters::LossyWrap};
...
+
+    /// A trait defining the data model for a language display name (Standard vs. Menu).
+    ///
+    /// <div class="stab unstable">
+    /// 🚫 This trait is sealed; it cannot be implemented by user code. If an API requests an item that implements this
+    /// trait, please consider using a type from the implementors listed below.
+    /// </div>
+    pub trait LanguageIdentifierDisplayNameModel: private::Sealed {
+        /// The payload storage type for this model.
+        type LanguageMarkerUnstable: DynamicDataMarker<
```
robertbastian, line 43: #[doc(hidden)]

components/experimental/src/displaynames/single/language.rs
```diff
@@ -296,20 +342,23 @@ impl LanguageIdentifierDisplayNameOwned {
         // Step 5: Load essentials
         let essentials_payload = provider
             .load(DataRequest {
-                id: DataIdentifierBorrowed::for_locale(&formatting_locale),
+                id: DataIdentifierBorrowed::for_locale(&LocaleNamesEssentialsV1::make_locale(
```
robertbastian, line 345: why aren't you passing `formatting_locale` into this function?

components/experimental/src/displaynames/single/language.rs
```diff
@@ -180,7 +221,96 @@ impl LanguageIdentifierDisplayNameOwned {
...
+    {
+        let formatting_locale = LocaleNamesLanguageMediumV1::make_locale(prefs.locale_preferences);
+
+        // Step 1: Load language name
+        // Try the menu name
+        let language_payload =
+            Self::load_language_menu_name(provider, &formatting_locale, subject.language)?;
+
+        // If the language name is not loaded yet, try loading it from the language subtag alone.
+        let language_payload = match language_payload {
```
robertbastian, line 283: why isn't this done as part of `load_language_menu_name`? or why isn't `load_language_menu_name` inlined? both are menu-specific code afaict

components/experimental/src/displaynames/single/language.rs
```diff
@@ -374,6 +527,27 @@ impl LanguageIdentifierDisplayNameOwned {
...
+            Ok(MenuNamePartsOrString::String(string)) => NameOrFallback(Ok(string.as_ref())),
+            Err(lang) => NameOrFallback(Err(lang.as_str())),
+        };
+
+        LanguageIdentifierDisplayName(LossyWrap(LanguageIdentifierDisplayNameInner {
+            base_name,
+            qualifiers,
+        }))
+    }
+}
```
robertbastian, line 549: can you make this generic instead of having two `as_borrowed` functions, which will both show up in docs? just move the match statement into the trait

components/experimental/src/displaynames/provider.rs
```diff
@@ -281,6 +282,25 @@ icu_provider::data_marker!(
...
+    /// A data struct that is [`MenuNameParts`]
+    MenuNameParts(VarZeroCow<'a, MenuNamePartsULE>),
+    /// A data struct that is a string
+    String(VarZeroCow<'a, str>),
+}
+
+/// A marker for [`MenuNamePartsOrString`]
+#[derive(Debug)]
+#[allow(clippy::exhaustive_structs)] // marker
+pub struct MenuNamePartsOrStringErased;
```
robertbastian, line 298: this is `ErasedDataMarker<MenuNamePartsOrString<'static>>`

robertbastian, line 298: does this need to be public?

components/experimental/src/displaynames/single/language.rs
```diff
@@ -534,7 +534,9 @@ impl LanguageIdentifierDisplayNameOwned<models::Menu> {
         let mut qualifiers = self.qualifiers.as_borrowed();
         let base_name = match self.language_payload.get() {
             Ok(MenuNamePartsOrString::MenuNameParts(parts)) => {
-                qualifiers.menu_extension = Some(parts.extension());
+                if !parts.extension().is_empty() {
+                    qualifiers.menu_extension = Some(parts.extension());
```
robertbastian, line 538: issue: add documentation to the `extension` field that it might be empty. the current docs are not sufficient (given that you stumbled over this)

components/experimental/tests/displaynames/tests.rs
```diff
@@ -372,10 +386,22 @@ fn test_concatenate() {
...
         }
         if matches!(cas.display_type, DisplayType::Any | DisplayType::Menu) {
+            let dname_menu_owned = LanguageIdentifierDisplayNameOwned::try_new_menu(
+                locale.clone().into(),
+                lang_id,
+                single_options,
+            )
+            .unwrap();
             let borrowed = dname_menu_owned.as_borrowed();
             // "Hindi (Latin)" is a literal string in data,
```
robertbastian, line 406: why is that? even if CLDR says `hi-Latn`: "Hindi (Latin)", that looks like redundant data to me and we shouldn't store it.

robertbastian (changes requested):
I don't think I agree with using the generic-model pattern here. You're converting to an erased `MenuNamePartsOrString` in the `as_borrowed` function, and the borrowed type isn't generic. so why not just do that in the constructors?

components/experimental/tests/displaynames/tests.rs
sffc: agreed, good follow-up to remove the redundant data

components/experimental/src/displaynames/single/language.rs
sffc: I thought you would prefer implementing concrete functions on concrete types over adding more trait functions, given the design of CurrencyFormatter and TimeZoneName

sffc:
> I don't think I agree with using the generic-model pattern here. You're converting to an erased `MenuNamePartsOrString` in the `as_borrowed` function, and the borrowed type isn't generic. so why not just do that in the constructors?

I thought you would prefer using this model since it avoids adding the enum branch and extra stack data to the common non-menu case, and it is the model you advocated for in CurrencyFormatter and TimeZoneInfo.

Also, this design is what was in the README.md spec, which Manish approved, and which you didn't explicitly review. The point of the spec is twofold: to help AI-assisted coding, and to agree on architecture ahead of time to avoid wasting effort on PRs.

But, a design where we just shove everything in the same LanguageDisplayNameOwned type would be simpler. It sounds like you are okay with that design, even though it (slightly) reduces the performance of the non-menu case.

N.B. `MenuNamePartsOrString` is not processed in `as_borrowed`. It is stored in `LanguageDisplayNameOwned<Menu>`.

sffc:
I implemented the non-generic solution in https://github.com/unicode-org/icu4x/pull/8216

components/experimental/src/displaynames/single/language.rs
robertbastian: not if they have a name collision. CurrencyFormatter does not have a `format` function per represenetation

robertbastian:
> it is the model you advocated for in CurrencyFormatter and TimeZoneInfo.

In Currency Formatter I advocated the generic model because there is significant _code and data_ that we want to pull in generically: compact vs decimal formatter. This is really not the case here, having an extra enum to store the menu extensions is a small cost to pay to avoid the API complexity that comes with the generic pattern.

> The point of the spec is twofold: to help AI-assisted coding, and to agree on architecture ahead of time to avoid wasting effort on PRs.

And even if I had reviewed it I might not have realised that we don't actually need the generic model. I stand by my decision to review code, not AI instructions.

PR #8207, opened 2026-07-09
Refactor language display names with QualifiersOwned

sffc:
Split from #8205

I pulled out the first commit to make it easier to review. However, if you also approve #8205, I will just merge that all at once.

## Changelog

N/A

### DIFF
--- a/components/experimental/src/displaynames/single/language.rs
+++ b/components/experimental/src/displaynames/single/language.rs
@@ -189,7 +189,7 @@ impl LanguageIdentifierDisplayNameOwned {
     /// - Language + Script (e.g., "zh-Hant")
     /// - Language + Region (e.g., "en-GB")
     ///
-    /// We then "consume"  the corresponding subtags from the input LanguageIdentifier
+    /// We then "consume"  the corresponding subtags from the input `LanguageIdentifier`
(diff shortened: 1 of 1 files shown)


components/experimental/src/displaynames/single/language.rs
```diff
@@ -296,20 +342,23 @@ impl LanguageIdentifierDisplayNameOwned {
         // Step 5: Load essentials
         let essentials_payload = provider
             .load(DataRequest {
-                id: DataIdentifierBorrowed::for_locale(&formatting_locale),
+                id: DataIdentifierBorrowed::for_locale(&LocaleNamesEssentialsV1::make_locale(
```
robertbastian, line 345: my only question on the other PR was: why not use the same `formatting_locale` as for the other markers? what if the data is inconsistent? should these be kept in sync somehow?

sffc: I had a slight preference to use each marker's own `make_locale` instead of assuming that they all have the same fallback strategy.

`load_language_dialect_name` takes a `formatting_locale` because all of the loads are from the same marker, `LocaleNamesLanguageMediumV1`

robertbastian (approved):

PR #8210, opened 2026-07-09
Adopt RSCD naming

### LINKED ISSUE #7510: Finalize name of ucd-tag/unihan-root CLI args

robertbastian:
Fixes #7510

## Changelog

`icu_provider_source`
    - New items: `SourceDataProvider::with_unicode_rscd_for_tag()`, `SourceDataProvider::with_unicode_rscd()`, `SourceDataProvider::TESTED_UNICODE_TAG`, `SourceDataProvider::is_missing_rscd_error()`,
    - Deprecated `SourceDataProvider::with_ucd_for_tag()`, `SourceDataProvider::with_unihan_for_tag()`, `SourceDataProvider::with_ucd()`, `SourceDataProvider::with_unihan()`, `SourceDataProvider::TESTED_UCD_TAG`, , `SourceDataProvider::is_missing_ucd_error()`


provider/icu4x-datagen/src/main.rs
```diff
@@ -173,25 +173,29 @@ struct Cli {
     #[cfg(feature = "provider")]
     icuexport_root: Option<PathBuf>,
 
-    #[arg(long, value_name = "TAG", default_value = "17.0.0")]
+    #[arg(long, value_name = "TAG", default_value = "latest", alias = "ucd-tag")]
     #[arg(
-        help = "Download versioned UCD from unicode.org (`https://www.unicode.org/Public/{tag}/`). \
+        help = "Download versioned RCSD from unicode.org (`https://www.unicode.org/Public/{tag}/`). \
```
eggrobin:
```suggestion
        help = "Download versioned RSCD from unicode.org (`https://www.unicode.org/Public/{tag}/`). \
```

provider/icu4x-datagen/src/main.rs
```diff
@@ -462,8 +466,10 @@ fn run(cli: Cli) -> eyre::Result<()> {
             eyre::bail!(
                 "Segmentation LSTM data is required for this invocation, set --segmenter-lstm-root or --segmenter-lstm-tag"
             );
-        } else if SourceDataProvider::is_missing_ucd_error(e) {
-            eyre::bail!("UCD data is required for this invocation, set --ucd-root or --ucd-tag");
+        } else if SourceDataProvider::is_missing_rscd_error(e) {
+            eyre::bail!(
+                "RSCD data is required for this invocation, set --unicode-rscd-root or --unicode-tag"
```
eggrobin, line 471: ATM machine (in UCD the D stands for Database, but here it just stands for Data).

Might be clearest to expand (Repertoire-Synchronized Character Data is required…)

eggrobin (commented):
(Alias=whatchamacallit)

provider/icu4x-datagen/src/main.rs
robertbastian: hm, RSC data?

eggrobin: That’s not really a thing you can do to acronyms (AT machine? A file in PD format?)…

robertbastian: I'm comfortable with saying PIN number here

provider/icu4x-datagen/src/main.rs
```diff
@@ -173,25 +173,29 @@ struct Cli {
...
+        help = "Download versioned RSCD from unicode.org (`https://www.unicode.org/Public/{tag}/`). \
                   Use 'latest' for the latest version verified to work with this version of the binary, \
-                  and 'latest-tag' for the literal tag 'latest' on unicode.org."
+                  and 'latest-tag' for the literal tag 'latest' on unicode.org. \
+                  Ignored if '--unicode-rscd-root' is present. Requires binary to be built with `networking` Cargo feature (enabled by default)."
     )]
     #[cfg_attr(not(feature = "networking"), arg(hide = true))]
     #[cfg(feature = "provider")]
-    ucd_tag: String,
+    unicode_tag: String,
```
sffc, line 185: Why `--unicode-tag` and not `--rscd-tag` or `--unicode-rscd-tag`

On SourceDataProvider, both the tag and the root have "rscd" in the fn names

sffc (approved):

provider/icu4x-datagen/src/main.rs
robertbastian: The _tag_ is the Unicode (The Unicode Standard) version, because RSCD is versioned with it (that's the S).


> On SourceDataProvider, both the tag and the root have "rscd" in the fn names

The fns have "rscd" in the name because they set the RSCD data source. The tag constant is called `TESTED_UNICODE_TAG`.

Issue #8206, opened 2026-07-08
currency: Decouple number style from currency style in compact currency constructors (6-function matrix)


younies:
### Summary

**Related PR**
https://github.com/unicode-org/icu4x/pull/8198

In our current experimental compact currency formatting implementation (`CurrencyFormatter<Compact>`), constructor names conflate two orthogonal formatting dimensions into a single name:
1. **The Number Style:** Compact / compact-short (`1.2K`) vs. Verbose / compact-long (`1.2 thousand`)
2. **The Currency Style:** symbol (Short as ICU) (`$`), symbol narrow (narrow as ICU) (`$`), vs. Long display name (`US dollars`)

### The Ambiguity
Because we use the terms "short" and "long" for *both* dimensions (i.e., short/long compact numbers and short/long currency names), a constructor named `try_new_compact_long` is ambiguous—does "long" refer to the number or the currency? Currently, `try_new_compact_long` formats *both* as long, producing `"1.2 thousand US dollars"`, leaving users with no way to get `"1.2K US dollars"`.

In fact, in our original design doc, we planned to introduce a `struct Verbose` alongside `struct Decimal` and `struct Compact` as a distinct `ValueRepresentation` marker to prevent this exact ambiguity:

```rust
/// Representation for compact long formatting (verbose notation)
/// (planned for when CLDR adds verbose compact currency support).
pub struct Verbose;
impl ValueRepresentation for Verbose {}
```

To eliminate this confusion, we should use the wording **`verbose`** for compact-long numbers (`1.2 thousand`) and reserve **`compact`** strictly for compact-short numbers (`1.2K`). This allows us to cleanly distinguish between the two number styles without clashing with currency width names.

### Proposed 6-Function Architecture (2x3 Matrix)
In this architecture, the prefix (`compact` vs `verbose`) controls the **number**, and the suffix (`_short`, `_narrow`, `_long`) controls the **currency width**:

| Constructor | Number Style | Currency Style | Example Output (`en-US`, `USD`) | Status |
| :--- | :--- | :--- | :--- | :--- |
| **`try_new_compact_short`** | Compact (`1.2K`) | Short (`$`) | `"$1.2K"` | Already correct |
| **`try_new_compact_narrow`** | Compact (`1.2K`) | Narrow (`$`) | `"$1.2K"` | Already correct |
[...]

PR #8211, opened 2026-07-09
feat(currency): Clarify currency and compact constructor naming with 9-function matrix

younies:
## Summary
This PR decouples number notation style from currency display width across both standard (`CurrencyFormatter<FixedDecimalFormatter>`) and compact (`CurrencyFormatter<CompactDecimalFormatter>`) currency formatting by introducing a systematic, unambiguous 9-function naming matrix.

## Motivation & Architecture
Previously, constructor names like `short`, `narrow`, and `long` conflated two orthogonal formatting dimensions into a single adjective:
1. **Number Notation Style:** Standard (`1,200`), Compact Short (`1.2K`), vs. Compact Long (`1.2 thousand`)
2. **Currency Display Width:** Short Symbol (`$`), Narrow Symbol (`$`), vs. Display Name (`US dollars`)

For example, in `try_new_compact_long`, the word `long` was ambiguous—did it refer to the compact number width (`1.2 thousand`), the currency width (`US dollars`), or both (`12 thousand US dollars`)?

To eliminate all ambiguity, this PR renames and organizes all constructors so that:
- **Number style prefixes:** No prefix = standard number (`1,200`), `compact_` = compact short (`1.2K`), `compact_long_` = compact long (`1.2 thousand`).
- **Currency style suffixes:** `_symbol` = short symbol (`$`), `_symbol_narrow` = narrow symbol (`$`), `_name` = full currency display name (`US dollars`).

#### 1. Standard Currency (`CurrencyFormatter<FixedDecimalFormatter>`)
| Previous Name | New Constructor Name | Number Style | Currency Style | Example (`en-US`, `USD`) |
| :--- | :--- | :--- | :--- | :--- |
| `try_new_short` | **`try_new_symbol`** | Standard (`1,200`) | Short Symbol (`$`) | `"$1,200.00"` |
| `try_new_narrow` | **`try_new_symbol_narrow`** | Standard (`1,200`) | Narrow Symbol (`$`) | `"$1,200.00"` |
| `try_new_long` | **`try_new_name`** | Standard (`1,200`) | Display Name (`US dollars`) | `"1,200.00 US dollars"` |

#### 2. Compact Currency — Short Numbers (`CurrencyFormatter<CompactDecimalFormatter>`)
| Previous Name | New Constructor Name | Number Style | Currency Style | Example (`en-US`, `USD`) |
| :--- | :--- | :--- | :--- | :--- |
| `try_new_compact_short` | **`try_new_compact_symbol`** | Compact (`1.2K`) | Short Symbol (`$`) | `"$1.2K"` |
| `try_new_compact_narrow` | **`try_new_compact_symbol_narrow`** | Compact (`1.2K`) | Narrow Symbol (`$`) | `"$1.2K"` |
| `try_new_compact_long` | **`try_new_compact_name`** | Compact (`1.2K`) | Display Name (`US dollars`) | `"1.2K US dollars"` |

#### 3. Compact Currency — Long Numbers (`CurrencyFormatter<CompactDecimalFormatter>`)
| Previous Name | New Constructor Name | Number Style | Currency Style | Example (`en-US`, `USD`) |
| :--- | :--- | :--- | :--- | :--- |
| `try_new_verbose_short` | **`try_new_compact_long_symbol`** | Compact Long (`1.2 thousand`) | Short Symbol (`$`) | `"$1.2 thousand"` |
| `try_new_verbose_narrow` | **`try_new_compact_long_symbol_narrow`** | Compact Long (`1.2 thousand`) | Narrow Symbol (`$`) | `"$1.2 thousand"` |
| `try_new_verbose_long` | **`try_new_compact_long_name`** | Compact Long (`1.2 thousand`) | Display Name (`US dollars`) | `"1.2 thousand US dollars"` |

TAG=agy
CONV=043e35c5-474c-4f4c-88cd-e1ddd3cb5704

## Changelog

icu_experimental/currency:

- Renamed standard currency constructors (`try_new_short` -> `try_new_symbol`, `try_new_narrow` -> `try_new_symbol_narrow`, `try_new_long` -> `try_new_name`).
- Renamed compact short constructors (`try_new_compact_short` -> `try_new_compact_symbol`, `try_new_compact_narrow` -> `try_new_compact_symbol_narrow`, `try_new_compact_long` -> `try_new_compact_name`).
- Added/renamed compact long constructors (`try_new_compact_long_symbol`, `try_new_compact_long_symbol_narrow`, `try_new_compact_long_name`).
- Updated FFI / unstable macro wrappers, unit tests, and doctests across all 9 variants.

### DIFF
--- a/components/experimental/src/dimension/currency/compact_format.rs
+++ b/components/experimental/src/dimension/currency/compact_format.rs
@@ -14,7 +14,7 @@ mod tests {
     pub fn test_en_us() {
         let prefs = locale!("en-US").into();
         let currency_code = CurrencyCode(tinystr!(3, "USD"));
-        let fmt = CurrencyFormatter::try_new_compact_short(prefs, &currency_code).unwrap();
+        let fmt = CurrencyFormatter::try_new_compact_symbol(prefs, &currency_code).unwrap();
 
         // Positive case
         let positive_value = "12345.67".parse().unwrap();
@@ -31,7 +31,7 @@ mod tests {
     pub fn test_fr_fr() {
         let prefs = locale!("fr-FR").into();
         let currency_code = CurrencyCode(tinystr!(3, "EUR"));
-        let fmt = CurrencyFormatter::try_new_compact_short(prefs, &currency_code).unwrap();
+        let fmt = CurrencyFormatter::try_new_compact_symbol(prefs, &currency_code).unwrap();
 
         // Positive case
         let positive_value = "12345.67".parse().unwrap();
@@ -48,7 +48,7 @@ mod tests {
     pub fn test_zh_cn() {
         let prefs = locale!("zh-CN").into();
         let currency_code = CurrencyCode(tinystr!(3, "CNY"));
-        let fmt = CurrencyFormatter::try_new_compact_short(prefs, &currency_code).unwrap();
+        let fmt = CurrencyFormatter::try_new_compact_symbol(prefs, &currency_code).unwrap();
 
         // Positive case
         let positive_value = "12345.67".parse().unwrap();
@@ -65,7 +65,7 @@ mod tests {
     pub fn test_ar_eg() {
         let prefs = locale!("ar-EG").into();
         let currency_code = CurrencyCode(tinystr!(3, "EGP"));
-        let fmt = CurrencyFormatter::try_new_compact_short(prefs, &currency_code).unwrap();
+        let fmt = CurrencyFormatter::try_new_compact_symbol(prefs, &currency_code).unwrap();
 
         // Positive case
         let positive_value = "12345.67".parse().unwrap();
@@ -87,7 +87,7 @@ mod tests {
         let prefs = locale!("en-US").into();
 
         let currency_code = CurrencyCode(tinystr!(3, "USD"));
-        let fmt = CurrencyFormatter::try_new_compact_long(prefs, &currency_code).unwrap();
+        let fmt = CurrencyFormatter::try_new_compact_long_name(prefs, &currency_code).unwrap();
 
         // Positive case
         let positive_value = "12345.67".parse().unwrap();
@@ -105,7 +105,7 @@ mod tests {
         let prefs = locale!("en-US").into();
 
         let currency_code = CurrencyCode(tinystr!(3, "USD"));
-        let fmt = CurrencyFormatter::try_new_compact_long(prefs, &currency_code).unwrap();
+        let fmt = CurrencyFormatter::try_new_compact_long_name(prefs, &currency_code).unwrap();
 
         // Positive case
         let positive_value = "12345000.67".parse().unwrap();
@@ -123,7 +123,7 @@ mod tests {
         let prefs = locale!("fr-FR").into();
 
         let currency_code = CurrencyCode(tinystr!(3, "USD"));
-        let fmt = CurrencyFormatter::try_new_compact_long(prefs, &currency_code).unwrap();
+        let fmt = CurrencyFormatter::try_new_compact_long_name(prefs, &currency_code).unwrap();
 
         // Positive case
         let positive_value = "12345.67".parse().unwrap();
@@ -141,7 +141,7 @@ mod tests {
         let prefs = locale!("fr-FR").into();
 
         let currency_code = CurrencyCode(tinystr!(3, "USD"));
-        let fmt = CurrencyFormatter::try_new_compact_long(prefs, &currency_code).unwrap();
+        let fmt = CurrencyFormatter::try_new_compact_long_name(prefs, &currency_code).unwrap();
 
         // Positive case
         let positive_value = "12345000.67".parse().unwrap();
@@ -160,8 +160,8 @@ mod tests {
         let usd = CurrencyCode(tinystr!(3, "USD"));
         let sek = CurrencyCode(tinystr!(3, "SEK"));
 
-        let fmt_usd = CurrencyFormatter::try_new_compact_short(prefs, &usd).unwrap();
-        let fmt_sek = CurrencyFormatter::try_new_compact_short(prefs, &sek).unwrap();
+        let fmt_usd = CurrencyFormatter::try_new_compact_symbol(prefs, &usd).unwrap();
+        let fmt_sek = CurrencyFormatter::try_new_compact_symbol(prefs, &sek).unwrap();
 
         // Small number (magnitude < 3, no compact suffix): should fall back cleanly
         let small_value = "123".parse().unwrap();
@@ -171,5 +171,25 @@ mod tests {
         // Compact number with alphabetical currency code: should use alpha_next_to_number pattern with non-breaking space
         let compact_value = "12345.67".parse().unwrap();
         assert_writeable_eq!(fmt_sek.format_fixed_decimal(&compact_value), "SEK\u{a0}12K");
+    }
+
+    #[test]
+    pub fn test_compact_name_and_compact_long_symbol() {
+        let prefs = locale!("en-US").into();
+        let usd = CurrencyCode(tinystr!(3, "USD"));
+
+        let fmt_compact_name = CurrencyFormatter::try_new_compact_name(prefs, &usd).unwrap();
+        let fmt_compact_long_symbol =
+            CurrencyFormatter::try_new_compact_long_symbol(prefs, &usd).unwrap();
+
+        let val = "12345.67".parse().unwrap();
+        assert_writeable_eq!(
+            fmt_compact_name.format_fixed_decimal(&val),
+            "12K US dollars"
+        );
+        assert_writeable_eq!(
+            fmt_compact_long_symbol.format_fixed_decimal(&val),
+            "$12 thousand"
+        );
     }
 }
--- a/components/experimental/src/dimension/currency/formatter.rs
+++ b/components/experimental/src/dimension/currency/formatter.rs
@@ -49,7 +49,7 @@ pub(crate) enum CurrencyFormatterData {
         essential: DataPayload<CurrencyEssentialsV1>,
         currency: CurrencyCode,
     },
-    Essential {
+    Symbol {
         essential: DataPayload<CurrencyEssentialsV1>,
         symbol: DataPayload<CurrencySymbolsV1>,
     },
@@ -103,7 +103,7 @@ impl<V: AbstractFormatter> CurrencyFormatter<V> {
         )
         .allow_identifier_not_found()?
         {
-            Some(res) => CurrencyFormatterData::Essential {
+            Some(res) => CurrencyFormatterData::Symbol {
                 essential,
                 symbol: res.payload,
             },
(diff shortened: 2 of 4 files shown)


components/experimental/src/dimension/currency/compact_format.rs
```diff
@@ -83,75 +83,122 @@ mod tests {
...
+    #[test]
+    pub fn test_en_us_verbose_short_and_narrow() {
+        let prefs = locale!("en-US").into();
+        let currency_code = CurrencyCode(tinystr!(3, "USD"));
+        let fmt_short = CurrencyFormatter::try_new_verbose_short(prefs, &currency_code).unwrap();
+        let fmt_narrow = CurrencyFormatter::try_new_verbose_narrow(prefs, &currency_code).unwrap();
+
+        let val = "12345.67".parse().unwrap();
+        assert_writeable_eq!(fmt_short.format_fixed_decimal(&val), "$12 thousand");
+        assert_writeable_eq!(fmt_narrow.format_fixed_decimal(&val), "$12 thousand");
```
younies: Add a todo(https://github.com/unicode-org/icu4x/issues/8206) , while compact-* works perfectly in all of its cases, `compact-narrow` , `compact-short` and `compact-long`, it seems that the `verbose-narrow` and `verbose-short` is not making a lot of sense in formatting (e.g. `$12 thousand"` which I think even it should be `$12 thousand**s**` 

Add a comment to the issue with this note too

younies: Done! The note was added to [Issue #8206](https://github.com/unicode-org/icu4x/issues/8206#issuecomment-4925239827), and TODO(#8206) comments referencing it were added in both `compact_formatter.rs` and `compact_format.rs` (commit 278324556a).

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -430,7 +430,10 @@ impl<V: AbstractFormatter> CurrencyFormatter<V> {
     /// let currency_code = CurrencyCode(tinystr!(3, "USD"));
     /// let fmt = CurrencyFormatter::try_new_compact_long(currency_prefs, &currency_code).unwrap();
     /// let value = "12345.67".parse().unwrap();
-    /// assert_writeable_eq!(fmt.format_fixed_decimal(&value), "12 thousand US dollars");
+    /// assert_writeable_eq!(fmt.format_fixed_decimal(&value), "12K US dollars");
```
robertbastian: nit: probably want short, compact-short, and verbose-short here. remove compact-long

components/experimental/src/dimension/currency/compact_format.rs
```diff
@@ -83,75 +83,126 @@ mod tests {
...
+            fmt.format_fixed_decimal(&negative_val),
+            "-12 mille dollars des États-Unis"
+        );
+
+        // Millions
+        let millions_val = "12345000.67".parse().unwrap();
+        assert_writeable_eq!(
+            fmt.format_fixed_decimal(&millions_val),
+            "12 millions dollars des États-Unis"
+        );
```
robertbastian: suggestion: use this for an interesting plurals test case
```suggestion
        let millions_val = "1034500.67".parse().unwrap();
        assert_writeable_eq!(
            fmt.format_fixed_decimal(&millions_val),
            "1 million dollars des États-Unis"
        );
```

components/experimental/src/dimension/currency/compact_format.rs
```diff
@@ -83,75 +83,126 @@ mod tests {
...
+    #[test]
+    pub fn test_en_us_verbose_short_and_narrow() {
+        let prefs = locale!("en-US").into();
+        let currency_code = CurrencyCode(tinystr!(3, "USD"));
+        let fmt_short = CurrencyFormatter::try_new_verbose_short(prefs, &currency_code).unwrap();
+        let fmt_narrow = CurrencyFormatter::try_new_verbose_narrow(prefs, &currency_code).unwrap();
+
+        let val = "12345.67".parse().unwrap();
+        assert_writeable_eq!(fmt_short.format_fixed_decimal(&val), "$12 thousand");
+        assert_writeable_eq!(fmt_narrow.format_fixed_decimal(&val), "$12 thousand");
```
robertbastian: the plural form is a bug, please investigate

robertbastian (commented):
I'm not convinced we want to add the verbose forms, but they don't hurt for now

components/experimental/src/dimension/currency/formatter.rs
younies: Updated in commit 1f2a1ebb85: The doctests on `format_fixed_decimal` now explicitly illustrate `short` (`"$12,345.67"`), `compact-short` (`"$12K"`), `compact-long` (`"12K US dollars"`), and `verbose-long` (`"12 thousand US dollars"`).

components/experimental/src/dimension/currency/compact_format.rs
younies: Applied in commit c9c950f359: Added the singular millions test case (`"1034500.67"` -> `"1 million dollars des États-Unis"`) alongside plural millions to verify grammatical agreement in French.

components/experimental/src/dimension/currency/compact_format.rs
younies: The grammatical ambiguity of juxtaposing short symbols (`$`) with spelled-out words (`thousand`) is documented in [Issue #8206](https://github.com/unicode-org/icu4x/issues/8206#issuecomment-4925239827).

In English, cardinal number words do not pluralize when quantifying (e.g., `"$12 thousand"` or `"12 thousand dollars"`, never `"$12 thousands"`). When a language's grammar does require plural inflection (such as French millions), CLDR and ICU4X correctly apply it—as demonstrated by the French test case (`"12 millions\u{a0}€"`) added in commit c8e653c098.

However, because symbol-word juxtaposition reads unnaturally in many contexts, whether `verbose-short` and `verbose-narrow` should be supported is under evaluation in #8206. A `TODO(#8206)` referencing this discussion was added above this test in commit 278324556a.

components/experimental/src/dimension/currency/formatter.rs
robertbastian: I asked to remove compact-long because it's a weird format. we shouldn't have too much example code here

younies: Agreed! Removed `compact-long` from the `format_fixed_decimal` doctests in commit 0c753c1161, leaving `short` (`"$12,345.67"`), `compact-short` (`"$12K"`), and `verbose-long` (`"12 thousand US dollars"`) as the core reference examples.

components/experimental/src/dimension/currency/formatter.rs
```diff
@@ -414,9 +414,13 @@ impl<V: AbstractFormatter> CurrencyFormatter<V> {
     ///
     /// let locale = locale!("en-US").into();
     /// let currency_code = CurrencyCode(tinystr!(3, "USD"));
-    /// let fmt = CurrencyFormatter::try_new_compact_short(locale, &currency_code).unwrap();
     /// let value = "12345.67".parse().unwrap();
-    /// assert_writeable_eq!(fmt.format_fixed_decimal(&value), "$12K");
+    ///
+    /// let fmt_short = CurrencyFormatter::try_new_short(locale, &currency_code).unwrap();
```
robertbastian: this duplicates the previous example

components/experimental/src/dimension/currency/compact_format.rs
robertbastian: The answer to my question is that it's not a bug because "thousand" is the correct value for `other`: https://github.com/unicode-org/cldr/blob/c8264332b43d317b98a3d05aaf86e5c18e37610e/common/main/en.xml#L5429

robertbastian (commented):
you have not addressed all previous comments

sffc (dismissed):
Let's please stick to the CLDR, ICU, and ECMA names for these things.

Compact decimal notation has two versions, "short" and "long". If you don't specify which, "short" is the default.

Currency symbols are called "[short] symbol", "narrow symbol", "[iso] code", or "[full] name". If you don't specify which, "short" is the default.

I thought you had packed the narrow symbol in with the short symbol, and made it a runtime option in the options bag. You can also do this with the formal symbol. There are not many currencies with narrow or formal symbol overrides, so I think this design still makes sense. It also means your formatter can format all currencies, without needing a BoundLocaleDataProvider.

Full Name needs the currency in the constructor, or else BoundLocaleDataProvider, which we can add later. Because of this, it needs to be a separate type, since the format function signature is different.

So your constructors should be called:

`CurrencyFormatter` (the `format` function takes the currency code):

| | Symbol (short or narrow) | ISO Code |
|---|---|---|
| Non-Compact | `try_new` | `try_new_iso_code` |
| Compact Short | `try_new_compact` | `try_new_iso_code_compact` |
| Compact Long | `try_new_compact_long` | `try_new_iso_code_compact_long` |

`CurrencyNameFormatter` (the currency code is in the constructor):

| | Long Name |
|---|---|
| Non-Compact | `try_new` |
| Compact Short | `try_new_compact` |
| Compact Long | `try_new_compact_long` |

robertbastian:
> Full Name needs the currency in the constructor, or else BoundLocaleDataProvider, which we can add later. Because of this, it needs to be a separate type, since the format function signature is different.

We just merged these types in https://github.com/unicode-org/icu4x/pull/8150 to streamline the API. This request is orthogonal to this PR and should not block it.

I agree with Younies that "long compact" is a misnomer.

sffc:
> > Full Name needs the currency in the constructor, or else BoundLocaleDataProvider, which we can add later. Because of this, it needs to be a separate type, since the format function signature is different.
> 
> We just merged these types in #8150 to streamline the API. This request is orthogonal to this PR and should not block it.

Ok, we can discuss in a follow-up issue.

> I agree with Younies that "long compact" is a misnomer.

I'm a hard no on inventing new terminology like "verbose". [CLDR](https://www.unicode.org/reports/tr35/tr35-numbers.html#Compact_Number_Formats) uses "long", and so does [ICU](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4j/com/ibm/icu/number/Notation.html#compactLong()) and [ECMA](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#compactdisplay), which all went through similar API reviews, and no one seems to have been concerned about this.

robertbastian:
This PR has a very simple path forward and should have been merged days ago. It's blocking all other currency work. @younies please do the renames and let's merge this.

younies:
As suggested by @sffc, we are now using the naming from CLDR across the constructors, and I have updated the currency description. The branch has also been rebased directly on top of upstream `main` as a standalone PR focused solely on the 9-function matrix.

components/experimental/src/dimension/currency/formatter.rs
robertbastian: this should be `Symbol` then

robertbastian (approved):

components/experimental/src/dimension/currency/formatter.rs
younies: sure, as the data name associated with the usage then.

robertbastian (approved):

PR #8215, opened 2026-07-09
Remove preferences serde impls

### LINKED ISSUE #8171: Should locale preferences serialize as strings
Nit/Issue: the serialized forms of these should probably be the BCP-47. It's hard to justify something other than that. Currently they are the EnumVariant names.

_Originally posted by @sffc in https://github.com/unicode-org/icu4x/pull/8102#discussion_r3504097473_

robertbastian:
Fixes #8171 

## Changelog

N/A

### DIFF
--- a/components/calendar/src/provider.rs
+++ b/components/calendar/src/provider.rs
@@ -82,6 +82,15 @@ pub struct CalendarPreference {
     pub default_hijri_algorithm: HijriCalendarAlgorithm,
 }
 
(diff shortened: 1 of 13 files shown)


sffc (approved):
I pushed some changes to:

- Add debug assertions in GIGO
- Add human-readable serialization
- Don't use discriminant zero (a nitpick of mine)

PR #8216, opened 2026-07-10
Implement language display menu names without using a trait

sffc:
Replaces https://github.com/unicode-org/icu4x/pull/8205

Based on the same refactor, https://github.com/unicode-org/icu4x/pull/8207

## Changelog

icu_experimental/displaynames:

- New constructors `LanguageIdentifierDisplayNameOwned::try_new_menu[_unstable|_with_buffer_provider]`


components/experimental/src/displaynames/single/language.rs
```diff
@@ -98,7 +110,13 @@ size_test!(
 #[derive(Debug)]
 pub struct LanguageIdentifierDisplayNameOwned {
     /// Either the language display name or the subtag as fallback
-    language_payload: DataPayloadOr<LocaleNamesLanguageMediumV1, Language>,
+    language_payload: DataPayloadOr<ErasedMarker<MenuNamePartsOrString<'static>>, Language>,
+    /// All other fields (shared between Standard and Menu)
+    qualifiers: QualifiersOwned,
```
robertbastian, line 115: actually now you don't need `QualifiersOwned` as an abstraction, and it's slightly confusing because one qualifier is still in the `LanguagePayload`.

components/experimental/src/displaynames/single/language.rs
```diff
@@ -143,94 +161,250 @@ impl LanguageIdentifierDisplayNameOwned {
...
-                    script,
-                    region,
-                    &mut buffer,
-                );
-                let id = DataIdentifierBorrowed::for_marker_attributes_and_locale(
-                    attrs,
+        // Only try dialect if requested (or default)
+        let language_payload =
+            if options.language_display.unwrap_or_default() == LanguageDisplay::Dialect {
+                Self::load_language_dialect_name(provider, &formatting_locale, &mut subject)?
```
robertbastian, line 167: I don't like that you outlined this into a function that is below the other constructors. this code is only used here and it does not improve readability to have to jump around the file to figure out what this method is doing

components/experimental/src/displaynames/single/language.rs
```diff
@@ -143,94 +161,250 @@ impl LanguageIdentifierDisplayNameOwned {
...
+            + DataProvider<LocaleNamesRegionMediumV1>
+            + DataProvider<LocaleNamesVariantMediumV1>
+            + DataProvider<LocaleNamesEssentialsV1>,
+    {
+        let formatting_locale = LocaleNamesLanguageMediumV1::make_locale(prefs.locale_preferences);
+
+        // Step 1: Load language name
+        // Try the menu name
+        let language_payload =
+            Self::load_language_menu_name(provider, &formatting_locale, subject.language)?;
```
robertbastian, line 257: same here

components/experimental/src/displaynames/single/language.rs
```diff
@@ -143,94 +161,250 @@ impl LanguageIdentifierDisplayNameOwned {
...
+                    subject.region = None;
+                }
+                return Ok(Some(response));
+            }
+        }
+        Ok(None)
+    }
+
+    /// Loads the name for an individual language subtag.
+    fn load_language_subtag_name<P>(
```
robertbastian, line 350: why extract `load_language_subtag_name` into a helper, but not `load_script_subtag_name` et al?

robertbastian (approved):

components/experimental/src/displaynames/single/language.rs
sffc: `load_language_subtag_name` has 2 call sites.

components/experimental/src/displaynames/single/language.rs
sffc: I was aiming to make `try_new_unstable` consistently call down to helper functions, rather than have a mix of helper functions and direct data loads. Also, I find that `load_language_dialect_name` ends up being more readable for a combination of reduced indentation and early return (which I could do inline but it adds another extra layer of indentation).

components/experimental/src/displaynames/single/language.rs
sffc: I still need a helper function for both constructors to use, and it needs a return type.

sffc:
Merging based on previous approval.

PR #8217, opened 2026-07-10
Model segmenter tailoring data like CLDR will

robertbastian:
#7962 

Best reviewed commit-by-commit.

## Changelog

N/A


sffc (approved):
Looks plausible; I'll assume @eggrobin will do a more detailed review of the segmenter datagen logic

robertbastian:
merging this as there really isn't any interesting segmenter logic that would need more review, it's just datagen plumbing

eggrobin:
(Yes, and we looked at the basic idea of factoring the tailorings together, so that looks as expected. Sorry about the delayed response, I have been stuck in the 19.0 propertizing mines.)

PR #8221, opened 2026-07-12
Add allowed-failure Miri CI job

spydix:


Manishearth:
Perhaps we should only run this on the cronjob CI, we don't need to run this every PR.

sffc:
Thanks for the contribution. CI is currently failing. Please reopen this if you are able to fix CI.

PR #8225, opened 2026-07-13
Add UnvalidatedTinyAsciiStr::default()

sffc:
Split from #7621

## Changelog

tinystr: add `UnvalidatedTinyAsciiStr::DEFAULT`

### DIFF
--- a/utils/tinystr/src/unvalidated.rs
+++ b/utils/tinystr/src/unvalidated.rs
@@ -27,9 +27,6 @@ impl<const N: usize> fmt::Debug for UnvalidatedTinyAsciiStr<N> {
 }
 
 impl<const N: usize> UnvalidatedTinyAsciiStr<N> {
-    /// The empty string as a default value.
-    pub const DEFAULT: Self = Self([0u8; N]);
-
     #[inline]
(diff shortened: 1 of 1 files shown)


Manishearth (approved):

robertbastian (changes requested):
I think what we're doing everywhere now is implement `Default` and shadow with `const fn default()`

utils/tinystr/src/unvalidated.rs
```diff
@@ -53,6 +53,18 @@ impl<const N: usize> UnvalidatedTinyAsciiStr<N> {
     pub const fn from_utf8_unchecked(bytes: [u8; N]) -> Self {
         Self(bytes)
     }
+
+    #[inline]
+    /// Returns the empty string.
+    pub const fn default() -> Self {
+        Self([0u8; N])
```
robertbastian: of course we don't use `Default` for `TinyAsciiStr`...

```suggestion
        TinyAsciiStr::EMPTY.to_unvalidated()
```

robertbastian (approved):

robertbastian (approved):

PR #8232, opened 2026-07-15
Install `cargo-rdme` from GitHub

robertbastian:
Because it doesn't build.

## Changelog

N/A


sffc (approved):
I assume the main difference here is that this uses the locked transitive dependencies?

robertbastian:
This downloads a binary

Issue #8234, opened 2026-07-16
[Currency Formatter] Calculate alpha-next-to-number pattern selection dynamically at runtime


younies:
In the Currency Formatter, we currently store whether each symbol should use the standard pattern or the standard-alpha-next-number pattern. @robertbastian has raised an excellent point: we could remove the pattern selector and calculate this in real time during formatting.

I suggest we store the directional value of the number relative to the symbol (`LEFT`, `RIGHT`, or `NONE`). This would indicate if the number appears to the left of the symbol, the right of the symbol, or never adjacent to it (e.g., if there is a space). This would make it easier to determine if a symbol like `"CA$"` should be placed next to a number for a specific pattern. Essentially, we could assign the direction values `FIRST`, `LAST`, or `NONE` based on which character of the symbol we need to check.

With this approach, we could quickly determine at runtime whether a currency symbol should be alpha-next-to-number. What do you think?

younies:
Here is a concrete suggested implementation for calculating this at runtime using an enum (`SymbolCharCheck` / `NumberAdjacency`) and helper method:

```rust
/// Indicates which character of the currency symbol needs to be checked for its alphabetic property
/// based on the adjacency direction (`LEFT`, `RIGHT`, or `NONE`) of the number relative to the symbol.
#[derive(Debug, Copy, Clone, PartialEq, Eq)]
pub enum SymbolCharCheck {
    /// Number is immediately to the left of the symbol without a space (e.g., `100CA$`).
    /// We inspect the FIRST character of the symbol.
    First,
    /// Number is immediately to the right of the symbol without a space (e.g., `CA$100`).
    /// We inspect the LAST character of the symbol.
    Last,
    /// Number is separated by a space or not adjacent to the symbol.
    None,
}

impl SymbolCharCheck {
    /// Maps pattern adjacency characteristics to the required character check.
    pub fn from_pattern_adjacency(number_is_left: bool, has_space: bool) -> Self {
        if has_space {
            Self::None
        } else if number_is_left {
            Self::First
        } else {
            Self::Last
        }
    }

    /// Determines at runtime during formatting whether the currency symbol requires
    /// the alpha-next-to-number pattern spacing/selection.
    pub fn requires_alpha_next_to_number(self, symbol: &str) -> bool {
        match self {
            Self::First => symbol.chars().next().is_some_and(char::is_alphabetic),
            Self::Last => symbol.chars().last().is_some_and(char::is_alphabetic),
            Self::None => false,
[...]

PR #8235, opened 2026-07-16
Remove pattern selection from currency symbols data

robertbastian:
#8234 

Once I did that, I could use data marker attributes for the symbols data. This reduces the data by 90%:

```diff
-currency/symbols/v1, <lookup>, 2463B, 434 identifiers
-currency/symbols/v1, <total>, 354150B, 330593B, 302 unique payloads
+currency/symbols/v1, <lookup>, 11386B, 1590 identifiers
+currency/symbols/v1, <total>, 25881B, 7297B, 808 unique payloads
```
## Changelog

N/A


sffc (commented):
ok, if this makes data that much smaller, it seems well motivated. We have better data provider infrastructure now (the VarULE optimization and better attributes) than we did when we implemented the current design. It means that the currency code needs to go into the constructor instead of the format function, unless/until we implement BoundLocaleDataProvider.

I'll let @younies look at the currency-specific code.

robertbastian:
> We have better data provider infrastructure now (the VarULE optimization and better attributes) than we did when we implemented the current design

The VarULE optimisation does not decrease data size. I think this could have been done like this from the beginning.

sffc (approved):
I'll approve this before @younies's review since it seems fine and unblocks other PRs.

The old design was good under the constraint that we have limited numbers of payloads per data key. The new design is better when we relax that constraint.

robertbastian:
Well not really. The old design was bad for deduplication, it put all
symbols in one struct, and while each locale uses the same symbols for most
currencies, even one different currency (e.g. a local one) would break
deduplication and store a full 1kB+ data struct. Even under the old
constraint (not sure where that comes from), we should have deduplicated
with a root marker like for time zone names. However with this approach, we
also get deduplication between narrow and short symbols, and there's zero
struct overhead.

PR #8237, opened 2026-07-16
Update MinuteOptional format behavior in h23 patterns

### LINKED ISSUE #7462: Update behavior of MinuteOptional format in h23 patterns

sffc:
Fixes #7462

Update `TimePrecision::MinuteOptional` formatting in `icu_datetime` so that 24-hour hour cycles (`h23`/`h24`) preserve zero minutes (e.g., `15:00`), while 12-hour hour cycles (`h11`/`h12`) continue to drop zero minutes (e.g., `3 PM`), following CLDR resolution CLDR-19327.

## Changelog
icu_datetime: Preserve zero minutes in MinuteOptional format for 24-hour hour cycles


sffc:
The semver error is

```
     Checked [   0.253s] 215 checks: 214 pass, 1 fail, 0 warn, 38 skip

     Summary semver requires new major version: 1 major and 0 minor checks failed

--- failure enum_variant_missing: pub enum variant removed or renamed ---

Description:
A publicly-visible enum has at least one variant that is no longer available under its prior name. It may have been renamed or removed entirely.
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.48.0/src/lints/enum_variant_missing.ron

Failed in:
  variant TimeGranularity::None, previously in file /home/runner/work/icu4x/icu4x/semver-checks/target/semver-checks/git-origin_main/32a9c42f67bde1261c8b5592d4fed39eb8fc135c/components/datetime/src/provider/pattern/mod.rs:40
  variant TimeGranularity::Hours, previously in file /home/runner/work/icu4x/icu4x/semver-checks/target/semver-checks/git-origin_main/32a9c42f67bde1261c8b5592d4fed39eb8fc135c/components/datetime/src/provider/pattern/mod.rs:42
```

I added aliases as associated constants. Is this sufficient? I think it is but I'm not sure about if there are weird edge cases on how matching works, etc.

components/datetime/src/provider/pattern/mod.rs
```diff
@@ -35,51 +35,74 @@ pub use item::{GenericPatternItem, PatternItem};
 #[cfg_attr(feature = "serde", derive(serde::Deserialize))]
 #[non_exhaustive]
 pub enum TimeGranularity {
-    /// No time is in the pattern.
+    /// No time is in the pattern, or hours are 23/24-hour style.
     #[default]
-    None,
-    /// Smallest time unit = hours.
-    Hours,
+    Hours23OrNone = 0,
```
robertbastian: nit: that's an ugly name because the `O` looks like a zero

robertbastian: this being a provider type, I'd prefer if we keep the current names. this doesn't need to be super ergonomic, and it would remove a lot of the diff of this PR

components/datetime/src/raw/neo.rs
```diff
@@ -408,7 +409,14 @@ impl TimePatternSelectionData {
     ) -> Option<TimePatternDataBorrowed<'_>> {
         let payload = self.payload.get_option()?;
         let time_precision = options.time_precision.unwrap_or_default();
-        let (variant, subsecond_digits) = input.resolve_time_precision(time_precision);
+        let prefer_keep_minutes = matches!(time_precision, TimePrecision::MinuteOptional)
```
robertbastian: I think this should be moved into `resolve_time_precision`, and `TimeGranularity` should be passed in there. This code is part of resolving the time precision

you wouldn't even need the `prefer_keep_minutes` bool, you could match directly on `TimeGranularity`, which would make the condition more readable

components/datetime/tests/simple_test.rs
```diff
@@ -483,3 +483,79 @@ fn test_adar_narrow() {
...
+    let fmt_fr_h12 = NoCalendarFormatter::try_new(locale!("fr-u-hc-h12").into(), fs).unwrap();
+    assert_writeable_eq!(fmt_fr_h12.format(&time_zero), "7\u{202f}AM");
+    assert_writeable_eq!(fmt_fr_h12.format(&time_nonzero), "7:12\u{202f}AM");
+
+    // fr with -u-hc-h11 override: zero minutes omitted
+    let fmt_fr_h11 = NoCalendarFormatter::try_new(locale!("fr-u-hc-h11").into(), fs).unwrap();
+    assert_writeable_eq!(fmt_fr_h11.format(&time_zero), "7\u{202f}AM");
+    assert_writeable_eq!(fmt_fr_h11.format(&time_nonzero), "7:12\u{202f}AM");
+
+    // DateTimeFormatter
```
robertbastian: why are you testing with different formatters? they're not using different code paths afaict

I don't like duplicated tests because I want to carefully review tests. there are too many tests here

robertbastian (commented):
I really don't like that we have all this datagen code in the datetime crate, and that you can produce a data diff without touching `icu_provider_source`. When reviewing, I don't know which code is datagen code and which code is runtime code, which matters a lot.

components/datetime/src/provider/pattern/mod.rs
sffc: :robot: Done! Reverted the variant names back to `None` and `Hours`, and expanded the doc comments on both variants to clearly document their 24-hour vs 12-hour mappings and `MinuteOptional` formatting rules.

components/datetime/src/raw/neo.rs
sffc: :robot: Done! Refactored `resolve_time_precision` to take `TimeGranularity` directly and match on `minute.is_zero() && time_granularity == TimeGranularity::Hours`.

components/datetime/tests/simple_test.rs
sffc: :robot: Done! Simplified `test_minute_optional_hour_cycle` to test only using `NoCalendarFormatter`.

sffc (commented):
:robot:  Addressed code review comments from @robertbastian.

components/datetime/src/raw/neo.rs
sffc: I disagree with this change. I am overloading `TimeGranularity` to mean two different things for more efficient data storage and bitpacking. The condition `minute.is_zero() && time_granularity == TimeGranularity::Hours` is technically correct but extremely misleading.

components/datetime/src/provider/pattern/runtime/helpers.rs
```diff
@@ -29,6 +29,8 @@ pub fn maybe_replace_first(pattern: &mut Pattern, f: impl Fn(&PatternItem) -> Op
     if let Some((i, result)) = result {
         pattern.items.to_mut_slice()[i] = result.to_unaligned();
     }
+    // Keep the pattern metadata up-to-date since replacing a field could change it
+    pattern.metadata = PatternMetadata::from_iter_items(pattern.items.iter());
```
sffc, line 33: Note: I don't love that we need to manually do this, because it raises questions about where else we need to manually update the metadata. I'm not interested in doing a full refactoring right now, but it would be a good cleanup for the future.

components/datetime/src/raw/neo.rs
sffc: I did it similarly but slightly differently, passing down a `CoarseHourCycle`.

components/datetime/src/provider/pattern/runtime/helpers.rs
robertbastian: should be fairly simple to audit the callers of `to_mut_slice`?

components/datetime/src/raw/neo.rs
robertbastian: > I am overloading `TimeGranularity` to mean two different things for more efficient data storage and bitpacking.

elaborate? this is not clear to me

pulling in a *third* hour type does not make this more understandable

robertbastian (commented):
.

components/datetime/src/raw/neo.rs
sffc: Previously, TimeGranularity meant one thing: the least significant time field in a pattern.

- "yMd" => None
- "E h" => Hours
- "H:mm" => Minutes
- …

I am splitting h12 and h23 into different fields now:

- "yMd" => None
- "H" => None (previously Hours) -- this is why I wanted to rename the variant
- "h" => Hours
- "H:mm" => Minutes

And I'm using that split to decide when to use MinuteOptional. However, the split is only observable in patterns that have only an hour, not ones that have minutes and seconds. This means I need to check the hours-only pattern even if I'm formatting with minutes and seconds in order to get the hour cycle in constant time.

I could have instead added a new bit to the pattern metadata to store the h12/h23 on all patterns regardless of their time granularity. I didn't do that because it isn't necessary and because it uses the last bit we have from the 4-bit metadata.

robertbastian: thanks for the explanation. it might be worth modelling this more in the type system and less in comments, like having the variant be `Hour(CoarseHourCycle)`

robertbastian (approved):

PR #8238, opened 2026-07-17
Add Writeable Replace adapter

sffc:
This is a zero-allocation string replacement adapter for Writeables.

I wrote the architecture and let an AI write the implementation. It used Knuth-Morris-Pratt. You can see which commits are human and which commits are AI.

Note: the needle needs to be `impl AsRef<str>` instead of `impl Writeable` in order to implement this in a no-allocation fashion, since it is sometimes necessary to backtrack. However, both the source and the replacement can be `impl Writeable`.

For now I'm only adding an adapter struct. In the future we could add an extension trait to allow things like

```rust
my_writeable.replace("a", "b")
```

## Changelog

writeable: new struct `writeable::adapters::Replace`

### DIFF
--- a/utils/writeable/src/replace.rs
+++ b/utils/writeable/src/replace.rs
@@ -140,7 +140,7 @@ where
         // KMP State Transition:
         // While we have a mismatch and we are not at the start of the needle,
         // backtrack using the prefix function.
-        while matched > 0 && self.remaining_needle.chars().next() != Some(c) {
+        while matched > 0 && !self.remaining_needle.starts_with(c) {
             let old_j = matched;
             matched = get_pi_bytes(self.needle, old_j);
             // Since we backtracked, the prefix of length `old_j - j` is no longer
@@ -152,7 +152,7 @@ where
         }
 
         // If the character matches the next character in the needle, advance the match state.
-        if self.remaining_needle.chars().next() == Some(c) {
+        if self.remaining_needle.starts_with(c) {
             // Advance remaining_needle by the matched character.
             self.remaining_needle = self
                 .remaining_needle
@@ -172,7 +172,7 @@ where
     }
 }
 
-impl<'a, A, C> Writeable for Replace<A, &'a str, C>
+impl<A, C> Writeable for Replace<A, &str, C>
 where
     A: Writeable,
     C: Writeable,
--- a/utils/writeable/src/replace.rs
+++ b/utils/writeable/src/replace.rs
@@ -29,7 +29,7 @@ use core::fmt;
 /// assert_writeable_eq!(replace, "I 💖 Rust and Rust loves me!");
 /// ```
 #[derive(Debug)]
-#[allow(clippy::exhaustive_structs)] // designed for nesting
+#[allow(clippy::exhaustive_structs)] // well-defined 3-tuple
 pub struct Replace<A, B, C> {
     /// The source writeable.
     pub source: A,
--- a/utils/writeable/src/replace.rs
+++ b/utils/writeable/src/replace.rs
@@ -4,14 +4,11 @@
 
 use crate::{impl_display_with_writeable, LengthHint, Writeable};
 use core::fmt;
-use core::ops::Range;
 
 /// A [`Writeable`] adapter that replaces occurrences of a needle with a replacement.
 ///
 /// This adapter performs the replacement in a streaming fashion during `write_to`,
 /// requiring zero allocations.
-///
-/// All occurrences of the needle are replaced.
 ///
 /// # Examples
 ///
@@ -40,13 +37,13 @@ pub struct Replace<A, B, C> {
     pub replacement: C,
 }
 
-/// Computes the Knuth-Morris-Pratt (KMP) prefix function (failure function) value
-/// for the character prefix ending at byte index `matched_bytes` in `needle`.
-///
-/// Returns the byte length of the longest proper prefix of `needle[0..matched_bytes]`
-/// that is also a suffix of `needle[0..matched_bytes]`.
-///
-/// This is computed on the fly without allocation by iterating over char boundaries.
+// Computes the Knuth-Morris-Pratt (KMP) prefix function (failure function) value
+// for the character prefix ending at byte index `matched_bytes` in `needle`.
+//
+// Returns the byte length of the longest proper prefix of `needle[0..matched_bytes]`
+// that is also a suffix of `needle[0..matched_bytes]`.
+//
+// This is computed on the fly without allocation by iterating over char boundaries.
 fn get_pi_bytes(needle: &str, matched_bytes: usize) -> usize {
     let s = match needle.get(0..matched_bytes) {
         Some(s) => s,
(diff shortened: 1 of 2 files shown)


utils/writeable/src/replace.rs
```diff
@@ -0,0 +1,318 @@
...
+    replacement: &'a C,
+    // The number of bytes of `needle` matched so far.
+    // This is always aligned to a character boundary.
+    matched_bytes: usize,
+}
+
+impl<'a, W, B, C> ReplaceWriter<'a, W, B, C>
+where
+    W: fmt::Write + ?Sized,
+    B: AsRef<str>,
```
robertbastian: I would want this to become [`Pattern`](https://doc.rust-lang.org/std/str/pattern/trait.Pattern.html) in the future, and that's not compatible with `AsRef<str>`. Just take `str` for now.

utils/writeable/src/replace.rs
```diff
@@ -0,0 +1,318 @@
...
+struct ReplaceWriter<'a, W: ?Sized, B, C> {
+    // The underlying sink to write to.
+    sink: &'a mut W,
+    // The needle we are searching for.
+    needle: &'a B,
+    // The replacement to write when the needle is matched.
+    replacement: &'a C,
+    // The number of bytes of `needle` matched so far.
+    // This is always aligned to a character boundary.
+    matched_bytes: usize,
```
robertbastian: you can encode the "always aligned to a character boundary" by storing a `&'a str` instead

utils/writeable/src/replace.rs
```diff
@@ -0,0 +1,318 @@
...
+            if replacement_hint.0 == replacement_len && needle_len == replacement_len {
+                return source_hint;
+            }
+        }
+
+        LengthHint::undefined()
+    }
+
+    // TODO: how should we implement write_to_parts? What if the needle spans a part?
+    // TODO: should we implement writeable_borrow, if a replacement doesn't occur? It is O(N)
```
robertbastian: `writeable_borrow` is meant to be constant, I thought we had documented this somewhere

utils/writeable/src/replace.rs
```diff
@@ -0,0 +1,318 @@
...
+        let needle_len = self.needle.as_ref().len();
+        let replacement_hint = self.replacement.writeable_length_hint();
+
+        if let Some(replacement_len) = replacement_hint.1 {
+            if replacement_hint.0 == replacement_len && needle_len == replacement_len {
+                return source_hint;
+            }
+        }
+
+        LengthHint::undefined()
```
robertbastian: that's not correct. if the needle is smaller than the replacement, we have a lower bound, if it's larger we have an upper bound

utils/writeable/src/replace.rs
```diff
@@ -0,0 +1,318 @@
...
+impl<A, B, C> Writeable for Replace<A, B, C>
+where
+    A: Writeable,
+    B: AsRef<str>,
+    C: Writeable,
+{
+    fn write_to<W: fmt::Write + ?Sized>(&self, sink: &mut W) -> fmt::Result {
+        let mut writer = ReplaceWriter::new(sink, &self.needle, &self.replacement);
+        self.source.write_to(&mut writer)?;
+        writer.flush()
```
robertbastian: I confused this with the `flush` method on `fmt::Write` (sinks get flushed). call this something else, like `finalize`

utils/writeable/src/replace.rs
```diff
@@ -0,0 +1,318 @@
...
+
+impl<'a, W, B, C> fmt::Write for ReplaceWriter<'a, W, B, C>
+where
+    W: fmt::Write + ?Sized,
+    B: AsRef<str>,
+    C: Writeable,
+{
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        for c in s.chars() {
+            self.write_char_buffered(c)?;
```
robertbastian: style: make this call `write_char`, and inline `write_char_buffered` into `write_char`

utils/writeable/src/replace.rs
```diff
@@ -0,0 +1,318 @@
...
+        self.needle
+            .as_ref()
+            .get(index..)
+            .and_then(|s| s.chars().next())
+    }
+
+    // Processes a single character from the source stream.
+    fn write_char_buffered(&mut self, c: char) -> fmt::Result {
+        let needle_str = self.needle.as_ref();
+        let needle_len_bytes = needle_str.len();
```
robertbastian: style: don't create a variable for such constant time operations. `needle_str.len()` is both shorter than `needle_len_bytes`, and always guaranteed to be the length of `needle_str`. seeing `needle_len_bytes` 20 lines below is less clear

robertbastian: in fact you also use `needle_str.len()` below. I'm gonna stop my review here and let you do an AI cleanup pass

sffc: Robert, I did do an AI cleanup pass, which you can see from the commit history. I just didn't notice it or think this is worth my time to fix. But fixing it is less time than writing this comment so I will fix it.

utils/writeable/src/replace.rs
sffc: Praise: It's nice that this gives a tight bound when needle and replacement are the same length.

utils/writeable/src/replace.rs
```diff
@@ -74,146 +71,154 @@ fn get_pi_bytes(needle: &str, matched_bytes: usize) -> usize {
...
+        // If needle and replacement have same exact length, length is unchanged.
+        if let Some(r_upper) = replacement_hint.1 {
+            if replacement_hint.0 == r_upper && needle_len == r_upper {
                 return source_hint;
             }
         }
 
-        LengthHint::undefined()
-    }
+        let mut lower = 0;
```
robertbastian, line 200: eh, just make `source_hint` mutable

utils/writeable/src/replace.rs
robertbastian: you only resolved the second comment

robertbastian (approved):
I'm at a point where I'm just giving up on readability and just pray I never have to touch this code.

utils/writeable/src/replace.rs
sffc: fixed in #8242

robertbastian: > I just didn't notice it or think this is worth my time to fix.

fixing it while you're already cleaning up this code takes maybe 10 seconds, because you know that `needle_str` isn't reassigned later, and these are actually always the same value. me having to figure this out takes more time than that. 

as the PR author, who is asking for my time to review, you should go out of your way to make the review straightforward. that means simplifying code and removing any potential of misunderstandings. this is especially important with AI generated code; as the reviewer I want to see that you fully understand what the AI is doing. if my AI had generated this code, I would have removed the variable so that my reviewer does not have to repeat the effort of figuring out what the AI is doing, and especially whether the AI introduced this variable for a reason and then forgot about it.

if you think your time as the author is worth more than the time of your reviewer, we have a problem

